PR #22, opened 2019-07-22
Spike

aeneasr:


cmd/tools.go
```diff
@@ -0,0 +1,17 @@
+// +build tools
+
+package cmd
+
+import (
```
ermik: Go best practices advise to comment on `_` imports to inform the reader of their... importance.

driver/registry_memory.go
```diff
@@ -0,0 +1,308 @@
...
+type postHooks []interface {
+	selfservice.HookLoginPostExecutor
+	selfservice.HookRegistrationPostExecutor
+}
+
+func (m *RegistryMemory) hooksPost(credentialsType identity.CredentialsType, configs []configuration.SelfServiceHook) postHooks {
+	var i postHooks
+
+	for _, hook := range configs {
+		switch hook.Run {
```
ermik: Could these switch cases be further refactored into case-specific functions? A lot is happening in one switch.

identity/pool_memory.go
```diff
@@ -0,0 +1,167 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/ory/go-convenience/stringslice"
+
+	"github.com/ory/herodot"
+	"github.com/ory/x/pagination"
+)
+
+var _ Pool = new(PoolMemory)
```
ermik: Is this an interface conformance check?

identity/pool_memory.go
```diff
@@ -0,0 +1,167 @@
...
+	defer p.RUnlock()
+
+	start, end := pagination.Index(limit, offset, len(p.is))
+	return p.is[start:end], nil
+}
+
+func (p *PoolMemory) Update(_ context.Context, i *Identity) (*Identity, error) {
+	p.RLock()
+	for k, ii := range p.is {
+		if ii.ID == i.ID {
```
ermik: How about:
```go
if ii.ID != i.ID {
        continue
}
// code  currently inside the conditional
```

identity/pool_memory.go
aeneasr: Yes

driver/registry_memory.go
aeneasr: Yeah most likely

PR #146, opened 2019-12-10
courier: Implement message templates and SMTP delivery

### LINKED ISSUE #99: hermes: Notification architecture
Several system components require the ability to notify a identity about something:

- Account activation / verification must send out an email or SMS containing a code for activating the account.
- Account recovery must send out an email with a password reset link
- Sending out a welcome email
- ...

To support this we need a system capable of delivering notifications using several protocols:

- SMTP
- SMS (Maybe Twillio or something similar?)

As well as being able to render the notification message itself with localization.

One possibility to solve templating would be to provide default (english) templates for all protocols and workflows and leave open the possibility to "mount" a directory (think Kubernetes) to load localized and customized templates:

```
+ default
  + email
    + account_recovery.html.tmpl (for html templates)
    + account_recovery.tmpl
    + ...
  + sms
    + account_recovery.tmpl
    + ...
```

plus a config variable such as `hermes.templates` pointing at a directory with a structure such as:

```
+ default
  + email
    + account_recovery.html.tmpl (for html templates)
    + account_recovery.tmpl
    + ...
  + sms
    + account_recovery.tmpl
    + ...
+ i18n
  + de_DE
    + email
      + ...
```

aeneasr:
This patch adds a message templates (with override capabilities)
and SMTP delivery.

Integration tests using MailHog test fault resilience and e2e email
delivery.

This system is designed to be extended for SMS and other use cases.

Closes #99


merlindorin:
@aeneasr why this implementation focus on smtp and not on a generic approach like a notification webhook?

With a webhook, we are implementation "free" and it will be easy to cover cases like "sms", "telegram notif" or whatever... maybe it's on the roadmap?

aeneasr:
Because webhooks would require a specific format depending on the upstream service (telegram works different from whatsapp from facebook from ...). SMTP and email are standards however.

But the general idea is to support multiple senders here (SMS for example) and maybe stuff like AWS SNS or similar.

merlindorin:
I was more thinking about a generic webhook that require an external adapter that will talk, if necessary, to an external service. The main goal is to be free of `ory` implementation (smtp, template, etc.) and do what we need through a custom service. In that case, we only need to be able to read json in order to set the execution context and do what we want.

For example:

```text
+-------------------+    +---------+    +--------------+    +-------------+
| user registration +--->| webhook +--->| email-sender +--->| smtp server |
+-------------------+    +---------+    +--------------+    +-------------+
```

Thank you for the quick reply :) <3

aeneasr:
Yeah a generic webhook also makes sense, but it would obviously increase the amount of time required to adopt ORY Kratos as one would have to implement the email adapter. I generally think though that a generic webhook additionally to SMTP would be a good idea! Would you be open to create an issue for that, or - even better - make a contribution towards this? :)

merlindorin:
Clearly, it's more complicated for a quick use. Will open an issue... not sure I will have enough time for a contribution. thx again!

PR #148, opened 2019-12-11
Optionally allow only one active session per identity 

### LINKED ISSUE #139: Optionally allow only one active session per identity

In some cases, it is considered good practice to have only one simoultaneous logon session. This may apply to online banking and other highly sensitive services.

It should be possible to deactivate all active logon session for that particular user on a successful login.

This could easily be achieved with the workflow engine (just write a workflow that removes all active sessions for the user).

**Context**

It should be made clear that this behavior may cause issues when trying to save data (e.g. submit form on a now logged out session). This should be used with care and consideration.

evalsocket:
## Related issue
This PR will close #139 

## Proposed changes

- [x] Add session_destroyer hook 
- [x] Add session_destroyer hook testcase
- [x] Create a new key for hook `	KeySessionDestroyer = "revoke_active_sesions" `
- [x] Register hook in `registry_default_hooks.go `

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation within the code base (if appropriate)
- [x] I have documented my changes in the [developer guide](https://github.com/ory/docs) (if appropriate)

### DIFF
--- a/docs/.kratos.yaml
+++ b/docs/.kratos.yaml
@@ -68,14 +68,14 @@ selfservice:
     after:
       password:
         - run: session
-        - run: revoke_active_sesions
+        - run: revoke_active_sessions
         - run: redirect
           config:
             default_redirect_url: http://test.kratos.ory.sh:4000/
             allow_user_defined_redirect: true
       oidc:
         - run: session
-        - run: revoke_active_sesions
+        - run: revoke_active_sessions
         - run: redirect
           config:
             default_redirect_url: http://test.kratos.ory.sh:4000/
--- a/docs/.kratos.yaml
+++ b/docs/.kratos.yaml
@@ -68,12 +68,14 @@ selfservice:
     after:
       password:
         - run: session
+        - run: revoke_active_sesions
         - run: redirect
           config:
             default_redirect_url: http://test.kratos.ory.sh:4000/
             allow_user_defined_redirect: true
       oidc:
         - run: session
+        - run: revoke_active_sesions
         - run: redirect
           config:
             default_redirect_url: http://test.kratos.ory.sh:4000/
--- a/driver/registry_default_hooks.go
+++ b/driver/registry_default_hooks.go
@@ -13,7 +13,6 @@ import (
 
 func (m *RegistryDefault) hooksPost(credentialsType identity.CredentialsType, configs []configuration.SelfServiceHook) postHooks {
 	var i postHooks
-
 	for _, h := range configs {
 		switch h.Run {
 		case hook.KeySessionIssuer:
--- a/driver/registry_default_hooks.go
+++ b/driver/registry_default_hooks.go
@@ -21,6 +21,11 @@ func (m *RegistryDefault) hooksPost(credentialsType identity.CredentialsType, co
 				i,
 				hook.NewSessionIssuer(m),
 			)
+		case hook.KeySessionDestroyer:
+			i = append(
+				i,
+				hook.NewSessionDestroyer(m),
+			)
 		case hook.KeyRedirector:
 			var rc struct {
 				R string `json:"default_redirect_url"`
--- a/driver/registry_default_hooks.go
+++ b/driver/registry_default_hooks.go
@@ -13,6 +13,7 @@ import (
 
 func (m *RegistryDefault) hooksPost(credentialsType identity.CredentialsType, configs []configuration.SelfServiceHook) postHooks {
 	var i postHooks
+
 	for _, h := range configs {
 		switch h.Run {
 		case hook.KeySessionIssuer:
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/gofrs/uuid"
-	"github.com/pkg/errors"
 
 	"github.com/ory/x/sqlcon"
 
@@ -30,12 +29,9 @@ func (p *Persister) DeleteSession(ctx context.Context, sid uuid.UUID) error {
 }
 
 func (p *Persister) DeleteSessionsFor(ctx context.Context, sid uuid.UUID) error {
-	count, err := p.c.RawQuery("DELETE FROM sessions WHERE identity_id =?", sid).ExecWithCount()
+	count, err := p.c.RawQuery("DELETE FROM sessions WHERE identity_id =?", sid).Exec()
 	if err != nil {
 		return sqlcon.HandleError(err)
-	}
-	if count == 0 {
-		return errors.WithStack(sqlcon.ErrNoRows)
 	}
 	return nil
 }
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -29,8 +29,7 @@ func (p *Persister) DeleteSession(ctx context.Context, sid uuid.UUID) error {
 }
 
 func (p *Persister) DeleteSessionsFor(ctx context.Context, sid uuid.UUID) error {
-	count, err := p.c.RawQuery("DELETE FROM sessions WHERE identity_id =?", sid).Exec()
-	if err != nil {
+	if err := p.c.RawQuery("DELETE FROM sessions WHERE identity_id =?", sid).Exec(); err != nil {
 		return sqlcon.HandleError(err)
 	}
 	return nil
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/gofrs/uuid"
+	"github.com/pkg/errors"
 
 	"github.com/ory/x/sqlcon"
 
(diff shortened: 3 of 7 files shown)


selfservice/hook/session_destroyer.go
```diff
@@ -0,0 +1,36 @@
...
+	return &SessionDestroyer{r: r}
+}
+
+func (e *SessionDestroyer) ExecuteRegistrationPostHook(w http.ResponseWriter, r *http.Request, a *registration.Request, s *session.Session) error {
+
+	return nil
+}
+
+func (e *SessionDestroyer) ExecuteLoginPostHook(w http.ResponseWriter, r *http.Request, a *login.Request, s *session.Session) error {
+	if err := e.r.SessionPersister().DeleteSession(r.Context(), s.Identity.ID); err != nil {
```
aeneasr: `DeleteSession` accepts the session ID, so you can not use `s.Identity.ID` here. Instead, you need to add a new method to [`session.Persister`](https://github.com/ory/kratos/blob/master/session/persistence.go#L31) such as `DeleteSessionsFor(ctx context.Context, identityID uuid.UUID) error`.

Next, you need to implement a test for that function [around here](https://github.com/ory/kratos/blob/master/session/persistence.go#L67).

Next, you need to implement the actual storage logic [here](https://github.com/ory/kratos/blob/master/persistence/sql/persister_session.go#L22).

Once that is done, you should run:

```
go test -short -tags sqlite ./...
```

in the root directory of the project and see if tests pass.

Then, you will use that new method you implemented here.

selfservice/hook/session_destroyer_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/ory/kratos/driver/configuration"
+	"github.com/ory/kratos/identity"
+	"github.com/ory/kratos/internal"
+	"github.com/ory/kratos/selfservice/hook"
+	"github.com/ory/kratos/session"
+	"github.com/ory/kratos/x"
+)
+
+func TestSessionIssuer(t *testing.T) {
```
aeneasr: This can not work because the function name is already reserved by another test. This should be something like `TestSessionDestroyer`

driver/registry_default_hooks.go
```diff
@@ -13,7 +13,6 @@ import (
 
 func (m *RegistryDefault) hooksPost(credentialsType identity.CredentialsType, configs []configuration.SelfServiceHook) postHooks {
 	var i postHooks
-
```
aeneasr: Revert this change

selfservice/hook/session_destroyer.go
```diff
@@ -0,0 +1,36 @@
...
+type SessionDestroyer struct {
+	r sessionDestroyerDependencies
+}
+
+func NewSessionDestroyer(r sessionDestroyerDependencies) *SessionDestroyer {
+	return &SessionDestroyer{r: r}
+}
+
+func (e *SessionDestroyer) ExecuteRegistrationPostHook(w http.ResponseWriter, r *http.Request, a *registration.Request, s *session.Session) error {
+
```
aeneasr: remove newline

selfservice/hook/session_destroyer_test.go
```diff
@@ -0,0 +1,43 @@
...
+	_, reg := internal.NewRegistryDefault(t)
+	viper.Set(configuration.ViperKeyURLsSelfPublic, "http://localhost/")
+	viper.Set(configuration.ViperKeyDefaultIdentityTraitsSchemaURL, "file://./stub/stub.schema.json")
+
+	var r http.Request
+	h := hook.NewSessionIssuer(reg)
+
+	t.Run("method=sign-in", func(t *testing.T) {
```
aeneasr: This should be `method=ExecuteLoginPostHook`

selfservice/hook/session_destroyer_test.go
```diff
@@ -0,0 +1,43 @@
...
+		w := httptest.NewRecorder()
+		sid := x.NewUUID()
+
+		i := identity.NewIdentity("")
+		require.NoError(t, reg.IdentityPool().CreateIdentity(context.Background(), i))
+		require.NoError(t, h.ExecuteLoginPostHook(w, &r, nil, &session.Session{ID: sid, Identity: i}))
```
aeneasr: This test isn't actually testing the functionality of what you've implemented. Instead, you need to do something along the lines of:

1. Create a new identity
2. Create 2 or more sessions for that identity
3. Run `ExecuteLoginPostHook`
4. Check if these 2 (or more) sessions have been removed from the store. To prove that they're gone, `GetSession` should return an error of type `sqlcon.ErrNoRows`.

driver/registry_default_hooks.go
evalsocket: done

selfservice/hook/session_destroyer.go
evalsocket: done

selfservice/hook/session_destroyer.go
evalsocket: done

selfservice/hook/session_destroyer_test.go
evalsocket: done

docs/.kratos.yaml
```diff
@@ -68,12 +68,14 @@ selfservice:
     after:
       password:
         - run: session
+        - run: revoke_active_sesions
```
aeneasr: Typo, this should be `revoke_active_sessions`

docs/.kratos.yaml
```diff
@@ -68,12 +68,14 @@ selfservice:
...
         - run: redirect
           config:
             default_redirect_url: http://test.kratos.ory.sh:4000/
             allow_user_defined_redirect: true
       oidc:
         - run: session
+        - run: revoke_active_sesions
```
aeneasr: See above

persistence/sql/persister_session.go
```diff
@@ -27,3 +28,14 @@ func (p *Persister) CreateSession(ctx context.Context, s *session.Session) error
 func (p *Persister) DeleteSession(ctx context.Context, sid uuid.UUID) error {
 	return p.c.Destroy(&session.Session{ID: sid}) // This must not be eager or identities will be created / updated
 }
+
+func (p *Persister) DeleteSessionsFor(ctx context.Context, sid uuid.UUID) error {
+	count, err := p.c.RawQuery("DELETE FROM sessions WHERE identity_id =?", sid).ExecWithCount()
+	if err != nil {
+		return sqlcon.HandleError(err)
+	}
+	if count == 0 {
```
aeneasr: We don't really care if count == 0

persistence/sql/persister_session.go
```diff
@@ -27,3 +28,14 @@ func (p *Persister) CreateSession(ctx context.Context, s *session.Session) error
 func (p *Persister) DeleteSession(ctx context.Context, sid uuid.UUID) error {
 	return p.c.Destroy(&session.Session{ID: sid}) // This must not be eager or identities will be created / updated
 }
+
+func (p *Persister) DeleteSessionsFor(ctx context.Context, sid uuid.UUID) error {
+	count, err := p.c.RawQuery("DELETE FROM sessions WHERE identity_id =?", sid).ExecWithCount()
```
aeneasr: We don't need `ExecWithCount`, `Exec` should be enough

selfservice/hook/session_destroyer_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	t.Run("method=ExecuteLoginPostHook", func(t *testing.T) {
+		w := httptest.NewRecorder()
+		sid := x.NewUUID()
+
+		i := identity.NewIdentity("")
+		require.NoError(t, reg.IdentityPool().CreateIdentity(context.Background(), i))
+		err := reg.SessionPersister().CreateSession(context.Background(), &session.Session{ID: sid, Identity: i}))
+		require.NoError(t, err)
+		err = reg.SessionPersister().CreateSession(context.Background(), &session.Session{ID: sid, Identity: i}))
```
aeneasr: This creates the same session twice, instead, it would be better to create two individual sessions with individual IDs

session/persistence.go
```diff
@@ -75,5 +78,17 @@ func TestPersister(p interface {
 			_, err := p.GetSession(context.Background(), expected.ID)
 			require.Error(t, err)
 		})
+
+		t.Run("case=delete session for", func(t *testing.T) {
+			var expected Session
+			require.NoError(t, faker.FakeData(&expected))
+			require.NoError(t, p.CreateIdentity(context.Background(), expected.Identity))
+			require.NoError(t, p.CreateSession(context.Background(), &expected))
+			require.NoError(t, p.CreateSession(context.Background(), &expected))
```
aeneasr: Here too, it would be better to create two individual sessions and create them :) That ensures that really ALL sessions are revoked, not just one

selfservice/hook/session_destroyer_test.go
evalsocket: are you saying that i need two sid ?
```
err := reg.SessionPersister().CreateSession(context.Background(), &session.Session{ID: sid, Identity: i}))
err := reg.SessionPersister().CreateSession(context.Background(), &session.Session{ID: sid2, Identity: i}))
```

persistence/sql/persister_session.go
evalsocket: done

persistence/sql/persister_session.go
evalsocket: done

docs/.kratos.yaml
evalsocket: done

docs/.kratos.yaml
evalsocket: done

selfservice/hook/session_destroyer_test.go
evalsocket: done

session/persistence.go
```diff
@@ -29,6 +29,9 @@ type Persister interface {
 
 	// Delete removes a session from the store
 	DeleteSession(ctx context.Context, sid uuid.UUID) error
+
+	// Delete removes a session the store
```
aeneasr: This should describe what the function is doing, so more something like:

```suggestion
	// DeleteSessionsFor removes all active session from the store for the given identity.
```

aeneasr:
Awesome, thank you for your hard work! Well done :)

PR #157, opened 2019-12-19
Serve Json Schemas

### LINKED ISSUE #86: Schemas should be mirrored by hive at some well-known url

zepatrik:
## Related issue

closes #86 

## Proposed changes

Schemas will be organized in their own table and mirrored by kratos.

## Checklist


zepatrik:
ping @aeneasr pls have a look

aeneasr:
hm why does this have the README patch in it?

contrib/sql/migrations/20191000000000_json_schemas.up.fizz
```diff
@@ -0,0 +1,6 @@
+create_table("json_schemas") {
+	t.Column("id", "uuid", {primary: true})
```
aeneasr: indent

persistence/sql/persister_identity.go
```diff
@@ -115,6 +116,10 @@ func createIdentityCredentials(ctx context.Context, tx *pop.Connection, i *ident
 	return nil
 }
 
+func (p *Persister) injectTraitsUrl(i *identity.Identity) {
+	i.TraitsSchemaURL = fmt.Sprintf("%s/schemas/%s", p.cf.PublicListenOn(), i.TraitsSchemaID)
```
aeneasr: PublicListenOn() returns e.g. `0.0.0.0:1234` - so this will probably not work. Instead, use `SelfURLsPublic` (I think that's the name)

persistence/sql/persister_identity.go
```diff
@@ -139,7 +144,13 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
 
 func (p *Persister) ListIdentities(ctx context.Context, limit, offset int) ([]identity.Identity, error) {
 	is := make([]identity.Identity, 0)
-	return is, sqlcon.HandleError(p.c.RawQuery("SELECT * FROM identities LIMIT ? OFFSET ?", limit, offset).All(&is))
+	err := sqlcon.HandleError(p.c.RawQuery("SELECT * FROM identities LIMIT ? OFFSET ?", limit, offset).All(&is))
+
+	for i := range is {
+		p.injectTraitsUrl(&is[i])
```
aeneasr: This definitely needs a test. Pointers behave unexpectedly in loops, I think this is fine as it is but make sure to test it!

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"errors"
+	"github.com/gobuffalo/pop"
+	"github.com/gofrs/uuid"
+	"github.com/ory/kratos/schema"
+	"github.com/ory/x/sqlcon"
+)
+
+func (p *Persister) GetSchema(id uuid.UUID) (*schema.JsonSchema, error) {
+	var s schema.JsonSchema
```
aeneasr: I think we need to decide if this is `json_schema` or `schema`. I think we should just go for `schema` - creates less confusion in the code base.

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,27 @@
...
+	if err := p.c.Find(&s, id); err != nil {
+		return nil, sqlcon.HandleError(err)
+	}
+	return &s, nil
+}
+
+func (p *Persister) CreateSchema(s schema.JsonSchema) error {
+	if s.Url == "" {
```
aeneasr: In Go, we capitalize abbreviations -> `s.URL`

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,27 @@
+package sql
+
+import (
+	"errors"
```
aeneasr: Make sure to import `github.com/pkg/errors` instead - it adds stack traces!

zepatrik:
I don't know... rebased as always but could not get rid of it

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,27 @@
...
+	var s schema.JsonSchema
+	if err := p.c.Find(&s, id); err != nil {
+		return nil, sqlcon.HandleError(err)
+	}
+	return &s, nil
+}
+
+func (p *Persister) CreateSchema(s schema.JsonSchema) error {
+	if s.Url == "" {
+		return errors.New("URL must be provided")
```
aeneasr: We can add more context here. We have the herodot package which allows us to return errors with extra info - such as HTTP status code. This mistake would case a 400 Bad Request (because the client sent invalid payload):

```
return errors.WithStack(herodot.ErrBadRequest.WithReason("The schema is missing the URL property."))
```

schema/handler.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (h *Handler) get(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	s, err := h.r.SchemaPool().GetSchema(x.ParseUUID(ps.ByName("id")))
+	if err != nil {
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
+
+	h.r.Writer().Write(w, r, s)
```
aeneasr: I think we should only respond with the actual schema content, not the wrapper? So probably something like `Write(w, r, s.Schema)`

aeneasr: So I just saw that we don't have the schema yet - only it's url. We should probably get that somewhere. Either here (with caching), or store it in the database directly. In both cases we need a strategy to update the schema if it changes - but we don't need to do that now (but please create an issue if you skip the refresh here).

schema/pool.go
```diff
@@ -0,0 +1,19 @@
+package schema
+
+import "github.com/gofrs/uuid"
+
+type (
+	JsonSchema struct {
```
aeneasr: Rename to Schema

schema/pool.go
```diff
@@ -0,0 +1,19 @@
...
+		ID  uuid.UUID `json:"-" faker:"uuid" db:"id" rw:"r"`
+		Url string    `json:"-" faker:"-" db:"url"`
```
aeneasr: Rename to URL

Also I think it's safe to return the URL in json even if it's not really used anywhere.

schema/pool.go
```diff
@@ -0,0 +1,19 @@
...
+	}
+
+	Pool interface {
```
aeneasr: This should be renamed to `Persister`. We use `Pool` only in identity and it's something I wanted to rename for some while...

zepatrik:
How should client creation work? Should there be the `traits_schema_url` field that will be translated into an ID if it exists or create a new schema entry in the table if it doesn't exist?

aeneasr:
So we need a `unique` constraint on the `url` field in SQL (this is already done). Then we would check against that unique constraint to see if it already exists and if it doesn't, we create a new row in the table. The ID is generated by the database system, not by kratos!

We need to do this every time we store an identity.

aeneasr:
let me know when you want me to take another look

zepatrik:
Now I just have to fix all the tests and write new ones.

identity/handler.go
```diff
@@ -105,6 +107,8 @@ func (h *Handler) update(w http.ResponseWriter, r *http.Request, ps httprouter.P
 	}
 
 	i.ID = x.ParseUUID(ps.ByName("id"))
+	// Make sure the TraitsSchemaURL is only set by kratos
+	i.TraitsSchemaURL = ""
```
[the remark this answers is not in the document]
aeneasr: In fact, please use `decoderx.NewHTTP()` as seen in selfservice/profile/handler

identity/handler_test.go
```diff
@@ -22,16 +22,22 @@ import (
...
+	reg.SchemaHandler().RegisterAdminRoutes(adminRouter)
+	reg.SchemaHandler().RegisterPublicRoutes(publicRouter)
+	tsa := httptest.NewServer(adminRouter)
+	tsp := httptest.NewServer(publicRouter)
+	defer tsa.Close()
+	defer tsp.Close()
+
+	viper.Set(configuration.ViperKeyURLsSelfAdmin, tsa.URL)
+	viper.Set(configuration.ViperKeyURLsSelfPublic, tsp.URL)
+	_, _ = reg.SchemaPersister().RegisterDefaultSchema("file://./stub/identity.schema.json")
```
aeneasr: Error handling

identity/handler_test.go
```diff
@@ -22,16 +22,22 @@ import (
...
-	defer ts.Close()
-
-	viper.Set(configuration.ViperKeyURLsSelfAdmin, ts.URL)
-	viper.Set(configuration.ViperKeyDefaultIdentityTraitsSchemaURL, "file://./stub/identity.schema.json")
+	adminRouter := x.NewRouterAdmin()
+	publicRouter := x.NewRouterPublic()
+	reg.IdentityHandler().RegisterAdminRoutes(adminRouter)
+	reg.SchemaHandler().RegisterAdminRoutes(adminRouter)
+	reg.SchemaHandler().RegisterPublicRoutes(publicRouter)
+	tsa := httptest.NewServer(adminRouter)
```
aeneasr: You probably don't need admin I guess

identity/handler_test.go
```diff
@@ -77,11 +83,12 @@ func TestHandler(t *testing.T) {
...
 
-	t.Run("case=should fail to create an entity because schema url does not exist", func(t *testing.T) {
+	t.Run("case=should fail to create an entity because schema ID does not exist", func(t *testing.T) {
 		var i identity.Identity
-		i.TraitsSchemaURL = "file://./stub/does-not-exist.schema.json"
-		res := send(t, "POST", "/identities", http.StatusInternalServerError, &i)
-		assert.Contains(t, res.Get("error.reason").String(), "does-not-exist.schema.json")
+		i.TraitsSchemaID = x.NewUUID()
+		// TODO what should the status code/message be? (deps not found)
+		res := send(t, "POST", "/identities", http.StatusNotFound, &i)
```
aeneasr: StatusBadRequest - 400

identity/identity.go
```diff
@@ -86,6 +92,10 @@ func (i Identity) TableName() string {
 	return "identities"
 }
 
+func (i *Identity) InjectTraitsUrl(host string) {
+	i.TraitsSchemaURL = urlx.MustJoin(host, "schemas", i.TraitsSchemaID.String())
```
aeneasr: Make this a const or move this function (partially) to `schema` package, zb: `schema.SchemaURL(id uuid.UUID)`

identity/identity.go
```diff
@@ -160,12 +170,12 @@ func (i *Identity) CopyWithoutCredentials() *Identity {
...
 	return &Identity{
-		ID:              x.NewUUID(),
-		Credentials:     map[CredentialsType]Credentials{},
-		Traits:          Traits(json.RawMessage("{}")),
-		TraitsSchemaURL: traitsSchemaURL,
-		l:               new(sync.RWMutex),
+		ID:             x.NewUUID(),
+		Credentials:    map[CredentialsType]Credentials{},
+		Traits:         Traits(json.RawMessage("{}")),
+		TraitsSchemaID: traitsSchemaID,
```
aeneasr: inject url if id is not empty - or make `SchemaURL` so that it returns an empty string if uuid is nil

identity/pool.go
```diff
@@ -54,25 +53,35 @@ type (
...
+		schema.Persister
+	}
 )
 
-func TestPool(p Pool) func(t *testing.T) {
+func TestPool(p testPool) func(t *testing.T) {
 	return func(t *testing.T) {
-		viper.Set(configuration.ViperKeyDefaultIdentityTraitsSchemaURL, "file://./stub/identity.schema.json")
+		viper.Set(configuration.ViperKeyURLsSelfPublic, "http://mock-server.com")
+		defaultSchema, _ := p.RegisterDefaultSchema("file://./stub/identity.schema.json")
```
aeneasr: error

identity/pool.go
```diff
@@ -54,25 +53,35 @@ type (
...
+		secondSchema := schema.Schema{
+			URL: "file://./stub/identity-2.schema.json",
+		}
+		_ = p.RegisterSchema(&secondSchema)
```
aeneasr: error

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,59 @@
...
+	if s.URL == "" {
+		return errors.WithStack(herodot.ErrBadRequest.WithReason("The schema is missing the URL property."))
+	}
+
+	return sqlcon.HandleError(p.c.Transaction(func(tx *pop.Connection) error {
+		return tx.Create(s)
+	}))
+}
+
+func (p *Persister) RegisterDefaultSchema(url string) (*schema.Schema, error) {
```
aeneasr: Change this to:

```suggestion
func (p *Persister) RegisterDefaultSchema() (*schema.Schema, error) {
```

and get the default schema from `p.c.DefaultTraitsSchemaURL()`

identity/pool.go
```diff
@@ -123,13 +132,13 @@ func TestPool(p Pool) func(t *testing.T) {
...
 		t.Run("case=create and keep set values", func(t *testing.T) {
-			expected := passwordIdentity("file://./stub/identity-2.schema.json", "id-2")
+			expected := passwordIdentity(secondSchema.ID, "id-2")
 			require.NoError(t, p.CreateIdentity(context.Background(), expected))
 			createdIDs = append(createdIDs, expected.ID)
 
 			actual, err := p.GetIdentity(context.Background(), expected.ID)
 			require.NoError(t, err)
-			assert.Equal(t, "file://./stub/identity-2.schema.json", actual.TraitsSchemaURL)
+			assert.Regexp(t, "^http://mock-server.com/schemas/*", actual.TraitsSchemaURL)
```
aeneasr: Probably check for the `TraitsSchemaID` as well.

And test the concrete value.

identity/pool.go
```diff
@@ -99,15 +108,15 @@ func TestPool(p Pool) func(t *testing.T) {
...
+			expected := passwordIdentity(uuid.Nil, "id-1")
 			require.NoError(t, p.CreateIdentity(context.Background(), expected))
 			createdIDs = append(createdIDs, expected.ID)
 
 			actual, err := p.GetIdentity(context.Background(), expected.ID)
 			require.NoError(t, err)
 
 			assert.Equal(t, expected.ID, actual.ID)
-			assert.Equal(t, "file://./stub/identity.schema.json", actual.TraitsSchemaURL)
+			assert.Regexp(t, "^http://mock-server.com/schemas/*", actual.TraitsSchemaURL)
```
aeneasr: And test the concrete value.

identity/pool.go
```diff
@@ -203,11 +212,11 @@ func TestPool(p Pool) func(t *testing.T) {
 		})
 
 		t.Run("case=fail to update an identity because credentials changed but update was called", func(t *testing.T) {
-			initial := oidcIdentity("", x.NewUUID().String())
+			initial := oidcIdentity(defaultSchema.ID, x.NewUUID().String())
 			require.NoError(t, p.CreateIdentity(context.Background(), initial))
 			createdIDs = append(createdIDs, initial.ID)
 
-			assert.Equal(t, "file://./stub/identity.schema.json", initial.TraitsSchemaURL)
+			assert.Regexp(t, "^http://mock-server.com/schemas/*", initial.TraitsSchemaURL)
```
aeneasr: And test the concrete value.

identity/pool.go
```diff
@@ -224,17 +233,17 @@ func TestPool(p Pool) func(t *testing.T) {
 
 			actual, err := p.GetIdentityConfidential(context.Background(), toUpdate.ID)
 			require.NoError(t, err)
-			assert.Equal(t, "file://./stub/identity.schema.json", actual.TraitsSchemaURL)
+			assert.Regexp(t, "^http://mock-server.com/schemas/*", actual.TraitsSchemaURL)
```
aeneasr: And test the concrete value.

identity/pool.go
```diff
@@ -224,17 +233,17 @@ func TestPool(p Pool) func(t *testing.T) {
...
 		})
 
 		t.Run("case=update an identity and set credentials", func(t *testing.T) {
-			initial := oidcIdentity("", x.NewUUID().String())
+			initial := oidcIdentity(defaultSchema.ID, x.NewUUID().String())
 			require.NoError(t, p.CreateIdentity(context.Background(), initial))
 			createdIDs = append(createdIDs, initial.ID)
 
-			assert.Equal(t, "file://./stub/identity.schema.json", initial.TraitsSchemaURL)
+			assert.Regexp(t, "^http://mock-server.com/schemas/*", initial.TraitsSchemaURL)
```
aeneasr: And test the concrete value.

identity/pool.go
```diff
@@ -243,20 +252,20 @@ func TestPool(p Pool) func(t *testing.T) {
...
 			})
 			expected.Traits = Traits(`{"update":"me"}`)
-			expected.TraitsSchemaURL = "file://./stub/identity-2.schema.json"
+			expected.TraitsSchemaID = secondSchema.ID
 			require.NoError(t, p.UpdateIdentityConfidential(context.Background(), expected))
 
 			actual, err := p.GetIdentityConfidential(context.Background(), expected.ID)
 			require.NoError(t, err)
-			assert.Equal(t, "file://./stub/identity-2.schema.json", actual.TraitsSchemaURL)
+			assert.Regexp(t, "^http://mock-server.com/schemas/*", actual.TraitsSchemaURL)
```
aeneasr: And test the concrete value.

identity/validator.go
```diff
@@ -35,9 +41,16 @@ func (v *Validator) Validate(i *Identity) error {
 		NewValidationExtensionIdentifier().WithIdentity(i),
 	}
 
-	err := v.v.Validate(
+	v.d.SchemaPersister()
```
aeneasr: cleanup

identity/validator.go
```diff
@@ -35,9 +41,16 @@ func (v *Validator) Validate(i *Identity) error {
...
+	s, err := v.d.SchemaPersister().GetSchema(i.TraitsSchemaID)
+	if err != nil {
+		return err
+	}
+
+	err = v.v.Validate(
 		stringsx.Coalesce(
-			i.TraitsSchemaURL,
+			s.URL,
 			v.c.DefaultIdentityTraitsSchemaURL().String(),
```
aeneasr: There is no need for this any more because TraitsSchemaID is always set. If it is not set, use to the default _from the database_.

identity/validator_test.go
```diff
@@ -50,9 +48,9 @@ func TestSchemaValidator(t *testing.T) {
 	ts := httptest.NewServer(router)
 	defer ts.Close()
 
-	conf, _ := internal.NewRegistryDefault(t)
-	viper.Set(configuration.ViperKeyDefaultIdentityTraitsSchemaURL, ts.URL+"/schema/firstName")
-	v := NewValidator(conf)
+	conf, reg := internal.NewRegistryDefault(t)
+	_, _ = reg.SchemaPersister().RegisterDefaultSchema(ts.URL + "/schema/firstName")
```
aeneasr: error

identity/validator_test.go
```diff
@@ -104,6 +102,18 @@ func TestSchemaValidator(t *testing.T) {
 		},
 	} {
 		t.Run(fmt.Sprintf("case=%d", k), func(t *testing.T) {
+			if tc.i.TraitsSchemaURL != "" {
+				s, err := reg.SchemaPersister().GetSchemaByUrl(tc.i.TraitsSchemaURL)
+				if err != nil {
+					s = &schema.Schema{
+						URL: tc.i.TraitsSchemaURL,
+					}
+					_ = reg.SchemaPersister().RegisterSchema(s)
```
aeneasr: error

aeneasr: I would add a function called "Create if not exists".

persistence/sql/persister_identity.go
```diff
@@ -116,8 +116,13 @@ func createIdentityCredentials(ctx context.Context, tx *pop.Connection, i *ident
 }
 
 func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) error {
-	if i.TraitsSchemaURL == "" {
-		i.TraitsSchemaURL = p.cf.DefaultIdentityTraitsSchemaURL().String()
+	if uuid.Nil == i.TraitsSchemaID {
+		s, err := p.GetSchemaByUrl(p.cf.DefaultIdentityTraitsSchemaURL().String())
```
aeneasr:
```suggestion
		s, err := p.GetSchemaByURL(p.cf.DefaultIdentityTraitsSchemaURL().String())
```

aeneasr: I would do `GetDefaultSchema() (Schema, error)`

persistence/sql/persister_identity.go
```diff
@@ -139,10 +146,26 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
 
 func (p *Persister) ListIdentities(ctx context.Context, limit, offset int) ([]identity.Identity, error) {
 	is := make([]identity.Identity, 0)
-	return is, sqlcon.HandleError(p.c.RawQuery("SELECT * FROM identities LIMIT ? OFFSET ?", limit, offset).All(&is))
+	err := sqlcon.HandleError(p.c.RawQuery("SELECT * FROM identities LIMIT ? OFFSET ?", limit, offset).All(&is))
```
aeneasr:
```suggestion
	if err := sqlcon.HandleError(p.c.RawQuery("SELECT * FROM identities LIMIT ? OFFSET ?", limit, offset).All(&is)); err != nil {
	  return nil, err
	}
```

persistence/sql/persister_identity.go
```diff
@@ -139,10 +146,26 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
...
+
+	for i := range is {
+		is[i].InjectTraitsUrl(p.cf.SelfPublicURL().String())
+	}
+
+	return is, err
```
aeneasr:
```suggestion
	return is, nil
```

persistence/sql/persister_identity.go
```diff
@@ -161,6 +184,16 @@ func (p *Persister) UpdateIdentityConfidential(ctx context.Context, i *identity.
 }
 
 func (p *Persister) UpdateIdentity(ctx context.Context, i *identity.Identity) error {
+	// Inject old TraitsSchemaID if none is provided
+	if uuid.Nil == i.TraitsSchemaID {
```
aeneasr: Move this to its own function:

```
if err := p.setTraitsSchemaID(i); err != nil ...}
```

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	if err := p.c.Find(&s, id); err != nil {
+		return nil, sqlcon.HandleError(err)
+	}
+	return &s, nil
+}
+
+func (p *Persister) GetDefaultSchema() (*schema.Schema, error) {
+	if !viper.IsSet(configuration.ViperKeyDefaultIdentityTraitsSchemaURL) {
```
aeneasr: Go via the config, not viper itself.

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil, errors.WithStack(herodot.ErrNotFound.WithReason("The default identity traits schema is not set."))
+	}
+
+	return p.GetSchemaByUrl(viper.GetString(configuration.ViperKeyDefaultIdentityTraitsSchemaURL))
```
aeneasr: Via config

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if err := p.c.Find(&s, id); err != nil {
+		return nil, sqlcon.HandleError(err)
+	}
+	return &s, nil
+}
+
+func (p *Persister) GetDefaultSchema() (*schema.Schema, error) {
+	if !viper.IsSet(configuration.ViperKeyDefaultIdentityTraitsSchemaURL) {
+		return nil, errors.WithStack(herodot.ErrNotFound.WithReason("The default identity traits schema is not set."))
```
aeneasr: This is an internal server error

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	return p.GetSchemaByUrl(viper.GetString(configuration.ViperKeyDefaultIdentityTraitsSchemaURL))
+}
+
+func (p *Persister) GetSchemaByUrl(url string) (*schema.Schema, error) {
```
aeneasr: Capitalize URL

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,59 @@
...
+	var s schema.Schema
+	if err := p.c.Where("url = ?", url).First(&s); err != nil {
+		return nil, err
```
aeneasr: sqlcon.HandleErr

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,59 @@
...
+	return &s, nil
+}
+
+func (p *Persister) RegisterSchema(s *schema.Schema) error {
+	if s.URL == "" {
+		return errors.WithStack(herodot.ErrBadRequest.WithReason("The schema is missing the URL property."))
+	}
+
+	return sqlcon.HandleError(p.c.Transaction(func(tx *pop.Connection) error {
+		return tx.Create(s)
```
aeneasr: No transaction needed

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,59 @@
...
+	}))
+}
+
+func (p *Persister) RegisterDefaultSchema(url string) (*schema.Schema, error) {
+	ds := schema.Schema{
+		URL: url,
+	}
+	viper.Set(configuration.ViperKeyDefaultIdentityTraitsSchemaURL, ds.URL)
```
aeneasr: No viper.Set!

schema/handler.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/ory/x/urlx"
+	"github.com/pkg/errors"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+)
+
+type (
+	Configuration interface {
```
aeneasr: Rename this to: `handlerConfiguration`

persistence/sql/persister_schema.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (p *Persister) GetSchemaByUrl(url string) (*schema.Schema, error) {
+	var s schema.Schema
+	if err := p.c.Where("url = ?", url).First(&s); err != nil {
+		return nil, err
+	}
+	return &s, nil
+}
+
+func (p *Persister) RegisterSchema(s *schema.Schema) error {
```
aeneasr: `CreateSchema`

schema/handler.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if err := h.r.SchemaPersister().RegisterSchema(&s); err != nil {
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
+
+	h.r.Writer().WriteCreated(w, r,
+		urlx.AppendPaths(
+			h.c.SelfPublicURL(),
+			"schemas",
```
aeneasr: Use a const here

schema/handler.go
```diff
@@ -0,0 +1,112 @@
...
+	r handlerDependencies,
+) *Handler {
+	return &Handler{
+		c: c,
+		r: r,
+	}
+}
+
+func (h *Handler) RegisterPublicRoutes(public *x.RouterPublic) {
+	public.GET("/schemas/:id", h.get)
```
aeneasr: You can use the const to construct this here too

schema/handler.go
```diff
@@ -0,0 +1,112 @@
...
+				return
+			}
+		}
+
+		if info.IsDir() {
+			h.r.Writer().WriteError(w, r, herodot.ErrNotFound)
+			return
+		}
+
+		http.ServeFile(w, r, fp)
```
aeneasr:
```
	f, err := os.Open("")
	if err != nil {
		// writer - WithReason("blabla").WithDebug("")
	}
	defer f.Close()
	
	var w http.ResponseWriter
	if _, err := io.Copy(w, f); err != nil {
		// ...
	}
```

schema/handler.go
```diff
@@ -0,0 +1,112 @@
...
+		r: r,
+	}
+}
+
+func (h *Handler) RegisterPublicRoutes(public *x.RouterPublic) {
+	public.GET("/schemas/:id", h.get)
+}
+
+func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
+	admin.PUT("/schemas", h.registerSchema)
```
aeneasr: Remove this - this should only be done from the config file.

PR #167, opened 2020-01-15
Improve dev flag

### LINKED ISSUE #162: Improve `--dev` flag

`--dev` is currently only used for csrf but we need to use it in other areas as well.

We should also check if self urls are https and if not disallow the startup!

zepatrik:
## Related issue

closes #162

## Proposed changes

Adds a `InsecureDevMode() bool` function to the config provider. 

## Checklist


driver/configuration/provider.go
```diff
@@ -78,4 +78,6 @@ type Provider interface {
 	TracingServiceName() string
 	TracingProvider() string
 	TracingJaegerConfig() *tracing.JaegerConfig
+
+	InsecureDevMode() bool
```
aeneasr: Maybe rename this to `IsInsecureDevMode()`?

driver/driver_default.go
```diff
@@ -14,12 +14,22 @@ type DefaultDriver struct {
...
 }
 
-func NewDefaultDriver(l logrus.FieldLogger, version, build, date string) (Driver, error) {
+func NewDefaultDriver(l logrus.FieldLogger, version, build, date string, dev bool) (Driver, error) {
 	if l == nil {
 		l = logrusx.New()
 	}
 
-	c := configuration.NewViperProvider(l)
+	if dev {
```
aeneasr: Let's move this warning to cmd/daemon instead - otherwise we'll get a lot of warnings during testing.

zepatrik: The logger is not initialized in the deamon yet but that should be possible

aeneasr: You can call it in "ServeAll", then it will be initialized

zepatrik:
done @aeneasr

aeneasr:
One last place where this needs to be updated: https://github.com/ory/kratos/blob/master/cmd/daemon/serve.go#L109

Issue #244, opened 2020-02-11
Investigate MySQL empty timestamp issue on session


aeneasr:
**Describe the bug**

https://community.ory.sh/t/kratos-using-mysql-fails-to-create-a-session-when-sigining-in/1508/2

philipjkim:
I'm trying to run kratos on localhost, using remote MySQL rather than SQLite.

What I've done are:

- Schema migration on the remote MySQL DB
- Run kratos using `docker-compose` with customized `quickstart.yml`

I used `latest` tag of `oryd/kratos` docker image instead of `latest-sqlite` since I'm not using SQLite. Here are my `quickstart.yml`:

```yaml
version: "3.7"

services:
  kratos-selfservice-ui-node:
    image: oryd/kratos-selfservice-ui-node:latest
    ports:
      - "4435:4435"
    environment:
      - PORT=4435
      - KRATOS_PUBLIC_URL=http://kratos:4433/
      - KRATOS_ADMIN_URL=http://kratos:4434/
      - KRATOS_BROWSER_URL=http://127.0.0.1:4455/.ory/kratos/public
      - JWKS_URL=http://oathkeeper:4456/.well-known/jwks.json
    networks:
      - intranet
    restart: on-failure

  oathkeeper:
    image: oryd/oathkeeper:v0.35.5-beta.1
    depends_on:
      - kratos
    ports:
      - "4455:4455"
      - "4456:4456"
    command: serve proxy -c "/etc/config/oathkeeper/.oathkeeper.yml"
    environment:
      - LOG_LEVEL=debug
    restart: on-failure
    networks:
      - intranet
    volumes:
      - type: bind
        source: ./contrib/quickstart/oathkeeper
        target: /etc/config/oathkeeper

  kratos:
    image: oryd/kratos:latest
    ports:
      - "4433:4433" # public
      - "4434:4434" # admin
    restart: unless-stopped
    environment:
      - DSN=mysql://$MYSQL_USERNAME:$MYSQL_PASSWORD@tcp($MYSQL_ADDRESS)/$MYSQL_DATABASE?parseTime=true&multiStatements=true
    command: serve -c /etc/config/kratos/.kratos.yml --dev
    volumes:
      - type: bind
        source: ./contrib/quickstart/kratos/email-password
        target: /etc/config/kratos
    networks:
      - intranet

  mailhog:
    image: mailhog/mailhog:v1.0.0
    ports:
      - "8025:4436"
    networks:
      - intranet

networks:
[...]

PR #246, opened 2020-02-13
fix: Set AuthenticatedAt in session issuer hook

aeneasr:
Closes #224


aeneasr:
Whoops, this was supposed to reference #244

PR #254, opened 2020-02-17
refactor(persistence/sql): move connection to context to enable transactions

zepatrik:
## Checklist

### DIFF
--- a/persistence/sql/persister_test.go
+++ b/persistence/sql/persister_test.go
@@ -155,6 +155,9 @@ func TestPersister(t *testing.T) {
 }
 
 func getErr(args ...interface{}) error {
+	if len(args) == 0 {
+		return nil
+	}
 	lastArg := args[len(args)-1]
 	if e, ok := lastArg.(error); ok {
 		return e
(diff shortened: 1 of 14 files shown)


persistence/sql/persister_test.go
```diff
@@ -146,3 +153,53 @@ func TestPersister(t *testing.T) {
 		t.Logf("DSN: %s", dsn)
 	}
 }
+
+func getErr(args ...interface{}) error {
```
aeneasr: Wow this is actually really useful. We should probably add that to ory/x!

persistence/sql/persister_test.go
```diff
@@ -146,3 +153,53 @@ func TestPersister(t *testing.T) {
...
+	lastArg := args[len(args)-1]
```
aeneasr: panics if `len(args) == 0`

aeneasr:
```suggestion
	if len(args) == 0 {
		return nil
	}
	lastArg := args[len(args)-1]
```

PR #256, opened 2020-02-18
fix: Set SeenAt on ErrorContainer init

philipjkim:
## Related issue

#244

## Proposed changes

Added `SeenAt` field when initializing `ErrorContainer` (MySQL complains when a `DATETIME` column is empty)

## Checklist

## Further comments

Feel free to dispose this PR if it's not the best solution for given issue.

### DIFF
--- a/persistence/sql/persister_errorx.go
+++ b/persistence/sql/persister_errorx.go
@@ -30,6 +30,7 @@ func (p *Persister) Add(ctx context.Context, csrfToken string, errs ...error) (u
 	c := &errorx.ErrorContainer{
 		CSRFToken: csrfToken,
 		Errors:    buf.Bytes(),
+		SeenAt:    time.Now().UTC(),
 		WasSeen:   false,
 	}
 


persistence/sql/persister_errorx.go
```diff
@@ -30,6 +30,7 @@ func (p *Persister) Add(ctx context.Context, csrfToken string, errs ...error) (u
 	c := &errorx.ErrorContainer{
 		CSRFToken: csrfToken,
 		Errors:    buf.Bytes(),
+		SeenAt:    time.Now().UTC(),
```
aeneasr: SeenAt should be set when the error has been read. I think we might have to use a `*time.Time` pointer to fix this.

philipjkim: @aeneasr OK, then please close this PR and handle the issue by yourself.

aeneasr: Ok, I will do that - thank you for your contribution :) Fixing this was actually quite a lot of work as I had to fix several issues in an upstream project for migrations to properly work as `seen_at` was actually NOT NULL (though it should be NULLable) 😅

aeneasr:
I've tackled this as https://github.com/ory/kratos/pull/259

PR #284, opened 2020-03-13
feat: add `dsn: memory` shorthand

### LINKED ISSUE #228: Have `dsn: memory` as an alias for SQLite in memory DSN

SQLite supports memory with `sqlite://foo.db?mode=memory&_fk=true`. Using `dsn: memory` should be an alias for that.

zepatrik:
## Related issue
closes #228 

This will also run the migrations on every startup


zepatrik:
We probably need an integration test for this, can you give me hints how I should do that @aeneasr ?
Or should we do that with cypress at some later point

driver/registry.go
```diff
@@ -107,15 +109,23 @@ type selfServiceStrategy interface {
...
 
 	registry, ok := driver.(Registry)
 	if !ok {
 		return nil, errors.Errorf("driver of type %T does not implement interface Registry", driver)
 	}
 
+	// if dsn is memory we have to run the migrations on every start
+	if c.DSN() == "memory" {
+		registry.Logger().Print("Kratos is running migrations on every startup as DSN is memory.\n")
+		registry.Logger().Print("This means your data are lost when Kratos terminates.\n")
```
aeneasr:
```suggestion
		registry.Logger().Print("This means your data is lost when Kratos terminates.\n")
```

driver/configuration/provider_viper.go
```diff
@@ -153,6 +153,15 @@ func (p *ViperProvider) DSN() string {
 	return ""
 }
 
+func (p *ViperProvider) DSNAddress() string {
+	dsn := p.DSN()
+	if dsn == "memory" {
+		return "sqlite://:memory:?_fk=true"
```
aeneasr:
```suggestion
		return "sqlite://mem.db?mode=memory?_fk=true&cache=shared"
```

https://github.com/mattn/go-sqlite3/issues/204, https://github.com/mattn/go-sqlite3/issues/511, https://www.sqlite.org/sharedcache.html#shared_cache_and_in_memory_databases

driver/registry.go
```diff
@@ -107,15 +109,23 @@ type selfServiceStrategy interface {
...
+		return nil, errors.WithStack(err)
 	}
 
 	registry, ok := driver.(Registry)
 	if !ok {
 		return nil, errors.Errorf("driver of type %T does not implement interface Registry", driver)
 	}
 
+	// if dsn is memory we have to run the migrations on every start
+	if c.DSN() == "memory" {
```
aeneasr: Let's do this for all DSN that have prefix `sqlite://` and a URL query with `mode=memory`, that way this also works with custom SQLite DSN strings that do not use `memory` (could be useful for testing other SQLite modifiers for example).

Also removes the need for another function `DSNAddress`

driver/registry.go
```diff
@@ -107,15 +110,28 @@ type selfServiceStrategy interface {
 }
 
 func NewRegistry(c configuration.Provider) (Registry, error) {
-	driver, err := dbal.GetDriverFor(c.DSN())
+	dsn, err := url.Parse(c.DSN())
```
aeneasr: This will fail for mysql connection strings. Only parse the URL if it has prefix `sqlite://`. It's also possible that this fails for `sqlite://:memory:` which is I think a valid connection string but not a valid URL. Maybe just split at `?` and parse the query only.

driver/configuration/provider_viper.go
```diff
@@ -145,7 +145,13 @@ func (p *ViperProvider) PublicListenOn() string {
 }
 
 func (p *ViperProvider) DSN() string {
-	if dsn := viperx.GetString(p.l, ViperKeyDSN, ""); len(dsn) > 0 {
+	dsn := viperx.GetString(p.l, ViperKeyDSN, "")
+
+	if dsn == "memory" {
```
aeneasr: Please add a test for this

zepatrik:
ready for review @aeneasr

aeneasr:
good job!

PR #299, opened 2020-03-18
Document the schema API and serve it in the admin API

### LINKED ISSUE #287: Serve the schemas in the common API and have it documented
It can be useful to apps to be able to fetch the schema associated with an identity. Right now, the identity admin API has a `traits_schema_id` and a `traits_schema_url`. The issue is that the URL is the public one and the schemas API is not served by the admin API. 
In my case, my app only has access to the admin API, which means that I have no simple way to fetch the traits schema associated with an identity.

Also the schemas endpoints are not included in the swagger API doc, and I think they could be.

sandhose:
## Related issue

Fixes #287

## Proposed changes

The API to get a traits schema was only exposed on the public endpoint. In some cases it can make sense to have it available on the admin endpoint. This PR does that, as well as adding the endpoint to the swagger documentation.
I also noticed this endpoint responded with a `text/plain` content-type, so I changed the header to `application/json`.

## Checklist

## Further comments

I first tried to find ways to reference the jsonschema schema as the response type but did not find a good way to do it. With the empty interface as response type, the go-swagger api client properly parses the response to a `interface{}`, which then can be cast to a map.

I also regenerated the SDK


schema/handler.go
```diff
@@ -66,6 +102,7 @@ func (h *Handler) get(w http.ResponseWriter, r *http.Request, ps httprouter.Para
 		src = resp.Body
 	}
 
+	w.Header().Add("Content-Type", "application/json")
```
aeneasr: /cc @zepatrik just a quick reminder that we need to set the content type explicitly when not using herodot :)

aeneasr:
Awesome, thank you for your contribution!

PR #303, opened 2020-03-20
feat: allow configuring same-site for session cookies

### LINKED ISSUE #257: Allow configuration of same-site cookie
**Describe the bug**

It's currently not possible to set SameSite for session cookies. This should be possible.

zepatrik:
## Related issue
closes #257 

## Proposed changes

It is now possible to set SameSite for the session cookie via the key `security.session.cookie.same_site`.


driver/configuration/provider_viper.go
```diff
@@ -370,3 +373,15 @@ func (p *ViperProvider) SelfServiceVerificationReturnTo() *url.URL {
 func (p *ViperProvider) SelfServicePrivilegedSessionMaxAge() time.Duration {
 	return viperx.GetDuration(p.l, ViperKeySelfServicePrivilegedAuthenticationAfter, time.Hour)
 }
+
+func (p *ViperProvider) SessionSameSiteMode() http.SameSite {
+	switch viperx.GetString(p.l, ViperKeySessionSameSite, "Lax") {
```
aeneasr: Let's make this case-insensitive!

zepatrik: I thought we might respect the standard https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#SameSite
but is not really necessary

aeneasr: Yeah I guess you're right, I mean it is being validated in the JSON Schema so it would be easily detectable. I think both ways are ok. What do you think?

zepatrik: Let's just leave it this way, validation will anyway only allow the capitalized version.

PR #305, opened 2020-03-21
chore: update ory/x dependency and add test case

zepatrik:


cmd/validate_viper_test.go
```diff
@@ -0,0 +1,44 @@
...
+	require.NoError(t, os.Setenv("DSN", "memory"))
+	require.NoError(t, os.Setenv("IDENTITY_TRAITS_DEFAULT_SCHEMA_URL", "file://./stub.schema.json"))
+	require.NoError(t, os.Setenv("SELFSERVICE_LOGOUT_REDIRECT_TO", "https://example.com"))
+	require.NoError(t, os.Setenv("COURIER_SMTP_CONNECTION_URI", "https://example.com"))
+	require.NoError(t, os.Setenv("URLS_PROFILE_UI", "https://example.com"))
+	require.NoError(t, os.Setenv("URLS_MFA_UI", "https://example.com"))
+	require.NoError(t, os.Setenv("URLS_LOGIN_UI", "https://example.com"))
+	require.NoError(t, os.Setenv("URLS_REGISTRATION_UI", "https://example.com"))
+	require.NoError(t, os.Setenv("URLS_ERROR_UI", "https://example.com"))
+	require.NoError(t, os.Setenv("URLS_VERIFY_UI", "https://example.com"))
```
aeneasr: These other values are set but not validated/tested - any particular reason to keep them around?

zepatrik: Validation fails as exactly these values are required. Validation errors are not in a deterministic order and because of that I decided it is the most reliable way to test the validation with some default input that passes validation so only the errors relevant for the test case will be returned.

aeneasr: Ah I see!

PR #307, opened 2020-03-23
chore: moved watchAndValidateViper to viperx

zepatrik:


cmd/serve.go
```diff
@@ -50,7 +47,11 @@ DON'T DO THIS IN PRODUCTION!
 `)
 		}
 
-		watchAndValidateViper()
+		schema, err := schemas.Find("config.schema.json")
```
aeneasr: Please move this back to `watchAndValidateViper` to reduce duplication in cases where we need to load the schema again

aeneasr:
🎉

PR #433, opened 2020-05-26
Added short guide for connecting kratos to AzureAD

ernax78:
## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [x] I have added or changed [the documentation](docs/docs).

PR #434, opened 2020-05-26
feat: Add support for Multitenant Azure AD as an OIDC provider

jakhog:
## Proposed changes

This PR adds a new OIDC provider called `microsoft`, that adds support for the multi-tenant features of the Microsoft identity platform. The platform can already be used as an OIDC provider when authenticating towards a specific Tenant - as it behaves as a generic OIDC provider - but when using it to authenticate e.g. with personal Microsoft accounts the endpoints behave in a somewhat special way, described in [this blog post](https://www.cloudidentity.com/blog/2014/08/26/the-common-endpoint-walks-like-a-tenant-talks-like-a-tenant-but-is-not-a-tenant/).

The proposed code differs from the generic OIDC provider by getting the issuer from the returned JWT `id_token`, but subsequently verifies the returned token in the exact same way. It also adds another property in the OIDC provider configuration called `Tenant` that is to be used with the `microsoft` provider.

## Checklist

      the changes.
      works.

## Further comments

I have not added any tests, as I couldn't get the current test suite to run with `go test ./...`. I get a lot of errors like these
```
...
internal/testhelpers/selfservice_settings.go:140:3: t.Cleanup undefined (type *testing.T has no field or method Cleanup)
internal/testhelpers/selfservice_settings.go:140:3: too many errors
...
time="2020-05-26T21:08:10+02:00" level=warning msg="Unable to connect to database, retrying." error="unsupported dialect 'sqlite3'"
time="2020-05-26T21:08:11+02:00" level=warning msg="Unable to connect to database, retrying." error="unsupported dialect 'sqlite3'"
...
```
So any help with that would be greatly appreciated, and allow me to write a few tests for the new provider.

The `Tenant` configuration property was added directly to the OIDC provider configuration. It would perhaps be nicer to have per-provider configuration schemas, but that didn't seem like a reasonable thing to do just now.

Also, I have not written any new documentation about this feature, but just noticed PR #433, and assume that it would make sense to build upon that if that is merged.

### DIFF
--- a/docs/docs/guides/sign-in-with-github-google-facebook-linkedin.mdx
+++ b/docs/docs/guides/sign-in-with-github-google-facebook-linkedin.mdx
@@ -132,9 +132,15 @@ local claims = std.extVar('claims');
 {
   identity: {
     traits: {
-      [if "email" in claims then "email" else null]: claims.email,
+      // Allowing unverified email addresses enables account
+      // enumeration attacks, especially if the value is used for
+      // e.g. verification or as a password login identifier.
+      //
+      // If connecting only to your organization (one tenant), claims.email is safe to use if you have not actively disabled e-mail verification during signup.
+      //
       // The email might be empty if the account is not linked to an email address.
       // For a human readable identifier, consider using the "preferred_username" claim.
+      [if "email" in claims then "email" else null]: claims.email,
     },
   },
 }
--- a/docs/docs/guides/sign-in-with-github-google-facebook-linkedin.mdx
+++ b/docs/docs/guides/sign-in-with-github-google-facebook-linkedin.mdx
@@ -105,46 +105,42 @@ This will enable you to log in using any Azure AD directory - Multitenant and
 personal Microsoft accounts (e.g. Skype, Xbox) depending on the settings made
 when creating the application in Azure AD.
 
-To set up "Sign in with Microsoft" you must first register an application
-[Register an application with the Microsoft identity platform](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app).
 
-It might be nessecary to enable the implicit grant flow for access tokens and ID
-tokens.
+#### Creating an Application in Azure AD
+To set up "Sign in with Microsoft" you must first [register an application with
+the Microsoft identity platform](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app).
 
-Set the "Authorization callback URL" to:
-
-```
-https://public.server.fqdn.example.org/.ory/kratos/public/self-service/browser/flows/strategies/oidc/callback/azuread
-```
-
-The pattern of this URL is:
+Select "Web" as the "Redirect URI" type, and set the URI to:
 
 ```
 http(s)://<domain-of-ory-kratos>:<public-port>/self-service/browser/flows/strategies/oidc/callback/<provider-id>
 ```
 
-```json title="contrib/quickstart/kratos/email-password/oidc.azuread.jsonnet"
-# claims contains all the data sent by the upstream.
+After the "App Registration" is created, make note of the `Application ID` and
+`Directory ID` on top of the Overview page. To create the client secret,
+navigate to "Certificates & secrets" and click "+ New client secret". Remember
+to copy the secret value as it will only be shown once.
+
+#### Configuring Kratos
+Create a Jsonnet claims mapper as described in
+[OpenID Connect and OAuth2 Credentials](../concepts/credentials/openid-connect-oidc-oauth2.mdx).
+Save the code in
+`<kratos-directory>/contrib/quickstart/kratos/email-password/oidc.microsoft.jsonnet`.
+
+```json title="contrib/quickstart/kratos/email-password/oidc.microsoft.jsonnet"
 local claims = std.extVar('claims');
 {
   identity: {
     traits: {
-      // Allowing unverified email addresses enables account
-      // enumeration attacks, especially if the value is used for
-      // e.g. verification or as a password login identifier.
-      //
-      // If connecting only to your organization (one tenant), claims.email is safe to use if you have not actively disabled e-mail
-      //verification during signup.
-      email: claims.email
+      [if "email" in claims then "email" else null]: claims.email,
+      // The email might be empty if the account is not linked to an email address.
+      // For a human readable identifier, consider using the "preferred_username" claim.
     },
   },
 }
 ```
 
-Also see ["Disable email verification during customer sign-up in Azure Active
-Directory B2C"] for warnings.
-
-Enable the Microsoft Azure provider in the ORY Kratos config located at
+Enable the Microsoft provider in the ORY Kratos config located at
 `<kratos-directory>/contrib/quickstart/kratos/email-password/.kratos.yml`.
 
 ```yaml title="contrib/quickstart/kratos/email-password/.kratos.yml"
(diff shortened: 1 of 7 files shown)


selfservice/strategy/oidc/provider_microsoft.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, errors.WithStack(ErrIDTokenMissing)
+	}
+
+	parser := new(jwt.Parser)
+	unverifiedClaims := microsoftUnverifiedClaims{}
+	if _, _, err := parser.ParseUnverified(raw, &unverifiedClaims); err != nil {
+		return nil, err
+	}
+
+	p, err := gooidc.NewProvider(context.Background(), unverifiedClaims.Issuer)
```
aeneasr, line 70: I believe this opens a very serious and real attack surface. Using the issuer from the JWT to verify the JWT means that an attacker can easily forge a JWT with his own issuer URL which has all the information (link cryptographic keys). This can be used to impersonate users.

The IssuerURL **must** come from the configuration.

jakhog:
Thanks for the tips regarding tests, I'll give it a go and see how that works.

Regarding end-to-end tests, that would be very cool. Were you thinking of a full sign in process, or just checking that the `/auth` endpoint works? For a full sign in, I assume we would need some kind of browser testing thing that completes the user flow, and I don't really have any experience with that...

selfservice/strategy/oidc/provider_microsoft.go
jakhog: Hmm, that is true. If an attacker spoofs the `https://login.microsoftonline.com/.../token` endpoint to return a token that contains an issuer pointing to a completely unrelated OIDC provider, that would be verified as valid.

What if we verify that the issuer is on the form `https://login.microsoftonline.com/{tenant-id}/v2.0`? The `tenant-id` is in the claims, so we can use that.

That would ofcourse still potentially be an attack vector, but as far as I can see, the same vulnerability (spoofing an issuer) would be present on any OIDC provider.

jakhog: We don't really need to use the issuer url from the claims at all, the only thing that is needed for the multi-tenant flows to work is to get the actual tenant id from the JWT. So that would be another (pretty much the same) solution.

aeneasr: If we don't need to use the claims issuer, then let's just use the "hardcoded" issuer! I think that's the safest. Extracting the tenant-id from the claims is totally valid, as long as the validation chain is kept in tact (which is ensured by hardcoding/hard-configuring the issuer URL)

aeneasr:
Sorry for my late reply, last few days was very busy.

Yeah good question with the tests. Honestly, this could probably be a HTTP replay? Alternatively yeah, this would work with our e2e test suite. We've done this for Hydra already. To add Microsoft there, you would:

- Include the microsoft config + microsoft jsonnet here: https://github.com/ory/kratos/tree/master/test/e2e/profiles/oidc
- Maybe use something like `envsubst` to inject e.g. `TEST_OIDC_MICROSOFT_CLIENT_ID` into the config file (we already modify the config file [here](https://github.com/ory/kratos/blob/master/test/e2e/run.sh#L98) so adding an `envsubst` line there would be very easy) or figure out how to set this using an env var (might be tricky)
- finally more or less copy this [e2e test](https://github.com/ory/kratos/blob/master/test/e2e/cypress/integration/profiles/oidc/registration/success.spec.js#L20-L66) and use I think `Cypress.env('TEST_OIDC_MICROSOFT_EMAIL')` and `Cypress.env('TEST_OIDC_MICROSOFT_PASSWORD')` for the username/password form

selfservice/strategy/oidc/provider_microsoft.go
jakhog: Sorry, forgot about this for a little while, and I my wording wasn't the best in my previous reply. What I meant when I said "we don't really need to use the issuer url from the claims", is that in stead of accepting the issuer url blindly, we can generate it using a template like `https://login.microsoftonline.com/{tenant-id}/v2.0` where `tenant-id` is present in the claims.

The tenant id will always be a guid, so by doing a bit of validation on the string from the claims. Does that sound like a viable option?

As far as I can tell, this is the only way we can get it to work with the multi-tenant flows for Microsoft. The initial auth and token endpoints are known in advance, but the issuer (with the keys to verify the JWT) is not known until the user has actually signed in with an account. So, in a way, it's not really an implementation of OIDC, but more like a mixture of OAuth2.0 + a bit of OIDC at the end.

aeneasr: Ok, sounds good to me! :)

aeneasr:
Ok, setting up tests is probably a lot of effort right now. Let's scrap that and we'll figure out internally how to do it!

Could you still add some docs over here please?

https://github.com/ory/kratos/edit/master/docs/docs/guides/sign-in-with-github-google-facebook-linkedin.mdx

jakhog:
So, I've done some things:

1. I implemented the change that we discussed, using the `tid` claim (verifying that it is a UUID) to construct the final issuer URL.
2. Split out two methods on the generic oidc provider to enable reuse in the microsoft provider. See https://github.com/ory/kratos/blob/a89a969f82a691ef6109eadaa9fdfaa8c67ebfc5/selfservice/strategy/oidc/provider_generic_oidc.go.
Don't know if you want to do it that way or if you like the copy-paste better, and I didn't spend a lot of time thinking about the method names. So let me know if you'd like me the revert or change that.
3. Updated the "Sign in with Microsoft" documentation to reflect the changes.

I spent some time looking in to the verification of email when adding users to an Azure AD, and it seems that Microsoft does not send the "email" claim if the account is not linked to a verified email. I was at least not able to reproduce the scenario described in the previous warning (in the Jsonnet snippet in the documentation), so I removed that warning. I tried quite a few things to add a "fake" email to users (with new instances of Azure AD or a B2C AD) but was not able to trigger a sign in with those users. However I did not find any documentation about the behaviour from Microsoft, so maybe there should be some kind of warning still?

jakhog:
Seems like the linting check failed because of a timeout, but I'm not allowed to trigger a re-run. So maybe you could give it a go?

aeneasr:
Yeah sure, rerunning the CI now :)

docs/docs/guides/sign-in-with-github-google-facebook-linkedin.mdx
```diff
@@ -105,46 +105,42 @@ This will enable you to log in using any Azure AD directory - Multitenant and
...
 local claims = std.extVar('claims');
 {
   identity: {
     traits: {
-      // Allowing unverified email addresses enables account
-      // enumeration attacks, especially if the value is used for
-      // e.g. verification or as a password login identifier.
-      //
-      // If connecting only to your organization (one tenant), claims.email is safe to use if you have not actively disabled e-mail
-      //verification during signup.
```
aeneasr, line 137: Does AD not keep track of verified email addresses?

docs/docs/guides/sign-in-with-github-google-facebook-linkedin.mdx
```diff
@@ -105,46 +105,42 @@ This will enable you to log in using any Azure AD directory - Multitenant and
...
-      email: claims.email
+      [if "email" in claims then "email" else null]: claims.email,
+      // The email might be empty if the account is not linked to an email address.
+      // For a human readable identifier, consider using the "preferred_username" claim.
```
aeneasr, line 137:
```suggestion
      // Allowing unverified email addresses enables account
      // enumeration attacks, especially if the value is used for
      // e.g. verification or as a password login identifier.
      //
      // If connecting only to your organization (one tenant), claims.email is safe to use if you have not actively disabled e-mail verification during signup.
      //
      // The email might be empty if the account is not linked to an email address.
      // For a human readable identifier, consider using the "preferred_username" claim.
      [if "email" in claims then "email" else null]: claims.email,
```

aeneasr:
Awesome 🎉 

PR #467, opened 2020-06-02
docs: minor fixes

gmazzamuto:
These are minor fixes for just a few errors in the documentation that I've spotted while learning Kratos. Hope this helps!

* Remove some repeated sentences
* Fix some REST paths
* Fix `<input disabled>`

## Checklist

      the changes.
      works.

### DIFF
--- a/selfservice/form/fields.go
+++ b/selfservice/form/fields.go
@@ -34,7 +34,7 @@ type Field struct {
 	// Pattern is the equivalent of `<input pattern="{{.Pattern}}">`
 	Pattern string `json:"pattern,omitempty"`
 
-	// Disabled is the equivalent of `<input disabled>`
+	// Disabled is the equivalent of `<input disabled="{{.Disabled}}">`
 	Disabled bool `json:"disabled,omitempty"`
 
 	// Required is the equivalent of `<input required="{{.Required}}">`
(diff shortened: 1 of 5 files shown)


selfservice/form/fields.go
```diff
@@ -34,7 +34,7 @@ type Field struct {
 	// Pattern is the equivalent of `<input pattern="{{.Pattern}}">`
 	Pattern string `json:"pattern,omitempty"`
 
-	// Disabled is the equivalent of `<input disabled="{{.Disabled}}">`
+	// Disabled is the equivalent of `<input disabled>`
```
aeneasr, line 37: This is actually correct, because `.Disabled` is `true` or `false`, leading to `<input disabled="true">` or `<input disabled="false">` which are both valid. This makes controlling the input field easy :)

```suggestion
	// Disabled is the equivalent of `<input disabled="{{.Disabled}}">`
```

gmazzamuto: Actually I think I got it right, I should have explained it better. The [specification](https://developer.mozilla.org/it/docs/Web/HTML/Element/input#disabled) doesn't allow for a value to be specified. These three are all equivalent (they all produce a disabled input):
```html
  <input type="password" disabled="false">
  <input type="password" disabled="true">
  <input type="password" disabled>
```

aeneasr: You're right! Would you be up for another patch to fix this? Thank you! :)

aeneasr: https://github.com/ory/kratos/pull/486

PR #508, opened 2020-06-14
docs: Fixed username & password use case example by replacing email with username

mikemilano:
This is a documentation correction

url: https://www.ory.sh/kratos/docs/concepts/credentials/username-email-password#use-case-username-and-password

**Problem**:
The Username and Password use case example shows an email example instead of a username example.

**Fix**:
This PR renames the `email` to `username` and removes `format` based on the username block in the following use case.


aeneasr:
Thank you!!

PR #509, opened 2020-06-14
docs: Replaced mailhog references with mailslurper in current documentation.

mikemilano:
The docs referenced Mailhog but the quickstart uses MailSlurper.

This PR replaces references to Mailhog with Mailslurper in the latest docs. I did not update the versioned docs as I do not know when Mailhog was actually used.

urls:
- https://www.ory.sh/kratos/docs/quickstart
- https://www.ory.sh/kratos/docs/guides/zero-trust-iap-proxy-identity-access-proxy


aeneasr:
Thank you!

PR #526, opened 2020-06-22
fix: account recovery can't use recovery token

### LINKED ISSUE #525: New account recovery feature can't use recovery token
**Describe the bug**

The `persister_recovery.go` file contains a bug in the [UseRecoveryToken](https://github.com/ory/kratos/blob/8bf9a1ac4162c677a455c2f02de658bd5d146905/persistence/sql/persister_recovery.go#L83) function. The for loop is written in such a way that even if the recovery token is found in the database, the loop continues. If there are multiple session secrets configured, a no-rows-found exception will be thrown, and the outcome of the loop is that the `err` variable will contain an error, which will make the function return the error rather than the token.

So, for example, **if you have two session keys, and the first was used to hash the token**, then the first iteration in the loop will produce a match, but instead of breaking the loop, it continues with the next session secret. Then, a now rows found error will be thrown at the second iteration, and then the loop stops. So, the outcome of the loop is that the `err `variable will have an error, and therefore the following if-statement becomes true and the error is returned from the function:

```
if err != nil {
    return err
}
```

Same problem if the second session secret is the one that was used to hash the token. Then, the first iteration in the loop would produce an error, and the second iteration a match. But still, the `err `variable would contain an error so the if-statement above would make the function return the error instead of the token.

**Reproducing the bug**

Follow the intended recovery flow.

_Server configuration_

Relevant section:

```yaml
secrets:
  session:
  - session-key-123-456
  - session-key-234-567
```
-->

**Expected behavior**

Instead of continuing the for-loop when a matching token has been found, **the `err `variable should be cleared and the loop exited with a break statement**.  This fixed the issue for me, at least. 

**Environment**

- Version: current master branch

**Additional context**

avamonitoring:
## Related issue

This PR fixes #525. 

## Proposed changes

The PR contains 3 changes:

1. When a matching token has been found in the database, breaks the for loop and clears the `err `variable, so the matching token can be returned
2. Makes sure that if a `sql: no rows in result set` error is thrown, it is recognized as intended
3. Makes sure only the matching database row is updated instead of the entire table

## Checklist

      the changes.
      works.


persistence/sql/persister_recovery.go
```diff
@@ -87,17 +88,19 @@ func (p *Persister) UseRecoveryToken(ctx context.Context, token string) (*link.T
 
 		for _, secret := range p.cf.SessionSecrets() {
 			if err = tx.Eager().Where("token = ? AND NOT used", p.hmacValueWithSecret(token, secret)).First(rt); err != nil {
-				if !errors.Is(err, sqlcon.ErrNoRows) {
+				if !errors.Is(err, sql.ErrNoRows) {
```
aeneasr, line 91:
```suggestion
				if !errors.Is(sqlcon.HandleError(err), sqlcon.ErrNoRows) {
```

persistence/sql/persister_recovery.go
```diff
@@ -87,17 +88,19 @@ func (p *Persister) UseRecoveryToken(ctx context.Context, token string) (*link.T
...
 					return err
 				}
+			} else {
+				err = nil
```
aeneasr, line 95: err is already nil and does not need to be re-assigned

```suggestion
```

avamonitoring:
I will give it a go writing a test case.

avamonitoring:
@aeneasr We need to remove the import I added as well. Can you request a change?

persistence/sql/persister_recovery.go
```diff
@@ -2,6 +2,7 @@ package sql
 
 import (
 	"context"
+  "database/sql"
```
aeneasr, line 5:
```suggestion
```

avamonitoring:
@aeneasr It appears as if there are already 2 relevant testcases [here](https://github.com/ory/kratos/blob/8bf9a1ac4162c677a455c2f02de658bd5d146905/selfservice/strategy/link/persister_conformity.go#L53) and. Are those cases sufficient, or do you want them moved/copied to the location you suggested?

aeneasr:
Those are not sufficient because they do not break when we revert the changes made :) A proper way to do this would be to write a failing test case for the observed bug first, and then implement the actual logic. If tests were not catching this prior, they will not be sufficient now, because reverting your code wouldn't make a difference.

avamonitoring:
@aeneasr OK, but if I follow the documentation for how to run the tests, and cd into the `selfservice/strategy/link` folder, and `$go test -tags sqlite -v .`, there is no evidence of the tests in `persister_conformity.go` ever being run. The same applies if I do a `$make test`. I'm a bit puzzled. Obviously these tests need a running database. But how, and from where are these tests being invoked?

aeneasr:
Oh I see - that's very unfortunate that you're hitting that issue. It appears to be a new problem. Here's what you could try:

```
docker rm -f kratos_test_database_mysql kratos_test_database_postgres kratos_test_database_cockroach kratos_test_hydra || true
docker run --name kratos_test_database_mysql -p 3444:3306 -e MYSQL_ROOT_PASSWORD=secret -d mysql:5.7
docker run --name kratos_test_database_postgres -p 3445:5432 -e POSTGRES_PASSWORD=secret -e POSTGRES_DB=postgres -d postgres:9.6 postgres -c log_statement=all
docker run --name kratos_test_database_cockroach -p 3446:26257 -d cockroachdb/cockroach:v20.1.1 start --insecure

# windows cmd.exe (NOT powershell):
set TEST_DATABASE_MYSQL="mysql://root:secret@(127.0.0.1:3444)/mysql?parseTime=true&multiStatements=true"
set TEST_DATABASE_POSTGRESQL="postgres://postgres:secret@127.0.0.1:3445/postgres?sslmode=disable"
set TEST_DATABASE_COCKROACHDB="cockroach://root@127.0.0.1:3446/defaultdb?sslmode=disable"

go run -test sqlite ./...
```

aeneasr:
I also realized that I broke tests on master: https://github.com/ory/kratos/pull/527

aeneasr:
> Edit: I'm starting to believe this error only happens when you have more than one session secret. From having run the tests (with added tracing), I can't provoke the exception to happen with the original code, which leads me to believe when running the tests there is only one session secret configured? So, basically I think the tests are fine, but the configuration is only using one session secret and that's why the tests are not catching what I (and everybody else using more than one session secret) ran into.

That's definitely possible! You can set a multiple for your test like so:

```
import "github.com/ory/kratos/driver/configuration"
import "github.com/ory/viper"

viper.Set(configuration.ViperKeySecretsSession, []string{"secret-a", "secret-b"})
```

avamonitoring:
@aeneasr I followed your instructions and added

`viper.Set(configuration.ViperKeySecretsSession, []string{"secret-a", "secret-b"})`

to `persister_conformity.go`, and then tested both the original code as well as the corrected one. With the original code, the error was now thrown correctly, and with the new code the test passes. So, the tests are fine, it was just the extra session secret that was needed to catch the exception. But, tests are still failing at some other tests, as shown below. Don't think it's related to this PR though:

```
--- FAIL: TestSettingsStrategy/suite=unlink (0.18s)
        --- PASS: TestSettingsStrategy/suite=unlink/case=should_not_be_able_to_unlink_the_last_remaining_connection (0.03s)
        --- PASS: TestSettingsStrategy/suite=unlink/case=should_not_be_able_to_unlink_an_non-existing_connection (0.03s)
        --- PASS: TestSettingsStrategy/suite=unlink/case=should_not_be_able_to_unlink_a_connection_not_yet_linked (0.04s)
        --- FAIL: TestSettingsStrategy/suite=unlink/case=should_unlink_a_connection (0.07s)
```

PR #530, opened 2020-06-25
feat: Add User ID to a header when calling whoami

koenmtb1:
## Related issue
We're using the AuthService provided by Ambassador to check if the user is authenticated but right now we're separately calling the whoami again. This means that for every request that a user does we're talking to Kratos twice. By adding the User ID to a header we can prevent this by using the `allowed_authorization_headers` option (see Ambassador doc: https://www.getambassador.io/docs/latest/topics/running/services/auth-service/). 

## Proposed changes
Set the UserID in a header, I've put it on "X-Authenticated-User" right now.

## Checklist
      the changes.
      works.

### DIFF
--- a/session/handler.go
+++ b/session/handler.go
@@ -82,7 +82,7 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, ps httprouter.P
 	s.Identity = s.Identity.CopyWithoutCredentials()
 	
 	// Set userId as the X-Authenticated-User header.
-	w.Header().Set("X-Authenticated-User", s.Identity.ID.String())
+	w.Header().Set("X-Kratos-Authenticated-Identity-Id", s.Identity.ID.String())
 
 	h.r.Writer().Write(w, r, s)
 }
--- a/session/handler.go
+++ b/session/handler.go
@@ -56,7 +56,8 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
 // Check who the current HTTP session belongs to
 //
 // Uses the HTTP Headers in the GET request to determine (e.g. by using checking the cookies) who is authenticated.
-// Returns a session object or 401 if the credentials are invalid or no credentials were sent.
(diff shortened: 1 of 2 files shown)


koenmtb1:
Is a documentation change and tests necessary for this change?

aeneasr:
Yes!

session/handler.go
```diff
@@ -80,6 +80,9 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, ps httprouter.P
 
 	// s.Devices = nil
 	s.Identity = s.Identity.CopyWithoutCredentials()
+	
+	// Set userId as the X-Authenticated-User header.
+	w.Header().Set("X-Authenticated-User", s.Identity.ID.String())
```
aeneasr, line 85: I think it would make sense to bind this field to the kratos domain, as `X-Authenticated-User` appears in a several places with different implementations! How about this?

```suggestion
	w.Header().Set("X-Kratos-Authenticated-Identity-Id", s.Identity.ID.String())
```

Alternatively we could also send the identity in the payload:


```suggestion
	w.Header().Set("X-Kratos-Authenticated-Identity", jsonify(s.Identity))
```

koenmtb1: I agree with the change of header name. 
For me it doesn't matter if the entire Identity is in there but the ID would be enough. Adding the Identity itself would add about 400 bytes of data (which is insignificant of course). But you might help someone else who does want the entire thing. I'll leave the second decision up to you.

aeneasr: Ok, let's keep the ID then only :)

aeneasr:
Could you add a test [here](https://github.com/ory/kratos/blob/master/session/handler_test.go#L74) for this please? :)

koenmtb1:
Looks like there's a new vulnerability that is causing issues. Should I rebase to fix it?

aeneasr:
All good, thank you!

Issue #316, opened 2020-04-05
Kratos offline, disable network checks


frederikhors:
I'm trying to use kratos offline.

Is there a way to disable password check on https://haveibeenpwnd.com and similar network constraints (eg. Kratos Public APIs)?

aeneasr:
You can currently not change the password policy, but we'll add that at a later stage.

frederikhors:
> You can currently not change the password policy, but we'll add that at a later stage.

This is very urgent for anyone who uses a completely offline environment.

So for now there is no way unless you edit the code and recompile it, right?

aeneasr:
> This is very urgent for anyone who uses a completely offline environment.
>
> So for now there is no way unless you edit the code and recompile it, right?

Why don't you create a PR instead?

frederikhors:
I would love to know where to start! :)

aeneasr:
Looking at the code, it looks like this should be the case already: https://github.com/ory/kratos/blob/master/selfservice/strategy/password/validator.go#L53-L68

Maybe there's a bug though. Would probably be best to write a test case for that and see if it works as intended :)

https://github.com/ory/kratos/blob/master/selfservice/strategy/password/validator_test.go#L30

frederikhors:
But after the call to `NewDefaultPasswordValidatorStrategy()` this line https://github.com/ory/kratos/blob/master/selfservice/strategy/password/validator.go#L72 set it to `false`.

```go
func NewDefaultPasswordValidatorStrategyStrict() *DefaultPasswordValidator {
	v := NewDefaultPasswordValidatorStrategy()
	v.ignoreNetworkErrors = false
	return v
}
```

Maybe is this?

jakhog:
I can confirm that the validator gets stuck in an infinite loop here https://github.com/ory/kratos/blob/d743ae8d1b91adb6e0eb0380c1b4d1f99940e754/selfservice/strategy/password/validator.go#L168-L178 if Kratos is not able to connect to the online password check. I'll try to make a test case and fix it (shouldn't be that hard).

Would it however also make sense to enable a configuration setting for not performing online checks at all? Not sure, but maybe in a given deployment it would not return an error right away but after a long timeout, which isn't that great for the UX...

frederikhors:
I agree with the option to disable it at all.

jakhog:
I fixed the bug in #589 so that it can work, but I have not added an option to disable the online check. I'll leave the decision if that is something you want to you @aeneasr.

[...]

PR #589, opened 2020-07-18
fix: Network errors during pwnedpasswords.com check causing recursive loop

jakhog:
## Related issue

On network errors, the password validator ends up in a recursion loop until either the fault disappears - or more likely the stack overflows. Related to #316, @aeneasr .

## Proposed changes

Change the validator code so that when `ignoreNetworkErrors = true`, a failure during the _pwnedpasswords.com_ check returns no error, but the validation of the same password later would retry the online check.

Added test cases that previously crashed or timed out the test suite (because of the loop), but they pass now. I also took the liberty to add a few more tests for the online password leak check.

## Checklist

      the changes.
      works.

## Further comments

I implemented a `FakeHTTPClient` to use for these tests and feel like it should really live somewhere else to potentially be reused, but I couldn't find a good home for it. The _internal/testhelpers_ package doesn't work as it causes an import cycle, any suggestions?

This is also the first Go test I've written, so happy with any feedback if something is not perfect 🙂

### DIFF
--- a/selfservice/strategy/password/validator.go
+++ b/selfservice/strategy/password/validator.go
@@ -99,23 +99,17 @@ func lcsLength(a, b string) int {
 	return greatestLength
 }
 
-func (s *DefaultPasswordValidator) fetch(hpw []byte) error {
+func (s *DefaultPasswordValidator) fetch(hpw []byte) (bool, error) {
 	prefix := fmt.Sprintf("%X", hpw)[0:5]
 	loc := fmt.Sprintf("https://api.pwnedpasswords.com/range/%s", prefix)
 	res, err := s.c.Get(loc)
 	if err != nil {
-		if s.ignoreNetworkErrors {
-			return nil
-		}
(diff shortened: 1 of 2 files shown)


selfservice/strategy/password/validator.go
```diff
@@ -172,7 +166,10 @@ func (s *DefaultPasswordValidator) Validate(identifier, password string) error {
 	s.RUnlock()
 
 	if !ok {
-		if err := s.fetch(hpw); err != nil {
+		if shouldIgnore, err := s.fetch(hpw); err != nil {
+			if shouldIgnore {
+				return nil
+			}
```
aeneasr, line 172: And then you do this:

```suggestion
		if err := s.fetch(hpw); errors.Is(err, ErrNetworkFailure) {
			return nil
		} else if err != nil {
			return err
		}
```

jakhog:
Thanks for the feedback! I was away on vacation so haven't seen this until now, but seems like you already fixed it so thank you 😄

aeneasr:
No problem, I hope you had a good vacation and were able to regenerate :)

PR #592, opened 2020-07-20
fix: improve pagination when listing identities

aeneasr:
## Checklist

      the changes.
      works.

### DIFF
--- a/persistence/sql/persister_identity.go
+++ b/persistence/sql/persister_identity.go
@@ -191,16 +191,16 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
 	})
 }
 
-func (p *Persister) ListIdentities(ctx context.Context, limit, offset int) ([]identity.Identity, error) {
+func (p *Persister) ListIdentities(ctx context.Context, page, limit int) ([]identity.Identity, error) {
 	is := make([]identity.Identity, 0)
 
 	/* #nosec G201 TableName is static */
 	if err := sqlcon.HandleError(p.GetConnection(ctx).
-		RawQuery(fmt.Sprintf("SELECT * FROM %s LIMIT ? OFFSET ?", new(identity.Identity).TableName()), limit, offset).
-		Eager("VerifiableAddresses", "RecoveryAddresses").All(&is)); err != nil {
+		Paginate(page, limit).Order("id DESC").
+		Eager("VerifiableAddresses", "RecoveryAddresses").
+		All(&is)); err != nil {
 		return nil, err
 	}
-
 	for i := range is {
 		if err := p.injectTraitsSchemaURL(&(is[i])); err != nil {
 			return nil, err
--- a/persistence/sql/persister_identity.go
+++ b/persistence/sql/persister_identity.go
@@ -199,16 +199,15 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
 	})
 }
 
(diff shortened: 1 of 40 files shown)


x/pagination.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"net/http"
+	"strconv"
+)
+
+const paginationMaxItems = 500
+
+// ParsePagination parses limit and page from *http.Request with given limits and defaults.
+func ParsePagination(r *http.Request) (page, limit int) {
+	if offsetParam := r.URL.Query().Get("page"); offsetParam == "" {
```
harsimranmaan, line 12: Can be refactored into something like

```suggestion
	const maxIntVal = int(^uint(0) >> 1)
```
```suggestion
	offsetParam := paginationParam(r.URL.Query().Get("page"), 0, maxIntVal)
	limitParam  := paginationParam(r.URL.Query().Get("limit"), paginationMaxItems, paginationMaxItems)
```

```suggestion
func parseParam(rawParamValue string, defaultValue int, maxValue int) (paramVal int) {
	paramVal := defaultValue
	if rawParamValue == "" {
		return
	} 

	if paramVal64, err := strconv.ParseInt(rawParamValue, 10, 64); err != nil {
		paramVal = defaultValue
	} else {
		paramVal = int(paramVal64)
	}
	
	if paramVal > maxValue {
		paramVal = maxValue
	}

	if paramVal < 0 {
		paramVal = defaultValue
	}
	return
}
```

Would be even better if the default for the "limit" is smaller than the maxValue for the limit as having smaller limits gives users the ability to bump up(to the maximum) when needed. Users would likely never set the limits lower than the default in practice.

x/pagination.go
```diff
@@ -0,0 +1,45 @@
...
+			limit = int(limit64)
+		}
+	}
+
+	if limit > paginationMaxItems {
+		limit = paginationMaxItems
+	}
+
+	if limit < 0 {
+		limit = 0
```
harsimranmaan, line 37: LIMIT 0 would return empty datasets. Is this desired?

harsimranmaan:
Also curious how the client would know if a next page is available?

x/pagination.go
aeneasr: Right this should be `1`, nice catch!

aeneasr:
Good question, we should use this: https://github.com/ory/x/blob/master/pagination/header.go#L23

Maybe abstract it a little further to make it easier to use (no manual counting for example)

x/pagination.go
aeneasr: All your remarks should have been addressed now :)

harsimranmaan: Yes, thanks. There's still scope for exacting a function to avoid code duplication.

persistence/sql/persister_identity.go
```diff
@@ -157,6 +157,14 @@ func createRecoveryAddresses(ctx context.Context, tx *pop.Connection, i *identit
 	return nil
 }
 
+func (p *Persister) CountIdentities(ctx context.Context) (int64, error) {
+	count, err := p.c.WithContext(ctx).Count(new(identity.Identity))
```
harsimranmaan, line 161: Counts can be very expensive to compute as the table size grows. A caching layer might serve well in the future for this kind of use case but that would require an across the board tuning once that scale is hit.

aeneasr: Thank you, that's definitely true. There are ways to improve that though (e.g. with caching). There are some PRs open in gobuffalo to help with this.

Issue #590, opened 2020-07-20
Add the config schema to schemastore.org


zepatrik:
**Is your feature request related to a problem? Please describe.**

[schemastore.org](https://www.schemastore.org/json/) provides json schemas via an API for editors like VSCode and JetBrains to offer code completion, validation and more. If we add our config.schema.json our users get editor support for config files.

**Describe the solution you'd like**

We can easily submit a PR to add our schema. One problem I see is that the config file per default is called `config.yml` which is not descriptive enough. We should rather change the default file name to be `kratos.yml` so we can add the pattern.

**Describe alternatives you've considered**

Add the schema but keep the default `config.yml`. This would result in most of the users not benefiting from the feature as they probably use the default file name.

**Additional context**

It is important to figure out how they do versioning of schemas, especially for us as it changes quite often.

zepatrik:
The versioning strategy of schemastore.org is to just postfix the file with the version number and load the burden of deciding which version to use to the editor (e.g. by examining the environment). As this will clearly not work with our config files my proposal is to add a top level version key that currently only is used for validation. Kratos will use a config file although the version does not match but spit out a warning. This key will be required then. Further we can write a "version meta" schema that uses `oneOf`/`allOf` directives to map the top level version key to the proper schema hosted on github. We can even automate the generation of that file when releasing :tada:

tacurran:
@zepatrik
JSON schema maps the structure of JSON files, so assume that the version meta is a JSON schema. Regarding the "add a top level version key that currently only is used for validation" Would you suggest (e.g. version meta - a JSON containing all versions and descriptions thereof) using a JSON Schema ($schema": "http://json.schemastore.org/kratosconfig) to describe the version and the use $ref to navigate to the actual schema.

In a JSON schema, a $ref keyword is a JSON Pointer to a schema, or a type or property in a schema. A JSON pointer takes the form of A # B in which: A is the relative path from the current schema to a target schema. ... B is the complete path from the root of the schema to a type or property in the schema. (http://niem.github.io/json/reference/json-schema/references/)
"Further we can write a "version meta" schema that uses oneOf/allOf directives to map the top level version key to the proper schema hosted on github." This would have to be either in the ory/kratos/ release config since schemastore.org does not seem to accept the type of "version meta" you imagine.

In addition, schemas may be combined with the following keywords:
*allOf*: must be valid against all of the subschemas
*anyOf*: must be valid against any of the subschemas
[...]

PR #608, opened 2020-07-29
feat: config versions meta schema

zepatrik:
## Related issue

Prepares #590 

## Proposed changes

The newly added schema detects the schema version and maps it to the corresponding release on GitHub. There will be a script to automatically add new entries in our release pipeline.


.schema/config.schema.json
```diff
@@ -4,6 +4,11 @@
   "title": "ORY Kratos Configuration",
   "type": "object",
   "definitions": {
+    "version": {
+      "type": "string",
+      "description": "SemVer according to https://semver.org/ prefixed with `v` as in our releases.",
+      "pattern": "^v(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
```
aeneasr, line 10: This won't work with `latest` or generally builds that have not been built as part of the CI build pipeline, right?

.schema/config.schema.json
```diff
@@ -972,7 +987,8 @@
   "required": [
     "identity",
     "dsn",
-    "selfservice"
+    "selfservice",
+    "version"
```
aeneasr, line 991: Does `version` really have to be required?

zepatrik: right we wanted to print a warning in case the version is not specified or does not match

.schema/config.schema.json
zepatrik: yes but it would not make sense to have `version: latest` in the config anyway. In case of latest/master/... we would ignore the key completly

aeneasr: That would be the idea!

.schema/version.schema.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$id": "https://github.com/ory/kratos/.schema/versions.config.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
zepatrik, line 3: One small thing is missing: this should be
```suggestion
  "$schema": "https://raw.githubusercontent.com/ory/cli/{{ release }}/.schema/version_meta.schema.json#",
```
but we have to release the cli therefore.

zepatrik:
Ready to be merged from my side

PR #614, opened 2020-07-30
feat: add the already declared (and settable) tracer as a middleware

err0r500:
## Proposed changes

It looks like the tracer is declared but not used even if one is allowed to set env vars to activate and configure it.

I just added it as a negroni middleware, exactly like it's already done on oathkeeper.

## Checklist

      the changes.
      works.


cmd/daemon/serve.go
```diff
@@ -45,6 +45,10 @@ func servePublic(d driver.Driver, wg *sync.WaitGroup, cmd *cobra.Command, args [
 	n.Use(NewNegroniLoggerMiddleware(l, "public#"+c.SelfPublicURL().String()))
 	n.Use(sqa(cmd, d))
 
+	if tracer := d.Registry().Tracer(); tracer.IsLoaded() {
```
aeneasr, line 48: This should also be added to `serveAdmin` :)

err0r500: sure

PR #625, opened 2020-08-05
docs: added sidebar item

vinckr:
Added Kratos Video Guides Transcripts

## Checklist

      the changes.
      works.

### DIFF
--- a/docs/sidebar.js
+++ b/docs/sidebar.js
@@ -73,7 +73,8 @@ module.exports = {
     "guides/login-session",
     "guides/secret-key-rotation",
     "guides/high-availability-ha",
-    "guides/docker"
+    "guides/docker",
+    "guides/kratos-video-tutorials"
   ],
   "Reference": [
     "reference/configuration",


docs/sidebar.js
```diff
@@ -73,7 +73,8 @@ module.exports = {
     "guides/login-session",
     "guides/secret-key-rotation",
     "guides/high-availability-ha",
-    "guides/docker"
+    "guides/docker",
+    "guides/kratos-video-tutorials"
```
aeneasr, line 77: Wouldn't this make more sense in the "Further Reading" section?

vinckr: It is not very essential true, I just thought put it there because it directly related to the guides above.

aeneasr: Yeah ok makes sense!

aeneasr: As we discussed this changed to `.json` format - could you update the PR or make a new one?

PR #632, opened 2020-08-05
feat: add version schema generation step to release automation

zepatrik:
## Proposed changes

This adds the version schema step to the release automation. Note that I also converted all `environment:` configurations to use key/value maps instead of an array. I noted this issue because of the JSON schema for cci. Here are the corresponding docs: https://circleci.com/docs/reference-2-1/#jobs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
       -
         image: circleci/golang:1.14
         environment:
-          GO111MODULE: on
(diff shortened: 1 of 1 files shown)


aeneasr:
@zepatrik have you tested the script? Does it actually work in the CI? We don't want to do a release which then fails in the last step because something in the script is broken (e.g. missing +x flag or other syntax errors).

PR #633, opened 2020-08-05
chore: update docusaurus template

aeneasr:
Updated docusaurus template to current https://github.com/ory/docusaurus-template.


zepatrik:
Consider first merging https://github.com/ory/docusaurus-template/pull/35

PR #634, opened 2020-08-06
fix: tell tls what the smtps server name is

alsuren:
## Related issue

As discussed with @aeneasr via email.

Previously there was no way to use smtps to send email without adding `?skip_ssl_verify=true` to the end of the courier.smtp.connection_uri. In practice this means that anyone with the ability to poison DNS records (for any kratos installation that supports password reset emails) will be able to register an ssl certificate for any-old-domain and receive all password reset emails (since there is no way to configure kratos to verify the ssl server name for the email provider).

> thank you for the disclosure! Given that misconfiguration results in an error (or enforces TLS verification skipping which results in raised eyebrows :) ) I would not classify this as a security vulnerability. Hence, it is ok (and encouraged) to just open a public PR for this. Thank you!

## Proposed changes

Allow the tls module to correctly validate the hostname of the smtp server, by telling it what hostname you're expecting to connect to.

## Checklist

      the changes.
      works.

Once this patch is applied, it is possible to send mail via `smtps://...@smtp.sendgrid.net:465/` without any warnings or errors (using their free tier). I'm not sure what other tests or documentation changes would be valuable.

### DIFF
--- a/courier/courier.go
+++ b/courier/courier.go
@@ -49,7 +49,7 @@ func NewSMTP(d smtpDependencies, c configuration.Provider) *Courier {
 		ssl = true
 		sslSkipVerify, _ := strconv.ParseBool(uri.Query().Get("skip_ssl_verify"))
 		// #nosec G402 This is ok (and required!) because it is configurable and disabled by default.
-		tlsConfig = &tls.Config{InsecureSkipVerify: sslSkipVerify}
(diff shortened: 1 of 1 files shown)


aeneasr:
Thank you! e2e tests fail for a different reason which I'll address in another PR

aeneasr:
Looks like we have to wait for NPM to fix their cache issues: https://status.npmjs.org

Issue #645, opened 2020-08-10
Logout does not use new cookie domain setting


wezzle:
**Describe the bug**

The browser logout flow does not use the newly configurable cookie domain added in #516. I'm guessing the domain and path options should be correctly set for new cookies in this file:
https://github.com/ory/kratos/blob/master/driver/registry_default.go#L369

The options set in the following file do not seem to be persisted in the gorilla/session's store:
https://github.com/ory/kratos/blob/master/session/manager_http.go#L66

On browser flow login the session.Options are as followed:
`&{Path:/ Domain:.happyitaly.localhost MaxAge:3600 Secure:false HttpOnly:true SameSite:2}`

On browser flow logout they are as followed:
`&{Path:/ Domain: MaxAge:2592000 Secure:false HttpOnly:true SameSite:0}`

The max age is then overwritten by: https://github.com/ory/kratos/blob/master/session/manager_http.go#L120

**Reproducing the bug**

1. Set custom cookie domain to `.domain.com` (notice the `.` so it is readable for all subdomains)
2. Start the logout flow
3. See the `Set-Cookie` header in the network panel, notice the wrong `Domain`

**Expected behavior**

Cookie domain is set correctly so logout works.

**Environment**

- Version: c13b1838
- Environment: Docker

fenix01:
Hi,

I was going to post a similar issue. I have the same bug with the logout flow. It doesn't set the Domain for the cookie.

PR #646, opened 2020-08-10
fix: correct cookie domain on logout

wezzle:
## Related issue

#645 

## Checklist

(Not sure what my commit message should've had as prefix?)
      the changes.
      works.


wezzle:
@aeneasr I've just created a seperate CookieManager for the continuity module. If you'd rather have it the other way around I can refactor it but I had no idea of a good name (sessionSessionStore?)

As for the test perhaps you can guide me in the right direction. As far as I can see it is hard to test cookies with a different domain setting as the server is running on `127.0.0.1` and the http library does not send or accept cookies if the domain does not match.

wezzle:
@aeneasr I know you're probably busy with the other projects but please let me know if there is anything I can do to get this issue fixed and merged.

driver/registry_default.go
```diff
@@ -368,11 +369,37 @@ func (m *RegistryDefault) CookieManager() sessions.Store {
...
 		m.sessionsStore = cs
 	}
 	return m.sessionsStore
 }
 
+func (m *RegistryDefault) ContinuityCookieManager() sessions.Store {
+	if m.continuitySessionStore == nil {
+		cs := sessions.NewCookieStore(m.c.SecretsSession()...)
+		cs.Options.Secure = !m.c.IsInsecureDevMode()
+		cs.Options.HttpOnly = true
```
aeneasr, line 397: Given that `SameSite: None` no longer works with `secure: false` I think we should set the samesite mode of this cookie to `Lax` to avoid any issues (given that it needs to work with cross-origin redirects).

aeneasr (approved):

aeneasr:
Awesome 🎉 

PR #671, opened 2020-08-25
chore: rename quickstart config files and path

zepatrik:
## Related issue

With our config schemas added to schemastore.org, we should use our proposed file names ourselves.
This will result in IDEs recognizing the files as kratos/oathkeeper configs.

## Proposed changes

Rename the quickstart config files and adapt the paths in the `quickstart*.yml` files.

## Further comments

The kratos config file is already renamed but the path was not up to date.


aeneasr (approved):
LGTM

PR #674, opened 2020-08-26
chore: refactor cmd to make commands importable

zepatrik:
## Related issue

Before the commands where not importable as they automatically registered at their parent.

## Proposed changes

All command modules have a `RegisterCommandRecursive` function that registers the command at a provided parent and also all it's subcommands.

## Further comments

I decided to add have modules for every single command to allow importing of e.g. jsonnet lint without format. Also the new cli helpers modules is required to avoid circular dependencies.


aeneasr:
> I decided to add have modules for every single command to allow importing of e.g. jsonnet lint without format.

I don't think there are many upsides to this (in e.g. compile size) because we will import a lot of those commands anyways. Having all the sub-commands for a parent command in the same directory makes it much easier to navigate the code base though IMO.

zepatrik:
But I should keep the top level packages right? Like `jsonnet` or `serve`

aeneasr:
Yes, absolutely! If we have many many subcommands it will help with readability :)

aeneasr (approved):

Issue #691, opened 2020-09-05
[I18n] Add i18n support for Messages


UkonnRa:
**Is your feature request related to a problem? Please describe.**

Now Messages are something like:

```json
[
  {
    "id": 4000006,
    "text": "The provided credentials are invalid, check for spelling mistakes in your password or username, email address, or phone number.",
    "type": "error",
    "context": {}
  }
]
```

So if `text` contains some variables, it will be impossable to do the i18n work

**Describe the solution you'd like**

Use [ICU format](http://site.icu-project.org/) to solve the problem

```json5
[
  {
    "id": 4000006, // This is a 1/1 mapping to `i18nText`, in another word, one ID can map to one and only one `i18nText`
    "text": "Hello world, Foo Bar", // A meanful description for developer, it should NOT be used as i18n directly
    "i18nText": "Hello, {firstName} {lastName}",
    "i18nData": {
        "firstName": "Foo",
        "lastName": "Bar"
    },
    "type": "error",
    "context": {}
  }
]
```

**Describe alternatives you've considered**

No

**Additional context**

React i18n-next: https://react.i18next.com/misc/using-with-icu-format

zepatrik:
Actually the parameters `firstName` and `lastName` would be under the `"context"` key. But I agree with you that we should use a standard instead of our own solution, especially when it does exactly the same thing. The refactoring would not be much work, just enforcing the standard at this type: https://github.com/ory/kratos/blob/master/text/message.go#L35-L40
Would you be up for a PR @UkonnRa ?

UkonnRa:
I wonder what is the current i18n solution right now? How Kratos handles variables, plural, gender right now? Or right now all Messages are static strings and have no variables at all?

zepatrik:
Right now we just have IDs so you can map messages to your own strings. You have to interpret the context on your own and we don't have a neat overview over all the IDs and associated contexts. The message you see is just a sensible default you can display if you don't want to bother rewriting all the messages. Until now, we had no complex messages that would require plural or gender forms, but we will get to that point eventually. Everything happens in the [text](https://github.com/ory/kratos/tree/master/text) package, including all the constants.

aeneasr:
[...]

PR #692, opened 2020-09-07
[WIP, discussion needed] Add `i18nText` and `i18nData`

UkonnRa:
## Related issue

https://github.com/ory/kratos/issues/691

## Proposed changes

Add `i18nText` and `i18nData` for i18n

## Checklist

      the changes.
      works.

### DIFF
--- a/text/message.go
+++ b/text/message.go
@@ -33,10 +33,12 @@ func (h *Messages) Clear() Messages {
 }
 
 type Message struct {
-	ID      ID              `json:"id"`
-	Text    string          `json:"text"`
-	Type    Type            `json:"type"`
-	Context json.RawMessage `json:"context,omitempty" faker:"-"`
+	ID       ID              `json:"id"`
+	Text     string          `json:"text"`
+	Type     Type            `json:"type"`
+	Context  json.RawMessage `json:"context,omitempty" faker:"-"`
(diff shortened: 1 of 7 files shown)


UkonnRa:
In most cases, `context` is `i18nData`, except when related to `expired_at` / `ago`. So we have to separate these two fields

zepatrik:
After having a more detailed look into all the available libraries and formats, I think we should stick with the current approach. We should provide an already formatted text to use for evaluation/development (`"text"`). To customize you will have to write your own translations for every message ID, but we provide the `"context"` in a way so you can use it right as is to pass to your translations. Is there anything we have to change to the format of `"context"` to allow easy usage?
Btw I think it would be awesome to have some community maintained default translations, we would be happy to feature them and provide support.

UkonnRa:
I get your point. If so I will just use `context` as the data

zepatrik:
I mean if we can refactor it in some way to improve the usability then we're definitely up for it.

aeneasr:
Hm I would be interested to see different approaches to i18n. I know the old systems that e.g gettext/PHP used but those were really hard to work with (.po files). I'm definitely open to improve this! Maybe we could have a human readable string and one with replacements?

PR #693, opened 2020-09-07
feat: client cli

### LINKED ISSUE #663: Design of the client cli

We want to add a client CLI with management capabilities similar to [Hydra](https://www.ory.sh/hydra/docs/cli/hydra/). This issue summarizes all the features it should provide for now.

Feel free to comment on any feature. The points denoted with a question mark are not going to be included if nobody wants them.

Output formats (flag `--output -o`):
- Table view (default)
- JSON pretty
- JSON
- ID only (nice for piping, similar to `docker ps -q`)
- ? file

Configuration of remote URL:
- env var
- flag
- ? config file in working dir
- ? globally defined context like with kubectl

Here is the list of proposed commands, output format is just exemplary:
- [ ] check status
    ```
    $ kratos remote status
    Property | Status
    ---------|-------
    alive    | true
    ready    | true
    ```
- [ ] check version
    ```
    $ kratos remote version
    v1.0.0
    ```
- [ ] list identities (filtering will be added when it is implemented in kratos). **Should we allow listing all without pagination? What information would you like to see in the table output?**
    ```
    $ kratos identities list --page 1 --per-page 100
    ID   | Identifier
    -----|------------
    ID_1 | foo@bar.com
    ```
- [ ] create identity (reading from `STD_IN` should be enough as one can pipe a file via cat)
    ```
    $ cat identity_peter.json | kratos identities create
    ID_PETER
    ```
- [ ] get a single identity (ID should be provided via `STD_IN` or argument). **What information would you like to see in the table output?**
    ```
    $ kratos identities get <id>
    {...}   # dump of the identity json
- [ ] update a single identity (reading from `STD_IN` should be enough as one can pipe a file via cat)
    ```
    $ cat identity_peter_updated.json | kratos identities update
    ID_PETER
    ```
- [ ] delete a single identity (ID should be provide via `STD_IN` or argument; skip question with `--force -f`). **What information would you like to see in the table output?**
    ```
    $ kratos identities delete <id>
    Do you really want to delete the identity with the identifier foo@bar.com? [yes/no]: y
    <id>
**Additional context**

The CLI will be a wrapper for the go SDK.

zepatrik:
## Related issue

closes #663 

## Proposed changes

This adds a client cli to manage kratos.

The available commands are (assuming `KRATOS_ADMIN_ENDPOINT` is set):
```
$ kratos remote version
v1.0.0
$ kratos remote status
ALIVE   true    
READY   true 
$ kratos identities delete <id>
<id>
$ kratos identities validate identity-file.json
exit code 0
$ kratos identities put identity-file.json
<id>
$ kratos identities list
ID                                      VERIFIED ADDRESS 1      RECOVERY ADDRESS 1      SCHEMA ID       SCHEMA URL                                                      
c00bbc8f-dfa3-40f1-9173-e270e8c274b8    foox@bar.com            foox@bar.com            default         http://127.0.0.1:4455/.ory/kratos/public/schemas/default        
a37a3e82-556e-482f-a8c8-20c4346a5f0c    foo@bar.com             foo@bar.com             default         http://127.0.0.1:4455/.ory/kratos/public/schemas/default        
79096f15-258e-42f8-a715-80f1af6fef38    foof@bar.com            foof@bar.com            default         http://127.0.0.1:4455/.ory/kratos/public/schemas/default
$ kratos identities get c00bbc8f-dfa3-40f1-9173-e270e8c274b8
ID                      c00bbc8f-dfa3-40f1-9173-e270e8c274b8                            
VERIFIED ADDRESSES      foox@bar.com                                                    
RECOVERY ADDRESSES      foox@bar.com                                                    
SCHEMA ID               default                                                         
SCHEMA URL              http://127.0.0.1:4455/.ory/kratos/public/schemas/default        
$ kratos identities delete <id> <invalid-id>
<id>
404: <invalid-id>
```

## Further comments

I think the best way to test this is actually in integration with an upstream kratos. What is the best way to set this up @aeneasr ?


zepatrik:
Note that there are 4 options for format:
- json (no `\n`)
- json-pretty (nice to read)
- quiet (ID only (like docker -q))
- table (default, as in above examples)

cmd/identities/list.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/ory/kratos/cmd/cliclient"
+	"github.com/ory/kratos/internal/httpclient/client/admin"
+	"github.com/ory/x/cmdx"
+)
+
+var listCmd = &cobra.Command{
+	Use: "list [<page> <per-page>]",
+	Args: func(cmd *cobra.Command, args []string) error {
+		// zero or exactly two args
+		if len(args) != 0 && len(args) != 2 {
```
aeneasr, line 21: cobra.Command has something called `Args` where you can use `cobra.ExactArgs` :)

zepatrik: Sure but it can not check for 0 *or* 2 args

cmd/identities/delete.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/ory/kratos/cmd/cliclient"
+	"github.com/ory/kratos/internal/httpclient/client/admin"
+)
+
+var deleteCmd = &cobra.Command{
+	Use:  "delete",
+	Args: cobra.MinimumNArgs(1),
```
aeneasr, line 16: Could you add some helper descriptions to these commands? Examples could also be very helpful!

aeneasr (commented):
Nice, this looks great. Could you add some helper texts and tests? :)

internal/clihelpers/printing.go
```diff
@@ -0,0 +1,143 @@
...
+		return formatJSONPretty
+	}
+
+	// output is a terminal, default is table
+	if info, _ := os.Stdout.Stat(); info.Mode()&os.ModeCharDevice != 0 {
+		return formatTable
+	}
+
+	// output is not a terminal, default is quiet
+	return formatQuiet
```
zepatrik, line 128: Do you think this is a good idea? I mean it is handy but maybe not intuitive, because this behavior is not common.

zepatrik: @aeneasr

zepatrik:
There is still the open question on how to best test this. I think starting a kratos with the in memory DB and then using the same binary to execute all the commands against itself would be the best idea. What do you think @aeneasr ?

cmd/identities/import.go
```diff
@@ -1,13 +1,108 @@
...
+	"github.com/ory/kratos/internal/httpclient/models"
 )
 
 // importCmd represents the import command
 var importCmd = &cobra.Command{
-	Use: "import <file.json [file-2.json [file-3.json] ...]>",
-	Run: client.NewIdentityClient().Import,
+	Use:   "import <file.json [file-2.json [file-3.json] ...]>",
+	Short: "Import identities from files or STD_IN",
+	Long: `Import identities from files or STD_IN:
```
aeneasr, line 24: I think the import command should per default be able to import lists of identities. I think it should be possible to detect wether it's a single identity (object) or an array of identities using `gjson`. I assume most people want to import more than one identity.

cmd/identities/validate.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/ory/jsonschema/v3"
+	"github.com/ory/kratos/cmd/cliclient"
+	"github.com/ory/kratos/internal/httpclient/client"
+	"github.com/ory/kratos/internal/httpclient/client/common"
+	"github.com/ory/x/viperx"
+)
+
+var validateCmd = &cobra.Command{
+	Use:   "validate <file.json [file-2.json [file-3.json] ...]>",
+	Short: "Validate local identity files",
```
aeneasr, line 26: Similarly to import, this should also support arrays. Please also add a description to make it a bit easier to understand what this command does. Maybe with an example output!

cmd/cliclient/identity.go
```diff
@@ -1,4 +1,4 @@
-package client
+package cliclient
```
aeneasr, line 1: This file can be removed!

aeneasr (commented):
This looks amazing! Checked it out locally, fixed a couple of descriptions but nothing big. I have two more ideas for improvement (see comments) but this is basically ready for merge IMO!

cmd/identities/import.go
```diff
@@ -20,8 +20,16 @@ import (
...
 var importCmd = &cobra.Command{
 	Use:   "import <file.json [file-2.json [file-3.json] ...]>",
-	Short: "import identities from files or STD_IN",
-	Long:  "Import identities from files or STD_IN. Files are expected to each contain a single identity. The validity of files can be tested beforehand using `... identities validate`. Importing credentials is not yet supported.",
+	Short: "Import identities from files or STD_IN",
+	Long: `Import identities from files or STD_IN:
+
+	kratos identities import file.json
+
+	kratos identities import <(cat file.json)
```
zepatrik, line 28: JSON via argument is not supported.
```suggestion
	cat file.json | kratos identities import
```

aeneasr:
Can you take a look at the build failures?

zepatrik:
all checks :heavy_check_mark: @aeneasr

aeneasr (approved):

PR #698, opened 2020-09-21
fix: keep HTML form type on registration error

### LINKED ISSUE #670: Email field type changes on second request for request context during registration flow
**Describe the bug**

The field type appears to change for the 'traits.email' field after the initial attempt to register an account is unsuccessful. The field starts of with a type of 'email' and then changes to having a type of 'text'. In my application I'm using the type value to make decisions about how to render input fields. The 'kratos-selfservice-ui-node' application appears to be not affected by this as it switches on the field name for making rendering decisions. 

**Reproducing the bug**

Using the same default schema as the kratos-selfservice-ui-node, the first time you start the registration flow and make a call to fetch the request context the form field descriptions for the password method come back as follows (note the type for 'traits.email' is 'email'):

```
0: {name: "csrf_token", type: "hidden", required: true, value: "<redacted>"}
1: {name: "password", type: "password", required: true}
2: {name: "traits.email", type: "email"}
```

If the user submits a request and is unsuccessful (due to using a common password as an example) and is redirected back to the registration form, the second request to fetch the request context then looks like the following (note the type for 'traits.email' is now 'text'):

```
0: {name: "csrf_token", type: "hidden", required: true, value: "<redacted>"}
1: {name: "password", type: "password", required: true, messages: Array(1)}
2: {name: "traits.email", type: "text", value: "q@q.com"}
```

**Expected behavior**

I think it is reasonable to expect the 'type' of a field to remain consistent across calls to retrieve a request context with the same request id. 

**Environment**

- Version: 0.4.6.alpha
- Environment: Docker Desktop on macOS Mojave

**Additional context**

N/A

zepatrik:
## Related issue
closes #670

## Proposed changes
Only set the value and not the whole field when generating the registration form with error messages.

### DIFF
--- a/selfservice/strategy/password/stub/profile.schema.json
+++ b/selfservice/strategy/password/stub/profile.schema.json
@@ -7,17 +7,18 @@
     "traits": {
       "type": "object",
       "properties": {
-    "email": {
-      "type": "string",
-      "ory.sh/kratos": {
-        "credentials": {
-          "password": {
-            "identifier": true
+        "email": {
+          "type": "string",
+          "format": "email",
+          "ory.sh/kratos": {
+            "credentials": {
+              "password": {
+                "identifier": true
+              }
+            }
           }
         }
       }
-    }
-    }
     }
   },
   "additionalProperties": false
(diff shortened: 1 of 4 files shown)


selfservice/strategy/password/stub/profile.schema.json
```diff
@@ -7,18 +7,19 @@
...
+              "password": {
+                "identifier": true
+              }
+            }
           }
         }
       }
     }
-    }
-    }
```
zepatrik, line 21: all of this is only formatting

aeneasr (approved):

PR #714, opened 2020-09-30
ci: bump closed reference notifier

zepatrik:


zepatrik:
I canceled CCI manually because this PR is only about GH actions. It was actually a bug with the action that resulted in the ignore argument to not work in CI but with the cli.... whatever, it works now.

zepatrik:
going to fix this with https://github.com/ory/meta/pull/29

PR #742, opened 2020-10-06
docs: Extract Prettier configuration file

kevgo:
## Related issue

no related issue

## Proposed changes

Extract the Prettier configuration into a dedicated configuration file. Advantages are that the Prettier configuration is now:
- redundancy-free in a single place (later we could even [share the same formatting style across all ORY projects](https://prettier.io/docs/en/configuration.html#sharing-configurations))
- available to [Prettier IDE plugins](https://prettier.io/docs/en/editors.html) (which everybody should use at this point)
- available when scripts or humans run Prettier manually on a file 
- easier to find

I have chosen JSON as the format since it is faster, safer, and I know how much @aeneasr loves YML 😉 

## Checklist

      the changes.
      works.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,5 @@
+{
+  "semi": false,
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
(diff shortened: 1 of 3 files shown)


zepatrik:
See #740 and https://github.com/ory/prettier-styles

kevgo:
Good idea!

PR #567, opened 2020-07-10
docs: Initial set of documentation tests

kevgo:
## Proposed changes

Checks technical parts of the documentation using a documentation testing framework ([Text-Runner](https://github.com/kevgo/text-runner)) and fixes a number of issues found.

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [x] I have added or changed [the documentation](docs/docs).

## Further comments

Keeping it simple for now to provide an idea of the capabilities of the tool. In later stages we can automate larger tutorials as well.

PR #749, opened 2020-10-06
docs: Replace ampersand

kevgo:
## Proposed changes

Replace `&` with `and` since "identity and access proxy" reads better than "identity & access proxy" to me. This is an extract from #567, it makes sense to discuss this separately. Not feeling strongly about this. If the `&` is part of the branding we can leave it in.

## Checklist

      the changes.
      works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ ORY Kratos is the first and only cloud native Identity and User Management Syste
   - [Installation](#installation)
 - [Ecosystem](#ecosystem)
   - [ORY Security Console: Administrative User Interface](#ory-security-console-administrative-user-interface)
-  - [ORY Oathkeeper: Identity & Access Proxy](#ory-oathkeeper-identity--access-proxy)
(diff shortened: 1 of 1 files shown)


kevgo:
Another reason for this change is that the old format leads to anchors containing double-dashes like
`ory-oathkeeper-identity--access-proxy`

aeneasr (changes requested):
Could you please address the merge conflicts? :)

aeneasr:
Also keep in mind that we're using `doctoc` to generate the ToC - not sure if that was what generated the output! Also not sure if this is currently automated...

zepatrik:
Pretty sure it's not automated....

kevgo:
Merge conflicts resolved. 

I was asking about doctoc automation in #748. Happy to look into automating it. Can we assume a Bash environment, i.e. `find` and `xargs`, or does this automation need to run on Windows as well?

aeneasr:
Ah sorry, I must have overlooked that. Yeah - this would run on a ubuntu instance. I think it would make sense to have doctoc execution as part of maybe the docs/build pipeline. @zepatrik what you think?

Issue #711, opened 2020-09-30
Building From Source fails


leadiato:
**Describe the bug**
Building Kratos from source fails when following [Kratos Building from source](https://www.ory.sh/kratos/docs/install/#building-from-source)

**Reproducing the bug**
1. run `go get -d -u github.com/ory/kratos`
2. `cd $(go env GOPATH)/src/github.com/ory/kratos`
3. GO111MODULE=on make install

**Output:**
make: *** No rule to make target '.bin/packr2', needed by 'install'.  Stop.

**Expected behavior**

`$(go env GOPATH)/bin/kratos help` runs successfully and returns proper version number

**Environment**

- Environment: Mac OS 10.14.6
- Make: GNU Make 4.3
- Go: go1.15.2 darwin/amd64

aeneasr:
https://www.ory.sh/kratos/docs/install#building-from-source

aeneasr:
Sorry, looks like you're following the guide!

leadiato:
To fix the issue I had to manually create 2 folders (`.bin/packr2` & `.bin/pack`) prior to start the build.

aeneasr:
Would you be open to contribute that change to the Makefile? :)

chris-rock:
The `make tools` mentioned in the docs is also not working.

aeneasr:
Could you please provide more details? What make version, what is "not working" (error message please), where in the docs - and so on! :) I can only help if sufficient details are provided, otherwise it's a guessing game.

chris-rock:
No problemo. There is a reference to `make tool` in https://github.com/ory/kratos/blob/master/README.md#install-tools It seems that there is no `tools` task in the latest Makefile. With the changes in #750 I got the binary compiled.

aeneasr:
Ok so `make tools` no longer exists - these things now compile automatically (e.g. "golangci-lint" or "goreturns"). Another problem was that we no longer use packr2 so it's not listed as a dependency and thus shouldn't be required!

PR #750, opened 2020-10-07
build: Created required directories during install

leadiato:
## Related issue

#711 

## Describe Issue:

Following [Kratos Building from source](https://www.ory.sh/kratos/docs/install/#building-from-source) instructions to build after cloning the repo fails with the error:
`make: *** No rule to make target '.bin/packr2', needed by 'install'. Stop.`

## Proposed changes

I'm creating the 2 required directories as part of `make install`

## Checklist

      the changes.
      works.
(No documentation added since the documentation is not listing those directories as pre requisite) 

Not sure how to write test for this change (it's my first time working with `make`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,8 @@ mocks: .bin/mockgen
 		mockgen -mock_names Manager=MockLoginExecutorDependencies -package internal -destination internal/hook_login_executor_dependencies.go github.com/ory/kratos/selfservice loginExecutorDependencies
 
 .PHONY: install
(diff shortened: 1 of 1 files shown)


aeneasr (changes requested):
I don't think this is correct. Here you would create two directories, but what we want is to run the two makefile tasks `.bin/packr2` and `./bin/pack` which installs two binaries `packr2` and `packr` into the `.bin` directory. Can you please detail your issue in the original thread? With error outputs etc. Thanks!

Issue #735, opened 2020-10-04
Add possibility to configure the "claims" query parameter in the auth_url of OIDC providers to request individial id_token claims


NickUfer:
**Is your feature request related to a problem? Please describe.**

I tried to configure https://twitch.tv/ as an OIDC provider, but the id_token does not contain the `email` or `email_verified` fields by default which I want to use. These fields are only present when the request to the authorization endpoint contains the `claims` query parameter with the value `{"id_token":{"email":null, "email_verified":null}}`. The `claims` parameter is by default optional, but you have to explicitly set the parameter at twitch to get both fields.

Therefore would it be nice to configure a provider to request specific claims in an id_token.

Specification: https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter

**Describe the solution you'd like**

The URL should contain the `claims` parameter if one is configured. Example: https://id.twitch.tv/oauth2/authorize?claims={"id_token":{"email":null,"email_verified":null}}&client_id=any_client_id&redirect_uri=http://127.0.0.1:7001/redirect&response_type=code&scope=openid+user:read:email&state=any_state

The configuration could simply look like this:

```yaml
strategies:
  oidc:
    enabled: true
    config:
      providers:
        - id: twitch
          provider: generic
          client_id: any_client_id
          client_secret: any_secret
          mapper_url: file:///etc/config/kratos/oidc.twitch.jsonnet
          issuer_url: https://id.twitch.tv/oauth2
          token_url: https://id.twitch.tv/oauth2/token
          scope:
            - openid
            - 'user:read:email' # required in the near future for email and email_verified claims
          requested_claims: '{"id_token":{"email":null,"email_verified":null}}'
```

NickUfer:
I've just looked at the code and it seems to be relatively easy to implement... adding a new field in the `Configuration` struct used by each provider and adding the `claims` `AuthURLParam` in `provider_generic_oidc.go`'s `AuthCodeURLOptions` function when it is not an empty string, correct?

If that's the case or something similar then I could create a PR if that's okay. I'm by the way new to go, trying to learn it (currently on day 4).

aeneasr:
Hey, yeah - this makes sense to me! Is the claims format always an unstructured JSON?

NickUfer:
Yes it is always. The claims can also be different for each provider.

aeneasr:
Ok, then I think the proposal makes sense! Would you be open to contribute this?

NickUfer:
Sure!

[...]

PR #753, opened 2020-10-10
feat: oidc provider claims config option

NickUfer:
## Related issue

https://github.com/ory/kratos/issues/735
@aeneasr 

## Proposed changes

Adds an option to oidc providers to request specific openid claims on an authorization request

## Checklist

      the changes.
      works.


.schema/config.schema.json
```diff
@@ -138,6 +138,15 @@
             "8eaef023-2b34-4da1-9baa-8bc8c9d6a490",
             "contoso.onmicrosoft.com"
           ]
+        },
+        "requested_claims": {
+          "title": "OpenID Connect claims",
+          "description": "The OpenID Connect claims and optionally their properties which should be included in the id_token or returned from the UserInfo Endpoint.",
+          "type": "string",
```
zepatrik, line 145: I would like to have this as an object and validate it further. That ensures at least it is valid JSON.

selfservice/strategy/oidc/provider_config.go
```diff
@@ -53,6 +53,12 @@ type Configuration struct {
 	//
 	// It can be either a URL (file://, http(s)://, base64://) or an inline JSONNet code snippet.
 	Mapper string `json:"mapper_url"`
+
+	// RequestedClaims string encoded json object that specifies claims and optionally their properties which should be
+	// included in the id_token or returned from the UserInfo Endpoint.
+	//
+	// More information: https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter
+	RequestedClaims string `json:"requested_claims"`
```
zepatrik, line 61:
```suggestion
	RequestedClaims json.RawMessage `json:"requested_claims"`
```

selfservice/strategy/oidc/provider_generic_test.go
```diff
@@ -12,16 +12,21 @@ import (
 	"github.com/ory/kratos/x"
 )
 
+func makeOIDCClaims() string {
+	return "{\"id_token\":{\"email\":{\"essential\":true},\"email_verified\":{\"essential\":true}}}"
```
zepatrik, line 16: It would be less error prone and better readable if you encode a `map[string]interface{]` here instead of having a json blob. Something like:
```suggestion
func makeOIDCClaims(t *testing.T) string {
    claims, err := json.Marshal(map[string]interface{}{
        "id_token": map[string]interface{}{...},
    }
	return claims
```

zepatrik (changes requested):
Thanks for your contribution :wink: 
I think it would be better to have the `requested_claims` as a JSON object. Is there some way to validate them further or are they proprietary?

NickUfer:
> Thanks for your contribution 
> I think it would be better to have the `requested_claims` as a JSON object. Is there some way to validate them further or are they proprietary?

The [spec](https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter) defines that on the top level can only be a `userinfo` and / or `id_token` JSON object. Once inside `userinfo` or `id_token` the keys are pretty much proprietary but the values are constrained. The values could be:
* `null`
* another JSON object with the keys:
     * `essential` which indicates if this claim is, well, essential to the requester. Value can only be a boolean.
     * `value` which indicates that the value in the id_token or the userinfo endpoint for this field must have that exact value
     * or `values` which is the same as `value` but only with multiple values

Example claims request:
```
  {
   "userinfo":
    {
     "given_name": {"essential": true},
     "nickname": null,
     "email": {"essential": true},
     "email_verified": {"essential": true},
     "picture": null,
     "http://example.info/claims/groups": null
    },
   "id_token":
    {
     "auth_time": {"essential": true},
     "acr": {"values": ["urn:mace:incommon:iap:silver"] }
    }
  }
```

If it is possible to define this in the schema without concrete key names then this should be possible.

selfservice/strategy/oidc/provider_config.go
NickUfer: Yeah, that's much better. I was also thinking about how I could validate that the parameter is indeed a JSON object, but my current Go knowledge hindered me from doing so :smile:

selfservice/strategy/oidc/provider_generic_test.go
NickUfer: Added in new commit

zepatrik:
Right, JSON schema can get a bit complicated. Here is my take:

```json
{
      "additionalProperties": false,
      "patternProperties": {
        "^userinfo$|^id_token$": {
          "type": "object",
          "additionalProperties": false,
          "patternProperties": {
            ".*": {
              "oneOf": [
                {
                  "const": null,
                  "description": "Indicates that this Claim is being requested in the default manner"
                },
                {
                  "allOf": [
                    {
                      "propertyNames": {
                        "$comment": "This is to implement 'additionalProperties: false'",
                        "enum": [
                          "essential",
                          "value",
                          "values"
                        ]
                      }
                    },
                    {
                      "properties": {
                        "essential": {
                          "description": "Indicates whether the Claim being requested is an Essential Claim",
                          "type": "boolean"
                        }
                      }
                    },
                    {
                      "$comment": "Although not required by the spec, it makes no sense to allow both at the same time",
                      "oneOf": [
                        {
                          "properties": {
                            "value": {
                              "description": "Requests that the Claim be returned with a particular value",
                              "$comment": "There seem to be no constrains on value"
                            }
                          }
                        },
                        {
                          "properties": {
                            "values": {
                              "description": "Requests that the Claim be returned with one of a set of values, with the values appearing in order of preference",
                              "type": "array",
                              "items": {
                                "$comment": "There seem to be no constrains on individual items"
                              }
                            }
                          }
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          }
        }
      }
    }
```

Maybe you can check if it works as expected by adding schema tests for it :wink: 
Have a look in `test/schema` and define above schema under the definition key. If you need help with that, let me know.
And maybe you can also improve the descriptions/add titles? I just copied the spec which is not super understandable.
Thanks a lot :wink:

zepatrik:
When https://github.com/ory/kratos/pull/757 is merged and you rebase, there should also be no timeout any more in the CI

zepatrik:
When everything passes, this looks good to me :+1:

NickUfer:
I would quickly merge [this PR](https://github.com/NickUfer/kratos/pull/1) from @zepatrik then I would be ready to merge too

.schema/config.schema.json
NickUfer: Done

selfservice/strategy/oidc/provider_config.go
NickUfer: Done.

zepatrik (approved):

aeneasr:
Awesome 🎉 

PR #766, opened 2020-10-14
fix: adds missing gitlab to list of supported oidc providers

NickUfer:
## Proposed changes

Adds gitlab to list of supported providers in error message.

## Checklist

      the changes.
      works.

### DIFF
--- a/selfservice/strategy/oidc/provider_config.go
+++ b/selfservice/strategy/oidc/provider_config.go
@@ -79,19 +79,25 @@ func (c ConfigurationCollection) Provider(id string, public *url.URL) (Provider,
 	for k := range c.Providers {
 		p := c.Providers[k]
 		if p.ID == id {
+			var providerNames []string
+			var addProviderName = func(pn string) string {
+				providerNames = append(providerNames, pn)
+				return pn
+			}
+
 			switch p.Provider {
-			case "generic":
+			case addProviderName("generic"):
 				return NewProviderGenericOIDC(&p, public), nil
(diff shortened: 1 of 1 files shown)


zepatrik:
Nice catch, could you automate that list generation though? That way it will be always up to date.
My suggestion would be to use a pattern similar to this: https://play.golang.org/p/ktNUWwnWb9R

zepatrik:
Sorry for the crude example but I hope it explains the idea.

zepatrik:
You can omit the custom type though and just go with strings, that should be fine.

zepatrik:
https://github.com/ory/kratos/blob/master/selfservice/strategy/oidc/provider_config.go#L19-L21 is also out of date, sorry for all the spam.

NickUfer:
> Nice catch, could you automate that list generation though? That way it will be always up to date.
> My suggestion would be to use a pattern similar to this: https://play.golang.org/p/ktNUWwnWb9R

Sure. I also thought about refactoring this but my second thought was maybe it's on purpose for simplicity or something.

zepatrik:
I guess it was because of laziness by @aeneasr :wink:

NickUfer:
I hope this is not too over engineered

zepatrik:
Looking good, one other idea I had was to use the [python switch approach](https://stackoverflow.com/a/11479840). You would have to iterate over the map but I think this is nicer 😉

zepatrik (approved):

PR #773, opened 2020-10-16
docs: improve schemas

zepatrik:
## Proposed Changes

- [x] weird version example in the documentation
- [x] remove pre release from version schema
- [x] add type to `session.cookie.domain`


aeneasr (changes requested):
Schema tests fail :D

PR #815, opened 2020-11-08
fix: convert all identifiers to lower case on login

### LINKED ISSUE #814: Unable to login with username identifier if username has upper case letters
**Describe the bug**

Users are unable to login with an identifier which is not an email and which has upper case letters at registration. They would have to enter their identifier in lowercase letters to login.

**Reproducing the bug**

* Define another identifier which is not an email. For this example use "username"
* Create an account with a username with upper case letters, for example "TestAccount"
* Try to login with "TestAccount" as the identifier. This will not work, but "testaccount" will

Identity Schema:
```json
{
  "$id": "https://schemas.ory.sh/presets/kratos/quickstart/email-password/identity.schema.json",
  "title": "Person",
  "type": "object",
  "properties": {
    "traits": {
      "type": "object",
      "properties": {
        "email": {
          "type": "string",
          "format": "email",
          "title": "E-Mail",
          "minLength": 3,
          "ory.sh/kratos": {
            "credentials": {
              "password": {
                "identifier": true
              }
            },
            "verification": {
              "via": "email"
            },
            "recovery": {
              "via": "email"
            }
          }
        },
        "username": {
          "type": "string",
          "title": "Username",
          "minLength": 3,
          "ory.sh/kratos": {
            "credentials": {
              "password": {
                "identifier": true
              }
            }
          }
        }
      },
      "required": [
        "email",
        "username"
      ],
      "additionalProperties": false
    }
  }
}

```

**Expected behavior**

Users should be able to login with all their identifiers regardless of case.

**Environment**

- Version: ddcffff6806d3aae9b6fe090fab842b81facb10d
- Environment: Ubuntu, Docker

NickUfer:
## Related issue

Closes #814 

## Proposed changes

On registration all identifiers are transformed to lowercase before inserting them into the database. This happens while validation: https://github.com/ory/kratos/blob/d32413a1fcd0ce1a82d2529f18b5d4334a490a2a/identity/extension_credentials.go#L38-L40
This PR chances the code to not only transform all emails but every identifier to lowercase before searching for them in the database.

## Checklist

      the changes.
      works.


aeneasr:
I am not sure if we really want to merge this - what if the identifier is a randomly generated string? How do other systems behave here?

aeneasr:
Oh and sorry, thank you for the contrib of course :)

zepatrik:
:thinking: and reading a bit about this, in theory only the domain of the email address is case insensitive. I could not find any email provider supporting differentiation by casing though. There is also always this potential security issue when dealing with casing: https://eng.getwisdom.io/hacking-github-with-unicode-dotless-i/

zepatrik:
Maybe it would be best to make this configurable? So add another key to our schema extension, e.g. 
```json
{
        "username": {
          "type": "string",
          "title": "Username",
          "minLength": 3,
          "ory.sh/kratos": {
            "caseSensitive": true,
            "credentials": {
              "password": {
                "identifier": true
              }
            }
          }
        }
}
```

aeneasr:
Yeah I guess a setting would make most sense here!

NickUfer:
> thinking and reading a bit about this, in theory only the domain of the email address is case insensitive. I could not find any email provider supporting differentiation by casing though. There is also always this potential security issue when dealing with casing: https://eng.getwisdom.io/hacking-github-with-unicode-dotless-i/

As every identifier is converted to lowercase at registration and is unique, there could not be the case that someone registers with for example `Tasse@example.com` und `taße@example.com`, as they both would be, according to the article, the identifier `tasse@example.com` which would not be possible.

> I am not sure if we really want to merge this - what if the identifier is a randomly generated string? How do other systems behave here?

As it is an identifier it would internally be saved in the lowercase version. So it would not be possible to have this random string in different case variations.

> Maybe it would be best to make this configurable? So add another key to our schema extension, e.g.
> 
> ```json
> {
>         "username": {
>           "type": "string",
>           "title": "Username",
>           "minLength": 3,
>           "ory.sh/kratos": {
>             "caseSensitive": true,
>             "credentials": {
>               "password": {
>                 "identifier": true
>               }
>             }
>           }
>         }
> }
> ```

With this solution the problem would still exist as an identifier is something you, by my understanding, always use for login. I actually don't know a platform which doesn't treat their identifiers as case insensitive and where you wouldn't be able to login with `testaccount` when your real name is `TestAccount`.

zepatrik:
Ok, I get your point. The article I posted was just a reminder that we have to lowercase **everywhere**. I think it would make sense to always lowercase identifiers, so this change would be good IMO.

NickUfer:
This change only helps to find the correct identifier in the database. It is virtually impossible to find the wrong identity, because of the already forced uniqueness by lowercasing all identifiers at registration:

We could actually omit these changes:
https://github.com/ory/kratos/blob/d52a6562737355178515dd103d8043650190fbb3/persistence/sql/persister_identity.go#L121-L125

As they are still applied here at validation and before persisting:
https://github.com/ory/kratos/blob/d32413a1fcd0ce1a82d2529f18b5d4334a490a2a/identity/extension_credentials.go#L38-L40

aeneasr:
Good points @NickUfer - I did not remember that we actually lowercase all of them. I think it would make sense to add a test or two (or rather failing test cases to prevent regression) to ensure that this behavior works correct. If you need help identifying the right places let me know!

NickUfer:
@aeneasr yeah I think I need some guidance here for the tests :D

And another thing, should I remove this:

> We could actually omit these changes in function `createIdentityCredentials`:
> https://github.com/ory/kratos/blob/d52a6562737355178515dd103d8043650190fbb3/persistence/sql/persister_identity.go#L121-L125

as this is also always done at:
https://github.com/ory/kratos/blob/d32413a1fcd0ce1a82d2529f18b5d4334a490a2a/identity/extension_credentials.go#L38-L40

The only added value of keeping this in is more clarity that all identifiers are transformed to lowercase. I had to search some time to find out that this transformation to lowercase is happening in the validation stage.

aeneasr:
Sorry @NickUfer, somehow your reply missed my radar.

> The only added value of keeping this in is more clarity that all identifiers are transformed to lowercase. I had to search some time to find out that this transformation to lowercase is happening in the validation stage. I only found out about this by checking the database and then searching in the code for it.

Let's keep it in as it is just safer :)

So for the regression tests it would probably make sense to just add tests to https://github.com/ory/kratos/blob/d52a6562737355178515dd103d8043650190fbb3/identity/pool.go#L90 and https://github.com/ory/kratos/blob/d32413a1fcd0ce1a82d2529f18b5d4334a490a2a/identity/extension_credentials_test.go

You could then also add tests integration tests similar to this one https://github.com/ory/kratos/blob/master/selfservice/strategy/password/login_test.go#L368 where you register with an upper-case email/username and log in with lowercase representation. Similarly you could add such a test to https://github.com/ory/kratos/blob/master/selfservice/strategy/password/registration_test.go as well.

Similarly, you could add a similar test to the cypress e2e tests: https://github.com/ory/kratos/blob/master/test/e2e/cypress/integration/profiles/email/login/success.spec.js

aeneasr:
Are you still up for the changes? :) If you need any help, let us know!

NickUfer:
Sure we can still change stuff :) I'm just very busy with all sorts of things right now

aeneasr (approved):
Awesome 🎉 Thank you for the changes - this looks very good!

PR #833, opened 2020-11-17
docs: update oidc provider with json conf support

bmonkman:
Also bump x lib version, add test for json in config, add support for showing env var blocks for specific objects in the config schema.

## Related issue

@zepatrik re: the conversation in https://github.com/ory/x/pull/232

## Proposed changes

Updated docs to reflect the changes made to config handling in the x library.
Since objects can now be provided via env vars as JSON I also added support in the docs to add the env var block for specific items by adding `"showEnvVarBlock": true`
The other option would be to just remove the check that is preventing that block from being added for all objects, but that seems like it would add a lot of verbosity to the logs, where in most cases there would be limited use for someone to use this new functionality. I'm specifically calling out the oidc section because it contains secret values and in a k8s environment it would be beneficial to provide these as an environment variable.
Happy to change that, though, if you all had a different vision for it.

Also Please let me know if that test I added is in the wrong location.

## Checklist

      the changes.
      works.


selfservice/strategy/oidc/strategy_test.go
```diff
@@ -484,6 +484,42 @@ func TestStrategy(t *testing.T) {
 		actual := sr.Methods[identity.CredentialsTypeOIDC]
 		assert.EqualValues(t, expected.Config.FlowMethodConfigurator.(*oidc.FlowMethod).HTMLForm, actual.Config.FlowMethodConfigurator.(*oidc.FlowMethod).HTMLForm)
 	})
+
```
zepatrik, line 487: This test should be located in `driver/configuration/provider_viper_test.go` :wink:

docs/scripts/config.js
```diff
@@ -87,7 +87,8 @@ const enhance = (schema, parents = []) => (item) => {
     })
   }
 
-  if (!hasChildren) {
+  const showEnvVarBlock = pathOr('', [...path, 'showEnvVarBlock'], schema)
+  if (!hasChildren || showEnvVarBlock) {
```
zepatrik, line 91: I like this, but can you add it here: https://github.com/ory/docusaurus-template/blob/master/scripts/config.js ? 
Sorry for all these separate PRs, but it will then be available in all projects :wink:

zepatrik: Maybe we can call the key `showEnvVarRawJSON` and also add the text you specified in the description?

zepatrik (changes requested):
Awesome, thank you :tada: 
I'm really sorry for all these separate PRs :sweat_smile:

selfservice/strategy/oidc/strategy_test.go
bmonkman: Moved and rewrote that test.

docs/scripts/config.js
bmonkman: Made the name a bit clearer, though I figured `showEnvVarBlockForObject` would be a bit clearer because it's about the type of the config option rather than the content of the value. I also changed the logic to automatically show the text if it's an object and this field is set. (Ported this change back from my PR in docusaurus. Not sure if that's necessary but I figured I'd just keep them in sync.

zepatrik: I will ping you once everything propagated to master

zepatrik: Don't expect it today though, I will go to bed soon :sweat_smile: 
When CI passed in https://github.com/ory/kratos/pull/836 and it is merged, we can merge yours. I will update your branch in the morning, as you are probably behind us a view hours.

bmonkman: Thanks, that's no problem!

zepatrik: Ok merged, if you want you can rebase. Also, can you maybe run the config docs gen script and commit the updated file?

bmonkman: There seem to be quite a few other changes as well. Should I go ahead and commit them here or do you typically merge a branch in and then regenerate on `master`?

zepatrik: I just merged it and ran `cd docs; npm run gen` so we can see your change in action :wink:

zepatrik:
Lets also wait for https://github.com/ory/docusaurus-template/pull/57 to propagate, so we can have a nice diff in this PR :wink:

zepatrik (approved):

aeneasr (approved):
Awesome 🎉 

PR #930, opened 2020-12-21
fix: use csrf none mode only when --dev is enabled

aeneasr:
Do not merge this yet. `SameSite` should be `Strict` for CSRF unless the TLD of ORY Kratos and the TLD of the UI do not share a domain. Instead, we should probably allow customizing the CSRF cookie:

- Allow setting the domain and path of the CSRF cookie
- Allow setting samesite of the CSRF cookie
- Fail setting `SameSite=None` when `--dev` is used.

### DIFF
--- a/docs/sidebar.json
+++ b/docs/sidebar.json
@@ -174,6 +174,31 @@
         "cli/kratos-serve", 
         "cli/kratos-version"
       ]
+    }, 
+    {
+      "items": [
+        "cli/kratos", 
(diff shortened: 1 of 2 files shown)


aeneasr:
As discussed on slack, removing the `--dev` flag will set `SameSite=None` which allows cross-domain deployments. Setting `None` is ok because the CSRF value can only be extracted from the Flow ID, which is only available to the UI, which is explicitly set in the ORY Kratos config.

Issue #938, opened 2020-12-24
Swagger specification for security is wrong


dmayle:
**Describe the bug**
The swagger specification for security is incomplete.  It only shows one of the three supported authentication mechanisms.

**Reproducing the bug**
https://github.com/ory/kratos/blob/master/session/manager_http.go#L98 Shows the code that checks security

**Expected behavior**
The swagger spec is correct

aeneasr:
Unfortunately this is an upstream/cantfix issue: https://github.com/ory/kratos/pull/939#discussion_r548850163

PR #939, opened 2020-12-24
docs: Update security mechanism specification to match code

dmayle:
As seen in https://github.com/ory/kratos/blob/master/session/manager_http.go#L98 the session token can be extracted in one of three different ways from the request.  This change updates the swagger definition for security to show all three possibilities.

## Related issue
#938 

## Proposed changes
After this, the security sections of the various apis will need to be updated to support the three mechanisms

## Checklist
      the changes.
      works.

## Further comments
This matters because of things like auto-generated rest clients based on the swagger specification

### DIFF
--- a/swagger_meta.go
+++ b/swagger_meta.go
@@ -19,6 +19,14 @@
 //          type: apiKey
 //          name: X-Session-Token
 //          in: header
+//     sessionBearerToken:
+//          type: apiKey
+//          name: Authorization
+//          in: header
+//     sessionCookie:
+//          type: apiKey
+//          name: Cookie
+//          in: header
 //
 //     Extensions:
 //     ---


swagger_meta.go
```diff
@@ -19,6 +19,14 @@
 //          type: apiKey
 //          name: X-Session-Token
 //          in: header
+//     sessionBearerToken:
+//          type: apiKey
+//          name: Authorization
+//          in: header
+//     sessionCookie:
+//          type: apiKey
```
aeneasr, line 27: Thank you for the submission! Unfortunately, this notation does not work as it sends `Cookie: the-session-token` while the correct notation is `Cookie: key=value;` or rather `Cookie: ory_kratos_session=...;`. In my research, I could not find a possibility to set this using Swagger 2.0. Session cookies are only supported starting with Swagger 3.0 (https://swagger.io/docs/specification/authentication/cookie-authentication/) but go-swagger has no plants of moving to 3.0 unfortunately: https://github.com/go-swagger/go-swagger/issues/1122#issuecomment-575968499

aeneasr (changes requested):

PR #961, opened 2021-01-05
feat: hasher load testing CLI

### LINKED ISSUE #955: Improve Argon2 helpers and defaults and documentation

* [ ] The argon2 choose should start with much less memory. In Stackoverflow and others I regularly see 128MB or 64MB as the default with 3+ iterations. It's also a good idea to check out libsodium argon2 defaults or other software which uses argon2 and see where their parameters are at.
* [ ] CLI argon2 helper should
  * [ ] start with less memory (I remember that we can't get the system memory, correct? otherwise half of system memory would be a good starting point)
  * [ ] should have a `--max-concurrent` argument which defines the maximum amount of concurrent password hashing processes. you could use go routines to dispatch the amount of concurrent subroutines. since this is the maximum, there should also be something which checks for the low and average or median amount of concurrent users. the deviation of the target time should not be too high for low, mid, high concurrent users. in general, I think the system memory / 2 (to leave room for other processes) / amount of concurrent users should be the maximum memory allowed to be requested.
  * [ ] is it possible to detect OOM errors in golang? if so, argon2 should not crash the process when we run out of memory. not sure if that is possible though but I saw that libsodium can do that
* [ ] Update the blog post with the new learnings
* [ ] **config.schema.json** needs a description for the config fields and sane defaults:
```

    "hashers": {
      "type": "object",
      "properties": {
        "argon2": {
          "title": "Configuration for the Argon2id hasher.",
          "type": "object",
          "properties": {
            "memory": {
              "title": "Memory consumption in bytes",
              "description": "Sets the memory (RAM) consumption in bytes. Minimum value is 16MB (16384 / 1024 = 16) and default value is 64MB. For more information check out https://www.ory.sh/kratos/docs/debug/performance-out-of-memory-password-hashing-argon2",
              "type": "integer",
              "minimum": 16384,
              "default": 65536
            },
            "iterations": {
              "type": "integer",
              "minimum": 1,
              "default": 3
            },
            "parallelism": {
              "type": "integer",
              "minimum": 1
            },
            "salt_length": {
              "type": "integer",
              "minimum": 16,
              "default": 32
            },
            "key_length": {
              "type": "integer",
              "minimum": 16,
              "default": 32
            }
          },
          "additionalProperties": false
        }
      },
      "additionalProperties": false
    },
```
* [ ] https://github.com/ory/kratos/blob/master/driver/config/provider_viper.go#L87 these defaults are insane (4GB RAM per default? no wonder all the k8s deployments crash!)
* [ ] We need to add some more details to https://github.com/ory/kratos/blob/master/docs/docs/debug/performance-out-of-memory-password-hashing-argon2.md or write a docs guide for choosing the argon2 arguments and have the debug document for possible error debugging (a bit duplicate but might help people find stuff).  
Here's what I wrote:
````

## Manual Configuration

You may also choose the Argon2 parameters manually.

:::note

Please keep in mind that your host machine is probably doing more than just
computing Argon2 hashes, so choose these parameters wisely. It is also important
that ORY Kratos will probably compute several hashes in parallel, depending
on how many concurrent logins or registrations you have.

:::

To configure Argon2, edit the ORY Kratos configuration file:

```yaml title="path/to/my/kratos/config.yml"
# $ kratos -c path/to/my/kratos/config.yml serve
hashers:
  argon2:
    # Memory consumption per hashing process denoted in bytes.
    # Here: 131072 / 1024 = 128MB
    memory: 131072

    # We recommend not choosing these parameters manually but using
    # the Argon2 calibration command.
    parallelism: 1
    iterations: 3
```
````

A clear and concise description of any alternative solutions or features you've
considered.

**Additional context**

### PR DESCRIPTION
**Pull request by zepatrik**
## Related issue

Closes #955

## Proposed changes

This adds a load testing CLI that allows to adjust the hasher parameters under simulated load.

## Checklist

      the changes.
      works.

### CODE REVIEW

File: cmd/hashers/argon2/loadtest.go [hunk 1]
```diff
@@ -0,0 +1,257 @@
...
+func (r *resultTable) Interface() interface{} {
+	return r
+}
+
+func newLoadTestCmd() *cobra.Command {
+	flagConf := &argon2Config{}
+
+	cmd := &cobra.Command{
+		Use:  "load-test <authentication-requests-per-minute>",
+		Args: cobra.ExactArgs(1),
```
aeneasr, line 64: This is missing a description. The output is also not very clear. What parameters are being used here?

```
 $GOPATH/bin/kratos hashers argon2 load-test 30 
It takes about 20s to collect all necessary data, please be patient.
TOTAL SAMPLE TIME	16.406087875s	
MEDIAN			207.022782ms	
STANDARD DEVIATION	64.05794ms	
MIN			145.107023ms	
MAX			344.1785ms	
MEMORY USED		1.51GB	
```

cmd/hashers/argon2/root.go
```diff
@@ -1,6 +1,31 @@
...
+)
+
+const (
+	FlagIterations        = "iterations"
+	FlagParallelism       = "parallelism"
+	FlagSaltLength        = "salt-length"
+	FlagKeyLength         = "key-length"
+	FlagMemory            = "memory"
+	FlagDedicatedMemory   = "dedicated-memory"
+	FlagMinimalDuration   = "minimal-duration"
```
aeneasr, line 26: I think this can be abbreviated to `min-duration`.

driver/config/provider_viper.go
```diff
@@ -82,22 +84,31 @@ const (
...
 	ViperKeyHasherArgon2ConfigKeyLength                             = "hashers.argon2.key_length"
+	ViperKeyHasherArgon2ConfigMinimalDuration                       = "hashers.argon2.minimal_duration"
+	ViperKeyHasherArgon2ConfigExpectedDeviation                     = "hashers.argon2.expected_deviation"
+	ViperKeyHasherArgon2ConfigDedicatedMemory                       = "hashers.argon2.dedicated_memory"
 	ViperKeyPasswordMaxBreaches                                     = "password.max_breaches"
 	ViperKeyIgnoreNetworkErrors                                     = "password.ignore_network_errors"
 	ViperKeyVersion                                                 = "version"
-	Argon2DefaultMemory                                      uint32 = 4 * 1024 * 1024
-	Argon2DefaultIterations                                  uint32 = 4
+	Argon2DefaultMemory                                             = uint32(512 * bytesize.MB / bytesize.KB)
```
aeneasr, line 93: I think the default should be 64MB or 128MB as most instances of Kratos are deployed in a docker container with ~1GB of ram available.

```suggestion
	Argon2DefaultMemory                                             = uint32(128 * bytesize.MB / bytesize.KB)
```

driver/config/provider_viper.go
```diff
@@ -82,22 +84,31 @@ const (
...
+	Argon2DefaultIterations                                  uint32 = 1
 	Argon2DefaultSaltLength                                  uint32 = 16
 	Argon2DefaultKeyLength                                   uint32 = 32
+	Argon2DefaultDuration                                           = 500 * time.Millisecond
+	Argon2DefaultDeviation                                          = 500 * time.Millisecond
+	Argon2DefaultDedicatedMemory                                    = 4 * bytesize.GB
```
aeneasr, line 99: I would set this to 1:

```suggestion
	Argon2DefaultDedicatedMemory                                    = 1 * bytesize.GB
```

driver/config/provider_viper.go
```diff
@@ -82,22 +84,31 @@ const (
...
 	ViperKeyPasswordMaxBreaches                                     = "password.max_breaches"
 	ViperKeyIgnoreNetworkErrors                                     = "password.ignore_network_errors"
 	ViperKeyVersion                                                 = "version"
-	Argon2DefaultMemory                                      uint32 = 4 * 1024 * 1024
-	Argon2DefaultIterations                                  uint32 = 4
+	Argon2DefaultMemory                                             = uint32(512 * bytesize.MB / bytesize.KB)
+	Argon2DefaultIterations                                  uint32 = 1
 	Argon2DefaultSaltLength                                  uint32 = 16
 	Argon2DefaultKeyLength                                   uint32 = 32
+	Argon2DefaultDuration                                           = 500 * time.Millisecond
```
aeneasr, line 97: I think 500ms is too long, it should be more in the 200 range:

```suggestion
	Argon2DefaultDuration                                           = 200 * time.Millisecond
```

.schema/config.schema.json
```diff
@@ -1095,6 +1095,24 @@
...
+              "description": "The time a hashing operation (~login latency) should take.",
+              "type": "string",
+              "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
+              "default": "500ms"
+            },
+            "expected_deviation": {
+              "description": "The standard deviation expected for hashing operations. If this value is exceeded you will be warned in the logs to adjust the parameters.",
+              "type": "string",
+              "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
+              "default": "500ms"
```
aeneasr, line 1109: Please update the defaults if my suggested changes are accepted.

.schema/config.schema.json
```diff
@@ -1095,6 +1095,24 @@
             "key_length": {
```
aeneasr, line 1095: Please update the other defaults as well

aeneasr (changes requested):

driver/config/provider_viper.go
zepatrik: Here the question remains: should we provide defaults that are secure or that will definitely work in every environment? Are such small docker containers considered for production environments? In a dev environment you will likely not have concurrent login requests. I went for 512MB as it should be always available (for non-concurrent settings at least).
Libsodium uses 256MB as the `moderate` level. Their [guidelines](https://libsodium.gitbook.io/doc/password_hashing/default_phf#guidelines-for-choosing-the-parameters) are not very helpful as well and contradict the argon2 paper (recommend at least 3 iterations).
In my opinion the default should work with dev environments and be high enough to guarantee security. It is not possible to find good default values that work for everyone IMO. People will always have to calibrate and test different values for their requirements.

driver/config/provider_viper.go
zepatrik: This parameter is the point where we will queue hashing operations and warn in the logs when it is reached, not a hard limit. Same considerations as above.

driver/config/provider_viper.go
zepatrik: Libsodium (and most other sources) recommend up to 1s. This only affects login requests, so I wanted to stick with that (median 500ms + acceptable deviation 500ms => 0.5-1s).

cmd/hashers/argon2/loadtest.go
zepatrik: The ones defined in the config file/defaults/flags/... Just as kratos would use them if you would run `kratos serve`
Should they be printed as well, just to indicate and help to spot mistakes?

.schema/config.schema.json
zepatrik: Should we make the memory parameter a string as well (like the dedicated memory)? Nobody will be able to understand it currently without a calculator.

driver/config/provider_viper.go
aeneasr: > Libsodium uses 256MB as the moderate level.

From libsodium:

> For interactive, online operations, crypto_pwhash_OPSLIMIT_INTERACTIVE and crypto_pwhash_MEMLIMIT_INTERACTIVE provide base line for these two parameters. This currently requires 64 MiB of dedicated RAM. Higher values may improve security (see below).
>
> Alternatively, crypto_pwhash_OPSLIMIT_MODERATE and crypto_pwhash_MEMLIMIT_MODERATE can be used. This requires 256 MiB of dedicated RAM, and takes about 0.7 seconds on a 2.8 Ghz Core i7 CPU.

Most docker containers get about 128-256 MB ram and about 0.2-0.5 CPUs unless it's a Java application.

>  Their guidelines are not very helpful as well and contradict the argon2 paper (recommend at least 3 iterations).

This recommendation is for Argon2i, not Argon2id - therefore not contradicting imo:

> opslimit, the number of passes, has to be at least 3 when using Argon2i.crypto_pwhash() and crypto_pwhash_str() will fail with a -1 return code for lower values.

I think it is unrealistic to scale a user system with 512MB ram per password hash. The value used here will be the default value for all users until the admin changes suddenly realizes the cost. It is then not easily possible to retrospectively change the RAM use of the existing hashes, meaning that the "old" hashes still use 512MB. It is therefore sensible to set something that reflects the real world. libsodium, a very popular and well maintained library, sets this to 64MB for online services. If we go with that, we will not be wrong, and we can still do a "bit more security" with 128MB.

Keep in mind that yes, making stuff expensive is the goal of password hashing, but

- noone was ever able to find SHA-3 collisions as far as I know - and Argon2 is aeons more expensive than SHA familiy algorithms (BCrypt, Scrypt, PKBDF2 all are much more expensive).
- security must be economical at scale

driver/config/provider_viper.go
aeneasr: I see, I thought this would be the minimum duration of a password hashing attempt

driver/config/provider_viper.go
aeneasr: I see!

.schema/config.schema.json
aeneasr: YES!!

driver/config/provider_viper.go
zepatrik: The wording can definitely be improved around that, I will check it again.

driver/config/provider_viper.go
zepatrik: Collisions are not the main concern in my opinion (and increasing memory will not prevent collisions as well). It is more about preventing an attacker to bute force the clear text passwords when they happen to get access to the hashes.
I just assumed that production systems are way to different to have a somehow reliable default value. Setting this to a rather high value would force people to actually go and determine sensible values (instead of using defaults). We can also easily add a parameter update functionality in the hasher that recomputes the hash with the new parameters after a successful check and then overwrites it in the database (lazy computation after the request is handled). I will open an issue for that.

zepatrik: The contradiction I was referring to was in

> If this it is way too long for your application, reduce memlimit, but keep opslimit set to 3.

Their [default "interactive" values](https://github.com/jedisct1/libsodium/blob/ae4add868124a32d4e54da10f9cd99240aecc0aa/src/libsodium/include/sodium/crypto_pwhash_argon2id.h#L66-L70) are 64MB, 2 iterations

I would rather use more memory and less iterations (as all other best practices suggest). Therefore 128MB, 1 iteration should be good (and take about the same time).

aeneasr (commented):
Please test the quickstart - I assume it might not be working due to the memory changes. Please also provide a breaking change that follows conventional commit syntax and explain in the breaking change what the user has to do in order to upgrade!

zepatrik:
BREAKING CHANGE: The configuration value for `hashers.argon2.memory` is now a string representation of the memory amount including the unit of measurement. To convert the value divide your current setting (KB) by 1024 to get a result in MB or 1048576 to get a result in GB. Example: `131072` would now become `128MB`.

zepatrik:
Should I update the blog post as well?

Issue #998, opened 2021-01-19
E-mail verification problem


decker502:
**Describe the bug**

Pass the verification through another email address during email verification

**Reproducing the bug**

1. Register with 1@test.com
2. When verifying, fill in the email address 2@test.com, 2@test.com can receive the email
3. After clicking the verification link, 1@test.com can pass the verification.

**Expected behavior**

I think the verification of 1@test.com can only be sent to the mailbox of 1@test.com

**Environment**

- Version: docker-image: v0.5.5-alpha.1-sqlite
- Environment: Docker

zepatrik:
I'm not quite sure I got the steps there. Please correct me:

1. register with 1@t.com (don't click the verification link; account is not verified yet)
2. change your e-mail address to 2@t.com (still not verified)
3. request a new verification link that is delivered to 2@t.com
4. click that link received at 2@t.com
5. your account is now verified but with the first address (1@t.com)

decker502:
Sorry, I missed that, 2@t.com is verified, other steps are correct

decker502:
The steps that I do is:
1. register  and verify with 2@t.com, then logout (account 2@t.com is verified)
2. register with 1@t.com (don't click the verification link; account is not verified yet)
3. request a verification link that is delivered to 2@t.com
4. click that link received at 2@t.com
5. your account is now verified but with the first address (1@t.com)

zepatrik:
So you verify 2@t.com beforehand with another account? Or is 5. incorrect and the account  is verified with 2@t.com in the end?

zepatrik:
Ah sorry did not receive your message before sending. Will try to reproduce later.

aeneasr:
So is this resolved? Not sure if I follow 😅

decker502:
I tried repeatedly and couldn't reproduce this problem. Let me observe again. You can close the issue, thank you.

PR #999, opened 2021-01-21
test: add case to ensure correct behavior when verifying a different email address

zepatrik:
@decker502 can you please check if this is exactly your steps described in #998 ? If not, can you open a PR against this branch to correct it? For me, this is failing locally as expected. You can run this yourself:
```
$ npm i
$ ./test/e2e/run.sh --dev sqlite verification
```


decker502:
According to this test case, there is indeed no problem. I need some time to confirm whether the test case is consistent with the version of Kratos I am using. Thx!

aeneasr:
@zepatrik could you solve the e2e issue so that we can merge this and ensure that there is no future regression regarding this?

PR #1012, opened 2021-01-25
Umbrella PR for Ory Kratos v0.6 with MFA and improved flows (#961)

### LINKED ISSUE #955: Improve Argon2 helpers and defaults and documentation

aeneasr:
Description TBD

## Release notes

- Removing the method table makes things even faster (no joins)
- 

## Related issues

- https://github.com/ory/cli/pull/63
- https://github.com/ory/kratos-client-go/pull/6
- https://github.com/ory/cli/pull/62

## Todo

- [ ] write tests for domain aliasing, especially in session manager and form actions
- [ ] hooks exxecute per strat and after everything is done too
- [ ] make sure hooks dont execute twice then
- [x] Before applying strategy we need to be sure that it is actually enabled (e.g. login -> password)
- [ ] Update docs
  - [ ] SDK usage examples
  - [ ] Form rendering
  - [ ] Explain UI nodes
- [ ] In ORY CLI revert
     ```
 	//ctx = context.WithValue(ctx, cliclient.HTTPClientContextKey, func(cmd *cobra.Command) *http.Client {
	//	return remote.NewHTTPClient(cmd)
	//})
     ```
- [ ] Protect against ID guessing / sharing
- [x] // TODO replace "payload.Identifier"
- [ ] `testhelpers.EncodeFormAsJSON` no longer escapes `.`
- [ ] `// TODO merge`
- [ ] feature overview page? include login delay time for non-existing accounts
- [ ] `// TODO: check prefix only password group?`
- [x] document that individual redirect urls for login / reg methods no longer exist


nerrixDE:
How's this doing? 🙃

aeneasr:
Blocked by other stuff and some priorization shifts unfortunately.

nerrixDE:
> Blocked by other stuff and some priorization shifts unfortunately.

sad, is this just about coding work or also about the concept?

If later, I could maybe help by putting together how @nextcloud does TOTP / "Google Authenticator"-like 2FA, which is I think more important on a time-priority basis than Webauthn, which only very very few people use, unlike "Google Authenticator"-2FA.
If this could help speed up things, I would gladly do that

aeneasr:
> sad, is this just about coding work or also about the concept?

We're currently working on getting ory cloud up and running and once we have a stable prod env we can re-dedicate focus on implementing features. Currently we're busy with stabilizing a lot of things - so even if it does not improve features, it greatly improves prod stability which is probably even more important!

mattbonnell:
Hey, happy to lend a hand on this while you focus on getting the SaaS up and running. MFA is a high-priority item for us and I'd love to help push this forward in any way possible.

I saw in #26 that you thought it would be helpful to have someone dig into WebAuthn and make a proposal for its integration into the login flow. Would this still be the most helpful thing at this time? 

From that discussion, I gather that while WebAuthn would be provide an easy flow, it would be unsuitable as a primary factor because of the impossibility of porting the private key across browsers/devices ([from your comment here](https://github.com/ory/kratos/issues/26#issuecomment-746542935)), pointing to the need for another mechanism. This leads me to believe maybe WebAuthn is not the highest priority item as far as MFA is concerned.

Looking through your PR, I see that you added some TOTP-related changes here. What would the ideal MVP MFA implementation look like? TOTP delivered via SMS? Just want to get a sense of where you left things off when priorities changed and where your thinking is at currently with respect to this so that I could help drive your vision in any way I can.

Thanks for all your hard work!

nerrixDE:
> Hey, happy to lend a hand on this while you focus on getting the SaaS up and running. MFA is a high-priority item for us and I'd love to help push this forward in any way possible.
> 
> I saw in #26 that you thought it would be helpful to have someone dig into WebAuthn and make a proposal for its integration into the login flow. Would this still be the most helpful thing at this time?
> 
> From that discussion, I gather that while WebAuthn would be provide an easy flow, it would be unsuitable as a primary factor because of the impossibility of porting the private key across browsers/devices ([from your comment here](https://github.com/ory/kratos/issues/26#issuecomment-746542935)), pointing to the need for another mechanism. This leads me to believe maybe WebAuthn is not the highest priority item as far as MFA is concerned.
> 
> Looking through your PR, I see that you added some TOTP-related changes here. What would the ideal MVP MFA implementation look like? Just want to get a sense of where you left things off when priorities changed and where your thinking is at currently with respect to this so that I could help drive your vision in any way I can.
> 
> Thanks for all your hard work!

Awesome, same here. I would like to have 2FA sooner as well. I think TOTP should be the priority, WebAuthn is really not wide spread and also not the best way to go. TOTP should be the priority. 

I could help as well with concepting, whereas my go skills are limited.

aeneasr:
Thank you, your help would be greatly appreciated! Best issues to work on can be found here: https://github.com/ory/kratos/issues?q=is%3Aissue+is%3Aopen+milestone%3Av0.6.0-alpha.1+label%3A%22help+wanted%22

Solving any one of those brings us close to 0.6! :) :): :)

x/time.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var rnd = rand.New(rand.NewSource(time.Now().Unix()))
```
aeneasr, line 12: @mattbonnell the default rand source has always the same seed (1) which makes the delay pattern predictable.

x/time_test.go
```diff
@@ -8,11 +8,11 @@ import (
...
-	base := time.Millisecond * 20
-	deviation := time.Millisecond * 5
-	start := time.Now()
-	RandomDelay(base, deviation)
-	delay := time.Since(start)
-	require.LessOrEqual(t, delay, base+deviation)
-	require.GreaterOrEqual(t, delay, base-deviation)
+	base := time.Millisecond * 2
+	deviation := time.Millisecond
+	for i := 0; i < 100; i++ {
```
aeneasr, line 13: @mattbonnell I was getting flaky tests and changed this so that it returns the duration instead of sleeping which makes it harder to debug. It's now failing consistently:

```
=== RUN   TestRandomDelay
    assertion_compare.go:342: 
        	Error Trace:	time_test.go:15
        	Error:      	"3.219555ms" is not less than or equal to "3ms"
        	Test:       	TestRandomDelay
        	Messages:   	[]
```

Can you please take a look and fix this? Thank you!

mbonnell-wish: yep let me take a look!

mbonnell-wish: okay no, seems to be an issue with the math :) will fix this today.

mattbonnell: fix PR https://github.com/ory/kratos/pull/1131

aeneasr:
Merging this because I am fed up with all the conflicts...still needs some work though.

Issue #954, opened 2021-01-03
Admin Api /recovery/link giving 404


Benehiko:
**Describe the bug**

When attempting to do a `POST` request to `/recovery/link` - I get a 404 from the Kratos server.
Currently I am on version v0.5.5-alpha.1 in docker.

**Reproducing the bug**

1. Run docker-compose up --build --force-recreate --remove-orphans
2. Make an API request with curl
    curl -X POST http://127.0.0.1:4434/recovery/link \
    -H 'Content-Type: application/json' \
    -H 'Accept: application/json'
    
3. Request responds with `404 page not found`

Kratos server logs:

    kratos_1          | time=2021-01-03T17:59:30Z level=info msg=started handling request method=POST name=admin#http://kratos:4434/ remote=172.18.0.1:47534 request=/recovery/link

    kratos_1          | time=2021-01-03T17:59:30Z level=info msg=completed handling request method=POST name=admin#http://kratos:4434/ remote=172.18.0.1:47534 request=/recovery/link status=404 text_status=Not Found took=272.286µs

`docker-compose.yml`

    kratos:
    depends_on:
      - kratos-migrate
    image: oryd/kratos:v0.5.5-alpha.1
    ports:
      - "4433:4433" # public
      - "4434:4434" # admin
    restart: unless-stopped
    environment:
      - DSN=postgres://postgres@172.17.0.1:5432/kratos?sslmode=disable&max_conns=20&max_idle_conns=4
    command:
      serve -c "/etc/config/kratos.dev.yaml" --dev
    volumes:
      - type: bind
        source: ./kratos
        target: /etc/config
    networks:
      - intranet

`kratos.dev.yaml`
**please note I override the dsn with an environment variable pointing it to postgres**

    dsn: sqlite:///var/lib/sqlite/db.sqlite?_fk=true
    serve:
      public:
        base_url: http://127.0.0.1:4455/.auth/
        port: 4433
      admin:
        base_url: http://kratos:4434/
        port: 4434
    identity:
      default_schema_url: file:///etc/config/identity.schema.traits.dev.json
      schemas:
        - id: lite
          url: file://etc/config/identity.schema.traits.lite.dev.json

    selfservice:
      default_browser_return_url: http://127.0.0.1:8080
      
      methods:
        profile:
          enabled: true
        link:
          enabled: false
        password:
          enabled: true
      
      whitelisted_return_urls:
        - http://127.0.0.1:4455
        - http://127.0.0.1:8080
      
      flows:
[...]

PR #1025, opened 2021-01-31
fix: give specific message instead of only 404 when method disabled

lmcl90:
## Related issue

#954 

## Proposed changes

Return 400 with message `This endpoint was disabled by system administrator. Please check your url or contact the system administrator to enable it.` when user requests endpoints belonging to disabled methods. 

## Checklist

      the changes.
      works.

### DIFF
--- a/driver/registry_default.go
+++ b/driver/registry_default.go
@@ -256,9 +256,7 @@ func (m *RegistryDefault) RegistrationStrategies() registration.Strategies {
 	if len(m.registrationStrategies) == 0 {
 		for _, strategy := range m.selfServiceStrategies() {
 			if s, ok := strategy.(registration.Strategy); ok {
-				if m.c.SelfServiceStrategy(string(s.ID())).Enabled {
-					m.registrationStrategies = append(m.registrationStrategies, s)
-				}
+				m.registrationStrategies = append(m.registrationStrategies, s)
 			}
 		}
 	}
@@ -269,9 +267,7 @@ func (m *RegistryDefault) LoginStrategies() login.Strategies {
 	if len(m.loginStrategies) == 0 {
 		for _, strategy := range m.selfServiceStrategies() {
 			if s, ok := strategy.(login.Strategy); ok {
-				if m.c.SelfServiceStrategy(string(s.ID())).Enabled {
-					m.loginStrategies = append(m.loginStrategies, s)
-				}
+				m.loginStrategies = append(m.loginStrategies, s)
 			}
 		}
 	}
--- a/driver/registry_default.go
+++ b/driver/registry_default.go
@@ -144,15 +144,15 @@ func (m *RegistryDefault) RegisterPublicRoutes(router *x.RouterPublic) {
 	m.RegistrationHandler().RegisterPublicRoutes(router)
 	m.LogoutHandler().RegisterPublicRoutes(router)
 	m.SettingsHandler().RegisterPublicRoutes(router)
-	m.LoginStrategies().RegisterPublicRoutes(router)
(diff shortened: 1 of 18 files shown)


driver/registry_default.go
```diff
@@ -256,9 +256,7 @@ func (m *RegistryDefault) RegistrationStrategies() registration.Strategies {
 	if len(m.registrationStrategies) == 0 {
 		for _, strategy := range m.selfServiceStrategies() {
 			if s, ok := strategy.(registration.Strategy); ok {
-				if m.c.SelfServiceStrategy(string(s.ID())).Enabled {
-					m.registrationStrategies = append(m.registrationStrategies, s)
-				}
+				m.registrationStrategies = append(m.registrationStrategies, s)
```
aeneasr, line 259: The problem here is that these strategies are not only used for handlers but also in other areas (e.g. populating the flow methods). I think this needs to be kept as is!

driver/registry_default_test.go
```diff
@@ -68,10 +68,10 @@ func TestDriverDefault_Strategies(t *testing.T) {
 	}{
 		{prep: func(conf *config.Config) {
 			conf.MustSet(config.ViperKeySelfServiceStrategyConfig+".password.enabled", false)
-		}},
+		}, expect: []string{"password", "oidc"}},
 		{prep: func(conf *config.Config) {
 			conf.MustSet(config.ViperKeySelfServiceStrategyConfig+".password.enabled", true)
-		}, expect: []string{"password"}},
+		}, expect: []string{"password", "oidc"}},
```
aeneasr, line 74: The changes in this file should be reverted as this is still the expected behavior!

aeneasr (changes requested):

Particularly, I think we could use the IsDisabled logic here as well:

https://github.com/ory/kratos/blob/master/driver/registry_default.go#L154-L162

Then, you could change e.g. [this call](https://github.com/ory/kratos/blob/master/driver/registry_default.go#L159) to use a new method such as `GetAllLoginStrategies()` which would simply ignore the `Enabled` property but wouldn't break the existing behavior!

driver/registry_default.go
lmcl90: Yeah, I see it. I'll refactor this.

aeneasr: Awesome, thanks!

selfservice/strategy/handler.go
```diff
@@ -0,0 +1,25 @@
...
+const EndpointDisabledMessage = "This endpoint was disabled by system administrator. Please check your url or contact the system administrator to enable it."
+
+func IsDisabled(c interface {
+	config.Provider
+	x.WriterProvider
+}, strategy string, wrap httprouter.Handle) httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		enabled := c.Config(r.Context()).SelfServiceStrategy(strategy).Enabled
+		if !enabled {
+			c.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(EndpointDisabledMessage))
```
aeneasr, line 20: Let's make this `herodot.ErrNotFound.WithDescription(EndpointDisabledMessage)` :)

aeneasr (changes requested):
Great job! Just one little detail and it's good to go!!

selfservice/strategy/handler.go
lmcl90: OK

lmcl90: There is no method named `WithDescription` in `DefaultError`. I think it should be `WithReason`?

aeneasr:
Thank you so much for your work!

PR #1033, opened 2021-02-03
feat: Courier foreground worker with "kratos courier watch"

mbonnell-wish:
## Related issue
#652 #732

## Proposed changes

Following the discussion in #1024, this PR defines a new `kratos courier watch` command which runs the message courier in the foreground, as well as disabling the background courier by default when `kratos serve` is executed. 
The `--watch-courier` flag can be passed to the `serve` command to execute the courier in the background as is the current behavior.

## Checklist

      the changes.
      works.

### DIFF
--- a/cmd/courier/watch_test.go
+++ b/cmd/courier/watch_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/ory/kratos/internal"
+	"github.com/phayes/freeport"
 	"github.com/stretchr/testify/require"
 )
 
@@ -22,7 +23,9 @@ func TestStartCourier(t *testing.T) {
 	t.Run("case=with metrics", func(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.Background())
 		_, r := internal.NewFastRegistryWithMocks(t)
-		r.Config(ctx).Set("expose-metrics-port", 8080)
+		port, err := freeport.GetFreePort()
+		require.NoError(t, err)
+		r.Config(ctx).Set("expose-metrics-port", port)
 		go StartCourier(ctx, r)
 		time.Sleep(time.Second)
 		res, err := http.Get("http://" + r.Config(ctx).MetricsListenOn() + "/metrics/prometheus")


mattbonnell:
Hey @aeneasr, would you happen to know the source of this error?

zepatrik (commented):
The error you get is caused by https://github.com/ory/x/blob/87359d96f1cf9eb327949f693077968a50ecc283/configx/provider.go#L162-L166
Because of that, you have to add the flag to the config schema, just like it is done with e.g. the `dev` or `help` flag: https://github.com/ory/kratos/blob/af200d1396c883d0736c213251c30e0a205a79c0/.schema/config.schema.json#L1168-L1184
We should probably adjust the flag loading, but it will be possible to put that value in the config file as well :wink: 

Looks good otherwise :+1:

mattbonnell:
Thanks!

mbonnell-wish:
What would we be looking for in terms of added material in the user guide docs?

mbonnell-wish:
Hmm, I'm failing these `validate` job because the generated cli docs don't pass the formatting check. Is there some command i should be running other than `make docs/cli`?

zepatrik:
`make format` will format everything, `cd docs; npm run format` only the docs :wink:

aeneasr:
I believe you need to update the e2e config to reflect the new changes as the mail tests are probably failing!

https://github.com/ory/kratos/blob/master/test/e2e/run.sh#L141

cmd/courier/watch.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/ory/x/stringsx"
+)
+
+var watchCmd = &cobra.Command{
+	Use:   "watch",
+	Short: "Starts the ORY Kratos message courier",
+	Run: func(cmd *cobra.Command, args []string) {
+		d := driver.New(cmd.Context(), configx.WithFlags(cmd.Flags()))
+
+		go ServeHealth(d, cmd, args)
```
aeneasr, line 26: I think having health and metrics servers is honorable :) But we don't really need them IMO. It would make sense if this would handle HTTP API requests. Then, the LB needs to know if the service is up (health) and prometheus would want to know how much latency we have (for example).

So in my opinion, this can be removed!

aeneasr (commented):
This looks great!

I think it would make sense to explain this in the docs a bit:

- https://www.ory.sh/kratos/docs/concepts/email-sms
- https://www.ory.sh/kratos/docs/guides/account-activation-email-verification - a small sentence and a link will be enough - something along the lines of "set up mail worker (link)".
- https://www.ory.sh/kratos/docs/guides/high-availability-ha - explain that with multiple nodes you need one worker node to handle the mail queue

Thank you!

cmd/courier/watch.go
mbonnell-wish: Thanks for the review! 

Yeah, after doing some more reading, I think we can do without liveness probes in this case, and just rely on k8s restarting the pod if the courier crashes. I'll remove this

On the metrics front, you're quite right that latency metrics wouldn't be very relevant here since it isn't handling HTTP requests, but we would like to be able to monitor the go runtime related metrics, particularly since we'd like to be able to measure how well it's performing as we add load and potentially tune things like the number of concurrent goroutines and batch fetch size in response to our observations.

mbonnell-wish: I will work on these additional docs today

mbonnell-wish: Ok draft of doc updates are in, take a read and don't hesitate to suggest improvements

cmd/courier/watch.go
```diff
@@ -0,0 +1,73 @@
...
+	},
+}
+
+func ServeMetrics(r driver.Registry, cmd *cobra.Command, args []string) {
+
+	c := r.Config(cmd.Context())
+	l := r.Logger()
+	n := negroni.New()
+
+	router := x.NewRouterAdmin()
```
aeneasr, line 35: We can keep the metrics around if you have a concrete use case for it (e.g. watching the metrics for this task). However, I would like to make it optional. One issue is also that this is using the admin config port, which might cause issues when not running as containers but on bare-metal (port re-allocation). It would need several config files, and so on - so how about adding a CLI flag `--expose-metrics-port 123` which, if set, enables the metrics port?

aeneasr (commented):

cmd/courier/watch.go
mattbonnell: Perfect, I’ll add this.

docs/docs/cli/kratos-courier-watch.md
```diff
@@ -0,0 +1,40 @@
...
+Starts the ORY Kratos message courier
+
+```
+kratos courier watch [flags]
+```
+
+### Options
+
+```
+  -h, --help   help for watch
```
mbonnell-wish, line 28: hm, for some reason the cli docs aren't listing the `--expose-metrics-port` flag

aeneasr: Should be updated automatically by the CI pipeline!

driver/config/config.go
```diff
@@ -578,6 +579,18 @@ func (p *Config) IsInsecureDevMode() bool {
 	return p.Source().Bool("dev")
 }
 
+func (p *Config) IsBackgroundCourierEnabled() bool {
+	return p.Source().Bool("watch-courier")
+}
+
+func (p *Config) CourierExposeMetricsPort() int {
+	return p.Source().Int("courier.watch.expose-metrics-port")
```
aeneasr, line 587: I don't think this is correct, shouldn't it be just `expose-metrics-port`?

cmd/courier/watch.go
```diff
@@ -0,0 +1,82 @@
+package courier
```
aeneasr, line 1: Can you please add a test to see if this works (e.g. health port) as expected?

aeneasr (commented):
Great! Now we need a bit of testing and we're good to go! :)

driver/config/config.go
mbonnell-wish: Yes, you're right. Forgot to update this bit.

cmd/courier/watch.go
mbonnell-wish: Yes, much needed. Where would the best place be to add it?

aeneasr: Two options: you extract as much from the cobra handler as possible (e.g. as a function) and then set it up, or you use the cobra test framework to run the command.

I would opt for the first option. Regarding courier tests and the SMTP integration, check out:

- https://github.com/ory/kratos/blob/master/courier/courier_test.go#L41

Cobra commands can be controlled from tests using:

```
import "github.com/ory/x/cmdx"

// ...
ce := &cmdx.CommandExecuter{New: func() *cobra.Command {
		return cmd.RootCmd // might need to be exported in kratos
	}, Ctx: ctx}

ce.ExecNoErr(t, "your", "command")
```

```
```

mbonnell-wish: Added some tests in `cmd/courier/watch_test.go`, let me know what you think!

cmd/courier/watch.go
```diff
@@ -0,0 +1,90 @@
...
+	server := graceful.WithDefaults(&http.Server{
+		Addr:    c.MetricsListenOn(),
+		Handler: context.ClearHandler(n),
+	})
+
+	l.Printf("Starting the metrics httpd on: %s", server.Addr)
+	if err := graceful.Graceful(func() error {
+		go server.ListenAndServe()
+		<-ctx.Done()
+		return server.Shutdown(ctx)
```
mbonnell-wish, line 67: Added this so the server can be shutdown by cancelling the context (eg in unit tests).
Let me know if there's a better way.

mbonnell-wish: Yeah, this fails the validate check due to the unchecked error. Must be a better way

cmd/courier/watch.go
```diff
@@ -0,0 +1,95 @@
...
+	l.Printf("Starting the metrics httpd on: %s", server.Addr)
+	if err := graceful.Graceful(func() error {
+		go func() {
+			if err := server.ListenAndServe(); err != nil {
+				l.Fatalf("Failed to start the metrics httpd: %s\n", err)
+				return
+			}
+		}()
+		<-ctx.Done()
+		return server.Shutdown(ctx)
```
mbonnell-wish, line 72: Added this so the server can be shutdown by cancelling the context (eg in unit tests).
Let me know if there's a better way.

mbonnell-wish:
hm, the `GO111MODULE=off go get github.com/mattn/goveralls github.com/ory/go-acc` test job is failing with
```
cannot find package "github.com/hashicorp/hcl/hcl/printer" in any of:
	/usr/local/go/src/github.com/hashicorp/hcl/hcl/printer (from $GOROOT)
	/go/src/github.com/hashicorp/hcl/hcl/printer (from $GOPATH)

```
Can't imagine it's related to my PR

cmd/courier/watch.go
mbonnell-wish, line 72: Added this so the server can be shutdown by cancelling the context (eg in unit tests).
Let me know if there's a better way.

aeneasr: I think this looks ok

aeneasr:
@mattbonnell , great work! The current problem for the tests is not your fault. I'll be trying to fix that now!

Issue #1032, opened 2021-02-02
Hashtag in url converts to %23


Zippersk:
**Describe the bug**

when `SELFSERVICE_FLOWS_LOGIN_UI_URL` is set to URL with a hashtag (`http://example.com/#/login`), the hashtag is converted to `%23` (and location header from kratos response become `http://example.com/%23/login`). Hybrid applications (based on electron, cordova, capacitor...), that needs to have a hash-based router, does not work correctly with the current implementation of rewriting `#`->`%23`.

My simplified kratos docker service.
```
  kratos:
    image: oryd/kratos:latest
    environment:
      - DSN=postgres://kratos:kratos@db:5432/kratos?sslmode=disable&max_conns=20&max_idle_conns=4
      - SELFSERVICE_DEFAULT_BROWSER_RETURN_URL=${RETURN_TO}
      - SELFSERVICE_WHITELISTED_RETURN_URLS=${RETURN_TO}
      - SELFSERVICE_FLOWS_REGISTRATION_UI_URL=${UI_BASE}/#/auth/registration
      - SELFSERVICE_FLOWS_LOGIN_UI_URL=${UI_BASE}/#/auth
    command: serve -c /etc/config/kratos/kratos.yml
    volumes:
      - ./kratos:/etc/config/kratos
```

Request to `/self-service/login/browser` (`UI_BASE` is set to `http://localhost:9999` in this example)
```
:authority: kratos.*****
:method: GET
:path: /self-service/login/browser
:scheme: https
accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
accept-encoding: gzip, deflate, br
accept-language: en-US,en;q=0.9,sk-SK;q=0.8,sk;q=0.7,cs;q=0.6
cache-control: no-cache
cookie: csrf_token=HRYkllp56zy8XelKLX+8JpRcW5JMjybXAT8Rem+4e7I=
pragma: no-cache
referer: http://localhost:9999/
sec-fetch-dest: document
sec-fetch-mode: navigate
sec-fetch-site: cross-site
upgrade-insecure-requests: 1
user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.104 Safari/537.36
```

Response:
```
cache-control: private, no-cache, no-store, must-revalidate
content-length: 95
[...]

PR #1040, opened 2021-02-05
fix: Hashtag in url converts to %23

Zippersk:
## Related issue
#1032 @aeneasr 

## Proposed changes

Lots of hybrid applications that are built with [electron](https://nklayman.github.io/vue-cli-plugin-electron-builder/guide/commonIssues.html#blank-screen-on-builds-but-works-fine-on-serve), cordova, capacitor, etc. use hash-based routing. When we use URLs with a hash mark ([fragment](https://en.wikipedia.org/wiki/URI_fragment)), Kratos replaced it with `%23` because [ParseRequestURI](https://golang.org/pkg/net/url/#ParseRequestURI) assumed that URL dows not to have a #fragment suffix (see docs). [Parse](https://golang.org/pkg/net/url/#Parse) function parse URL correctly with #fragment.

## Checklist

      the changes.
      works.
- [x ] I have added or changed [the documentation](docs/docs).

### DIFF
--- a/driver/config/config.go
+++ b/driver/config/config.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/markbates/pkger"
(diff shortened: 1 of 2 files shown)


driver/config/config.go
```diff
@@ -562,7 +562,7 @@ func (p *Config) CourierTemplatesRoot() string {
 }
 
 func (p *Config) parseURIOrFail(key string) *url.URL {
-	u, err := url.ParseRequestURI(p.p.String(key))
+	u, err := url.Parse(p.p.String(key))
```
aeneasr, line 565: Looks good! I think we should still check if there are the following things set:

- scheme (`http(s)://`)
- host

as `url.Parse` only returns an error if the string is empty!

aeneasr (commented):

driver/config/config.go
```diff
@@ -561,13 +562,29 @@ func (p *Config) CourierTemplatesRoot() string {
...
 func (p *Config) parseURIOrFail(key string) *url.URL {
-	u, err := url.ParseRequestURI(p.p.String(key))
+	u, frag := splitUrlAndFragment(p.p.String(key))
+	url, err := url.ParseRequestURI(u)
 	if err != nil {
 		p.l.WithError(errors.WithStack(err)).
 			Fatalf("Configuration value from key %s is not a valid URL: %s", key, p.p.String(key))
 	}
-	return u
+	if url.Host == "" || (url.Scheme != "http" && url.Scheme != "https") {
```
aeneasr, line 580: We allow other schemes as well (which is why tests fail) - e.g. `base64` or `file`

Zippersk: I am a bit confused. [There ](https://github.com/ory/kratos/pull/1040#discussion_r571103306) you said that we should check for the scheme (`http(s)://`). So should I remove the scheme check? Sorry for the misunderstanding.

aeneasr: No my bad, I was just not careful enough in my comment. `url.Scheme == ""` should be enough!

Zippersk:
@aeneasr please can you look at this again?

aeneasr (approved):

aeneasr:
Sorry, I was deep down in Kubernetes! Thank you for the changes!

driver/config/config.go
zepatrik: The host is `""` on absolute file URLs btw

PR #803, opened 2020-11-02
feat: add Argon2 calibration CLI

zepatrik:
## Related issue
closes #723 
closes #572 
closes #647 

## Proposed changes
This adds a CLI that conducts measurements to allow choosing the best argon2 parameters.

## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ ] I have read the [security policy](../security/policy).
- [ ] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](docs/docs).

PR #1088, opened 2021-02-24
feat: add random delay to login flow

### LINKED ISSUE #832: Add randomized constant time to every login request
**Describe the bug**

Currently, login requests will fail faster if the user does not exist as the hash does not have to be computed. This can leave to timing attacks where an attacker can guess if a user exists or not, which defeats account enumeration defenses.

**Expected behavior**

Every login request should take a similar amount of time regardless of whether the user exists or not. The time should be configurable.

**Environment**

- Version: v0.5

**Additional context**

If anyone wants to tackle this let me know and I will give the right pointers!

mattbonnell:
## Related issue
Closes #832 @aeneasr 

## Proposed changes

Adds a configurable randomized delay to the login flow to mitigate timing attacks (as specified in #832)

## Checklist

      the changes.
      works.

### DIFF
--- a/x/time_test.go
+++ b/x/time_test.go
@@ -1,0 +1,19 @@
+package x
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRandomDelay(t *testing.T) {
+	base := time.Millisecond * 20
+	deviation := time.Millisecond * 5
+	s := time.Now()
+	RandomDelay(base, deviation)
+	e := time.Now()
+	elapsed := e.Sub(s)
+	require.LessOrEqual(t, elapsed, base+deviation)
+	require.GreaterOrEqual(t, elapsed, base-deviation)
+}
--- a/x/time_test.go
+++ b/x/time_test.go
@@ -10,10 +10,9 @@ import (
 func TestRandomDelay(t *testing.T) {
 	base := time.Millisecond * 20
 	deviation := time.Millisecond * 5
-	s := time.Now()
+	start := time.Now()
 	RandomDelay(base, deviation)
-	e := time.Now()
-	elapsed := e.Sub(s)
-	require.LessOrEqual(t, elapsed, base+deviation)
-	require.GreaterOrEqual(t, elapsed, base-deviation)
+	delay := time.Now().Sub(start)
+	require.LessOrEqual(t, delay, base+deviation)
+	require.GreaterOrEqual(t, delay, base-deviation)
 }
--- a/x/time_test.go
+++ b/x/time_test.go
@@ -12,7 +12,7 @@ func TestRandomDelay(t *testing.T) {
 	deviation := time.Millisecond * 5
 	start := time.Now()
 	RandomDelay(base, deviation)
-	delay := time.Now().Sub(start)
+	delay := time.Since(start)
 	require.LessOrEqual(t, delay, base+deviation)
 	require.GreaterOrEqual(t, delay, base-deviation)
 }
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -500,27 +500,7 @@
                 },
                 "after": {
                   "$ref": "#/definitions/selfServiceAfterLogin"
-                },
-		"random_delay": {
-		  "title": "Login Randomized Delay Configuration",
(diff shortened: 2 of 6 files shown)


zepatrik (changes requested):
I think it would make sense to use the config values introduced by #803 instead. Can you base your changes off of the v0.6 branch?

mbonnell-wish:
> I think it would make sense to use the config values introduced by #803 instead. Can you base your changes off of the v0.6 branch?

Sure thing, which config values are you referring to specifically? I looked through that PR to find some but couldn't find any related to time specifically.

mattbonnell:
> I think it would make sense to use the config values introduced by #803 instead. Can you base your changes off of the v0.6 branch?

Sure thing, which config values specifically?

zepatrik:
An sorry I linked the wrong PR, it is #961
It is not merged on master yet but on the v0.6 branch

aeneasr:
Ah - sorry - it makes sense to have master as the upstream branch, not the 0.6 - they have many merge conflicts and are a bit outdated!

zepatrik:
But half of the changes are done there already?

aeneasr:
Ah, you're right @zepatrik

mattbonnell:
Question for you @aeneasr: in #832, you said we should add the random delay whether the user exists or not, but I’m a bit confused by that. Won’t that make the “user does exist flow” take twice as long as the “user doesn’t exist”, leaving us vulnerable to timing attacks? Shouldn’t we only add the delay when the user doesn’t exist, before returning error? Maybe I’m missing something there.

zepatrik:
Agree @mattbonnell 
The goal would be for requests to take the same time regardless whether the identifier exists and also avoid hashing if necessary (because resources :moneybag:).

mattbonnell:
@zepatrik made the updates as discussed here 👍 

By the way, the go.mod seems broken on `v0.6`. To get it working locally I have to call `go get ./...`, I left go.mod/go.sum changes out because go.sum modifications seem to cause CI to fail as well, though.

aeneasr:
Yeah that's possible - 0.6 is currently in a broken CI stage. Thank you for the changes, they look really good! Maybe we can also add a test for this? :)

mattbonnell:
> Yeah that's possible - 0.6 is currently in a broken CI stage. Thank you for the changes, they look really good! Maybe we can also add a test for this? :)

added a test for the RandomDelay utility function that I introduce here; let me know if there are other tests you'd like added

x/time_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRandomDelay(t *testing.T) {
+	base := time.Millisecond * 20
+	deviation := time.Millisecond * 5
+	start := time.Now()
+	RandomDelay(base, deviation)
+	delay := time.Now().Sub(start)
```
zepatrik, line 15:
```suggestion
	delay := time.Since(start)
```
`gosimple` will complain otherwise :wink:

zepatrik (changes requested):

x/time_test.go
mattbonnell: thanks, fixed!

zepatrik (approved):

aeneasr:
Sorry to be that guy 😓 but I think we need a proper e2e test that checks this. Otherwise chances are high that this will break or regress in the future. The easiest would be to add a test case here:

1. https://github.com/ory/kratos/blob/master/selfservice/strategy/password/login_test.go#L230

Basically you check if the request is taking at least the configured amount of time - that should be enough to ensure this behavior does not regress in the future

Thank you!

mattbonnell:
> Sorry to be that guy 😓 but I think we need a proper e2e test that checks this. Otherwise chances are high that this will break or regress in the future. The easiest would be to add a test case here:
> 
> 1. https://github.com/ory/kratos/blob/master/selfservice/strategy/password/login_test.go#L230
> 
> Basically you check if the request is taking at least the configured amount of time - that should be enough to ensure this behavior does not regress in the future
> 
> Thank you!

Coming right up!

mattbonnell:
@aeneasr added

aeneasr (approved):
Great job @mattbonnell, thank you! :)

PR #1130, opened 2021-03-08
Disable pwnedpasswords breaches check

ro-tex:
## Proposed changes

This PR allows Kratos to be configured to not check passwords against the database of pwnedpasswords.com.

## Checklist

      the changes.
      works.

## Further comments

I understand that this check is helpful because it prevents users from using passwords which are vulnerable to dictionary attacks by people who have the list of compromised passwords held by pwnedpasswords. Nevertheless, in some use cases it is acceptable to let users be responsible for their own security and allow them to use any password they like. While this should certainly not be the default, it should be a configurable option. This is what this PR makes possible.

### DIFF
--- a/driver/config/config.go
+++ b/driver/config/config.go
@@ -89,6 +89,7 @@ const (
 	ViperKeyHasherArgon2ConfigKeyLength                             = "hashers.argon2.key_length"
 	ViperKeyPasswordMaxBreaches                                     = "selfservice.methods.password.config.max_breaches"
 	ViperKeyIgnoreNetworkErrors                                     = "selfservice.methods.password.config.ignore_network_errors"
+	ViperKeyDisableBreachesCheck                                    = "selfservice.methods.password.config.disable_breaches_check"
 	ViperKeyVersion                                                 = "version"
 	Argon2DefaultMemory                                      uint32 = 4 * 1024 * 1024
 	Argon2DefaultIterations                                  uint32 = 4
@@ -120,8 +121,9 @@ type (
 		URL string `json:"url"`
 	}
 	PasswordPolicy struct {
-		MaxBreaches         uint `json:"max_breaches"`
-		IgnoreNetworkErrors bool `json:"ignore_network_errors"`
+		MaxBreaches          uint `json:"max_breaches"`
+		IgnoreNetworkErrors  bool `json:"ignore_network_errors"`
+		DisableBreachesCheck bool `json:"disable_breaches_check"`
 	}
 	Schemas []Schema
 	Config  struct {
@@ -719,7 +721,8 @@ func (p *Config) ConfigVersion() string {
 
 func (p *Config) PasswordPolicyConfig() *PasswordPolicy {
 	return &PasswordPolicy{
-		MaxBreaches:         uint(p.p.Int(ViperKeyPasswordMaxBreaches)),
-		IgnoreNetworkErrors: p.p.BoolF(ViperKeyIgnoreNetworkErrors, true),
+		MaxBreaches:          uint(p.p.Int(ViperKeyPasswordMaxBreaches)),
+		IgnoreNetworkErrors:  p.p.BoolF(ViperKeyIgnoreNetworkErrors, true),
+		DisableBreachesCheck: p.p.BoolF(ViperKeyDisableBreachesCheck, false),
 	}
 }
--- a/selfservice/strategy/password/validator.go
+++ b/selfservice/strategy/password/validator.go
@@ -153,6 +153,10 @@ func (s *DefaultPasswordValidator) Validate(ctx context.Context, identifier, pas
 		return errors.Errorf("the password is too similar to the user identifier")
 	}
 
+	if s.reg.Config(ctx).PasswordPolicyConfig().DisableBreachesCheck {
+		return nil
+	}
+
 	/* #nosec G401 sha1 is used for k-anonymity */
 	h := sha1.New()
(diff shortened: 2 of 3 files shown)


zepatrik:
> Nevertheless, in some use cases it is acceptable to let users be responsible for their own security and allow them to use any password they like.
Can you elaborate that a bit? What is the case that made you come up with this PR? I am trying, but could not come up with any case outside of some testing environment.
This feature ensures Kratos is following best-practices and [NIST guidelines](https://pages.nist.gov/800-63-3/sp800-63b.html). If there is any problem that occurs because of this feature, it should not be an "easy fix" to just disable it. We want to make security easy, and to make it hard to do mistakes.

aeneasr:
>  Nevertheless, in some use cases it is acceptable to let users be responsible for their own security and allow them to use any password they like.

NIST and other security experts will disagree with this statement. We are trying to make secure things easy and insecure things hard, disabling such a safe guard is usually something inexperienced developers abuse to work around limitations in their project. For that reason, we typically do not merge features that weaken security without a valid and reasonable use case.

Could you elaborate why you want to disable a mechanism that prevents account takeover?

ro-tex:
I am part of a team working on a network of web portals which give users access to the Sia storage blockchain and act as an application layer on top of it. The solution is called Skynet. Skynet is accessible to anyone for free and users don't need accounts in order to use it. In the near future we are going to introduce user accounts which allow users to pay for premium features but the base offering will remain free.

Our main issues with the password policy enforced by Kratos are:
1. It adds friction to the user registration process which will make it harder convert users from anonymous to registered. Our solution is not something which forces people to register (like a bank, a company CMS, an online store), so removing any UX hiccups from the registration path is very important to us.
2. It feels opaque and frustrating when you come up with several passwords and, even though they are strong by themselves, they are rejected because they were leaked in a data breach somewhere.
3. The very notion that we have no control over the password policy we are imposing on our users is making me and my team uneasy. We want to be able to defend each individual limitation we impose on our users and the check against "Have I Been Pwned?" is not one of them.

Again, I am not trying to say that this approach is generally bad or that it shouldn't be the default. It might be a bit more aggressive that the usual but not excessively so. But it should be configurable.

When you talk about NIST guidelines, do you mean point "5.1.1.1 Memorized Secret Authenticators" which states: "If the CSP or verifier disallows a chosen memorized secret based on its appearance on a blacklist of compromised values, the subscriber SHALL be required to choose a different memorized secret"? That refers to the usual blocklists of common weak passwords which are checked during dictionary attacks. I am not convinced that the vast database of leaked passwords held by "Have I Been Pwned" could be considered a blacklist. First, there is nothing inherently wrong with the passwords there, as they are not there because they are weak but because they were leaked after an attack. Second, they are far too numerous to be used in a dictionary attack.

With the above I am not trying to argue against your chosen way of ensuring passwords are strong enough. I am just trying to understand the references to NIST guidelines.

> Could you elaborate why you want to disable a mechanism that prevents account takeover?

Most of all, I want to be in control over the mechanisms I impose on my users. I want to be able to choose the point of balance between security and accessibility.

aeneasr:
> It adds friction to the user registration process which will make it harder convert users from anonymous to registered. Our solution is not something which forces people to register (like a bank, a company CMS, an online store), so removing any UX hiccups from the registration path is very important to us.

Security always adds friction and I would argue that anything related to data storage (Skynet) or monetization (Sia coin) should do its best to prevent account takeover. However, this decision is of course up to you :)

> When you talk about NIST guidelines, do you mean point "5.1.1.1 Memorized Secret Authenticators" which states: "If the CSP or verifier disallows a chosen memorized secret based on its appearance on a blacklist of compromised values, the subscriber SHALL be required to choose a different memorized secret"? That refers to the usual blocklists of common weak passwords which are checked during dictionary attacks. I am not convinced that the vast database of leaked passwords held by "Have I Been Pwned" could be considered a blacklist. First, there is nothing inherently wrong with the passwords there, as they are not there because they are weak but because they were leaked after an attack. Second, they are far too numerous to be used in a dictionary attack.

Checking well-known password lists - or rather well-known bruteforce lists - is security practice from late 2000/early 2010. The security landscape has evolved since then and it is common and best practice to check against a list of well known (because leaked in breaches) passwords.

> Most of all, I want to be in control over the mechanisms I impose on my users. I want to be able to choose the point of balance between security and accessibility.

You can, there is an option to configure how often a password was leaked before it blocks registration: https://github.com/ory/kratos/blob/master/driver/config/.schema/config.schema.json#L663-L669

zepatrik:
I was referring to section 5.1.1.2:

> When processing requests to establish and change memorized secrets, verifiers SHALL compare the prospective secrets against a list that contains values known to be commonly-used, expected, or compromised. For example, the list MAY include, but is not limited to:
>
> 1. Passwords obtained from previous breach corpuses.
> 2. Dictionary words.
> 3. Repetitive or sequential characters (e.g. ‘aaaaaa’, ‘1234abcd’).
> 4. Context-specific words, such as the name of the service, the username, and derivatives thereof.

ro-tex:
> Checking well-known password lists - or rather well-known bruteforce lists - is security practice from late 2000/early 2010. The security landscape has evolved since then and it is common and best practice to check against a list of well known (because leaked in breaches) passwords.

You are right, my security practices seem to have gathered some dust. I did some extra reading and it seems that what I'm suggesting here is actually being discouraged by NIST as of early 2020. Well, I learned something. Thank you for your time and for the positive approach to this discussion.

aeneasr:
Any time, and welcome to the Ory family :)

PR #1131, opened 2021-03-08
fix: fix norm distribution math

mattbonnell:
## Checklist

      the changes.
      works.


x/time.go
```diff
@@ -19,6 +19,6 @@ func RequireEqualTime(t *testing.T, expected, actual time.Time) {
 	require.EqualValues(t, expected.UTC().Round(time.Second), actual.UTC().Round(time.Second))
 }
 
-func RandomDelay(base time.Duration, deviation time.Duration)time.Duration {
-	return time.Duration(rnd.NormFloat64()*float64(deviation) + float64(base))
+func RandomDelay(base, deviation time.Duration) time.Duration {
+	return time.Duration(rnd.NormFloat64()*(float64(deviation)/3) + float64(base))
```
aeneasr, line 23: What is the `/3` for? Maybe you could add a short description of the formula being used here :)

mattbonnell: Added some description in the PR description but yeah a comment here would be good, will add.

aeneasr: If I understand correctly (sorry, it's been a while since I had statistics at university) wouldn't this mean that we get a value of infinite if we try infinite amount of times? We will not do it infinitely, but it means that we could get excessively long times (e.g. a minute, an hour) some (if we run it to eternity) of the times? If so, I think a min/max cutoff would be in place here!

mattbonnell: Yep, came to the same realization 😅 . added exactly that.

aeneasr:
Thank you!

PR #1135, opened 2021-03-09
docs: FAQ improvement

vinckr:
Added some items to the FAQ as well as rewriting some existing ones, styling the links differently. 

tested with 
`npm run gen`
`npm run start`
and working.

PS: there was a section including {id} in one answer, this caused the FAQ not to load, I removed the section with that code segment, it was not crucial anyway.

## Checklist

      the changes.
      works.

### DIFF
--- a/docs/faq.yaml
+++ b/docs/faq.yaml
@@ -4,12 +4,14 @@
   q: We want to seperate our customers and employees, so we store them in different
     databases. But we would like to have them use the same login dialog for our portal.
     How can I achieve that with Ory Kratos?
-  a: You can deploy Ory Kratos two times, and use the same login/registration endpoints.
-    So you define the same login/registration UI URLs in both of the Kratos configurations.
+  a: You can deploy Ory Kratos two times, and use the same login UI pointing to two
+    different Kratos login endpoints - `/login/customer` or `/login/employee`, either
+    by having two different login routes, or by adding some logic to your login UI
+    that reroutes customers to `/login/customer` and employees to `/login/employee`.
+    So you define the same login or registration UI URLs in both of the Kratos configurations.
     You may need to tell your login/registration UI which Kratos it is supposed to
     talk to. The instances are cheap to deploy and the databases are completely isolated
-    from each other.  For example you can have something like `/login/customer` and
-    `/login/employee`.
+    from each other.  For example    something like `/login/customer` and `/login/employee`.
 - tags:
   - kratos
   - ecosystem
@@ -98,5 +100,6 @@
   q: I want to implement a single-page-app (SPA). Is this possible with Ory Kratos?
   a: 'Yes, definitely. We do not have an example-app (yet). around some page-reloads,
     as it is not possible to use the API flows in the browser. This is important to
-    prevent CSRF attacks. So hopefully you are not too strict with your SPA. A sample
-    app is being worked on here: [Implement React SPA sample app](https://github.com/ory/kratos/issues/668).'
+    prevent [CSRF attacks](https://en.wikipedia.org/wiki/Cross-site_request_forgery).
+    So hopefully you are not too strict with your SPA. A sample app is being worked
+    on here: [Implement React SPA sample app](https://github.com/ory/kratos/issues/668).'
--- a/docs/faq.yaml
+++ b/docs/faq.yaml
@@ -1,22 +1,23 @@
 - tags:
   - kratos
   - database
-  q: We want to seperate our customers and employees, so we store them in different
+  q: |
+    We want to seperate our customers and employees, so we store them in different
     databases. But we would like to have them use the same login dialog for our portal.
     How can I achieve that with Ory Kratos?
-  a: You can deploy Ory Kratos two times, and use the same login UI pointing to two
-    different Kratos login endpoints - `/login/customer` or `/login/employee`, either
-    by having two different login routes, or by adding some logic to your login UI
-    that reroutes customers to `/login/customer` and employees to `/login/employee`.
-    So you define the same login or registration UI URLs in both of the Kratos configurations.
+  a: |
+    You can deploy Ory Kratos two times, and use the same login/registration endpoints.
+    So you define the same login/registration UI URLs in both of the Kratos configurations.
     You may need to tell your login/registration UI which Kratos it is supposed to
     talk to. The instances are cheap to deploy and the databases are completely isolated
-    from each other.  For example    something like `/login/customer` and `/login/employee`.
+    from each other.  For example you can have something like `/login/customer` and
+    `/login/employee`.
 - tags:
   - kratos
   - ecosystem
   q: Is the code audited by an independent entity and is there a bug bounty program?
-  a: We will do that when the APIs and core are stable. A bug bounty program is in
+  a: |
+    We will do that when the APIs and core are stable. A bug bounty program is in
     the making but not done yet. If you are a security researcher and interested in
     working on Kratos, please reach out to [security@ory.sh](mailto:security@ory.sh).
 - tags:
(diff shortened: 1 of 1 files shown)


docs/faq.yaml
```diff
@@ -5,75 +5,97 @@
     databases. But we would like to have them use the same login dialog for our portal.
     How can I achieve that with Ory Kratos?
   a: You can deploy Ory Kratos two times, and use the same login/registration endpoints
-    if you like. You may need to tell you login/registration ui which Kratos it is
-    supposed to talk to.
+    if you like. You may need to tell your login/registration UI which Kratos it is
+    supposed to talk to. The instances are cheap to deploy and the databases are completely
+    isolated from each other.  For example you can have something like `/login/customer`
+    and `login/employee`.
```
zepatrik, line 11:
```suggestion
    and `l/ogin/employee`.
```

docs/faq.yaml
```diff
@@ -5,75 +5,97 @@
...
+- tags:
+  - kratos
+  - ecosystem
+  - bug
```
zepatrik, line 15: This tag makes little sense to me
```suggestion
```

docs/faq.yaml
```diff
@@ -5,75 +5,97 @@
...
+  q: Will you audit the code and/or provide bug bounties?
+  a: Yes, when we are near stable candidate we will audit the code and provide bug
+    bounties.
```
zepatrik, line 18:
```suggestion
  q: Is the code audited by an independent entity and is there a bug bounty program?
  a: We will do that when the APIs and core are stable. A bug bounty program is in the making but not done yet.
      If you are a security researcher and interested in working on Kratos, please reach out to [security@ory.sh](mailto:security@ory.sh).
```

docs/faq.yaml
```diff
@@ -5,75 +5,97 @@
...
+  - kratos
+  - registration
+  - cors
+  q: 'Hey guys, I am currently trying to get the Kratos registration flow working
+    on a react app that I am creating. I am having CORS issues with the completeSelfServiceRegistrationFlowWithPasswordMethod
+    call when using the @oryd/kratos-client JS package. '
+  a: You are using AJAX to submit the form. This is not supported right now. You need
+    to submit a regular HTML form. Do not use AJAX. Do not use fetch. Let the user
+    click the submit button and do not do anything. AJAX will be better supported
+    starting Kratos 0.6.  s
```
zepatrik, line 39:
```suggestion
    starting Kratos v0.6
```

docs/faq.yaml
```diff
@@ -5,75 +5,97 @@
...
-    to be edited by the end-user. Is there anything in the roadmap to support this
-    scenario?
-  a: 'There is an open issue for this feature: https://github.com/ory/kratos/issues/47.'
+  - feature
+  q: ' Is the ability to freeze certain identity fields after registration anywhere
+    on the roadmap? Also I want read-only or hidden user-specific data. Can I use
+    traits for that? Example: Some of our identity traits fields may be the subject
+    to identity validation ([know your customer](https://en.wikipedia.org/wiki/Know_your_customer)).
+    Once they are validated we do not allow the identity to be edited by the end-user.'
+  a: 'Currently traits are  visible to the user and also editable by him. It makes
```
zepatrik, line 49:
```suggestion
  a: 'Currently traits are visible to the users and also editable by themselves. It makes
```

docs/faq.yaml
```diff
@@ -5,75 +5,97 @@
...
-  - traits
-  q: I want read-only or hidden user-specific data. Can't i use traits for that?
-  a: Traits are currently visible to the user and also editable by him. Although it
-    make sense to have read-only and/or hidden traits for users, we can't give any
-    timing when such features will be available. [Issue](https://github.com/ory/kratos/issues/47)
+  q: I want to implement a single-page-app (SPA). Is this possible with Ory Kratos.
+  a: Yes, definitely. We do not have an example-app (yet). Also you might not come
+    around some page-reloads, as you definitely should avoid the API-flows and use
+    the browser-flows! So hopefully you are not too strict with your SPA. A sample
+    app is being worked on here:[Implement React SPA sample app](https://github.com/ory/kratos/issues/668).
```
zepatrik, line 101:
```suggestion
    app is being worked on here: [Implement React SPA sample app](https://github.com/ory/kratos/issues/668).
```

docs/faq.yaml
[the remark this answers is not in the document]
vinckr: Hm I understood it as, you have two different databases, two kratos instances and one central login page. 
so when you go to `/login` you can either choose for example `/login/customer` or `/login/employee`. So technically not the same login UI, just the same entry point for it. 

But you can also understand it as, you have one Login UI and then there is some logic in there to figure out if you are employee/customer and in which database you get saved. 

I feel like both is possible and feasible, just depends on your preferences, right? 

Not sure how to make it clearer in the question tho.

docs/faq.yaml
vinckr: I agree, will remove it.

docs/faq.yaml
vinckr: slipped on the keyboard there apparently...

docs/faq.yaml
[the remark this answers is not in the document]
vinckr: I see. 
So rephrase it to
 "Use the same login UI pointing to two different Kratos login endpoints - `/login/customer` or `/login/employee`, either by having two different login routes, or by adding some logic to your login UI that reroutes customers to `/login/customer` and employees to `/login/employee`. 
?

vinckr:
Commited your suggestions and added a more detailed explanation of the separate databases issue.
Can you check if it is more clear now? @zepatrik

docs/faq.yaml
```diff
@@ -4,76 +4,102 @@
...
     migrations? E.g. configure a callback to the legacy system when you cannot find
     the corresponding user, and store the identity on successful legacy system response.
-  a: No plans yet, but migration is an important piece of the puzzle. We will probably
-    focus on importing existing users first though. An alternative to callback and
-    custom code is probably fronting the legacy system with ORY Hydra (OAuth2/OIDC
-    Server) and then using that as an upstream in ORY Kratos.
+  a: No plans yet, but migration is an important piece of the puzzle. We will focus
+    on importing existing users first. An alternative to callback and custom code
+    is fronting the legacy system with Ory Hydra (OAuth2/OIDC Server) and then using
+    that as an upstream in Ory Kratos.
```
aeneasr, line 82: Please wrap **all** multi-line statements like so - what you have there is not valid YAML and will break (did you test rendering the FAQ?):

```suggestion
  a: |
    No plans yet, but migration is an important piece of the puzzle. We will focus
    on importing existing users first. An alternative to callback and custom code
    is fronting the legacy system with Ory Hydra (OAuth2/OIDC Server) and then using
    that as an upstream in Ory Kratos.
```

Please replace all `a: '` multi-line formatted with the pipe syntax as well

aeneasr (changes requested):

docs/faq.yaml
vinckr: Rendering the FAQ works fine. The YAML is valid too. 

`npm run gen`
`npm run start`
and working.

I just wanted to make a content update in this PR. 
We need to set up a proper generation pipeline for the FAQ too, it currently lives in one of Kims private repositories. I will open an issue tracking this, but its on my list too. 

As far as I could see the sentiment was that, when the FAQ is in the state we want it (and implement for all projects) we can set up a proper pipeline for outside contributors ect. too.

aeneasr: True, looks like it's valid YAML. However, when formatted, the `'` are introduced which need to be escaped if we do `it's` (`it\'s`) which is why it's better to pipe this as I proposed.

vinckr: Yea, you are right, that is why I removed all wordings like "doesn't" ect. 

Certainly not the most elegant solution 😅 . 
I'll fix it now by hand but
will review the script that builds the .yaml next week and make a proposal for a proper pipeline that we can use. 

And if it works well we can open it for outside contribs as well.

aeneasr (approved):

PR #1136, opened 2021-03-09
fix: simplify password breaches error message

mattbonnell:
## Proposed changes

This PR simplifies the error reason given when a password has appeared in data breaches to not include the actual number and rather just show "this password has appeared in data breaches and must not be used". We got some feedback from our design people that the current error message is probably too technical for the average user and I'm inclined to agree. Open to hearing other perspectives on this 😄 .

## Checklist

      the changes.
      works.


aeneasr:
LGTM however could you do this against master? then we see if test pass :)

PR #1151, opened 2021-03-15
fix: use numerical User ID instead of name to avoid security issues.

harnash:
## Related PR
* https://github.com/ory/hydra-login-consent-node/pull/91

## Proposed changes

Our docker image scanner does not allow running processes inside
container using non-numeric User spec (to determine if we are trying
to run docker image as root).## Checklist

      the changes.
      works.


aeneasr:
Is there a particular reason why numerical users are more secure than user aliases?

harnash:
> Is there a particular reason why numerical users are more secure than user aliases?

Probably I've phrased it wrong. It is more Quality of Life improvement than a security concern. When running `oryd/kratos:v0.5.5-alpha.1-sqlite` in our kubernetes cluster I get the following error:

```
 Error: container has runAsNonRoot and image has non-numeric user (ory), cannot verify user is non-root (pod: "kratos-dev-6cd4454c95-7s5wj │
│ _dev(61b6b10c-6b17-4984-a807-2f566669b938)", container: kratos)
```

source: https://github.com/kubernetes/kubernetes/blob/5648200571889140ad246feb82c8f80a5946f167/pkg/kubelet/kuberuntime/security_context.go#L88

aeneasr:
I see, that makes sense then! Thank you :)

Issue #1137, opened 2021-03-10
Bcrypt algorithm support


seremenko-wish:
**Is your feature request related to a problem? Please describe.**

1) Currently, Kratos only supports the Argon2id algorithm for password hashing. This algorithm is configured by 4 settings (parallelism, memory cost, number of iterations, and hash length), and there are no official recommendations about safe settings for this specific version of the algorithm. 

2) The fact that Argon2 has to be calibrated for a specific environment, makes it cumbersome to use.

2) Missing support of the bcrypt algorithm blocks the importing user credentials feature described here https://github.com/ory/kratos/issues/605.

**Describe the solution you'd like**

Bcrypt is a secure and well-established algorithm for password hashing, and it requires only one `cost` setting.

Wiki: https://en.wikipedia.org/wiki/Bcrypt

**Additional context**

The recommended minimum cost value is 10, but for the better tradeoff between security and user experience, that value can be increased to 12.

leovanhaaren:
Would be nice if Bcrypt was supported yes, for us this is also a must-have and its a common algorithm.

leovanhaaren:
@aeneasr alright thanks for the answer, we are currently evaluating ory/kratos, so If I understand it correctly then this should land soon? (the importing of bcrypt passwords and auto converting them)

The above would also be enough for us, we just dont want every user having to manually to reset their password.

aeneasr:
While I understand that a date of delivery is helpful when planning, we stopped giving out due dates for features or milestones. It puts maintainers in an unfair spot as

1. there are many things going on in the community which need time (reviewing PRs, addressing issues, answering questions) and demand is no not easily foreseeable;
2. maintainers do a lot of things in their free time and can not commit to deliver something;
3. internal priorization can shift at any point in time

When we do give out due dates, we often encounter people who are pissed off because something was promised and not delivered "on time" (as free software ;) ). Unfortunately that is the sad truth of many interactions in this context which is why we stopped doing it altogether.

Having said that, you are in the privileged position of being part in a global open source community! If you think that you can contribute towards the particular feature, maintainers will do their best to give you the right pointers and review PRs.

If that's not an option, you can consider to become a Ory Sponsor on [Open Collective](https://opencollective.com/ory) or [Patreon](https://www.patreon.com/_ory). This helps us to employ more maintainers and technical staff - increasing velocity in development, community, interaction, and other areas! All collected money goes directly into this.

seremenko-wish:
[...]

Issue #605, opened 2020-07-27
Add ability to import user credentials


aeneasr:
**Is your feature request related to a problem? Please describe.**

It should be possible to import user/identity credentials when creating identities using the `POST /identities` Admin API.

**Describe the solution you'd like**

There needs to be some type of value (or a dedicated API?) which allows the client to specify the credential. For the credential type, we should be able to work with passwords hashed with

- [ ] BCrypt;
- [ ] Argon2*;
- [ ] PBKDF1/2.

We could also allow other algorithms, such as SHA with rounds or similar methods. We should probably have a list of software platforms and their hashing algorithms to decide what to support. Custom or insecure mechanisms (salted md5?) should probably not be supported in order to keep the code base tidy.

It should also be possible to import OIDC linked accounts.

The idea would be to specify a payload, for example:

```
{
  "traits": ...
  "import_credentials": [
    {
      "type": "password",
      "hash": "...",
      "algorithm": "bcrypt"
    },
    {
      "type": "oidc",
      "provider": "google",
      "user_id": "123456"
    }
  ]
}
```

On the next login, the password would then be converted to a secure Argon2 hash and the original hash will be removed from the database.

**Describe alternatives you've considered**

Instead of providing the key `import_credentials` there would be a collection of endpoints at `/identities/<identity>/credentials` which would allow for PUT, POST, DELETE.

aeneasr:
@Jasrags do you have any preference regarding the API? We're currently thinking about having a dedicated endpoint for that which would mean that you need to do two requests. One to `POST /identities` and then one to `POST /identities/<identity>/credentials`. We'd probably add a CLI helper to help with this.

Jasrags:
@aeneasr I think that could work. I don't have any preferences at the moment.

aeneasr:
Sounds good @Jasrags if you have any other troubles with migrating over to ORY Kratos let me know :)

zcmack:
i'll pile on and say that i have a very, very similar use case and that having a dedicated endpoint / two requests would be just fine. this may be out of scope for this issue, but I figure I'll share for context: in an evaluation of FusionAuth this behaved similarly and worked very well, I was able to use their SDK to easily script a migration however the system would not allow for users to have duplicate email addresses by design. this doesn't work so well for a large public site in which having multiple accounts bound to a single email address is perfectly allowable.

ykuzma1:
[...]

PR #1169, opened 2021-03-22
feat: Bcrypt algorithm support

seremenko-wish:
## Related issue
#1137 @aeneasr 

## Proposed changes
- New `hashers.algorithm` option in config to control password hashing algorithm. By default, the value is `argon2`. In order to use the Bcrypt hashing algorithm, `bcrypt` value should be used. Also, added a setting to control the cost for Bcrypt algorithm (`hashers.bcrypt.cost`) with default value `12`.
- Bcrypt algorithm truncates all input strings to 72 bytes, so a corresponding check was added
- Tests for Bcrypt hasher

## Checklist
      the changes.
      works.

## Further comments
- Change of `hashers.algorithm`should be considered a breaking change since passwords hashed by Argon2 won't be properly validated after this change. In order to handle this case properly, hashers should be separated from classes that check hashes. Should we create a separate issue for this?
- Returning an error when a user tries to use very long password when the Bcrypt algorithm is active, seems like a user-friendly behavior. I would like to get feedback about this.


aeneasr (changes requested):
Thank you for working on this, it looks good! One thing we need to take care of however is that we can't change the algorithm if we have users signed up already, as those users would no longer be able to log in.

This means that we need a password upgrade here: https://github.com/ory/kratos/blob/master/selfservice/strategy/password/login.go#L149

The code for generating the password entry can be found at https://github.com/ory/kratos/blob/master/selfservice/strategy/password/registration.go#L177-L196

It would probably also make sense to add a section about bcrypt here: https://github.com/ory/kratos/blob/master/docs/docs/concepts/credentials/username-email-password.mdx

And lastly, it's actually fine by me if we use BCrypt per default, as the kratos-typical workloads are web based, meaning that Argon2 might underperform.

driver/config/config.go
```diff
@@ -732,3 +748,14 @@ func (p *Config) PasswordPolicyConfig() *PasswordPolicy {
 		IgnoreNetworkErrors: p.p.BoolF(ViperKeyIgnoreNetworkErrors, true),
 	}
 }
+
+func (p *Config) HasherPasswordHashingAlgorithm() string {
+	configValue := p.p.StringF(ViperKeyHasherAlgorithm, DefaultPasswordHashingAlgorithm)
+	switch configValue {
+	case "bcrypt":
+		return configValue
+	case "argon2":
```
zepatrik, line 757: Switches in go break by default :wink: 
```suggestion
	case "argon2":
		fallthrough
```

seremenko-wish:
Last commit fixes https://github.com/ory/kratos/issues/1175

seremenko-wish:
> Thank you for working on this, it looks good! One thing we need to take care of however is that we can't change the algorithm if we have users signed up already, as those users would no longer be able to log in.

I separated the hash comparator from hashers, so changing of hashing algorithm won't prevent users from logging in.

> This means that we need a password upgrade here: https://github.com/ory/kratos/blob/master/selfservice/strategy/password/login.go#L149
fixed

> The code for generating the password entry can be found at https://github.com/ory/kratos/blob/master/selfservice/strategy/password/registration.go#L177-L196
this code does not need any changes as it uses the hasher specified in the configuration

> It would probably also make sense to add a section about bcrypt here: https://github.com/ory/kratos/blob/master/docs/docs/concepts/credentials/username-email-password.mdx
added

> And lastly, it's actually fine by me if we use BCrypt per default, as the kratos-typical workloads are web based, meaning that Argon2 might underperform.
For backwards compatibility, I left Argon2id as the default hasher. We can create a separate task to make that switch in the future.

hash/hasher_bcrypt.go
```diff
@@ -0,0 +1,43 @@
...
+
+	return hash, nil
+}
+
+func validateBcryptPasswordLength(password []byte) error {
+	// Bcrypt truncates the password to the first 72 bytes, following the OpenBSD implementation,
+	// so if password is longer than 72 bytes, function returns an error
+	// See https://en.wikipedia.org/wiki/Bcrypt#User_input
+	if len(password) > 72 {
+		return errors.New("passwords are limited to a maximum length of 72 characters")
```
zepatrik, line 40: This error will be passed down to the UI I guess. Please add it to the messages to allow translations and mapping to other text: https://github.com/ory/kratos/blob/1940a679eb6b695e22cb939fb0d8d85cebb82a1e/text/message_registration.go

I guess you would then in the flow check for this error you return here similar to https://github.com/ory/kratos/blob/4bda1f09472bd65608cafe0d5d44287487da6f70/selfservice/flow/registration/error.go#L92

But maybe @aeneasr can give better pointers.

zepatrik (changes requested):
Nice, this looks really good :+1:

driver/config/.schema/config.schema.json
```diff
@@ -1294,6 +1294,12 @@
       "title": "Hashing Algorithm Configuration",
       "type": "object",
       "properties": {
+        "algorithm": {
+          "title": "Password hashing algorithm",
+          "description": "One of the values: argon2, bcrypt",
+          "type": "string",
+          "default": "argon2"
+        },
```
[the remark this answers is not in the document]
seremenko-wish: I would set the minimum to 12 as well. There is no official recommendation, and I found a few posts in StackOverflow that point out that the minimum cost in 2021 for web-based password hashing shouldn't be less than 12. See this post for example https://stackoverflow.com/a/61304956

hash/hash_comparator.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, nil, nil, ErrIncompatibleVersion
+	}
+
+	p = new(config.Argon2)
+	_, err = fmt.Sscanf(parts[3], "m=%d,t=%d,p=%d", &p.Memory, &p.Iterations, &p.Parallelism)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	salt, err = base64.RawStdEncoding.Strict().DecodeString(parts[4])
```
aeneasr, line 94: Will using `Strict()` here cause issues with existing hashes?

hash/hasher_bcrypt.go
aeneasr: Right, it would be an error like this one (actually we can just add it do that file for now): https://github.com/ory/kratos/blob/master/schema/errors.go#L59-L70

aeneasr (commented):
Awesome, just two tiny things but this is almost good to ship!

hash/hash_comparator.go
seremenko-wish: No, all correct existing hashes will be ok. Strict() just does not allow to decode invalid base64 encoded strings.

aeneasr: Great, thank you!

aeneasr (approved):

aeneasr:
The last thing is getting tests green - so running `make format` will resolve the one issue. The e2e tests have been failing for a while now on master so we need for someone to fix those first. I will probably not be able to do that this week :/

aeneasr:
I have resolved the e2e tests so this should work again - rerunning tests now.

seremenko-wish:
ok. all checks are green now!

aeneasr:
Awesome, thank you so much for your hard work!

hash/hash_comparator.go
```diff
@@ -0,0 +1,107 @@
...
+	return ErrMismatchedHashAndPassword
+}
+
+func isBcryptHash(hash []byte) bool {
+	res, _ := regexp.Match("^\\$2[abzy]?\\$", hash)
+	return res
+}
+
+func isArgon2idHash(hash []byte) bool {
+	res, _ := regexp.Match("^\\$argon2id\\$", hash)
```
harnash, line 69: Nitpicking: wouldn't `strings.HasPrefix()` be sufficient in this case?

seremenko-wish: Agree, that could be a good simplification. Not sure if it is worth doing it now. I used this approach as it will allow adding more algorithms in the future and user regexes for identifying them.

harnash: Nah. I think is not worth separate PR right now. When/if we tackle with #605 then that would be a good time to refactor this. 

Anyway @seremenko-wish thanks a lot for this PR, this will get us closer to get all critical features implemented. Is the #1175 closable now since hash comparison is fixed in this PR?

seremenko-wish: closed #1175

this definitely opens gates for more algorithm integrations. Looks like PBKDF1/2 integration is missing, but that would be easy to add into existing hashing framework.

aeneasr: Yup! Not sure if we want to support other algorithms as main hasher but definitely support them when importing users/hashes! This PR lays the foundation for doing that!

seremenko-wish:
@aeneasr would you want to add this feature to the next version release?

aeneasr:
Yes, this is planned for 0.6

I am currently working on pushing out 0.6 this month!

aeneasr:
Looks like OWASP is also not quite clear themselves as to which algorithm to support as they have recently updated their docs to encourage use of Argon2id whenever possible: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pre-hashing-passwords

seremenko-wish:
Yeah. That's fine. I saw that they recommend some argon2 settings. It would be interesting to measure performance with those settings.

btw, they encourage not to use any pre-hashing functions with bcrypt https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pre-hashing-passwords

aeneasr:
Yes, thank you! I also saw that in a talk someone mentioned in another issue ;)

PR #1179, opened 2021-03-27
fix: docs build

zepatrik:
## Related issue

Malformed sidebar

### DIFF
--- a/docs/sidebar.json
+++ b/docs/sidebar.json
@@ -61,8 +61,6 @@
     "reference/html-forms",
     "reference/api",
     {
-      "type": "category",
-      "label": "Command Line Interface (CLI)",
       "Command Line Interface (CLI)": [
         "cli/kratos",
         "cli/kratos-courier",
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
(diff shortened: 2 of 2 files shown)


docs/sidebar.json
```diff
@@ -61,8 +61,6 @@
     "reference/html-forms",
     "reference/api",
     {
-      "type": "category",
-      "label": "Command Line Interface (CLI)",
       "Command Line Interface (CLI)": [
```
aeneasr, line 64: This won't fix the docs. Revert the changes and:

```suggestion
      "items": [
```

aeneasr (changes requested):

docs/sidebar.json
zepatrik: Why would it not?

https://docusaurus.io/docs/sidebar#creating-a-hierarchy

> Note: it's possible to use the shorthand syntax to create nested categories:
> ```
> module.exports = {
>   docs: {
>     Guides: [
>       'creating-pages',
>       {
>         Docs: [
>           'docs-introduction',
>           'docs-sidebar',
>           'markdown-features',
>           'versioning',
>         ],
>       },
>     ],
>   },
> };
> ```

aeneasr: Looks like I learned something new ;)

Issue #1152, opened 2021-03-15
Routing library doesn't incorporate base_url


snoepkast:
**Describe the bug**

As discussed in Slack.

I am running Kratos (and others) behind a proxy (Traefik) which is configured to serve the components from different paths:

```
localhost/kratos
localhost/kratos-self-service-ui
localhost/hydra
etc
```

I noticed that none of the paths seem to be recognised when running on anything other than / so I configured Traefik to strip the path prefix before forwarding the request, now Kratos will serve the request like it should, but when it redirects, it redirects to / instead of /kratos/.

In Slack it was explained to me that the urls from the configuration are only used for explicit redirects and not in the routing library, which also needs to take the base_url into account.

**Reproducing the bug**

You can reproduce by checking:

```
localhost/kratos/sessions/whoami  -> 404
localhost/kratos/sessions/whoami (with traefik stripping /kratos path prefix) -> 200
localhost/kratos/sessions/whoami/ (notice the trailing slash) -> redirects to: localhost/sessions/whoami -> 404
```

**Server configuration**

```yaml
serve:
  public:
    base_url: http://localhost/kratos/
```

**Expected behavior**

When setting the `base_url` I expect Kratos to serve it's endpoints relative to this, and also take the base_url into account when redirecting.

aeneasr:
That would probably need some work with httprouter's APIs. It's a good issue to get into the code base if anyone's up for that!

snoepkast:
I was just thinking about this issue and while it's good to take the base_url into account, for this particular case it might be better to ignore trailing slashes? I mean, does it really matter if there's a trailing slash or not? Instead of a redirect we could just serve the request?

aeneasr:
Agreed, that's what I proposed here: https://github.com/ory/kratos/pull/1181#pullrequestreview-622991591

redanvil:
i ran into this as well and as a band-aid did something like this with traefik:

```
  kratos:
    image: oryd/kratos:v0.10.1
    labels:
      
... rest of labels

      - "traefik.http.routers.sessions.entrypoints=web"
      - "traefik.http.routers.sessions.rule=PathPrefix(`/sessions`)"
      - "traefik.http.routers.sessions.middlewares=redirect@docker,kratos-strip-prefix@docker"
      - "traefik.http.routers.sessions.service=sessions"
      - "traefik.http.services.sessions.loadbalancer.server.port=4433"
    
[...]

PR #1181, opened 2021-03-29
feat: strip base url path prefix while routing

mrosales:
Hello! This is my first kratos PR and I'm still relatively unfamiliar with the code base, so I'm happy to take suggestions about other ways to do this that might be more in line with existing ORY practices as well as feedback about style or code conventions.

## Related issue

#1152 

## Proposed changes

This change is aimed at supporting serving Kratos on a shared hostname behind a path prefix. If a `base_url` config value for the `public` or `admin` server contains a path, Kratos will assume that the entire server is running behind that path and strip the prefix from incoming requests. Requests that do not match the leading prefix will receive a 404 response.

## Checklist

      the changes.
      works.

## Further comments

Design assumptions:
* If the base url is specified as `https://my.domain.com/kratos`, Kratos can assume that requests **must** have a leading `/kratos` prefix.
* It is not desirable to interpret trailing slashes on a base url as a prefix to strip. (`https://my.domain.com/` does not result in `/` being stripped from incoming requests)
* It's acceptable to add a global middleware for this rather than changing the root `http.Handler` at launch.
    * It appeared as if a middleware was necessary because the config field was dynamic, but it looks like `base_url` is marked as immutable so `kratos` doesn't actually support changing this on the fly.

Open Questions

* I could use some guidance on if there are better places to set up tests for this behavior. I only added unit tests for the added middleware functions, but nothing that actually tests the full configuration. I only validated this by running the quickstart with a path prefix.
* There seems to be a mix of test styles/structures? Is there a preferred way to structure unit tests for this project?
* Would it be preferred to just read the base path prefix at startup and add a static middleware?
* Should this address the `domain_aliases` config in some way? I wasn't exactly sure how to interact with those properties.

### DIFF
--- a/cmd/daemon/middleware.go
+++ b/cmd/daemon/middleware.go
@@ -2,15 +2,18 @@ package daemon
 
 import (
 	"net/http"
+	"net/url"
+	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/negroni"
 
-	"github.com/ory/x/logrusx"
-
 	"github.com/ory/x/healthx"
+	"github.com/ory/x/logrusx"
 	"github.com/ory/x/reqlog"
+
+	"github.com/ory/kratos/driver/config"
 )
 
 func NewNegroniLoggerMiddleware(l *logrusx.Logger, name string) *reqlog.Middleware {
@@ -33,4 +36,34 @@ func NewNegroniLoggerMiddleware(l *logrusx.Logger, name string) *reqlog.Middlewa
 		})
 	}
 	return n
+}
+
+func NewStripPrefixMiddleware(prefixLoader func(r *http.Request) string) negroni.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+		prefix := strings.TrimRight(prefixLoader(r), "/")
+		if len(prefix) == 0 {
+			next.ServeHTTP(w, r)
+			return
+		}
+		http.StripPrefix(prefix, next).ServeHTTP(w, r)
+	}
+}
+
+func extractPrefixFromBaseURL(u *url.URL) string {
+	if u == nil {
+		return ""
+	}
+	return strings.TrimSuffix(u.Path, "/")
+}
+
+func publicURLPrefixExtractor(provider config.Provider) func(r *http.Request) string {
+	return func(r *http.Request) string {
+		return extractPrefixFromBaseURL(provider.Config(r.Context()).SelfPublicURL(r))
+	}
+}
+
+func adminURLPrefixExtractor(provider config.Provider) func(r *http.Request) string {
+	return func(r *http.Request) string {
+		return extractPrefixFromBaseURL(provider.Config(r.Context()).SelfAdminURL())
+	}
 }
--- a/cmd/daemon/middleware_test.go
+++ b/cmd/daemon/middleware_test.go
@@ -1,0 +1,212 @@
+package daemon
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/ory/x/configx"
+	"github.com/ory/x/logrusx"
+
+	"github.com/ory/kratos/driver/config"
+)
+
+func TestNewStripPrefixMiddleware(t *testing.T) {
+	tests := []struct {
+		name         string
+		target       string
+		prefixLoader func(r *http.Request) string
+		wantNext     bool
+		wantPath     string
+		wantStatus   int
+	}{
+		{
+			name:         "empty prefix 1",
+			target:       "http://127.0.0.1:1234",
+			prefixLoader: func(r *http.Request) string { return "" },
+			wantNext:     true,
+			wantPath:     "",
+		},
+		{
+			name:         "empty prefix 2",
+			target:       "http://127.0.0.1:1234/foo",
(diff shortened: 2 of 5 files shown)


cmd/daemon/middleware.go
```diff
@@ -34,3 +37,33 @@ func NewNegroniLoggerMiddleware(l *logrusx.Logger, name string) *reqlog.Middlewa
...
+func extractPrefixFromBaseURL(u *url.URL) string {
+	if u == nil {
+		return ""
+	}
+	return strings.TrimSuffix(u.Path, "/")
+}
+
+func publicURLPrefixExtractor(provider config.Provider) func(r *http.Request) string {
+	return func(r *http.Request) string {
+		return extractPrefixFromBaseURL(provider.Config(r.Context()).SelfPublicURL(r))
```
aeneasr, line 61: Wouldn't it make more sense to do this in `SelfPublicURL` / `SelfAdminURL` so that the result is the same across the code base?

aeneasr (changes requested):
Thank you for your work on this!

As it is right now, this will be hard to accept as it introduces a serious backwards incompatible change and it acts differently from the other ory software (e.g. Ory Hydra). Generally the idea is that you use something like `strip_path: true` in your reverse proxy (nginx) configuration which is supported by all reverse proxies I know.

The original issue was about things like double slashes `http://foo/bar//` and also a redirect feature in httprouter https://godoc.org/github.com/julienschmidt/httprouter#Router.RedirectTrailingSlash which was causing the incorrect redirect as httprouter does not know the full URL but instead only sees the http.Request path, which is - usually - the reverse proxy stripped path!

mrosales:
Thanks for the quick feedback @aeneasr ! I have a few clarifying questions and then I'll try to make another pass that hopefully addresses some of the issues here.

> Generally the idea is that you use something like `strip_path: true` in your reverse proxy (nginx) configuration which is supported by all reverse proxies I know.

Yep, makes sense, using the proxy to strip the prefix makes sense and is more broadly understood. If we're to rely on a third party to strip the path, what would the recommended way to set up tests for this be? Just to make sure that the redirects add the full base path and that the trailing slash routes get resolved correctly?

> The original issue was about things like double slashes `http://foo/bar//` and also a redirect feature in httprouter https://godoc.org/github.com/julienschmidt/httprouter#Router.RedirectTrailingSlash which was causing the incorrect redirect as httprouter does not know the full URL but instead only sees the http.Request path, which is - usually - the reverse proxy stripped path!

I think I misread that from the original issue and was just taking a stab at a more broad "make serving under a base url prefix work". After re-reading the original reporter's post and your reply here, it seems like the part that is misbehaving is that when serving kratos behind a path-stripping proxy is just the trailing slash redirect provided by `httprouter`, but not the actual API route if you configure path stripping properly. Does that sound correct?

aeneasr:
> I think I misread that from the original issue and was just taking a stab at a more broad "make serving under a base url prefix work". After re-reading the original reporter's post and your reply here, it seems like the part that is misbehaving is that when serving kratos behind a path-stripping proxy is just the trailing slash redirect provided by httprouter, but not the actual API route if you configure path stripping properly. Does that sound correct?

Yes exactly!

> If we're to rely on a third party to strip the path, what would the recommended way to set up tests for this be? Just to make sure that the redirects add the full base path and that the trailing slash routes get resolved correctly?

Do you mean tests in Ory Kratos itself or as an infrastructure test?

mrosales:
So @aeneasr I took a look at the `httprouter` API and am finding [the following snippet](https://github.com/julienschmidt/httprouter/blob/v1.2.0/router.go#L350) that redirects directly using the value of `r.URL.String()` as the redirect URL.

I think I'll need some guidance around what you think the cleanest next step would be. A few options might be:

1. Disable the trailing slash redirects (possibly only when a base url path prefix is set). If the path is being stripped by a proxy, the proxy could rewrite trailing slashes too. Alternatively just document that trailing slash redirects do not work with a path prefix and add a config option to disable that routing option.
2. Make the official recommendation to not strip the prefix in the proxy and match on the full prefix path in kratos. This might break existing integrations that rely on path-stripping.
3. (hacky because it assumes the path is already stripped) modify URL to contain the prefix and then route on the full path
4. Some set of options to implement a subset of the above.

aeneasr:
Thank you for the summary!

> Disable the trailing slash redirects (possibly only when a base url path prefix is set). If the path is being stripped by a proxy, the proxy could rewrite trailing slashes too. Alternatively just document that trailing slash redirects do not work with a path prefix and add a config option to disable that routing option.

I think this is probably the smartest thing to do! It's easy to debug because the error is pretty obvious - instead of e.g. some random redirect which you don't understand!

aeneasr:
I am closing this as per https://github.com/ory/kratos/pull/1181#pullrequestreview-622991591

abador:
@aeneasr I've been trying to prepare some backwards compatible version of this: https://github.com/Wikia/kratos/pull/61 Would it be acceptable for you? The changes are pretty straightforward

aeneasr:
To be honest, I'm not sure if we should add this. Whenever you introduce path-based routing, an API gateway will sit in front of Ory Kratos. That gateway is responsible for routing, and also responsible to appropriately construct the URL it sends to the upstream (path + host + port + scheme). Mixing this into Ory Kratos feels as if we are doing what the API gateway *should* be doing in the first place. Most server tools I know of delegate this responsibility to the router that sits in front. Do you see this differently?

aeneasr:
If the CDN supports path-based routing but not path stripping, I would raise this issue with the CDN provider to support path stripping, as it will affect other open source services as well (in particular all of Ory's system) :)

Issue #1188, opened 2021-03-30
Unicode-escaped ampersand in query string returned by Admin API recovery link endpoint


fenech:
**Describe the bug**

Ampersand `&` escaped to `\u0026` in recovery link returned from [Admin API recovery link endpoint](https://www.ory.sh/kratos/docs/reference/api/#create-a-recovery-link). The escape sequence must be replaced by `&` in order to use the link.

**Reproducing the bug**

1. Run Kratos following Quickstart instructions (SQLite DB can be used)
2. Add a user, making a note of user's ID
3. Send a cURL request to the recovery link endpoint of the admin API. replacing the `identity_id` value with the newly-created user's ID:
    ```sh
    curl -X POST http://localhost:4434/recovery/link -H 'Content-Type: application/json' \
    -d '{"identity_id": "57ff394d-4667-4f5f-95db-db2f6f153741", "expires_in":"12h"}'
4. The JSON response is of the form:
    ```json
    {
      "recovery_link":"http://127.0.0.1:4455/.ory/kratos/public/self-service/recovery/methods/link?flow=ac1ce7ac-76df-481e-84de-9df59ee5d33d\u0026token=OqgpDBnwpPNb36Sk4sRFlVrAkPUtVavF",
      "expires_at":"2021-03-30T21:01:50.10467534Z"
    }
    ```
_Server logs_

```
time=2021-03-30T09:11:52Z level=info msg=started handling request method=POST name=admin#http://kratos:4434/ remote=10.0.2.100:36828 request=/recovery/link
time=2021-03-30T09:11:52Z level=info msg=A recovery link has been created. audience=audit email_address=Value is sensitive and has been redacted. To see the value set config key "log.leak_sensitive_values = true" or environment variable "LOG_LEAK_SENSITIVE_VALUES=true". identity_id=57ff394d-4667-4f5f-95db-db2f6f153741 recovery_link_token=Value is sensitive and has been redacted. To see the value set config key "log.leak_sensitive_values = true" or environment variable "LOG_LEAK_SENSITIVE_VALUES=true". service_name=ORY Kratos service_version=v0.5.4-alpha.1 via=email
time=2021-03-30T09:11:52Z level=info msg=completed handling request method=POST name=admin#http://kratos:4434/ remote=10.0.2.100:36828 request=/recovery/link status=200 text_status=OK took=2.767399ms
```

_Server configuration_

```yaml
session:
  lifespan: 24h
serve:
  public:
[...]

PR #1212, opened 2021-04-06
fix: Avoid unicode-escaping ampersand in recovery URL query string

fenech:
## Related issue

https://github.com/ory/kratos/issues/1188
@aeneasr 

## Proposed changes

Specify option to output JSON without escaping HTML entities, to keep the `&` in the query string intact.

## Checklist

      the changes.
      works.

## Further comments

Tested manually using cURL, but I'm not sure about the best way to add automatic tests for this.

I ran `go get github.com/ory/herodot@v0.9.5` `go mod tidy`, which led to the changes in `go.mod` and `go.sum`.

### DIFF
--- a/selfservice/strategy/link/strategy_recovery.go
+++ b/selfservice/strategy/link/strategy_recovery.go
@@ -191,7 +191,8 @@ func (s *Strategy) createRecoveryLink(w http.ResponseWriter, r *http.Request, _ 
 			url.Values{
 				"token": {token.Token},
 				"flow":  {req.ID.String()},
-			}).String()})
+			}).String()},
+		herodot.UnescapedHTML)
 }
 
 // swagger:parameters completeSelfServiceRecoveryFlowWithLinkMethod
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/gobuffalo/uuid v2.0.5+incompatible
 	github.com/gofrs/uuid v3.2.0+incompatible
 	github.com/golang/gddo v0.0.0-20190904175337-72a348e765d2
-	github.com/golang/mock v1.4.4
+	github.com/golang/mock v1.5.0
 	github.com/google/go-github/v27 v27.0.1
 	github.com/google/go-jsonnet v0.16.0
 	github.com/google/uuid v1.2.0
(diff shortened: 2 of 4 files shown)


aeneasr (approved):

selfservice/strategy/link/strategy_recovery.go
```diff
@@ -191,7 +191,8 @@ func (s *Strategy) createRecoveryLink(w http.ResponseWriter, r *http.Request, _
 			url.Values{
 				"token": {token.Token},
 				"flow":  {req.ID.String()},
-			}).String()})
+			}).String()},
+		herodot.UnescapedHTML)
```
aeneasr, line 195: Could you please add a test for this?

fenech: As I mentioned in the PR description, I wasn't sure about the best way to do it, do you have any suggestions?

Since `"\u0026" == "&"` in Go, I'm not sure how we could test it from within `checkLink`, defined here https://github.com/ory/kratos/blob/master/selfservice/strategy/link/strategy_recovery_test.go#L56.

I guess that the way to do it would involve looking at the raw body of the response, before it is decoded into `*admin.CreateRecoveryLinkOK`, but I'm not sure how to go about doing that in the current tests.

aeneasr: Hm, it also appears that current tests fail due to this? Basically `\u0026` is `&` in UTF-8 speak so all clients that are capable of decoding UTF8 should know what to do here. What library where you facing issues with here?

fenech: The original issue was observed using `curl` on the command line: the link returned was not clickable.

aeneasr: I see, could you try `curl ... | jq`? I guess curl is returning the raw payload

fenech: Yeah, that was what I did to get around it in my script :smile:

aeneasr: Ok, unless we find a good solution on how to fix the tests or figure out why they fail in the first place we might have to stick with the UTF8 encoded variant, although I agree that it is a bit counterintuitive

fenech:
@aeneasr the tests are passing now - the issue was that Encode adds a newline to the end of the JSON output.

aeneasr (approved):

aeneasr:
Thank you!

PR #1239, opened 2021-04-16
feat: enable concurrent courier workers

mattbonnell:
## Proposed changes

Use `FOR UPDATE` row-locking during message fetching transactions to enable the use of multiple concurrent courier workers.

The `FOR UPDATE` clause will cause the DB to lock the given rows for the duration of the fetching transaction, preventing other workers from selecting the same message rows before one is able to update their status to "processing".

This will make workers trying to fetch at the same time form an orderly queue. 

If CockroachDB supported `SKIP LOCKED`, we could use that clause too, which would cause workers to just ignore locked rows instead of waiting, but support for this has yet to be added to CockroachDB.

`FOR UPDATE` is supported by MySQL, Postgres and CockroachDB.

Sqlite does not support `FOR UPDATE` so it is not used when Sqlite is being used.

## Checklist

      the changes.
      works.

### DIFF
--- a/persistence/sql/persister_courier.go
+++ b/persistence/sql/persister_courier.go
@@ -29,7 +29,7 @@ func (p *Persister) NextMessages(ctx context.Context, limit uint8) (messages []c
 		var m []courier.Message
 		query := "SELECT * FROM %s WHERE status = ? ORDER BY created_at ASC LIMIT ?"
 		if !p.isSQLite {
-			query += " FOR UPDATE SKIP LOCKED"
+			query += " FOR UPDATE"
 		}
 		if err := tx.
 			RawQuery(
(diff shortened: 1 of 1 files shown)


persistence/sql/persister_courier.go
```diff
@@ -26,10 +26,17 @@ func (p *Persister) AddMessage(ctx context.Context, m *courier.Message) error {
 func (p *Persister) NextMessages(ctx context.Context, limit uint8) ([]courier.Message, error) {
 	var m []courier.Message
 	if err := p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) error {
+		query := "SELECT * FROM %s WHERE status = ? ORDER BY created_at ASC LIMIT ?"
```
aeneasr, line 29: This makes sense to me! One question would be if you could add a test for this. My proposal is:

1. Create 50 messages
2. Run 50 `NextMessages` in goroutines with limit of 1
3. Expect to receive 50 different messages

mattbonnell: Yeah will do!

aeneasr:
Whoops, missed this - should I take another look?

aeneasr:
Are you still up for the changes? :monocle_face: If you need any help, let us know!

I'll convert this to a draft in the meanwhile. When it's ready for review again, please remove the draft indicator! 👀

aeneasr:
I'm closing this to clean up our backlog. Feel free to reopen if you come back to this :)

Issue #271, opened 2020-03-01
Feature Request: Webhooks


driverpt:
**Is your feature request related to a problem? Please describe.**
In a Microservices based environment, when a new user is registered and/or activated, we have the need to provision resources in other services for the newly created/activated user.

**Describe the solution you'd like**
Everytime a user is created and/or activated, we'd like support to have Webhooks to other services with a well-known JSON Structure.

e.g.:
User A registers in Kratos. (Webhook triggered for Services A which posts in a Message Queue the event). Service D, E and F, provision resources for this user (e.g.: Storage, Security and ACL).
User A is activated in Kratos (Email Activation). All Resources are ready to go.

**Describe alternatives you've considered**

I've considered creating a microservice and have and endpoint for everytime a user logs in, but seems a bit overkill to have an endpoint +1 redirect just to do these checks and provisioning when it can be done via WebHook

**Additional context**

None

aeneasr:
Yes, that was on the roadmap :) Triaging.

This is btw also supported with after/before hooks - all we need is a JSON RPC handler there, contributions welcome!

zepatrik:
I think we should only send the identity's id though so you can gather any data yourself.

aeneasr:
It's also important to note that there are two types of webhooks:

- Broadcast which would just send out an event such as "user.signedup" with some identity data
- RPC which would send and also receive data to modify the identity for example

The second is pretty difficult to achieve as it also has a certain toll on request latency, the first one is pretty easy to achieve.

driverpt:
OK. I'll give it a look and try to make a PR. Is there any internal message queue in the code for these events so we can just create Event Handlers?

zepatrik:
Nope, the only messaging we have right now is the email verification. Webhooks should be implemented like the other "hooks" (sorry for the bad wording here). That way you can easily configure them as pre/post registration/login/...
Have a look in the `selfservice/hook` module.

aeneasr:
Here's the list of relevant code lines:

- Hook names: https://github.com/ory/kratos/blob/master/selfservice/hook/hooks.go#L7
- Hook interface for post/pre sign up: https://github.com/ory/kratos/blob/master/selfservice/flow/registration/hook.go#L17-L22
- Hook interface for post/pre login: https://github.com/ory/kratos/blob/master/selfservice/flow/login/hook.go#L12-L17
[...]

PR #1240, opened 2021-04-16
feat: web hooks support

dadrus:
## Related issue

@aeneasr: This is the promised WIP for #271 (Feature Request: Webhooks)

## Proposed changes

As discussed in the linked ticket this PR describes the idea of a possible implementation for web hooks. The support is currently added to `verification` and `registration` flows only and should be the basis for further discussion. As of now, the implementation is very simple - no retries, no asynchrony. The actual implementation of how the configuration is parsed and applied is also very simple and should be changed to be more concise. 

The configuration for the web-hooks goes into the `hooks` property and has the following structure:

```.yaml
hooks:
  - hook: web-hook
    config:
      url: https://test.kratos.ory.sh/after_verification_hook
      method: POST # GET, DELETE, etc
      body: /link/to/my/jsonnet/file
      auth:
        type: <some-type>
        config:
          <type-specific-config>
```

The `flow` and `session` objects are currently bound into the jsonnet template. E.g. to create a body looking like `{ user_id: <some-id> }` to be send to the web hook end point, the jsonnet template would look like this: `function(flow, session) { user_id: session.identity.id }`

For `auth` following types are implemented:
* api-key
* basic-auth

For `api-key` the config looks as follows:

```.yaml
name: Some-Name
value: The-Value-of-My-Key
in: header # alternatively cookie
```

For `basic-auth` the config looks as follows:
```.yaml
user: My-User
password: My-Pass-Value
```

The `after` property of `registration` and `verification` has been extended to contain a `global` property, which consists only out of a `hooks` property, which can contain as many web-hooks in the above said format as desired.

The `hooks` property of `oidc` and `password` properties of the aforesaid flows were extended to contain the web-hook item in addition to already defined hook types.

## Checklist

      the changes.
      works.

## Further comments

As the support for web-hooks touches many flows, I rather would like to start small and discuss the desired changes in the implementation and usage first, before adopting the support to all other flows.

I used the word `auth` for the configuration of "security" settings of a web-hook by intention. What has been done so far is IMHO just a small part of what might be required. I can see the need for custom certificate trust stores to verify the servers certificate, as well as the need to support mTLS. Other mechanisms, or combinations might be required as well.


.docker/Dockerfile-build
```diff
@@ -28,9 +28,9 @@ COPY --from=builder /usr/bin/kratos /usr/bin/kratos
 
 # By creating the sqlite folder as the ory user, the mounted volume will be owned by ory:ory, which
 # is required for read/write of SQLite.
-RUN mkdir -p /var/lib/sqlite
-RUN chown ory:ory /var/lib/sqlite
-VOLUME /var/lib/sqlite
+#RUN mkdir -p /var/lib/sqlite
+#RUN chown ory:ory /var/lib/sqlite
+#VOLUME /var/lib/sqlite
```
dadrus, line 34: Please ignore the changes to this file for now. I'll update it to the original values, when all required changes are implemented. We had to update this file to be able to build kratos locally. Please don't ask why, but it didn't work without these changes on our side.

cmd/jsonnet/lint.go
```diff
@@ -29,12 +32,10 @@ var jsonnetLintCmd = &cobra.Command{
...
+				var outBuilder strings.Builder
+				errorsFound := linter.LintSnippet(vm, &outBuilder, file, string(content))
 
-				ew := &linter.ErrorWriter{Writer: os.Stderr}
-				linter.Lint(node, ew)
-				if ew.ErrorsFound {
+				if errorsFound {
 					_, _ = fmt.Fprintf(os.Stderr, "Linter found issues.")
 					os.Exit(1)
 				}
```
dadrus, line 41: I had to apdate jsonnet to 0.17.0 from 0.16.0 -> breaking API change

driver/config/config.go
```diff
@@ -518,7 +519,8 @@ func (p *Config) SelfPublicURL(r *http.Request) *url.URL {
 
 	var aliases []domainAlias
 	if err := json.NewDecoder(bytes.NewBufferString(raw)).Decode(&aliases); err != nil {
-		p.l.WithError(err).WithField("config", raw).Errorf("Unable to unmarshal domain alias configuration, falling back to primary domain.")
+		// this is happening too often to log it
+		// p.l.WithError(err).WithField("config", raw).Warnf("Unable to unmarshal domain alias configuration, falling back to primary domain.")
 		return primary
```
dadrus, line 524: Will be uncommented as soon as the functionality is complete. It just filled up our logs. It was easier to see what is going on by disabling this log line

selfservice/hook/web_hook.go
```diff
@@ -0,0 +1,187 @@
...
+
+var _ registration.PostHookPostPersistExecutor = new(WebHook)
+var _ verification.PostHookExecutor = new(WebHook)
+
+type (
+	webHookDependencies interface {
+		x.LoggingProvider
+	}
+
+	emptyAuthConfig struct {}
```
dadrus, line 24: This and the following types are actually strategies and not configs.

selfservice/hook/web_hook.go
```diff
@@ -0,0 +1,187 @@
...
+	WebHook struct {
+		r webHookDependencies
+		c json.RawMessage
+	}
+)
+
+func (c *emptyAuthConfig) apply(_ *http.Request) {}
+
+func (c *webHookConfig) Unmarshal(bytes []byte) error {
+	if err := json.Unmarshal(bytes, c); err != nil {
```
dadrus, line 63: This is the place I meant in the PR comment related to how configuration parsing happens and which should be changed.

selfservice/hook/web_hook.go
```diff
@@ -0,0 +1,187 @@
...
+	switch c.In {
+	case "cookie":
+		req.AddCookie(&http.Cookie{ Name: c.Name, Value: c.Value })
+	default:
+		req.Header.Set(c.Name, c.Value)
+	}
+}
+
+func (a *Auth) UnmarshalJSON(bytes []byte) error {
+	type auth Auth
```
dadrus, line 87: And this part as well

selfservice/hook/web_hook.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	var body io.Reader
+	if len(conf.Body) != 0 {
+		body, err = e.createBody(conf.Body, f, s)
+		if err != nil {
+			return fmt.Errorf("failed to create web hook body: %w", err)
+		}
+	}
+	if err = e.doHttpCall(conf, body); err != nil {
+		return fmt.Errorf("failed to call web hook %w", err)
```
dadrus, line 138: One could add an additional strategy here, defining how the actual execution of the Web hook happens, like synchronously, or asynchronously, with or without retries, whether the response should somehow be handled or not, etc.

selfservice/hook/web_hook.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return nil
+}
+
+func (e *WebHook) createBody(jsonnetFile string, f interface{}, s interface{}) (io.Reader, error) {
+	vm := jsonnet.MakeVM()
+
+	buf := new(bytes.Buffer)
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
```
dadrus, line 148: Might be an additional configuration option

selfservice/hook/web_hook.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, err
+	}
+	vm.TLACode("flow", buf.String())
+
+	buf.Reset()
+	if err := enc.Encode(s); err != nil {
+		return nil, err
+	}
+	vm.TLACode("session", buf.String())
+
```
dadrus, line 161: Which further objects should be made available to the jsonnet template?

driver/registry_default.go
```diff
@@ -106,9 +106,10 @@ type RegistryDefault struct {
 	selfserviceSettingsErrorHandler *settings.ErrorHandler
 	selfserviceSettingsExecutor     *settings.HookExecutor
 
-	selfserviceVerifyErrorHandler *verification.ErrorHandler
-	selfserviceVerifyManager      *identity.Manager
-	selfserviceVerifyHandler      *verification.Handler
+	selfserviceVerifyErrorHandler   *verification.ErrorHandler
+	selfserviceVerifyManager        *identity.Manager
+	selfserviceVerifyHandler        *verification.Handler
+	selfserviceVerificationExecutor *verification.HookExecutor
```
dadrus, line 113: Was reported as an issue by golangci-lint

selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -90,7 +90,7 @@ func (s *Strategy) processRegistration(w http.ResponseWriter, r *http.Request, a
 
 	vm := jsonnet.MakeVM()
 	vm.ExtCode("claims", jsonClaims.String())
-	evaluated, err := vm.EvaluateSnippet(provider.Config().Mapper, jn.String())
+	evaluated, err := vm.EvaluateAnonymousSnippet(provider.Config().Mapper, jn.String())
 	if err != nil {
```
dadrus, line 94: Was marked as an error by golangci-lint (because of go-jsonnet update from 0.16.0 to 0.17.0)

.docker/Dockerfile-build
```diff
@@ -10,13 +10,13 @@ ADD internal/httpclient/go.* internal/httpclient/
 ADD corp/go.* corp/
```
aeneasr, line 10: Note to self: revert this file

dadrus:
@aeneasr: In the mean time I've an update for https://github.com/ory/kratos/pull/1240#pullrequestreview-638223601 and https://github.com/ory/kratos/pull/1240#pullrequestreview-638223631. Just let me know, when you did the review of the current state and which further changes you would like to see. I'll update the PR then.

internal/.kratos.yaml
```diff
@@ -98,15 +104,37 @@ selfservice:
...
       ui_url: http://test.kratos.ory.sh/register
       lifespan: 98m
       after:
         default_browser_return_url: https://self-service/registration/return_to
         password:
           hooks:
-            -
-              hook: session
+            - hook: session
+            - hook: web-hook
```
aeneasr, line 116: I would probably rename this to json-rpc :)

internal/.kratos.yaml
```diff
@@ -98,15 +104,37 @@ selfservice:
...
+              config:
+                url: https://test.kratos.ory.sh/after_registration_hook
+                method: POST
+                auth:
+                  type: api-key
+                  config:
+                    name: My-API-Key
+                    value: secret
+                    in: header
+        global:
```
aeneasr, line 126: How about:

```
    registration:
       ui_url: http://test.kratos.ory.sh/register
       lifespan: 98m
       after:
         default_browser_return_url: https://self-service/registration/return_to
         hooks:
           - hook: session
         password:
           hooks: # overrides `hooks` (does NOT merge!)
             - hook: session
             - hook: web-hook
```

aeneasr (commented):
This is great stuff, thank you! I think this looks very good. I've added some ideas that could improve the usability a bit.

Still need to do a thorough code review, plus we will need some documentation for this :)

But great job! And sorry for being a bit late on this one...

internal/.kratos.yaml
dadrus: Please, let us not do this. JSON-RPC is defined in https://www.jsonrpc.org/specification. it defines, how request and response payloads should look like. E.g. Request: `{"jsonrpc": "2.0", "method": "subtract", "params": [42, 23], "id": 1}`¸ Response: `{"jsonrpc": "2.0", "result": 19, "id": 1}`. This is not what we do, but we can easily mimic that by using a proper jsonnnet template. For a request it would look like: `function(flow, session) {jsonrpc:2.0, method: foo, params: flow.id}`. But the response content would anyway be ignored by the current implementation.

internal/.kratos.yaml
dadrus: Can do this. This will however require some more changes as the
```.go
func HookStrategyKey(key, strategy string) string {
	return fmt.Sprintf("%s.%s.hooks", key, strategy)
}
```
cannot be used as is anymore for "global" hooks

dadrus: > And sorry for being a bit late on this one...
Thank you for taking your time for review! :)

dadrus: > ... plus we will need some documentation for this :)

Definitely. I would update the documentation as soon as all configuration details are clarified and you consider the implementation as complete in sense of this PR

dadrus: > ... so maybe just pass a few things like header, url, method

Will add it to the next update to this PR.

dadrus:
```.yaml
password:
           hooks: # overrides `hooks` (does NOT merge!)
```

Are you sure, you don't want to merge? This would mean config duplication. If a hook should run always, one would have to define it for e.g. `password` and `oidc`

dadrus: On the other hand overriding would solve a particular question, which was open: In which sequence should the hooks be executed?
I mean one could define
```.yaml
registration:
       ui_url: http://test.kratos.ory.sh/register
       lifespan: 98m
       after:
         default_browser_return_url: https://self-service/registration/return_to
         hooks: <-------- defined before strategy
           - hook: session
         password:
           hooks: # overrides `hooks` (does NOT merge!)
             - hook: session
             - hook: web-hook
```

as well as 

```.yaml
registration:
       ui_url: http://test.kratos.ory.sh/register
       lifespan: 98m
       after:
         default_browser_return_url: https://self-service/registration/return_to
         password:
           hooks: # overrides `hooks` (does NOT merge!)
             - hook: session
             - hook: web-hook
         hooks: <-------- defined after strategy
           - hook: session
```
If the hooks are merged, the sequence is not defined, if a strategy always overrides a global hook, the sequence is clear.

dadrus:
The above commits
* rearrange the web hook implementation code. It resolves from my point of view my own comments regarding how the config is parsed and how it is applied.
* enable access to the request method, url and headers from the jsonnnet template. For this a ctx object was made available for the template, which have all the available properties.
* remove the need to configure "global" web hooks under the property named global. So hooks can be added directly underneath of the after property in the supported flows. (before properties and support for other flows will come later)
* revert the changes we did just for us to build kratos locally and not have our logs filled in with error (which actually are no errors)

selfservice/hook/web_hook.go
dadrus: Request header, url and method are now available (added in the recent commits)

dadrus:
Still open tasks:
* hooks should not be merged. a specific strategy should always override the "global" hooks
* add support for all strategies for before and after properties.
* update documentation

@aeneasr: Do you see additional topics?

driver/config/config.go
```diff
@@ -146,7 +147,11 @@ type (
 var Argon2DefaultParallelism = uint8(runtime.NumCPU() * 2)
 
 func HookStrategyKey(key, strategy string) string {
-	return fmt.Sprintf("%s.%s.hooks", key, strategy)
+	if strategy == "none" {
```
harnash, line 150: Nitpick: This took me a while to track. It may be only me but probably more useful name for this could be something like `global` or `all`? I get the intention but for me it was a bit misleading.

harnash: I see this is being discussed already 😉

selfservice/hook/web_hook.go
```diff
@@ -0,0 +1,250 @@
...
+		Flow:           flow,
+		RequestHeaders: req.Header,
+		RequestMethod:  req.Method,
+		RequestUrl:     req.URL.String(),
+		Session:        session,
+	})
+}
+
+func (e *WebHook) execute(data *templateContext) error {
+	conf, err := newWebHookConfig(e.c)
```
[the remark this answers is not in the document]
dadrus: actually yes and I would love to change this, but how should an error be properly logged and propagated during a hook creation?

I mean this is the code responsible for hook creation:

```.go
func (m *RegistryDefault) getHooks(credentialsType string, configs []config.SelfServiceHook) (i []interface{}) {
	for _, h := range configs {
		switch h.Name {
		case hook.KeySessionIssuer:
			i = append(i, m.HookSessionIssuer())
		case hook.KeySessionDestroyer:
			i = append(i, m.HookSessionDestroyer())
		case hook.KeyWebHook:
			i = append(i, hook.NewWebHook(m, h.Config))
		default:
			var found bool
			for name, m := range m.injectedSelfserviceHooks {
				if name == h.Name {
					i = append(i, m(h))
					found = true
					break
				}
			}
			if found {
				continue
			}
			m.l.
				WithField("for", credentialsType).
				WithField("hook", h.Name).
				Errorf("A unknown hook was requested and can therefore not be used")
		}
	}

	return i
}
```

I could have made `hook.NewWebHook` return an error as well, but it just didn't feel right for me just to log it here and continue.

selfservice/hook/web_hook_test.go
```diff
@@ -0,0 +1,214 @@
...
+		Session:        &td,
+	}
+
+	b, err := createBody("test_body.jsonnet", data)
+	assert.NoError(t, err)
+
+	buf := new(strings.Builder)
+	io.Copy(buf, b)
+
+	fmt.Println(buf.String())
```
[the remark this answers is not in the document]
dadrus: Ups. Sorry ;) Will remove this.

driver/config/config.go
dadrus: ok. Will change it back to global ;)

selfservice/hook/web_hook.go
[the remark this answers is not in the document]
harnash: Got it! Makes sense.

dadrus:
Last commit removes the debug statement from the test (https://github.com/ory/kratos/pull/1240#discussion_r618603977) and renames the "name" of the strategy for global hooks from "none" to "global" (https://github.com/ory/kratos/pull/1240#discussion_r618630428)

harnash:
> Last commit removes the debug statement from the test ([#1240 (comment)](https://github.com/ory/kratos/pull/1240#discussion_r618603977)) and renames the "name" of the strategy for global hooks from "none" to "global" ([#1240 (comment)](https://github.com/ory/kratos/pull/1240#discussion_r618630428))

Thank you for this @dadrus! This is important piece for us as well. I'll keep watching and do some more passes over this. If you need any help let me/us know!

dadrus:
> > Last commit removes the debug statement from the test ([#1240 (comment)](https://github.com/ory/kratos/pull/1240#discussion_r618603977)) and renames the "name" of the strategy for global hooks from "none" to "global" ([#1240 (comment)](https://github.com/ory/kratos/pull/1240#discussion_r618630428))
> 
> Thank you for this @dadrus! This is important piece for us as well. I'll keep watching and do some more passes over this. If you need any help let me/us know!

dadrus:
I somho

harnash:
> I somehow pressed something wrong, so the PR was closed. Sorry for that.
> 
> @harnash: Thank you also for your review! If you are familiar with the code, it would be helpful to understand why some ci/circle: test, tbr and ci/circleci: test-e2e tests are failing. I cannot understand how the changes from this PR influence these.

I'm not too familiar but I will take a look. Perhaps I can spot something.

selfservice/strategy/link/strategy_verification_test.go
```diff
@@ -252,5 +252,77 @@ func TestVerification(t *testing.T) {
...
+				assert.Contains(t, verificationLink, "token=")
+
+				cl := testhelpers.NewClientWithCookies(t)
+				res, err := cl.Get(verificationLink)
+				require.NoError(t, err)
+				defer res.Body.Close()
+
+				assert.Equal(t, http.StatusBadRequest, res.StatusCode)
+				// TODO: assert retry verification after webhook failure
+				// 2nd Try
```
harnash, line 280: Why this test tries to send verification link twice and it suppose to fail on the first try? I fail to see the difference between those two calls and my local test runs seems to fail at this point. I might miss something obvious here...

driver/config/.schema/config.schema.json
```diff
@@ -536,13 +581,7 @@
...
                 },
                 "after": {
-                  "type": "object",
-                  "properties": {
-                    "default_browser_return_url": {
-                      "$ref": "#/definitions/defaultReturnTo"
-                    }
-                  },
-                  "additionalProperties": false
+                  "$ref": "#/definitions/selfServiceAfterVerification"
```
harnash, line 584: This change is causing schema tests to fail. You would need to change `test/schema/fixtures/config.schema.test.success/root.full.yml` to:
```yaml
 verification:
      enabled: true
      ui_url: /verification
      lifespan: 1h
      after: "#/definitions/selfServiceAfterVerification"
```

harnash: PR for this https://github.com/innoq/kratos/pull/1

harnash:
@dadrus regarding e2e tests you should update your branch with upstream master since there were quite a few fixes to those tests (specifically: https://github.com/ory/kratos/commit/a6cc4294bb46fe3551b8974a7bf959a74587c709#diff-9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08).

selfservice/hook/web_hook.go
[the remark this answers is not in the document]
dadrus: @harnash: After thinking a bit about your question: 
Kratos supports config reload on the fly. That way the configuration should always be parsed on every hook execution and not be cached. So whether the config is deserialized during the hook creation or execution, doesn't matter at all as the execution happens directly after the creation.

driver/config/.schema/config.schema.json
dadrus: Merged

selfservice/hook/web_hook.go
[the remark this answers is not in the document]
dadrus: @abador: Maybe we should take it as a TODO for the future as an enhancement of the implementation

selfservice/strategy/link/strategy_verification_test.go
dadrus: It was an own attempt to get familiar with the code, which was forgotten to be removed. Last commit (after rebase) removes it. My own fault. Sorry for stealing your time

selfservice/hook/web_hook.go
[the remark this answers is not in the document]
harnash: Sounds good!

internal/.kratos.yaml
aeneasr: Right, web-hook is probably common enough that people understand that this is a JSON request

selfservice/hook/web_hook.go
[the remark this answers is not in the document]
aeneasr: Yeah this could definitely be improved here. I think for now this is fine though as these hooks only run when people sign up, sign in (both of which use BCrypt so they're slow anyways) or change their settings. I think the performance impact is negligible until you hit 50% Google scale

internal/.kratos.yaml
aeneasr: > Are you sure, you don't want to merge? This would mean config duplication. If a hook should run always, one would have to define it for e.g. password and oidc.
>
> I think the current implementation is more user friendly as there is no additional semantic: global hooks run always, strategy specific hooks run only for the particular strategy.

So the problem is the order in which things are executed can't be changed. For example:

```
       after:
         default_browser_return_url: https://self-service/registration/return_to
         password:
           hooks:
             - hook: a
             - hook: b
         hooks:
           - hook: c
           - hook: d
```

If we merge these, then what is being executed in what order?

- a, b, c, d
- c, d, a, b
- a, c, b, d
- what if you want: b, d, a, c?

This is important because some hooks might cancel the request and thus order is important. By not merging these keys, but making this explicit, you have to write a bit more config but at least the result is predictable. I am strongly in favor of not merging these configs.

aeneasr (commented):
Wow good job! I also saw that you're from innoq - someone on our end will reach out to you soon as we think that there's some things we  can do together :)

Regarding the PR, I added another comment regarding the config format. Other than that and  docs this looks good to merge to me!

dadrus:
A colleague of my has by accident unfortunately deleted the innoq/kratos repository. We've now forked the repository again and pushed the latest stand of the `feature/web-hooks-support` branch (the last commit from above). Is there a way to connect this PR with the aforesaid branch again? Or is the only way to recover to create a new PR?

aeneasr:
Shit, I'm not sure if that's possible...

harnash:
``

> HELP: A colleague of my has by accident unfortunately deleted the innoq/kratos repository. We've now forked the repository again and pushed the latest stand of the `feature/web-hooks-support` branch (the last commit from above). Is there a way to connect this PR with the aforesaid branch again? Or is the only way to recover to create a new PR?

I have a local copy of your branch until commit `272ac2373925fd73b03a8204279c1028281ea242`. It might help to force push this to your current branch (if possible) and then cherry-pick missing commits. This way Github might pickup this as the same branch... maybe. Not sure if this is worth the hassle.

aeneasr:
I believe it should be possible to get the commit b1bb6e923461a5e941fa46af437966170eeab5f3 right? Or is that not possible?

harnash:
> I believe it should be possible to get the commit [b1bb6e9](https://github.com/ory/kratos/commit/b1bb6e923461a5e941fa46af437966170eeab5f3) right? Or is that not possible?

Probably contacting Github support would be best course of actions if we want to salvage this PR.

aeneasr:
https://github.com/ory/kratos/compare/recover-1240?expand=1

aeneasr:
^ has all the changes :)

dadrus:
The changes are all there. It would be just great not to have to create a new PR, but be able to continue here. As I've written in my previous comment, I've pushed the relvant commit to the new forked innoq/kratos repository including the proper branch name (innoq:feature/web-hooks-support), but it is not linked to this PR any more.

BTW. Sorry for silence. Since more or less the entire kratos API has changed compared to v0.5..., I'm now updating our PoC code base to be able to see the effects and have kind of "system" tests before implementing support for other flows.

dadrus:
Given the last commit from you @aeneasr, should I then push my changes to https://github.com/ory/kratos/tree/recover-1240 when these are ready?

aeneasr:
Ah I see - I don't think we can salvage this PR. But from my side that's ok. I'll delete the recover-1240 branch then - you won't be able to push your changes there because you do not have write access to this repo :)

harnash (approved):

aeneasr:
Just FYI, planning a release of Kratos this week :)

dadrus:
Let us continue the discussion in #1289.

@aeneasr: I'll try to finalize all the changes as fast as possible. So the web-hook support can hopefully make into the planned release

aeneasr:
Closing in favor of #1289

PR #1244, opened 2021-04-18
fix: marking whoami auhorization parameter as 'in header'

### LINKED ISSUE #1215: client api Whoami not work with Header `Authorization`
Hi team,

Whoami not support query authorization, but code generate with query.
No any Header `Authorization` is set

Bug in here https://github.com/ory/kratos/blob/95b33768138352427d87b6e147698142b955c3b3/internal/httpclient/client/public/whoami_parameters.go#L156

Please hotfix it thanks.

yholkamp:
## Related issue

I believe this resolves #1215; for the `whoami` endpoint, the Authorization string was included as 'query' parameter in the OpenAPI spec as the used 'in' location was not valid. As a result, generated API clients pass the authorization to Kratos via a query param, which is not supported.

## Proposed changes

PR includes a small patch to change the location from 'query' to 'header', which should let users generate a working API client from the spec.

## Checklist

      the changes.
      works.

## Further comments

Regenerating the specification via `make` resulted in a number of unrelated changes for me, which I've not included in this PR for brevity.

### DIFF
--- a/session/handler.go
+++ b/session/handler.go
@@ -125,7 +125,7 @@ type whoamiParameters struct {
 	// in: header
 	Cookie string `json:"Cookie"`
 
-	// in: authorization
+	// in: header
 	Authorization string `json:"Authorization"`
 }
 
--- a/spec/swagger.json
+++ b/spec/swagger.json
@@ -1557,9 +1557,8 @@
           },
           {
             "type": "string",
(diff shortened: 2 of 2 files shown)


aeneasr:
Thank you, I think this looks correct!

Issue #820, opened 2020-11-11
More meta information about the managed identity


dadrus:
**Is your feature request related to a problem? Please describe.**

Feature Fequest:

When the user accesses its profile, it should see 
* which authentication methods are configured, 
* when was the last password change
* ...

**Describe the solution you'd like**

All, the browser based self-service flows, API based flows as well as the implementation of the admin API to retrieve the information about the identity (`GET /identities/{id}`) should not only return the traits, but also the aforesaid additional information

**Additional context**

![image](https://user-images.githubusercontent.com/10072595/98856190-8234f800-245d-11eb-88c5-2d0993ae2cb5.png)

aeneasr:
Good ideas!

PR #1274, opened 2021-04-26
feat: Include Credentials Metadata in admin api.

martinei:
## Related issue
#820

## Checklist
      the changes.
      works.

## Further comments

This is a simple attempt to add the information of the last changed date to the admin api for "get identity".
I'd like to see what you think about it.

### DIFF
--- a/identity/identity.go
+++ b/identity/identity.go
@@ -38,7 +38,7 @@ type Identity struct {
 	ID uuid.UUID `json:"id" faker:"-" db:"id"`
 
 	// Credentials represents all credentials that can be used for authenticating this identity.
-	Credentials map[CredentialsType]Credentials `json:"-" faker:"-" db:"-"`
+	Credentials map[CredentialsType]Credentials `json:"credentials,omitempty" faker:"-" db:"-"`
 
 	// SchemaID is the ID of the JSON Schema to be used for validating the identity's traits.
 	//
@@ -192,4 +192,37 @@ func (i Identity) GetID() uuid.UUID {
 
 func (i Identity) GetNID() uuid.UUID {
 	return i.NID
+}
+
+func (i Identity) MarshalJSON() ([]byte, error) {
+	type localIdentity Identity
+	i.Credentials = nil
+	result, err := json.Marshal(localIdentity(i))
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+func (i *Identity) UnmarshalJSON(b []byte) error {
+	type localIdentity Identity
+	err := json.Unmarshal(b, (*localIdentity)(i))
+	i.Credentials = nil
+	return err
+}
+
+type IdentityWithCredentialsMetadataInJSON Identity
+
+func (i IdentityWithCredentialsMetadataInJSON) MarshalJSON() ([]byte, error) {
+	type localIdentity Identity
+	for k, v := range i.Credentials {
+		v.Config = nil
+		i.Credentials[k] = v
+	}
+
+	result, err := json.Marshal(localIdentity(i))
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
 }
--- a/identity/identity_test.go
+++ b/identity/identity_test.go
@@ -1,6 +1,9 @@
 package identity
 
 import (
+	"encoding/json"
+	"github.com/gofrs/uuid"
+	"github.com/ory/x/sqlxx"
 	"testing"
 
 	"github.com/ory/kratos/driver/config"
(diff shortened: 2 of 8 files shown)


zepatrik (changes requested):
You will also have to adjust the swagger annotations here https://github.com/ory/kratos/pull/1274/files#diff-461d37b91ce0b27e82db7c21166a58a7f8bea1446854e488822f8b79ad95a4e6R161

aeneasr (changes requested):
Thank you! I think it would make more sense to add `json:"created_at"` here: https://github.com/ory/kratos/blob/master/identity/identity.go#L79 :)

zepatrik:
This PR adds information about the last credential update, not the identity creation date :thinking:

aeneasr:
My bad, in that case here: https://github.com/ory/kratos/blob/master/identity/credentials.go#L50

martinei:
I think that does not work because of: https://github.com/ory/kratos/blob/master/identity/identity.go#L42
I think it is/was a security decission to not include credentials in identities JSON representation.

That's why I created a separate representation model here to have control over what is exposed via the API.

martinei:
> You will also have to adjust the swagger annotations here https://github.com/ory/kratos/pull/1274/files#diff-461d37b91ce0b27e82db7c21166a58a7f8bea1446854e488822f8b79ad95a4e6R161

Can you give me a hint how to generate  openapi/swagger document from the go code?

martinei:
Never mind. I found  `make sdk`.

martinei:
It Looks like the embedded Struct here does not survive the openapi spec and code generation circle:
```
type IdentityWithCredentialsMetadata struct {
	*Identity
	Credentials map[CredentialsType]CredentialRepresentation `json:"Credentials"`
}
```

Any Idea how I could easily add additional Properties (Like the credentials Meta Data) to the JSON representation of Identity  without breaking code generation?
Currenlty the Only Option I see would be duplicating everything into a transfer model.

identity/handler.go
```diff
@@ -133,6 +136,54 @@ type getIdentityParameters struct {
...
+	// RecoveryAddresses contains all the addresses that can be used to recover an identity.
+	//
+	// Extensions:
+	// ---
+	// x-omitempty: true
+	// ---
+	RecoveryAddresses []RecoveryAddress `json:"recovery_addresses,omitempty"`
+
+	// this should have an example json, here but I fail to produce on that go-swagger accepts
+	Credentials map[CredentialsType]CredentialRepresentation `json:"Credentials"`
```
aeneasr, line 184:
```suggestion
	Credentials map[CredentialsType]CredentialRepresentation `json:"credentials"`
```

identity/handler.go
```diff
@@ -133,6 +136,54 @@ type getIdentityParameters struct {
 	ID string `json:"id"`
 }
 
+// swagger:model Credentials
+type CredentialRepresentation struct {
+	UpdatedAt time.Time `json:"UpdatedAt"`
+}
+
+// swagger:model Identity
```
aeneasr, line 144: This will clash with the existing identity model - also, it should be lower case :)

```suggestion
// swagger:model identity
```

In order to not clash with the existing model, this would need to be changed to something else, e.g.


```suggestion
// swagger:model someOtherIdentity
```

identity/handler.go
```diff
@@ -53,7 +56,7 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
 type identityResponse struct {
 	// required: true
 	// in: body
-	Body *Identity
+	Body *IdentityRepresentation
```
aeneasr, line 59: Instead of creating a full, new model why not a composite struct?

```suggestion
	Body struct {
	  Identity
	  // any other field like:
	  Foo string `json:"foo"`
	}
```

aeneasr (commented):

identity/handler.go
martinei: Thanks for your feedback @aeneasr !
This is what I initially tried, but I could not get this to survive the go-swagger cycle. For me it looks like it's not possible to reproduce the embedded Type in the generated client code.

aeneasr: Ah, damn, that is possibly a problem. An alternative would also be to supply an openapi patch: https://github.com/ory/kratos/tree/master/.schema/openapi/patches

martinei: I wasn't aware of the patches and will have a look. Thank you!

martinei: @aeneasr I made it work with an embedded struct and also switched Naming to lower case. Some remarks:
 - I had to remove type ( ...) in identity with single lines since go-swagger generates different output in this case
 - The e2e test fail in circle-ci and I don't quite understand why. I am not sure this is related to my changes. Could you take  a look?

aeneasr: Sorry for the long delay, will take a look now!

aeneasr:
I thought about this approach a bit more - and sorry for iterating over this. I looked at the PR and I think that swagger will cause some issues - Discriminators are kinda difficult to use and the resulting schema might be problematic in some of the generators. I have however an idea for an alternative. The idea is as follows:

1. Change `Credentials map[CredentialsType]Credentials `json:"-" faker:"-" db:"-"`` to `Credentials map[CredentialsType]Credentials `json:"credentials,omitempty" faker:"-" db:"-"``
2. Add method `MarshalJSON()` and `UnmarshalJSON()` to `identity.Identity` which remove (encoding) / ignore (decoding) the `credentials` field per default.
3. Add method `MarshalJSONWithCredentials` to Identity which would include the credentials but remove the `credentials.config` part which includes the hashed passwords and other configs
4. Add tests to ensure that credentials are omitted for both pass-by-value (so `json.Marshall(identity.Identity{Credentials: /*...*/})` as well as pass-by-reference (`json.Marshall(identity.Identity{Credentials: /*...*/})`

That way we don't have to mess with swagger, we have a tried & tested set up, and it doesn't break the SDK APIs with the changed discriminators.

What do you think? Sorry for coming up with the idea just now but I had to look into the PR and try out the SDK to understand what could be improved :)

martinei:
Thank you for the review! I aggre that discriminator is kind of a hack (given that we don't even have/use polymorphic serialization here). 
I think the approach you outlined would work technically and I am willing to update the MR if this is the way you want to go. However first I want to raise the following concern and a question:

- This puts even more Details of Kratos API / Interface into the domain "Object" ("Identity") making it harder to understand the boundary behavior from code. 
A much cleaner solution would be to have dedicated structs and mapping methods to represent the api interface. This would decouple the core model (already coupled to the DB)  from the API representation (which is through code generation already coupled tightly to the SDKs.
- The "Credentials" struct already carries some JSON Tags. Where are credentials currently serialized to JSON? We would need to update `UpdatedAt time.Time `json:"-" db:"updated_at"`` to `UpdatedAt time.Time `json:"updatedAt" db:"updated_at"``. What side effects does this have?

aeneasr:
> This puts even more Details of Kratos API / Interface into the domain "Object" ("Identity") making it harder to understand the boundary behavior from code.
>
> A much cleaner solution would be to have dedicated structs and mapping methods to represent the api interface. This would decouple the core model (already coupled to the DB) from the API representation (which is through code generation already coupled tightly to the SDKs.

That's definitely true - however at the moment we do this in all of Ory's products as it is easier to maintain, review, and contribute to if changes do not need to touch several places. So this is fine from my side!

> The "Credentials" struct already carries some JSON Tags. Where are credentials currently serialized to JSON? We would need to update UpdatedAt time.Time json:"-" db:"updated_at" to `UpdatedAt time.Time `json:"updatedAt" db:"updated_at". What side effects does this have?

That doesn't have side effects and is ok / desired. I'm generally going to be exposing these fields in the future!

martinei:
> That's definitely true - however at the moment we do this in all of Ory's products as it is easier to maintain, review, and contribute to if changes do not need to touch several places. So this is fine from my side!

Ok. I'll give it a try (however I am busy today with other stuff).

martinei:
@aeneasr I pushed the changes with a small variation of you proposal: Instead of `MarshalJSONWithCredentials` I have implemented `MarshalJSON` on a Type Alias.

identity/identity.go
```diff
@@ -193,3 +193,36 @@ func (i Identity) GetID() uuid.UUID {
 func (i Identity) GetNID() uuid.UUID {
 	return i.NID
 }
+
+func (i Identity) MarshalJSON() ([]byte, error) {
+	type localIdentity Identity
+	i.Credentials = nil
```
aeneasr, line 199: Does this have side effects? Meaning that using `MarshalJSON` will set `Credentials` to `nil`? I'm always not sure with pass by ref / pass by value...

If it has side effects you could also do use `sjson.Delete()` to remove the key

identity/identity_test.go
```diff
@@ -15,3 +20,79 @@ func TestNewIdentity(t *testing.T) {
...
+	}
+
+	i.Credentials = credentials
+	jsonBytes, err := json.Marshal(i)
+	assert.Nil(t, err)
+	var jsonMap = map[string]json.RawMessage{}
+	err = json.Unmarshal(jsonBytes, &jsonMap)
+	assert.Nil(t, err)
+	assert.Nil(t, jsonMap["credentials"])
+	assert.Equal(t, credentials, i.Credentials, "Original credentials should not be touched by marshalling")
```
aeneasr, line 39: You could simply this with:

```suggestion
	var actual bytes.Buffer
	require.Nil(t, json.NewEncoder(&b).Encode(i))
	
	assert.False(t, gjson.Get(b.String(), "credentials").Exists(), "Original credentials should not be touched by marshalling")
	
	// To ensure the original identity is not changed / Unmarshal has no side effects:
	require.NotEmpty(t, i.Credentials)
```

aeneasr (commented):
Awesome, great idea with the alias! I have two small things for the tests but this is lookinglike it's going to be merged right after :)

identity/identity.go
martinei: I first had a test to prove it doesnt :-). "i" is passed by value here.

identity/identity_test.go
```diff
@@ -15,3 +20,79 @@ func TestNewIdentity(t *testing.T) {
...
+		},
+	}
+	i.Credentials = credentials
+	jsonBytes, err := json.Marshal(&i)
+	assert.Nil(t, err)
+	var jsonMap = map[string]json.RawMessage{}
+	err = json.Unmarshal(jsonBytes, &jsonMap)
+	assert.Nil(t, err)
+	assert.Nil(t, jsonMap["credentials"])
+	assert.Equal(t, credentials, i.Credentials, "Original credentials should not be touched by marshalling")
```
martinei, line 56: @aeneasr This one shows "credentials" is not set to nil.

aeneasr: Ah, nice, thank you!

aeneasr:

PR #1277, opened 2021-04-26
WIP: cookie decoder API

jess-sheneberger:
## Related issue

There is no related issue.

## Proposed changes

I'm integrating Kratos into our app and I've run into a problem I'm currently solving with the new API in this PR. I want to start a conversation and find the right solution to this problem, not necessarily just merge this code, which is why I've put the WIP prefix on this PR.

My app uses the API client, but in Kratos the OIDC flows are browser-only. That's fine, my client can launch a browser, start the OIDC flow, and intercept the result. At this point I have a cookie in hand. In the password flow, I'd have a token I can pass to my backend service. I tried invoking the Whoami API in Kratos from my backend service with the cookie both as a cookie and in the X-Session-Token header but that didn't work. This leads me to believe I need a token for the rest of my interaction with Kratos. So I put together this cookie decoder API that accepts the cookie and decodes it to the corresponding token which I'm able to use in my backend service to do the interaction with Kratos.

Is there a better way to do this?

## Checklist

      the changes.
      works.

### DIFF
--- a/session/handler.go
+++ b/session/handler.go
@@ -43,18 +43,21 @@ func NewHandler(
 }
 
 const (
-	RouteWhoami = "/sessions/whoami"
-	RouteRevoke = "/sessions"
+	RouteWhoami       = "/sessions/whoami"
+	RouteRevoke       = "/sessions"
+	RouteDecodeCookie = "/sessions/decodeCookie"
 	// SessionsWhoisPath  = "/sessions/whois"
 )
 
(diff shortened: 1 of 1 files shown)


zepatrik:
Could you debug at which point the `whoami` endpoint fails when you try to check your token there? Is it because there is no active session?

aeneasr:
I also don't really understand what this is for?

jess-sheneberger:
I thought I needed this but I was actually invoking the /session/whoami API incorrectly. Closing.

PR #1278, opened 2021-04-26
WIP: Login method api

jess-sheneberger:
## Related issue

There is no related issue.

## Proposed changes

In my app I want to allow users to register via one of several methods: two OIDC providers or a password.  When they return after sign-up, I want to provide a login experience that doesn't require them to remember how they registered. I want them to enter their username or email address and for the UI to start the correct OIDC provider flow or prompt for their password using the same method they registered with.

To make this possible, I've put together this login method API. It looks up the given email address and returns either the ID of the OIDC provider that should be used, the "password" method.  My app UI is able to interpret the result and either launch the correct OIDC flow or prompt for a password.

I also have a requirement that users who login via password need to have a username rather just than an email address.  Login via email address is fine for users who login via OIDC.

I understand this is a potential account enumeration attack vector, so maybe it should be turned on with a configuration option and left off for most users.

I'm open to different/better approaches to accomplish this. I'm marking it WIP because I implemented what seemed to me to be the most straightforward way, but maybe there is a different approach that would work better.

## Checklist

      the changes.
      works.

### DIFF
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -1,12 +1,13 @@
 package identity
 
 import (
-	"bytes"
+	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
-	"strings"
 
-	"github.com/ory/herodot"
+	"github.com/tidwall/gjson"
+
 	"github.com/ory/kratos/driver/config"
 
 	"github.com/julienschmidt/httprouter"
@@ -16,10 +17,11 @@ import (
 	"github.com/ory/x/urlx"
 
 	"github.com/ory/kratos/x"
+	"github.com/ory/x/sqlcon"
 )
 
 const RouteBase = "/identities"
-const RouteMethod = "/method/:id"
+const RouteKnownCredentials = "/credentials/known"
 
 type (
 	handlerDependencies interface {
@@ -49,17 +51,62 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
 	admin.POST(RouteBase, h.create)
 	admin.PUT(RouteBase+"/:id", h.update)
 
-	admin.GET(RouteMethod, h.method)
+	admin.POST(RouteKnownCredentials, h.knownCredentials)
 }
 
-type methodResponse struct {
+// A single login method and provider.
+type knownCredentialsMethod struct {
 	Method   string `json:"method"`
 	Provider string `json:"provider,omitempty"`
 }
 
-// swagger:route GET /method admin getMethod
+// Response object for knownCredentials
+type knownCredentialsResponse struct {
+	Found   bool                     `json:"found"`
+	Methods []knownCredentialsMethod `json:"methods,omitempty"`
+}
+
+// Request object for knownCredentials
+type knownCredentialsRequest struct {
+	Identifier string `json:"identifier"`
+	Method     string `json:"method"`
+}
+
+func (h *Handler) knownCredentialsOIDC(ctx context.Context, id string) (bool, []string, error) {
+	address, err := h.r.PrivilegedIdentityPool().FindVerifiableAddressByValue(ctx, VerifiableAddressTypeEmail, id)
+	if err != nil {
+		if errors.Is(err, sqlcon.ErrNoRows) {
+			return false, nil, nil
+		}
+		return false, nil, err
+	}
+
+	identity, err := h.r.PrivilegedIdentityPool().GetIdentityConfidential(ctx, address.IdentityID)
+	if err != nil {
+		if errors.Is(err, sqlcon.ErrNoRows) {
+			return false, nil, nil
+		}
+		return false, nil, err
+	}
+
+	creds, ok := identity.GetCredentials(CredentialsTypeOIDC)
+	if !ok {
+		return false, []string{}, nil
+	}
+
+	providers := gjson.Get(string(creds.Config), "providers")
+	result := []string{}
+	for _, provider := range providers.Array() {
+		result = append(result, provider.Get("provider").String())
+	}
+
+	return true, result, nil
+}
+
+// swagger:route GET /credentials/known admin knownCredentialsRequest
 //
-// Get login method for a previously registered email address or username
+// Check if a specified identifier has been previously registered with the specified method, or optionally search
+// for the method and provider if a method is not specified.
 //
 //     Produces:
 //     - application/json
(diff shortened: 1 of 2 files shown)


aeneasr:
If we put this into the admin endpoint I think it would be mergeable. It could also allow UIs to have a UI indicator if a username is already taken!

jess-sheneberger:
Sure, I can do that.

Right now in my configuration the admin endpoint isn't available to the client, so we'd need some way to expose this one admin API method to the client, right?

aeneasr:
Yes, you would usually do that in your routing / reverse proxy set up - exposing this specific endpoint to the public internet

aeneasr:
So should we keep this PR open or is it something that you were able to work around / do not intend on continue working on?

jess-sheneberger:
Let's keep it open, I'm still working on it.  I've got it moved over to the admin endpoint but I'm wrestling with my reverse proxy ATM so I can't test it yet.

jess-sheneberger:
OK, it's on the admin endpoint now. What else should I do before this can be merged?

identity/handler.go
```diff
@@ -44,6 +48,90 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
...
+//
+//     Schemes: http, https
+//
+//     Responses:
+//       200: loginMethod
+//       500: genericError
+func (h *Handler) method(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+
+	// if the credentials can be looked up directly by email address then they're type password
+	_, _, err := h.r.PrivilegedIdentityPool().FindByCredentialsIdentifier(r.Context(), CredentialsTypePassword, ps.ByName("id"))
```
aeneasr, line 75: If I understand correctly you want to check if a user is signed up using the provided credentials, correct? In that case it would probably make sense to have a `POST` request which accepts a payload like

```
{
  "identifier": "...",
  "method": "password" # or "oidc"
}
```

and returns a `true` or `false` if it was found or not found? I think this would make the logic below much easier!

If the `method` is omitted - which means we want to discover the method - we could still implement the search algorithm you have written and respond with:

```
{
  "found": true,
  "method": "oidc",
  "oidc_providers": ["google"] // if method = oidc
}
```

The method should probably be named something like `POST /credentials/known`.

What do you think?

identity/handler.go
```diff
@@ -44,6 +48,90 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
...
+		id, err := h.r.PrivilegedIdentityPool().GetIdentityConfidential(r.Context(), address.IdentityID)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, err)
+			return
+		}
+
+		creds, ok := id.GetCredentials(CredentialsTypeOIDC)
+
+		var cfg map[string]interface{}
+		if err := json.NewDecoder(bytes.NewBuffer(creds.Config)).Decode(&cfg); err != nil {
```
aeneasr, line 94: This will panic because ok is checked afterwards. It might also be simpler to use `gjson.Get(...)` instead of marshalling / type asserting

identity/handler.go
```diff
@@ -44,6 +48,90 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
...
+			return
+		}
+
+		providers, ok := cfg["providers"].([]interface{})
+		if ok == false {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReason("No providers found in oidc config").WithDebug(err.Error())))
+			return
+		}
+
+		provider, ok := providers[0].(map[string]interface{})
```
aeneasr, line 110: This only takes the first provider but what if we have more than one?

identity/handler.go
```diff
@@ -44,6 +48,90 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
...
+			return
+		}
+
+		h.r.Writer().Write(w, r, &methodResponse{Method: CredentialsTypeOIDC.String(), Provider: ps})
+		return
+
+	}
+
+	if strings.Contains(ps.ByName("id"), "@") {
+		h.r.Writer().Write(w, r, &methodResponse{Method: "need-username"})
```
aeneasr, line 130: `need-username` is not a method so I would just return `"method": "password"`

aeneasr (changes requested):
Thank you for the contribution! Unfortunately we can not accept contributions without tests. These are important to ensure that your feature does what it should, and it ensures that it keeps working even if others change your code!

If you need help with the tests feel free to ping a maintainer!

identity/handler.go
jess-sheneberger: Sure, that seems fine, but how about this for the response?
```
{
    "found": true,
    "methods": [
        {
            "method": "password"
        },
        {
            "method": "oidc",
            "provider": "google"
        },
        {
            "method": "oidc",
            "provider": "microsoft"
        } // ...
    ]
}
```

because there could be multiple methods to login as well as multiple OIDC providers.

identity/handler.go
jess-sheneberger: Cool, thanks for the `gjson.Get()` suggestion, I will do it that way, definitely simpler

identity/handler.go
aeneasr: Big brain! 🧠 :) That makes total sense

identity/handler.go
aeneasr: I love that tool :)

jess-sheneberger:
I believe I've incorporated all the review feedback now. I need a little help with writing some tests--It looks like this should be tested in handler_test.go, I see how to create identities but I don't see how to set credentials / associate with OIDC provider so I can exercise the new code. Can you point me in the right direction?

jess-sheneberger:
Ping - can anyone help me get started writing the tests? I need to create a couple identities, one associated with OIDC and one with password credentials.

zepatrik:
Sure, sorry for the late response from our side. I believe you can just use something similar to
https://github.com/ory/kratos/blob/103880c4fee41b1a6795eb031f921801891596f5/identity/handler_test.go#L180-L188
and
https://github.com/ory/kratos/blob/bce14c487450bd668859f362b98704644fa4c72a/identity/test/pool.go#L70-L74
and
https://github.com/ory/kratos/blob/7b8d59f48ed14a6d0672238645d8675d4bf7fd77/selfservice/strategy/oidc/strategy_settings_test.go#L114-L120

In general, these searches should give you more examples:
- https://github.com/ory/kratos/search?q=CredentialsTypeOIDC
- https://github.com/ory/kratos/search?l=Go&q=SetCredentials

aeneasr:

Thank you!

jess-sheneberger:
Accidentally closed this, I will reopen a new PR and link this one when I'm finished with the tests.  Sorry about that, I'm still figuring out the Github workflow :-(

Issue #882, opened 2020-12-07
Add option for disabling registration


zkonge:
**Is your feature request related to a problem? Please describe.**

No

**Describe the solution you'd like**

Add option for disabling registration.

**Describe alternatives you've considered**

For some private account system, it would be a useful option.
Now, I have to use web server path rule to redirect registration request.
An out of the box option would be better.

I'm happy to make contribution for it.

aeneasr:
Hello there, thank you for the feature suggestion. We definitely want to allow registration to be disabled. We have a system already partially in place which runs hooks before login / registration. We could write a hook which disables registration, or which requires an invite code (e.g. sent via a URL query parameter). What do you think?

radekg:
Not the reporter but a hook like this would be appreciated. I wonder if you have considered making a generic hook available. Something Traefik-forward-auth-reverse-proxy like or a grpc based hook?

aeneasr:
Yes, there are open issues for that already

LeonidChistov:
Would it be also possible to allow login via `oidc` method, but disallow registration of new users via `oidc` method? The reason why I suspect that this question could be different from the original one, is that it looks like new users may be created in scope of `login` flow when using `oidc` method.

aeneasr:
We were thinking of enabling methods on a per-flow basis. You usecase confirms that this is probably a good idea!

AUSBird:
By the looks of this issue's milestone changes, It's safe to assume this feature is not high on the priority.
I might put my feedback here as to my use case for this feature

I'm (Hopefully) releasing a beta version of my own software (web app) that I want to do a closed beta with a few invited people and slowly open up to more before opening up to public. It's not the end of the world for me since we can just move where this invitation system sits and make it part of the application's API.
The other use case I have is to limit access to the staging and development environments (E.g. Must register with a company email, or must be invited also)

aeneasr:
Contributions are always welcomed :) But of course, please sync on your plans beforehand!

jknlsn:
+ 1 for this, helps with migrating from Auth0 where similar exists!

Much the same position and use cases as @AUSBird describes.

aeneasr:
Adding an option to disable/enable self service registration should be straight forward :) We already have such an option for recovery and verification! Whoever wants to pick this up please go ahead :)

PR #1289, opened 2021-04-28
feat: web hooks support (recovery)

dadrus:
## Related issue

https://github.com/ory/kratos/issues/271

## Related PRs

Since unfortunately the original repository has been deleted, the https://github.com/ory/kratos/pull/1240 became disconnected. This PR is meant as a replacement. It contains all the changes done so far in the aforesaid PR.

## Proposed changes

What follows is copy & paste from https://github.com/ory/kratos/pull/1240 to give context.

As discussed in the linked ticket this PR describes the idea of a possible implementation for web hooks. The support is currently added to `verification` and `registration` flows only and should be the basis for further discussion. As of now, the implementation is very simple - no retries, no asynchrony. The actual implementation of how the configuration is parsed and applied is also very simple and should be changed to be more concise. 

The configuration for the web-hooks goes into the `hooks` property and has the following structure:

```.yaml
hooks:
  - hook: web-hook
    config:
      url: https://test.kratos.ory.sh/after_verification_hook
      method: POST # GET, DELETE, etc
      body: /link/to/my/jsonnet/file
      auth:
        type: <some-type>
        config:
          <type-specific-config>
```

The `flow` and `session` objects are currently bound into the jsonnet template. E.g. to create a body looking like `{ user_id: <some-id> }` to be send to the web hook end point, the jsonnet template would look like this: `function(flow, session) { user_id: session.identity.id }`

For `auth` following types are implemented:
* api-key
* basic-auth

For `api-key` the config looks as follows:

```.yaml
name: Some-Name
value: The-Value-of-My-Key
in: header # alternatively cookie
```

For `basic-auth` the config looks as follows:
```.yaml
user: My-User
password: My-Pass-Value
```

The `after` property of `registration` and `verification` has been extended to contain a `global` property, which consists only out of a `hooks` property, which can contain as many web-hooks in the above said format as desired.

The `hooks` property of `oidc` and `password` properties of the aforesaid flows were extended to contain the web-hook item in addition to already defined hook types.

## Checklist

      the changes.
      works.

## Further comments

As the support for web-hooks touches many flows, I rather would like to start small and discuss the desired changes in the implementation and usage first, before adopting the support to all other flows.

I used the word `auth` for the configuration of "security" settings of a web-hook by intention. What has been done so far is IMHO just a small part of what might be required. I can see the need for custom certificate trust stores to verify the servers certificate, as well as the need to support mTLS. Other mechanisms, or combinations might be required as well.


dadrus:
The https://github.com/ory/kratos/pull/1289/commits/876c8ea36d308e5bbef1925585f2d8293d3a0381 commit just adds a comment to see, whether it can salvage the #1240 PR. Unfortunately no.

aeneasr:
I will take a look tomorrow!

dadrus:
@aeneasr: The web-hook support has been added to the "after" property of all flows, but error and logout. I personally don't see any added value by adding the support for these two. What is your opinion?

Support for "before" properties has not been added yet. To be honest I personally would expect that the results of such hooks would somehow modify/affect the flow. This is however not yet discussed/designed. The only effect, which could be achieved with the given implementation, would be to cancel the further execution of the flow if the execution of a particular web-hook fails. If this is this desired behavior, I can add this support as well. However, even there is some code available, as well as some tests, the current schema does not define "before" hooks at all. So as of today, these cannot be defined in the kratos.yaml anyway, as kratos refuse to start due to config validation errors. So how should we proceed here?

driver/registry_default_test.go
```diff
@@ -295,12 +295,91 @@ func TestDriverDefault_Hooks(t *testing.T) {
...
+				expect: func(reg *driver.RegistryDefault) []settings.PostHookPostPersistExecutor { return nil },
+			},
+			{
+				uc: "Only verify hook configured for the strategy",
+				prep: func(conf *config.Config) {
+					conf.MustSet(config.ViperKeySelfServiceVerificationEnabled, true)
+					// I think this is a bug as there is a hook named verify defined for both profile and password
+					// strategies. Instead of using it, the code makes use of the property used above and which
+					// is defined in an entirely different flow (verification).
+				},
```
dadrus, line 316: @aeneasr: Is it really the desired this way?

selfservice/strategy/link/strategy_recovery.go
```diff
@@ -333,6 +333,10 @@ func (s *Strategy) recoveryIssueSession(w http.ResponseWriter, r *http.Request,
 		return s.handleRecoveryError(w, r, f, nil, err)
 	}
 
+	if err := s.d.RecoveryExecutor().PostRecoveryHook(w, r, f, sess); err != nil {
+		return s.handleRecoveryError(w, r, f, nil, err)
```
dadrus, line 337: @aeneasr: Is this the proper place for a web-hook, or should it be moved?

dadrus:
@aeneasr: There is one thing, which I personally don't like at all and this is the current "hook interface". I put it in braces, because, there is no defined (FMPOV) and this leads to a lot of code duplication. If you, for example, take a look at the `web_hook.go` file, you'll see, the `WebHook` struct has to implement different `PostHookExecutor` interfaces, which only differ in one argument - the passed in `Flow` object. This becomes more prominent in the implementation of the `hook.go` file of each strategy. It is almost all the same and the difference is again almost related to a specifically typed `Flow` argument. 

I think by introducing an interface for a `Flow` and by defining the hooks wiring and infrastructure types in a form of 

```.go
type (
	PreHookExecutor interface {
		ExecutePreFlowHook(w http.ResponseWriter, r *http.Request, a Flow) error
	}
	PreHookExecutorFunc func(w http.ResponseWriter, r *http.Request, a Flow) error

	PostHookPostPersistExecutor interface {
		ExecutePostFlowPostPersistHook(w http.ResponseWriter, r *http.Request, a Flow, s *session.Session) error
	}
	PostHookPostPersistExecutorFunc func(w http.ResponseWriter, r *http.Request, a Flow, s *session.Session) error

	PostHookPrePersistExecutor interface {
		ExecutePostFlowPrePersistHook(w http.ResponseWriter, r *http.Request, a Flow, i *identity.Identity) error
	}
	PostHookPrePersistExecutorFunc func(w http.ResponseWriter, r *http.Request, a Flow, i *identity.Identity) error

	HooksProvider interface {
		PreFlowHooks(ctx context.Context) []PreHookExecutor
		PostFlowPrePersistHooks(ctx context.Context, credentialsType identity.CredentialsType) []PostHookPrePersistExecutor
		PostFlowPostPersistHooks(ctx context.Context, credentialsType identity.CredentialsType) []PostHookPostPersistExecutor
	}
)
```

The entire hooks execution logic can be in one place and the WebHook implementation would only have to implement just the three methods defined in the `HooksProvider` from the above snippet.

I don't see this as part of this PR, but it would be great if we could refactor that to remove the duplications and make the code easier to understand.

aeneasr:
I don't think https://github.com/ory/kratos/pull/1289#issuecomment-829133224 should be addressed in this PR. There are good reasons for keeping this logic separated as it allows us to, for example, get full control over what hooks are available when where for which flows!

aeneasr:
It's a bit of duplication but it's not that bad IMO and it allows flexibility down the road :)

aeneasr:
@dadrus can you please enable write access for maintainers?

```
% git push innoq HEAD:feature/web-hooks-support
ERROR: Permission to innoq/kratos.git denied to aeneasr.
fatal: Konnte nicht vom Remote-Repository lesen.
```

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

driver/config/.schema/config.schema.json
```diff
@@ -60,6 +60,21 @@
...
       ]
     },
+    "selfServiceWebHook": {
+      "type": "object",
+      "properties": {
+        "hook": {
+          "const": "web-hook"
+        },
+        "config": {
+          "type": "object"
```
aeneasr, line 70: Please provide the actual config schema here :)

driver/config/config.go
```diff
@@ -181,7 +183,11 @@ func (c *Argon2) MarshalJSON() ([]byte, error) {
 var Argon2DefaultParallelism = uint8(runtime.NumCPU() * 2)
 
 func HookStrategyKey(key, strategy string) string {
-	return fmt.Sprintf("%s.%s.hooks", key, strategy)
+	if strategy == "global" {
```
aeneasr, line 186: As discussed in the other PR - please follow what I said here: https://github.com/ory/kratos/pull/1240#discussion_r619988905

driver/registry_default_login.go
```diff
@@ -29,6 +29,16 @@ func (m *RegistryDefault) PostLoginHooks(ctx context.Context, credentialsType id
 			b = append(b, hook)
 		}
 	}
+
+	if len(b) == 0 {
+		// since we don't want merging hooks defined in a specific strategy and global hooks
+		// global hooks are added only if no strategy specific hooks are defined
+		for _, v := range m.getHooks("global", m.Config(ctx).SelfServiceFlowLoginAfterHooks("global")) {
```
aeneasr, line 36: Replace global with https://github.com/ory/kratos/pull/1240#discussion_r619988905

driver/registry_default_recovery.go
```diff
@@ -41,3 +41,21 @@ func (m *RegistryDefault) AllRecoveryStrategies() (recoveryStrategies recovery.S
...
+func (m *RegistryDefault) RecoveryExecutor() *recovery.HookExecutor {
+	if m.selfserviceRecoveryExecutor == nil {
+		m.selfserviceRecoveryExecutor = recovery.NewHookExecutor(m)
+	}
+	return m.selfserviceRecoveryExecutor
+}
+
+func (m *RegistryDefault) PostRecoveryHooks(ctx context.Context) (b []recovery.PostHookExecutor) {
+
+	for _, v := range m.getHooks("global", m.Config(ctx).SelfServiceFlowRecoveryAfterHooks("global")) {
```
aeneasr, line 54: Replace global with https://github.com/ory/kratos/pull/1240#discussion_r619988905

driver/registry_default_test.go
aeneasr: Not sure I understand, could you elaborate?

aeneasr (commented):
Once the config is updated, this also needs some documentation:

- [ ] https://github.com/ory/kratos/blob/master/docs/docs/self-service/hooks.mdx
- [ ] Add a [guide](https://github.com/ory/kratos/tree/master/docs/docs/guides) with an example (e.g. for Mailchimp or Stripe or whatever you want as an example)

Thank you :)

driver/registry_default_recovery.go
dadrus: This is only in code to differentiate between strategy and "global" hooks. In the config the "hook" property can be configured directly within the "after" property.

driver/registry_default_login.go
dadrus: This is only in code to differentiate between strategy and "global" hooks. In the config the "hook" property can be configured directly within the "after" property.

driver/config/config.go
dadrus: This is only in code to differentiate between strategy and "global" hooks. In the config the "hook" property can be configured directly within the "after" property.

dadrus:
> @dadrus can you please enable write access for maintainers?

Unfortunately I don't see the option described in https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

dadrus:
@aeneasr : Try pushing again. GitHub offers a possibility to enable access to the entire repository for contributors.

dadrus:
Or should I just merge your PR https://github.com/innoq/kratos/pull/1 ?

driver/config/.schema/config.schema.json
dadrus: Add in https://github.com/ory/kratos/pull/1289/commits/65773bc8272f3ab75ac6c6c142e8c0a64409dde2

driver/registry_default_test.go
dadrus: I mean that there are hooks defined for the same purpose.

My understanding of the schema and the documentation is that one would have to configure a verification hook, like this:

```.yaml
settings:
  ...
  after:
    password:
      hooks:
        - hook: verify
```

But the code does not take this into account at all. Instead it checks, whether the `enabled` property of the `verification` flow: 

```.go
func (m *RegistryDefault) PostSettingsPostPersistHooks(ctx context.Context, settingsType string) (b []settings.PostHookPostPersistExecutor) {
	initialHookCount := 0
	if m.Config(ctx).SelfServiceFlowVerificationEnabled() {   <-------------------------------
		b = append(b, m.HookVerifier())
		initialHookCount = 1
	}
...
```

dadrus: Btw, some other flow implementations do the same

harnash:
> @aeneasr: The web-hook support has been added to the "after" property of all, but error and logout flows. I personally don't see any added value by adding the support for these two. What is your opinion?
> 
> Support for "before" properties has not been added yet. To be honest I personally would expect that the results of such hooks would somehow modify/affect the flow. This is however not yet discussed/designed. The only effect, which could be achieved with the given implementation, would be to cancel the further execution of the flow if the execution of a particular web-hook fails. If this is this desired behavior, I can add this support as well. However, even there is some code available, as well as some tests, the current schema does not define "before" hooks at all. So as of today, these cannot be defined in the kratos.yaml anyway, as kratos refuse to start due to config validation errors. So how should we proceed here? IMHO this is relevant for #882

If that helps. We do have  a scenario of `before` hook as you described. We run various checks on external services that basically do some checks and optionally prohibit user from logging in or dropping session altogether. It is nice to have but that is just our opinion :-)

dadrus:
Ok. I'll add support for `before` hooks in `registration` and `login` as it will at least partially solve #882 

Later on one could add support for an additional jssonnet template, which could modify flow and/or session object based on particular application requirements and on the result of a web-hook call (would be other PR).

dadrus:
@aeneasr: How would you like to proceed related to https://github.com/ory/kratos/pull/1289#issuecomment-829274471 and https://github.com/ory/kratos/pull/1289#issuecomment-829275068

harnash:
> @harnash: Take a look, does it cover your use cases?

Thank you @dadrus! In terms of hooks I think it covers our use cases.

driver/config/config.go
aeneasr: Ah, I overlooked that! Thank you

driver/registry_default_login.go
aeneasr: Sorry about that confusion!

driver/registry_default_test.go
aeneasr: Ah I see - the reasoning here is that we enable the verification hook per default when the flow is enabled as it would not really make sense otherwise to be enabled. I get however that this is a bit confusing here. Basically, the idea is: why would you NOT want to send an email when you have verification enabled?

I am however open to change this or at least remove the "verify" hook from the available config options.

aeneasr: Looking at the current docs I think it would make sense to remove the "verify" hook from the config schema (in a a separate PR) -> https://github.com/ory/kratos/issues/1302

selfservice/flow/recovery/hook.go
```diff
@@ -0,0 +1,87 @@
+package recovery
```
aeneasr, line 1: This one needs a test :) You can probably copy it from one of the other flows more or less 1:1

selfservice/flow/verification/hook.go
```diff
@@ -0,0 +1,89 @@
+package verification
```
aeneasr, line 1: This one needs a test :) You can probably copy it from one of the other flows more or less 1:1

selfservice/hook/web_hook.go
```diff
@@ -0,0 +1,307 @@
...
+		templatePath string
+		auth         AuthStrategy
+	}
+
+	webHookDependencies interface {
+		x.LoggingProvider
+	}
+
+	templateContext struct {
+		Flow           interface{} `json:"flow"`
```
aeneasr, line 61: I think it would make sense to add the flow name as well? So `login`, `registration`, and so on. That helps us differentiate in the webhook jsonnet!

selfservice/strategy/link/strategy_recovery.go
aeneasr: Yeah that LGTM

selfservice/hook/web_hook_test.go
```diff
@@ -0,0 +1,212 @@
+package hook
```
aeneasr, line 1: Could you add integration tests with `httptest.NewServer` for the web-hook receiver with different config options? :)

aeneasr (commented):
Yo, truly amazing job! :) 

This is good to merge from an implementation standpoint. We just need a few more test cases for the new hook handlers and an integration test for the web hook :) I have also made some minor changes which I again created as a PR against your fork. However, I am not able to create a pull request or push to your branch. Updating those settings would improve maintainer experience a bit ;D

Bildschirmfoto 2021-05-03 um 11 08 06

You can find my changes here: https://github.com/innoq/kratos/compare/feature/web-hooks-support...ory:feature/web-hooks-support

I was planning on releasing v0.6 today but will hold off the release until this is merged so that we can include the breaking change for the config in v0.6!

aeneasr:
````
feat: web hooks for self-service flows

This feature adds the ability to define web-hooks using a mixture of configuration and JsonNet. This allows integration with services like Mailchimp, Stripe, CRMs, and all other APIs that support REST requests. Additionally to these new changes, it is now possible to define hooks for verification and recovery as well!

The configuration for the web-hooks goes into the `hooks` property and has the following structure:

```.yaml
hooks:
  - hook: web_hook
    config:
      url: https://test.kratos.ory.sh/after_verification_hook
      method: POST # GET, DELETE, etc
      body: /link/to/my/jsonnet/file
      auth:
        type: <some-type>
        config:
          <type-specific-config>
```

The `flow` and `session` objects are currently bound into the jsonnet template. E.g. to create a body looking like `{ user_id: <some-id> }` to be send to the web hook end point, the jsonnet template would look like this: `function(flow, session) { user_id: session.identity.id }`

For `auth` following types are implemented:
* api-key
* basic-auth

For `api-key` the config looks as follows:

```.yaml
name: Some-Name
value: The-Value-of-My-Key
in: header # alternatively cookie
```

For `basic-auth` the config looks as follows:
```.yaml
user: My-User
password: My-Pass-Value
```

BREAKING CHANGE: As part of this change the configuration layout for hooks has changed. These are now to be defined under the `hook` key:

```patch
  selfservice:
    flows:
       <registration|login|settings|...>:
         after:
+          hooks:
+            - hook: some_hook
+              # ...
           <password|oidc|profile|...>:
-            - hook: some_hook
-              # ...
+            hooks:
+              - hook: some_hook
+                # ...
```

````

**Comment by dadrus** [comment]
@aeneasr: I now explicitly added you to the contributors list in the fork. You should now be able to do whatever you want in the fork ;)

**Comment by dadrus** [comment]
@aeneasr:
> ... We just need a few more test cases for the new hook handlers and an integration test for the web hook :)

Yep. I have started adding such tests last week (Not pushed yet as you can see).

File: selfservice/hook/web_hook.go (hunk 9 above)
Comment (dadrus) [reply selfservice/hook/web_hook.go:61]: Agree. IMHO, it would be better to have the corresponding property in the `Flow` object itself. One would then get the name of the Flow in the ctx object just like this: `ctx.flow.name`. This would however require an update of at least the open api spec. I mean there is no need to write that information to the db, as this information is static. There are however multiple places, which fetch the flow data from the db and would require an update. It would be great, if you could support here. I can then focus entirely on the tests for the web hooks.

Comment (aeneasr) [reply selfservice/hook/web_hook.go:61]: That makes sense! I think it would probably be easier to just do the type assertion in the JsonNet config here - but we can do that in a later PR.

**Comment by dadrus** [comment]
@aeneasr: Regarding https://github.com/ory/kratos/pull/1289#issuecomment-831130724
I assume, this is your preparation for the release notes. I would rather reference docs/docs/self-service/hooks.mdx as 
> The `flow` and `session` objects are currently bound into the jsonnet template. E.g. to create a body looking like `{ user_id: <some-id> }` to be send to the web hook end point, the jsonnet template would look like this: `function(flow, session) { user_id: session.identity.id }`

is not correct (at least the example).

File: selfservice/flow/verification/hook.go (hunk 8 above)
Comment (dadrus) [reply selfservice/flow/verification/hook.go:1]: Done

File: selfservice/flow/recovery/hook.go (hunk 7 above)
Comment (dadrus) [reply selfservice/flow/recovery/hook.go:1]: Done

**Comment by dadrus** [comment]
I'll add tests, which test the different web hook methods tomorrow. Thank you for waiting with the new release until I'm ready :)

**Comment by aeneasr** [comment]
> I don't know, how it was in the past, but already with v0.5 of kratos, hooks had to be configured in a hooks property of a method.

Oh, thank you for pointing that out! Looks like I remembered wrong 😅

File: selfservice/hook/web_hook_test.go (hunk 10 above)
Comment (dadrus) [reply selfservice/hook/web_hook_test.go:1]: Done

**Comment by dadrus** [comment]
Status update:
* Implementation is almost complete (see below)
* Documentation is almost complete (see below)
* Testing of the new functionality is pretty extensive.

Regarding the "almost": 
As already written in https://github.com/ory/kratos/pull/1289#issuecomment-829133224, I see some improvement related to how hooks in general should be designed in sense of the interface. Related to this, 
* some hooks enable access only the flow object (in addition to the request properties) in the jsonnet template. These are the "pre" method hooks. Since the user is not yet registered, or logged in, this is fine and expected.
* other give access to the session and identity objects in addition,
* yet other give only access to the identity object in addition to the flow and the request properties. So session object is not available. Affected flows are settings and verification flows. Since the corresponding two required interface were added by me, I assume I had to do something in addition, e.g. to fetch the session data from the db.

What I actually want to say is that the usage of the web hooks is not intuitive because of this. Btw. this specifics is not yet reflected in the documentation. My own expectation is, that the `before` hooks make only the `flow` object as well as the request properties accessible in the jsonnnet template. The `after` hooks should make everything accessible and it should not be dependent on a particular method. To resolve this, I would like to remove the `session` object from the `ctx` object passed to the template. As there is no admin API, or other way to make use of it, I personally don't see any added value by binding it to the template. So the `before` hooks would make the `flow`, `request_method`, `request_uri` and `request_headers` available and the `after` hooks would make the `identity` object available as well. This is easy to remember and less error prone compared to the current implementation. The changes are trivial and affect only the internal implementation of the web hook.

What do you think?

**Comment by aeneasr** [comment]
> What I actually want to say is that the usage of the web hooks is not intuitive because of this. Respectively these specifics are not yet reflected in the documentation. My own expectation is, that the before hooks make only the flow object as well as the request properties accessible in the jsonnnet template. The after hooks should make everything accessible and it should not be dependent on a particular method. To resolve this, I would like to remove the session object from the ctx object passed to the template. As there is no admin API, or other way to make use of it (session data), I personally don't see any added value by binding it to the template. So the before hooks would make the flow, request_method, request_uri and request_headers available and the after hooks would make the identity object available as well. This is easy to remember and less error prone compared to the current implementation. The changes are trivial and affect only the internal implementation of the web hook.

I would say that sounds like a plan!

File: driver/registry_default_test.go (hunk 1 above)
Comment (dadrus) [reply driver/registry_default_test.go:391]: > Basically, the idea is: why would you NOT want to send an email when you have verification enabled?

Because we now have web hooks ;) and one would maybe like to use other mechanisms? ¯\_(ツ)_/¯

**Comment by dadrus** [comment]
I appreciate if somebody could take a closer look at the tests. The failing test is defined in selfservice/flow/settings/handler_test.go. The assertion in the line 108 expects no cookies set, but there are two. Both are set in line handler.go:L100 `h.d.ContinuityManager().Abort(r.Context(), w, r, ContinuityKey(strategy.SettingsStrategyID()));`. I also don't see any side-effects from the web-hook implementation.

**Comment by dadrus** [comment]
@aeneasr: Yep. That was it: https://github.com/ory/kratos/commit/fb37337a7f5c0031e0048114a563456af80072f6 (regarding the failing tests). After merging master it is gone.

**Review by aeneasr — APPROVED** [review]
Thank you all, truly amazing work! This is a major improvement to Ory Kratos. Big love :)

**Comment by aeneasr** [comment]
I'll be waiting for all the tests to pass and then merge it!

**Comment by aeneasr** [comment]
Updated changelog

```

feat: web hooks for self-service flows

This feature adds the ability to define web-hooks using a mixture of configuration and JsonNet. This allows integration with services like Mailchimp, Stripe, CRMs, and all other APIs that support REST requests. Additionally to these new changes, it is now possible to define hooks for verification and recovery as well!

For more information, head over to the hooks documentation: https://www.ory.sh/kratos/docs/self-service/hooks
```

dadrus:
> Thank you all, truly amazing work! This is a major improvement to Ory Kratos. Big love :)

You're very welcome :)

PR #1300, opened 2021-04-30
Fix unexpected emails

cwei-wish:
## Checklist

      the changes.
      works.


selfservice/hook/verification.go
```diff
@@ -72,6 +74,12 @@ func (e *Verifier) do(r *http.Request, i *identity.Identity, f flow.Flow) error
 		if err := e.r.LinkSender().SendVerificationTokenTo(r.Context(), verificationFlow, address, token); err != nil {
```
aeneasr, line 74: I think we should update the status in this method instead!

selfservice/hook/verification.go
```diff
@@ -49,7 +50,8 @@ func (e *Verifier) do(r *http.Request, i *identity.Identity, f flow.Flow) error
 
 	for k := range i.VerifiableAddresses {
 		address := &i.VerifiableAddresses[k]
-		if address.Verified {
+		if address.Status == identity.VerifiableAddressStatusCompleted ||
```
aeneasr, line 53: Simplify: `if address.Status != VerifiableAddressStatusPending`

aeneasr (changes requested):

selfservice/hook/verification.go
cwei-wish: > Simplify: `if address.Status != VerifiableAddressStatusPending`

sounds good. fixed

selfservice/hook/verification.go
cwei-wish: > I think we should update the status in this method instead!

done.

selfservice/strategy/link/sender.go
```diff
@@ -137,13 +138,20 @@ func (s *Sender) SendVerificationTokenTo(ctx context.Context, f *verification.Fl
...
 		&templates.VerificationValidModel{To: address.Value, VerificationURL: urlx.CopyWithQuery(
 			urlx.AppendPaths(s.r.Config(ctx).SelfPublicURL(nil), verification.RouteSubmitFlow),
 			url.Values{
 				"flow":  {f.ID.String()},
 				"token": {token.Token},
-			}).String()}))
+			}).String()})); err != nil {
+		return err
+	}
+	address.Status = identity.VerifiableAddressStatusPending
```
aeneasr, line 150: Awesome! Could you now please add a test to verify that this is working as intended? Basically sending an email, checking if the status was updated, and then calling the hook again - and making sure the email is not sent twice!

cwei-wish: > Awesome! Could you now please add a test to verify that this is working as intended? Basically sending an email, checking if the status was updated, and then calling the hook again - and making sure the email is not sent twice!

added

cwei-wish:
@aeneasr test cases are added. would you please review it again

aeneasr (approved):

aeneasr:
Great job, thank you!

Issue #598, opened 2020-07-24
Implement identity state and administrative deactivation, deletion of identities


aeneasr:
**Is your feature request related to a problem? Please describe.**

[![](https://mermaid.ink/img/eyJjb2RlIjoiXG5zdGF0ZURpYWdyYW0tdjJcblx0WypdIC0tPiBBY3RpdmU6IGNyZWF0ZVxuICAgIEFjdGl2ZSAtLT4gQWN0aXZlOiB1cGRhdGVcblx0QWN0aXZlIC0tPiBEaXNhYmxlZDogZGlzYWJsZVxuXHREaXNhYmxlZCAtLT4gWypdOiBkZWxldGVcbiAgICBEaXNhYmxlZCAtLT4gQWN0aXZlOiBlbmFibGUiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ)](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiXG5zdGF0ZURpYWdyYW0tdjJcblx0WypdIC0tPiBBY3RpdmU6IGNyZWF0ZVxuICAgIEFjdGl2ZSAtLT4gQWN0aXZlOiB1cGRhdGVcblx0QWN0aXZlIC0tPiBEaXNhYmxlZDogZGlzYWJsZVxuXHREaXNhYmxlZCAtLT4gWypdOiBkZWxldGVcbiAgICBEaXNhYmxlZCAtLT4gQWN0aXZlOiBlbmFibGUiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ)

Issue #595, opened 2020-07-22
Send invite links directly via email


aeneasr:
**Is your feature request related to a problem? Please describe.**

It is currently not possible to send invite links directly to the user via email. Instead, the admin creates the link and somehow (email, sms, whatsapp) sends it to the user.

**Describe the solution you'd like**

It should be possible to make it easier for the administrator to send the link via email directly.

oliverpool:
Do you have any recommendation how to implement this?

Should a new endpoint be added?

Or simply a new optional `SendEmailTo` parameter to the `link.CreateRecoveryLink` struct. If not empty, an email will be sent to this address with the recovery link?

aeneasr:
I think something along those lines would be acceptable! One question though is if the admin can choose the email, or if the recovery email in the DB is used instead 🤔

oliverpool:
I would let the admin choose the email.

If the admin knows the UUID, then it is trivial for him to get the email.

And I think there is no guarantee that the user has an email associated (logging with username for instance?)

aeneasr:
That’s true - another concern though is that the admin „fatfingers“ the email which could send the recovery to the wrong person. Ot, a social engineering attack tricking service desk into sending someone else the recovery link?

oliverpool:
> service desk

Since Kratos is without UI, I think this is out of the scope of Kratos.

For instance as an admin, I except to be able to change the email of a user (if this user lost access to its email address), which could also lead to a "fatfinger" or social engineering attack.

Anyway it would be much better than the current solution (retrieve the link and send it with my private e-mail), which may leak the recovery link in undesirable places (my "sent" folder for instance).

aeneasr:
It’s not out of scope in my opinion. Ory Kratos‘ job is to offer reliable and secure methods to manage and secure identities.

We could have two functional endpoints or methods. One initiates sending the link to the user’s recovery address and the other allows overriding the email. We could also distinguish this using the POST body (e.g. email is not set -> use recovery address; email is set -> use override).

What do you think?

oliverpool:
> One initiates sending the link to the user’s recovery address

I guess that the self-service endpoint already covers this usecase, no?
https://www.ory.sh/kratos/docs/next/self-service/flows/account-recovery

aeneasr:
Yes, but sometimes you might not have this feature enabled for self-service

oliverpool:
Ok, fair enough.

[...]

PR #1312, opened 2021-05-04
feat(identities): add a state to identities

thomasruiz:
## Related issue

https://github.com/ory/kratos/issues/598

## Proposed changes

Add a `disabled_at` field in the identity model. If it is set, the account is disabled. Block session creation when the account is disabled.

## Checklist

      the changes.
      works.

## Further comments

This is draft just to make sure the implementation is correct. Tests and documentation are very much needed, but before going further, I'd rather have a validation from the maintainers.

### DIFF
--- a/identity/identity.go
+++ b/identity/identity.go
@@ -20,6 +20,22 @@ import (
 	"github.com/ory/kratos/x"
 )
 
+// swagger:model identityState
+type State string
+
+func (lt State) IsValid() error {
+	switch lt {
+	case StateActive, StateDisabled:
+		return nil
+	}
+	return errors.New("Identity state is not valid.")
+}
+
+const (
+	StateActive   State = "active"
+	StateDisabled State = "disabled"
+)
+
 // Identity represents an Ory Kratos identity
 //
 // An identity can be a real human, a service, an IoT device - everything that
@@ -50,6 +66,14 @@ type Identity struct {
 	// format: url
 	// required: true
 	SchemaURL string `json:"schema_url" faker:"-" db:"-"`
+
+	// State is the identity's state.
+	//
+	// required: true
+	State State `json:"state" faker:"-" db:"state"`
+
+	// StateChangedAt contains the last time when the identity's state changed.
+	StateChangedAt sqlxx.NullTime `json:"state_changed_at" faker:"-" db:"state_changed_at"`
 
 	// Traits represent an identity's traits. The identity is able to create, modify, and delete traits
 	// in a self-service manner. The input will always be validated against the JSON Schema defined
@@ -129,6 +153,10 @@ func (i *Identity) lock() *sync.RWMutex {
 	return i.l
 }
 
+func (i *Identity) IsActive() bool {
+	return i.State == StateActive
+}
+
 func (i *Identity) SetCredentials(t CredentialsType, c Credentials) {
 	i.lock().Lock()
 	defer i.lock().Unlock()
@@ -182,6 +210,8 @@ func NewIdentity(traitsSchemaID string) *Identity {
 		Traits:              Traits("{}"),
 		SchemaID:            traitsSchemaID,
 		VerifiableAddresses: []VerifiableAddress{},
+		State:               StateActive,
+		StateChangedAt:      sqlxx.NullTime(time.Now()),
 		l:                   new(sync.RWMutex),
 	}
 }
--- a/identity/identity.go
+++ b/identity/identity.go
@@ -20,21 +20,21 @@ import (
 	"github.com/ory/kratos/x"
 )
 
-// swagger:model identityState
+// swagger:enum State
 type State string
-
-func (lt State) IsValid() error {
-	switch lt {
-	case StateActive, StateDisabled:
-		return nil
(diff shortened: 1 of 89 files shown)


aeneasr:
Thank you for the contribution! I think what I meant in the original issue was more along the lines of adding a `State` parameter that is an enum of the different states. That way, we don't have to rely on nil values but instead have a clear cut singular state object :)

thomasruiz:
Would there be another state besides active and disabled? Having a date to know when the user was disabled is a good thing, it allows automatic cleanup and other things. We could have a `state` field and a `state_changed_at` but... I don't see any other state rather that active and disabled. :p

zepatrik:
`invited` might be a possible candidate later on (e.g. to implement #595). I am pretty sure there will be some more coming up :wink:

thomasruiz:
Alright, I see the point. This should be more what you guys had in mind. It still needs tests, and I'm thinking about adding a PATCH method to the identities controller. What do you think?

aeneasr:
Sorry @thomasruiz for dropping the ball on this. Would you still be open to work on this or have you proceeded to other work?

aeneasr:
To regenerate the SDK run `make sdk` in the root directory :)

identity/identity.go
```diff
@@ -20,6 +20,13 @@ import (
 	"github.com/ory/kratos/x"
 )
 
+type State uint8
```
aeneasr, line 23: As discussed on slack, if you do:

```
// swagger:model identityState
type State uint8
```

This should all work! Also, I think it would be a good idea to have the state as a string as that is easier for JSON clients to consume / understand than ints :)

aeneasr:
I think this is ready for review right? :) If so can you please mark it as ready for review? Thank you! :)

thomasruiz:
> I think this is ready for review right? :) If so can you please mark it as ready for review? Thank you! :)

There you go! I hope I didn't mess anything up :)

identity/identity.go
```diff
@@ -20,6 +20,22 @@ import (
 	"github.com/ory/kratos/x"
 )
 
+// swagger:model identityState
+type State string
```
zepatrik, line 24: Sorry to bother you with this, but there is even an enum definition
https://github.com/ory/keto/blob/0ac4fe1667ee1edd133fe3f117e204e3144f6c78/internal/expand/tree.go#L15-L23
You will have to move the const definitions right below the type definition IIRC, then it should work. FYI @aeneasr

identity/identity.go
```diff
@@ -52,6 +68,14 @@ type (
 		// required: true
 		SchemaURL string `json:"schema_url" faker:"-" db:"-"`
 
+		// State is the identity's state.
+		//
+		// required: true
```
aeneasr, line 73: I believe that for Enum you can use:

```suggestion
		// enum:
		// - active
		// - inactive
		// required: true
```

identity/identity.go
```diff
@@ -20,6 +20,22 @@ import (
...
+	switch lt {
+	case StateActive, StateDisabled:
+		return nil
+	}
+	return errors.New("Identity state is not valid.")
+}
+
+const (
+	StateActive   State = "active"
+	StateDisabled State = "disabled"
```
aeneasr, line 36: I would probably name this `inactive` as it is the opposite to `active` :)

identity/identity.go
```diff
@@ -20,6 +20,22 @@ import (
...
 
+// swagger:model identityState
+type State string
+
+func (lt State) IsValid() error {
+	switch lt {
+	case StateActive, StateDisabled:
+		return nil
+	}
+	return errors.New("Identity state is not valid.")
```
aeneasr, line 31: Errors in Go should be lower cased and no dots :)

```suggestion
	return errors.New("identity state is not valid")
```

identity/identity.go
aeneasr: We can use enum in the property field - commented in the PR below

aeneasr (commented):
Awesome! Just have two small things but this is basically good to merge! Could you also add a small section on the new identity [state to the docs](https://github.com/ory/kratos/blob/master/docs/docs/concepts/identity-data-model.mdx)? :)

identity/identity.go
zepatrik: Why would we not define it once for the `State` type and it just works everywhere?

aeneasr: I don't think go-swagger supports that

zepatrik: It does, the above referenced definition is translated into https://github.com/ory/keto/blob/0ac4fe1667ee1edd133fe3f117e204e3144f6c78/.schema/swagger.json#L1025-L1032
Added by this PR https://github.com/go-swagger/go-swagger/pull/2176

thomasruiz: I tried it and it does work. :)

aeneasr:
🙄 circle ci has an outage..

thomasruiz:
A

aeneasr:
🤣

aeneasr:
Thank you so much, a job well done! :)

Issue #1322, opened 2021-05-08
There is no way to tell the difference between a user who signed in and a user who is in the recovery mode


radekg:
User error.

radekg:
Something definitely isn't right here. I initialize the recovery from the browser and I am logged out. I receive the email with the flow and token. I click on the link from the email and am taken to the `selfservice.flows.settings.ui_url` with an unrelated ID of the flow to the one from the email. I can then fetch the flow user the `/self-service/settings/flows` but not `/self-service/recovery/flows` (http 404). My flow comes back as settings and you can see this here:

![Screen Shot 2021-05-09 at 01 12 09](https://user-images.githubusercontent.com/62946/117556001-af591080-b064-11eb-9a1a-0b5cbcc92c23.png)

When I have no cookies, as in, via cURL, I get the redirect to the `selfservice.flows.recovery.ui_url`. With cookies, in the browser I am being sent to `selfservice.flows.settings.ui_url` with an unrelated settings flow.

radekg:
Hi @aeneasr, would a simple flag like that work for you? If yes, then all the blocking of actions that should not be allowed to a `recovery` user could be implemented. Eventually, what a desired implementation could look like?

radekg:
The PR above introduces the recovery session. With that flag, it's really easy to distinguish between the intent to recover and a signed in user. The video attached shows the concept.

https://user-images.githubusercontent.com/62946/117588739-aaa86100-b125-11eb-9f8c-51142d1bc019.mp4

aeneasr:
Account recovery is essentially a privileged session which allows one to update their credentials, thus it is analogous to setting a session.

Resetting a password is not all that you can do here, it might also be to recover a broken link to an OIDC provider. Thus, the flow is more complex than showing a mask for changing the password!

However, adding a flag to indicate a recovery session is definitely a good idea!

aeneasr:
Once #1624 is merged you will get a list of authentication methods in the session where `link_recovery` is referring to a session issued as part of the recovery!

PR #1323, opened 2021-05-09
feat: Recovery session

radekg:
## Related issue

This is a naive implementation of https://github.com/ory/kratos/issues/1322.

## Proposed changes

Adds a `recovery` flag on the session object to allow differentiating between a signed in user and a user in recovery mode.

Screen Shot 2021-05-09 at 04 12 50

## Checklist

      the changes.
      works.

## Further comments

I had to render the migrations with `--replace`.


selfservice/flow/settings/handler.go
```diff
@@ -413,6 +413,22 @@ func (h *Handler) submitSettingsFlow(w http.ResponseWriter, r *http.Request, ps
...
+	// if the submission was from a recovery session, we have to end the recovery session and
+	// reissue a new session which isn't a recovery session anymore
+	if ss.Recovery {
+		if err := h.d.SessionManager().PurgeFromRequest(r.Context(), w, r); err != nil {
+			h.d.SettingsFlowErrorHandler().WriteFlowError(w, r, node.DefaultGroup, f, ss.Identity, err)
+			return
+		}
+		newSession := session.NewActiveSession(ss.Identity, h.d.Config(r.Context()), time.Now())
+
+		// this could sit nicely behind a flag to leave the user signed out after the password change
```
radekg, line 425: This could sit behind a config flag, for example `selfservice.flows.recovery.after.keep_logged_in: bool`.

aeneasr (commented):
Thank you, I understand that the flow is a bit weird right now. From a security perspective, account recovery is gaining access to a privileged session. The only difference is that most applications do not allow you to proceed without updating your credentials (password, oidc, 2fa, ...).

Generally I think we could have the session type set. If you find things that are relevant from a security perspective regarding this we can also look at changing the session behavior here!

selfservice/flow/settings/handler.go
radekg: I don't think this is a security related issue. From a user perspective, the latter one is definitely more ergonomic. Perhaps it might be worth asking why is it so common to require reauthenticating after recovery? This can always be done via some sort of reverse proxy setup but it is pretty complicated and non-ergonomic for the developer.

Having a flag gives the developer an option to choose. Maybe they do have certain ciso requirements to follow. Not having the option either disqualifies a product or puts a lot of effort on development team. Considering that the construct of having to reauth after recovery seems to be a common one and adds at most a few lines of code (setting flag), why not? I'm not fixed on this as being signed in works for me just fine.

aeneasr:
Thinking about this more, I agree that the current behaviour is a bit weird. I will try to think about some better approaches but don't have yet a good idea.

I think the PR is a good direction andI also agree that creating a session is kinda weird.

aeneasr:

Thank you!

aeneasr:
We’ll put a reworked recovery flow in the feature backlist!

g13013:
@aeneasr what's the plan for the rework ?

I thought that the config could allow different link redirect url for the recovery flow!?

aeneasr:
Primarily this: https://github.com/ory/kratos/issues/1451

PR #1332, opened 2021-05-11
feat: Dropbox like bcrypt-aes-256 hasher

harnash:
## Proposed changes

This is the proposal for adding a Dropbox-like hashing algorithm: https://dropbox.tech/security/how-dropbox-securely-stores-your-passwords. 

## Checklist

      the changes.
      works.

## Further comments
Since this is based on #1137 and touches some code introduced there I'm pinging @seremenko-wish to take a look as well if possible. We will do some internal tests after initial 👍🏻 .


hash/hash_comparator.go
```diff
@@ -60,24 +140,14 @@ func CompareArgon2id(_ context.Context, password []byte, hash []byte) error {
...
-
-func IsArgon2idHash(hash []byte) bool {
-	res, _ := regexp.Match("^\\$argon2id\\$", hash)
-	return res
-}
-
 func decodeArgon2idHash(encodedHash string) (p *config.Argon2, salt, hash []byte, err error) {
 	parts := strings.Split(encodedHash, "$")
-	if len(parts) != 6 {
+	if len(parts) != 5 {
```
zepatrik, line 145: How is this not breaking previously stored hashes? I don't think that the indexes should change here, the format strings did not change anything regarding `$` positions.

hash/hasher_bcrypt.go
```diff
@@ -24,15 +28,25 @@ func NewHasherBcrypt(c BcryptConfiguration) *Bcrypt {
...
 func (h *Bcrypt) Generate(ctx context.Context, password []byte) ([]byte, error) {
 	if err := validateBcryptPasswordLength(password); err != nil {
-		return nil, err
+		return nil, errors.WithStack(err)
 	}
 
 	hash, err := bcrypt.GenerateFromPassword(password, int(h.c.Config(ctx).HasherBcrypt().Cost))
 	if err != nil {
-		return nil, err
+		return nil, errors.WithStack(err)
```
zepatrik, line 36: Nice find :wink:

hash/hash_comparator.go
```diff
@@ -2,54 +2,134 @@ package hash
...
-func CompareArgon2id(_ context.Context, password []byte, hash []byte) error {
+func IsPasswordHash(hash []byte) bool {
+	_, _, err := ParsePasswordHash(hash)
+	return err == nil
+}
+
+func ParsePasswordHash(input []byte) (algorithm, hash string, err error) {
+	hashParts := strings.SplitN(string(input), "$", 3)
+	if len(hashParts) != 3 {
+		err = ErrUnknownHashFormat
```
zepatrik, line 64:
```suggestion
		err = errors.WithStack(ErrUnknownHashFormat)
```

hash/hash_comparator.go
```diff
@@ -2,54 +2,134 @@ package hash
...
 	"golang.org/x/crypto/argon2"
 	"golang.org/x/crypto/bcrypt"
+	"golang.org/x/crypto/sha3"
 
 	"github.com/ory/kratos/driver/config"
 )
 
 var ErrUnknownHashAlgorithm = errors.New("unknown hash algorithm")
+var ErrUnknownHashFormat = errors.New("unknown hash format")
+var ErrNonceSizeIncorrect = errors.New("nonce size is incorrect")
```
zepatrik, line 23: These errors are actually already annotated with a stack, but it is basically the `init` functions stack. IMO, we should change them to
```suggestion
var ErrUnknownHashAlgorithm = fmt.Errorf("unknown hash algorithm")
var ErrUnknownHashFormat = mt.Errorf("unknown hash format")
var ErrNonceSizeIncorrect = mt.Errorf("nonce size is incorrect")
```

hash/hash_comparator.go
```diff
@@ -2,54 +2,134 @@ package hash
...
 }
 
-func CompareArgon2id(_ context.Context, password []byte, hash []byte) error {
+func IsPasswordHash(hash []byte) bool {
+	_, _, err := ParsePasswordHash(hash)
+	return err == nil
+}
+
+func ParsePasswordHash(input []byte) (algorithm, hash string, err error) {
+	hashParts := strings.SplitN(string(input), "$", 3)
```
zepatrik, line 62: Previously stored bcrypt hashes will not have the alg ID prefixed right? We have to handle those legacy hashes and replace them in the database somehow.

hash/hasher_bcrypt.go
```diff
@@ -24,15 +28,25 @@ func NewHasherBcrypt(c BcryptConfiguration) *Bcrypt {
...
+	}
+
+	var b bytes.Buffer
+	if _, err := fmt.Fprintf(
+		&b,
+		"$%s%s",
```
zepatrik, line 42: Let's stay consistent?

```suggestion
		"$%s$%s",
```

I mean, `bcrypt` is a prefix of `bcryptAes`, so we should really have the delimiter here.

hash/hasher_bcrypt_aes.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	gcm, err := cipher.NewGCM(block)
+	if err != nil {
+		return nil, err
+	}
+
+	nonce := make([]byte, gcm.NonceSize())
+	if _, err := io.ReadFull(rand.Reader, nonce); err != nil {
+		return nil, err
```
zepatrik, line 47:
```suggestion
		return nil, errors.WithStack(err)
```

hash/hasher_bcrypt_aes.go
```diff
@@ -0,0 +1,83 @@
...
+func (h *BcryptAES) Generate(ctx context.Context, password []byte) ([]byte, error) {
+	cfg := h.c.Config(ctx).HasherBcryptAES()
+	sh := sha3.New512()
+	sh.Write(password)
+	bcryptPassword, err := bcrypt.GenerateFromPassword(sh.Sum(nil), int(cfg.Cost))
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	hash, err := h.aes256Encrypt(bcryptPassword, []byte(cfg.Key))
```
zepatrik, line 66: This key has to be always constant, otherwise previous hashes will fail right? I guess we need some way of rotating those keys, or make it clear that it can never be changed once used.

hash/hasher_test.go
```diff
@@ -122,30 +137,97 @@ func TestComparatorBcryptFail(t *testing.T) {
...
+			_, reg := internal.NewFastRegistryWithMocks(t)
 			mod := make([]byte, len(pw))
 			copy(mod, pw)
 			mod[len(pw)-1] = ^pw[len(pw)-1]
 
-			err := hash.CompareBcrypt(context.Background(), pw, mod)
+			err := hash.CompareBcrypt(context.Background(), reg.Config(context.Background()), pw, string(mod))
+			assert.Error(t, err)
+		})
+	}
```
zepatrik, line 148: Can you please add test cases for the "legacy" bcrypt format we had until now?

hash/hasher_test.go
```diff
@@ -122,30 +137,97 @@ func TestComparatorBcryptFail(t *testing.T) {
...
+		mkpw(t, 32),
+		mkpw(t, 64),
+		mkpw(t, 72),
+		mkpw(t, 96),
+		mkpw(t, 128),
+	} {
+		t.Run(fmt.Sprintf("case=%d", k), func(t *testing.T) {
+			mod := make([]byte, len(pw))
+			copy(mod, pw)
+			mod[len(pw)-1] = ^pw[len(pw)-1]
```
zepatrik, line 198: Just to make it obvious
```suggestion
			mod[len(pw)-1] = ^mod[len(pw)-1]
```

hash/hasher_test.go
```diff
@@ -122,30 +137,97 @@ func TestComparatorBcryptFail(t *testing.T) {
...
+	assert.Nil(t, hash.CompareBcrypt(context.Background(), p, []byte("test"), "$2a$15$GRvRO2nrpYTEuPQX6AieaOlZ4.7nMGsXpt.QWMev1zrP86JNspZbO"))
+	assert.Error(t, hash.Compare(context.Background(), p, []byte("test"), []byte("$bcrypt$2a$15$GRvRO2nrpYTEuPQX6AieaOlZ4.7nMGsXpt.QWMev1zrP86JNspZb1")))
+
+	assert.Nil(t, hash.Compare(context.Background(), p, []byte("test"), []byte("$argon2id$v=19$m=32,t=2,p=4$cm94YnRVOW5jZzFzcVE4bQ$MNzk5BtR2vUhrp6qQEjRNw")))
+	assert.Nil(t, hash.CompareArgon2id(context.Background(), p, []byte("test"), "$v=19$m=32,t=2,p=4$cm94YnRVOW5jZzFzcVE4bQ$MNzk5BtR2vUhrp6qQEjRNw"))
+	assert.Error(t, hash.Compare(context.Background(), p, []byte("test"), []byte("$argon2id$v=19$m=32,t=2,p=4$cm94YnRVOW5jZzFzcVE4bQ$MNzk5BtR2vUhrp6qQEjRN2")))
 
-	assert.Nil(t, hash.Compare(context.Background(), []byte("test"), []byte("$2a$15$GRvRO2nrpYTEuPQX6AieaOlZ4.7nMGsXpt.QWMev1zrP86JNspZbO")))
-	assert.Nil(t, hash.CompareBcrypt(context.Background(), []byte("test"), []byte("$2a$15$GRvRO2nrpYTEuPQX6AieaOlZ4.7nMGsXpt.QWMev1zrP86JNspZbO")))
-	assert.Error(t, hash.Compare(context.Background(), []byte("test"), []byte("$2a$15$GRvRO2nrpYTEuPQX6AieaOlZ4.7nMGsXpt.QWMev1zrP86JNspZb1")))
```
zepatrik, line 142: This case has to stay exactly as it was to ensure backwards compatibility.

zepatrik (changes requested):
I did not review the actual algorithm proposed by dropbox and if this implementation follows it, but there are some small issues I found. Especially key rotation gives me headaches and needs some thinking :sweat_smile:

driver/config/.schema/config.schema.json
```diff
@@ -1544,20 +1544,40 @@
...
               "maximum": 31,
               "default": 12
             }
           }
+        },
+        "bcryptaes": {
+          "title": "Configuration for the Bcrypt-AES hasher.",
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["cost"],
```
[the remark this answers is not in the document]
harnash: Yup. 
```suggestion
          "required": ["cost", "key"],
```

hash/hash_comparator.go
harnash: No they will not. In this case I will add a migration if that is OK.

hash/hash_comparator.go
harnash: If we assume that after migration all `bcrypt` passwords will have prefix it should work. In this implementation `Compare()` will strip prefixes (via `ParsePasswordHash()`) and this function will receive hash without prefix `argon2id`, `bcrypt` etc. Take a look at the tests - high level `Compare()` receives the same input, implementation of the lower level `Compre*()` differs. I did it this way to have unified input to those function and avoid parsing/splitting hashes multiple time in the process and pass only relevant part of the hash down the call stack.

hash/hasher_bcrypt.go
harnash: OK. That will need a bit of work. Will do that in separate commit.

hash/hasher_bcrypt_aes.go
harnash: Makes sense. I'll check how much work it would require to add this functionality (similarly to secrets).

hash/hasher_test.go
harnash: @zepatrik: could you explain what you want to test? Current approach does not allow "legacy" hash format and adding migration to the PR should convert "legacy" BCrypt hashes to "new" format with prefix. I could support "legacy" format if that desirable but then we would have additional complexity.

hash/hasher_test.go
harnash: Unless we migrate the data? I'm not sure what is the policy here. If we want to keep the compatibility (without doing migration) then I would need to add that.

hash/hash_comparator.go
```diff
@@ -2,54 +2,133 @@ package hash
...
+var ErrUnknownHashAlgorithm = fmt.Errorf("unknown hash algorithm")
+var ErrUnknownHashFormat = fmt.Errorf("unknown hash format")
 
-func Compare(ctx context.Context, password []byte, hash []byte) error {
-	if IsBcryptHash(hash) {
-		return CompareBcrypt(ctx, password, hash)
-	} else if IsArgon2idHash(hash) {
-		return CompareArgon2id(ctx, password, hash)
-	} else {
+func Compare(ctx context.Context, cfg *config.Config, password []byte, hash []byte) error {
```
rgruchalski-klarrio, line 24: The API here appears to be inconsistent. This function takes `hash []byte` but concrete implementation uses a `string`. Why not `[]byte` everywhere?

radekg, line 24: The API here appears to be inconsistent. This function takes `hash []byte` but concrete implementation uses a `string`. Why not `[]byte` everywhere?

harnash: Fix in 52b4daea

hash/hasher_test.go
aeneasr: Yes, definitely, we need backwards compatibility!

harnash: Alright. I'll add it then.

hash/hasher_bcrypt.go
harnash: Actually I remember now that we don't want to add this `$` here since BCrypt hashes have it already i.e. `$2a$15$GRvRO2nrpYTEuPQX6AieaOlZ4.7nMGsXpt.QWMev1zrP86JNspZbO` so it will be consistent with other functions.

hash/hasher_test.go
harnash: Done in 584fbbf

hash/hasher_test.go
harnash: Done in 584fbbf

harnash:
@zepatrik , @aeneasr  I have a question regarding passwords. Currently we have configuration for each hasher and we have an `algorithm` field which is used as default hashing algorithm for new credentials. There are other hashers that can be configured but they are used as "read-only". I have the following questions:
1. Do all hashers have to be defined when the are not configured as default one? For BCrypt there are sane default values but for BCrypt-AES `key` is required and is not provided by default. This would mean that this hasher will not function properly if there will be credentials in the DB which uses this hasher. It may be fine but I think user should be aware of that. It may also be set as required but it may be confusing if user does not plan to use this hasher at all.
2. When using multiple keys for BCrypt-AES hasher "older" keys will be automatically upgraded to a newest one similiarly to secrets. Is that OK or should it be optional?
3. Do we want to be able to "upgrade" from any legacy hasher to a newest/default one (configurable)?

hash/hasher_bcrypt_aes.go
harnash: Done in 66959d8

aeneasr:
> Do all hashers have to be defined when the are not configured as default one? For BCrypt there are sane default values but for BCrypt-AES key is required and is not provided by default. This would mean that this hasher will not function properly if there will be credentials in the DB which uses this hasher. It may be fine but I think user should be aware of that. It may also be set as required but it may be confusing if user does not plan to use this hasher at all.

I see - I think it is ok to require people to set an AES key if they intend on, or have used, this in the past.

> When using multiple keys for BCrypt-AES hasher "older" keys will be automatically upgraded to a newest one similiarly to secrets. Is that OK or should it be optional?

That is ok and expected for key rotation :) Nice :)

> Do we want to be able to "upgrade" from any legacy hasher to a newest/default one (configurable)?

The idea is that the hashers are always interoperable. If you use an BCrypt and want to move to BCrypt-AES it should be possible - vice versa as well! But there is always only "one" main hasher if that makes sense.

harnash:
> > Do all hashers have to be defined when the are not configured as default one? For BCrypt there are sane default values but for BCrypt-AES key is required and is not provided by default. This would mean that this hasher will not function properly if there will be credentials in the DB which uses this hasher. It may be fine but I think user should be aware of that. It may also be set as required but it may be confusing if user does not plan to use this hasher at all.
> 
> I see - I think it is ok to require people to set an AES key if they intend on, or have used, this in the past.

OK. So now the `key` parameter is optional unless `bcryptaes` is set as the main algorithm which should be in line with the above assumptions.

> > Do we want to be able to "upgrade" from any legacy hasher to a newest/default one (configurable)?
> 
> The idea is that the hashers are always interoperable. If you use an BCrypt and want to move to BCrypt-AES it should be possible - vice versa as well! But there is always only "one" main hasher if that makes sense.

OK. I'll leave this implementation to a separate PR then. Don't want to over complicate this more than necessary.

Now. I think implementation wise this PR is good to go. I would like to do some testing on our environment to check if the code does what it suppose to do. I'll let you know when that is done.

aeneasr:
> OK. So now the key parameter is optional unless bcryptaes is set as the main algorithm which should be in line with the above assumptions.


ok nice :)

docs/docs/concepts/credentials/username-email-password.mdx
```diff
@@ -52,6 +53,25 @@ hashers:
     cost: 12
 ```
 
+Similarly, Bcrypt wrapped with AES can be also configured. In addition to
+standard to Bcrypt `cost` parameter (default value is 12) it requires AES `key`
+to be specified. Since we want to use specific AES variant (namely AES-256) the
+`key` parameter should contain 32-bytes long strings (see:
```
aeneasr, line 59:
```suggestion
`key` parameter must contain 32-bytes long strings (see:
```

driver/registry_default.go
```diff
@@ -360,9 +360,12 @@ func (m *RegistryDefault) SessionHandler() *session.Handler {
 
 func (m *RegistryDefault) Hasher() hash.Hasher {
 	if m.passwordHasher == nil {
-		if m.c.HasherPasswordHashingAlgorithm() == "bcrypt" {
+		switch m.c.HasherPasswordHashingAlgorithm() {
+		case "bcrypt":
 			m.passwordHasher = hash.NewHasherBcrypt(m)
-		} else {
+		case "bcryptAes":
```
aeneasr, line 366: To stay consistent how about `bcryptaes`?

hash/hash_comparator.go
```diff
@@ -1,55 +1,160 @@
...
+	decoded := make([]byte, hex.DecodedLen(len(data)))
+	_, err = hex.Decode(decoded, data)
+	if err != nil {
+		return nil, err
+	}
+
+	nonce := decoded[:gcm.NonceSize()]
+	ciphertext := decoded[gcm.NonceSize():]
+
+	return gcm.Open(nil, nonce, ciphertext, nil)
```
aeneasr, line 114: I would feel more comfortable if we used https://github.com/gtank/cryptopasta/blob/master/encrypt.go

hash/hash_comparator.go
```diff
@@ -1,55 +1,160 @@
...
+	for _, key := range hasherCfg.Key {
+		aesDecrypted, lastError = aes256Decrypt(hash[1:], key)
+		if lastError == nil {
+			break
+		}
+	}
+
+	if lastError != nil {
+		return errors.WithStack(lastError)
+	}
```
aeneasr, line 136: See above

hash/hasher_bcrypt_aes.go
```diff
@@ -0,0 +1,88 @@
...
+	if _, err := io.ReadFull(rand.Reader, nonce); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	hash := append(nonce, gcm.Seal(nil, nonce, data, nil)...)
+	encoded := make([]byte, hex.EncodedLen(len(hash)))
+	hex.Encode(encoded, hash)
+
+	return encoded, nil
+}
```
aeneasr, line 56: https://github.com/gtank/cryptopasta/blob/master/encrypt.go

driver/registry_default.go
harnash: Done in 2520dd64

hash/hasher_bcrypt_aes.go
harnash: Done in c5b5e31e

hash/hash_comparator.go
harnash: Done in c5b5e31e

hash/hash_comparator.go
harnash: Done in c5b5e31e?

aeneasr:
@harnash sorry for the misunderstanding, what I meant was not to copy https://github.com/gtank/cryptopasta/blob/master/encrypt.go into this repo but instead use the import with their functions as they might be fixing things in the future and we don't want to keep checking with them :)

Regarding the PR itself, I'll have some down time the next week to read the Dropbox proposal and better understand how this PR implements it. I think we can get it merged in one max two weeks!

harnash:
> @harnash sorry for the misunderstanding, what I meant was not to copy https://github.com/gtank/cryptopasta/blob/master/encrypt.go into this repo but instead use the import with their functions as they might be fixing things in the future and we don't want to keep checking with them :)

Well I've decided to copy it instead of adding an another dependency due to:
1. `TL;DR- Copy & paste your crypto code from here instead of Stack Overflow.`
2. `little copy-paste is better than little dependency` following https://go-proverbs.github.io/ and looking at activity in this repo I highly doubt it will be updated unless there are some changes in the std library (which is also highly unlikely)

In the end it is your call @aeneasr.

> 
> Regarding the PR itself, I'll have some down time the next week to read the Dropbox proposal and better understand how this PR implements it. I think we can get it merged in one max two weeks!

No hurry there. We will be conducting some testing next week to see if everything is in order so please wait for green light from me if you can.

harnash:
@aeneasr I did tests on our dev cluster and it seems those changes are working. Did not found any apparent issues.

hash/hash_comparator.go
```diff
@@ -1,55 +1,169 @@
...
+var hashSeparator = []byte("$")
+var bcryptLegacyPrefix = regexp.MustCompile(`^2[abzy]?$`)
+
+func Compare(ctx context.Context, cfg *config.Config, password []byte, hash []byte) error {
+	algorithm, realHash, err := ParsePasswordHash(hash)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	switch {
```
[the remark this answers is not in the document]
aeneasr: Second guessing my previous thought we probably want to limit the amount of hashing methods to bcrypt family and argon2 for OS or highly confidential workloads. I think the current wrapping is fine and we probably want to have an onion model if we want to support more hash types for importing.

hash/hash_comparator.go
```diff
@@ -1,55 +1,169 @@
...
+	default:
+		err = ErrUnknownHashAlgorithm
+		return
+	}
+}
+
+// aes256Decrypt decrypts data using 256-bit AES-GCM.  This both hides the content of
+// the data and provides a check that it hasn't been altered. Expects input
+// form nonce|ciphertext|tag where '|' indicates concatenation.
+func aes256Decrypt(ciphertext []byte, key *[32]byte) (plaintext []byte, err error) {
```
aeneasr, line 97: I would move this to the aes hasher file and also mirror a comment I made earlier suggesting to use cryptopasta (as an import) instead of copy/pasting or writing our own here. I’ve implemented AES GCM before (with stdlib) and it‘s very easy to make small mistakes in the encoding or key preparations.

aeneasr (changes requested):
Thank you and sorry for the long review times. On a high level the Dropbox model looks interesting. I could not find a lot of literature on this topic and find the article from dropbox lacking of details (e.g. pseudo code implementation). Do you have any references or reference implementations you could point at?

Generally, pre-hashing passwords has the disadvantage that all passwords are of the same length. While not problematic in this context (imo), knowing the length of a password significantly decreases the amount of guesses one needs to take. However, the entropy is large enough for this not to be a problem as far as I can tell.

Regarding the implementation, I think this looks pretty good. I would definitely like to remove the AES GCM import and replace it with cryptopasta as I myself have made an error in the AES GCM implementation which was pretty annoying to fix for existing installations. It was related to the encoding iirc.

I think this will probably be the last hash mechanism we accept for some while. I think the hash algorithm guessing is already pretty complex and it would be good to find a better solution such as storing the password hash type alongside the hash in credentials config instead of the hash itself.

Also, is $bcryptaes$ a standard encoding for this type of hash? We want to stay compatible with other systems (e.g. laravel, BSD, …) that support this type of hashing - so if there is a standard we should follow it!

In any case, thank you for your great work so far!

aeneasr:
As this has been mentioned in another issue, using BCrypt on an unsalted/unpeppered SHA could potentially open the doors for so-called password shucking as explained in the defcon talk: https://youtu.be/OQD3qDYMyYQ

On OWASP it is explicitly discouraged to pre-hash passwords for BCrypt inputs due to this: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pre-hashing-passwords

With these in mind I am very mindful of actually merging this without further good counter arguments. I hope you understand!

harnash:
> Thank you and sorry for the long review times. On a high level the Dropbox model looks interesting. I could not find a lot of literature on this topic and find the article from dropbox lacking of details (e.g. pseudo code implementation). Do you have any references or reference implementations you could point at?

Sadly no. Back in a day when BCrypt was gaining traction and we decided to move away from older hashing algorithms we were looking at options that allows some kind of flexibility but is more secure than BCrypt alone and Dropbox's implementation was quite popular around that time. There are chances that other companies are using similar approach and I think it is a good alternative for plain Argon2id or BCrypt. It is also a good example of algorithm that uses pepper instead of just salt/nonce so other can use it as an template for their implementations if needed.

> 
> Regarding the implementation, I think this looks pretty good. I would definitely like to remove the AES GCM import and replace it with cryptopasta as I myself have made an error in the AES GCM implementation which was pretty annoying to fix for existing installations. It was related to the encoding iirc.
> 

Sure. I'll add imports and cleanup the code.

> I think this will probably be the last hash mechanism we accept for some while. I think the hash algorithm guessing is already pretty complex and it would be good to find a better solution such as storing the password hash type alongside the hash in credentials config instead of the hash itself.

I guess that would be the next logical step forward. Having constant prefixes that then can be extracted to a separate field should help in migrations.

> 
> Also, is $bcryptaes$ a standard encoding for this type of hash? We want to stay compatible with other systems (e.g. laravel, BSD, …) that support this type of hashing - so if there is a standard we should follow it!

I am not aware of any particular standard. We can choose whatever we want here I guess. It is a construct of multiple hashes so I guess [MCF](https://cryptopp.com/wiki/Modular_Crypt_Format) standard does not apply here.
> 
> In any case, thank you for your great work so far!

You're welcome!

driver/config/config.go
```diff
@@ -307,6 +313,26 @@ func (p *Config) HasherBcrypt() *Bcrypt {
 	return &Bcrypt{Cost: cost}
 }
 
+func (p *Config) HasherBcryptAES() *BcryptAES {
+	// warn about usage of default values and point to the docs
+	// warning will require https://github.com/ory/viper/issues/19
```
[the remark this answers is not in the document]
harnash: Fixed in 64456098.

driver/config/config.go
```diff
@@ -307,6 +313,26 @@ func (p *Config) HasherBcrypt() *Bcrypt {
...
+	cost := uint32(p.p.IntF(ViperKeyHasherBcryptAESCost, int(BcryptDefaultCost)))
+	if !p.IsInsecureDevMode() && cost < BcryptDefaultCost {
+		cost = BcryptDefaultCost
+	}
+
+	keys := p.p.Strings(ViperKeyHasherBcryptAESKey)
```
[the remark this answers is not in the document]
harnash: Fix in ff570033

hash/hash_comparator.go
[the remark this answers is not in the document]
harnash: I agree. We would probably need more flexible solution here so when dealing with very old systems that grew a number of hasher along the way it would not impact performance too much.

hash/hash_comparator.go
harnash: Done in b9aa9132.

driver/config/.schema/config.schema.json
```diff
@@ -1544,20 +1544,46 @@
...
+              "title": "Cost of the BCrypt-AES algorithm. Minimum is 4 when --dev flag is used and 12 otherwise.",
+              "type": "integer",
+              "minimum": 4,
+              "maximum": 31,
+              "default": 12
+            },
+            "key": {
+              "type": "array",
+              "title": "Encryption Keys for BCrypt-AES hasher",
+              "description": "The first key in the array is used for encrypting/verifying credentials while all other keys are used to verify older hashes that were encrypted with that old key.",
```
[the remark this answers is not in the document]
harnash: Fix in b60738bf8b2e5eab7f82952ded712884cf1ee712.

driver/config/.schema/config.schema.json
```diff
@@ -1544,20 +1544,46 @@
...
+          "additionalProperties": false,
+          "properties": {
+            "cost": {
+              "title": "Cost of the BCrypt-AES algorithm. Minimum is 4 when --dev flag is used and 12 otherwise.",
+              "type": "integer",
+              "minimum": 4,
+              "maximum": 31,
+              "default": 12
+            },
+            "key": {
```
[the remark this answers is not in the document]
harnash: Fix in e4cc6f66585497e008bc3b6332ec9ba2740267d7.

harnash:
> As this has been mentioned in another issue, using BCrypt on an unsalted/unpeppered SHA could potentially open the doors for so-called password shucking as explained in the defcon talk: https://youtu.be/OQD3qDYMyYQ
> 
> On OWASP it is explicitly discouraged to pre-hash passwords for BCrypt inputs due to this: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pre-hashing-passwords
> 
> With these in mind I am very mindful of actually merging this without further good counter arguments. I hope you understand!

Okay. So shall we then introduce pepper (again needs to be rotated as AES key) to mitigate this threat? Seems like a reasonable upgrade to the algorithm. I guess at the time Dropbox was introducing this approach password shucking wasn't yet widely known (or known at all).

radekg:
A question from an observer: what problem is this pr attempting to solve?

harnash:
> A question from an observer: what problem is this pr attempting to solve?

I would not call this a problem solving pr but rather new feat that introduces new hasher (layered with AES key rotation pepper that is somewhat different from the pure BCrypt and ArgonId) + some additional refactor around hashing (comparison and storing).

aeneasr:
> Okay. However having it all wrapped in AES should prevent shucking no? Anyway as a best practice we should probably introduce pepper (again needs to be rotated as AES key) to the password sha to discourage this behavior. Seems like a reasonable upgrade to the algorithm. I guess at the time Dropbox was introducing this approach password shucking wasn't yet widely known (or known at all).

Wrapping it in AES adds a different defense mechanism which, for example, prevents SQL injection of `SELECT hashes FROM pws`. It is not a mitigation for shucking where the vulnerability is using a "low" primitive unsalted hashing mechanism such as SHA1, MD5, or others which have been leaked en masse in the wild before. Unfortunately, the Dropbox model uses an unsalted hash primitive as the innermost hash methodology, making it vulnerable to exactly this attack. Of course, if you do not have the hashes in raw form, it doesn't apply - but the proper mitigation is salting / peppering the innermost hash, regardless of the outer functions.

Unfortunately I am not a password cracking professional so I do not know what the implications are of using salted SHA wrapped by BCrypt. However, the OWASP guidance is pretty clear in this regard and we should follow it. From the guide:

> An alternative approach is to pre-hash the user-supplied password with a fast algorithm such as SHA-256, and then to hash the resulting hash with bcrypt (i.e., bcrypt(base64(hmac-sha256(data:$password, key:$pepper)), $salt, $cost)). This is a dangerous (but common) practice that should be avoided due to password shucking and other issues when combining bcrypt with other hash functions.

This to me is a clear indicator that even a peppered or salted HMAC / SHA should be avoided: "is a dangerous (but common) practice that should be avoided". I also found more proof that pre-hashing is not a good idea unless every call within the library and framework are vetted:

https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html#The-Major-Problem

As such I can not merge the current PR with good consciousness even if I feel terribly sorry for all the work you have put in. I checked the git blame of the OWASP document and it appears to have changed from not-really-encouraging-but-also-not-really-discouraging pre-hashing for BCrypt a few months ago to strictly discouraging it. So this is a recent development in the light of new research.

My call is to remove this pre-hashing. It serves no purpose except for mapping a larger keyspace (e.g. 90 char pws) to a fixed smaller one and vice versa. It has been proven time and again that wrapping hashes in hashes is not a good idea. The cryptographers who initially defined BCrypt chose a raw password input for a purpose! This was unfortunately already my hunch when reading the Dropbox blog post - engineers trying to outsmart cryptographers.

Regarding AES encryption of hashes, one downside is managing the encryption key. If you lose that, your hashes are moot. We can probably leverage the `secrets.default` to allow administrators to reuse one secret here but it has some serious operational aspects to it.

Please also keep in mind that AES-GCM has also several limitations such as the amount of data you can encrypt: https://soatok.blog/2020/05/13/why-aes-gcm-sucks/

In Ory Hydra, we use AES-GCM to encrypt cryptographic keys (e.g. JWKS / RSA / ... for signing ID Tokens) which usually never exceed 100MB during the lifetime of Ory Hydra. This is explained better in the "short nonce" section of [this blogpost](https://soatok.blog/2020/05/13/why-aes-gcm-sucks/). Depending on how many users you have this can pose a threat.

I'll have to do some more thinking about AES-GCM for additional password security. It looks like OWASP is encouraging something along the lines of HMAC ( https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#peppering ) for peppering but they do not make a clear recommendation so I wouldn't bet my horse on it.

My mindspace right now is that I would avoid merging both SHA wrapping as well as AES-GCM without further research proving the effectiveness and security of the approach, or a positive sign from a well-known password cryptographer.

aeneasr:
I think [the conclusion from this blog post](https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html#The-Major-Problem) summarizes it very well:

> The underlying problem is that combining cryptographic operators that weren’t designed to be combined can be disastrous. Is it possible to do so safely? Yes. Is it a good idea to do it? No. This particular case is just one example where combining operations can be exceedingly dangerous.

zepatrik (commented):
In all of this discussion, I would like to bring you back to the basic idea behind all of this. To me it seems like the discussion went a bit too far into the details to see the big picture.

### Inner SHA

From the Dropbox post:

> First, the plaintext password is transformed into a hash value using SHA512. This addresses two particular issues with bcrypt. Some implementations of bcrypt truncate the input to 72 bytes, which reduces the entropy of the passwords. Other implementations don’t truncate the input and are therefore vulnerable to DoS attacks because they allow the input of arbitrarily long passwords. By applying SHA, we can quickly convert really long passwords into a fixed length 512 bit value, solving both problems.

For our plain bcrypt implementation introduced by #1169 we decided to return an error: https://github.com/ory/kratos/blob/103880c4fee41b1a6795eb031f921801891596f5/hash/hasher_bcrypt.go#L38-L49

So for this PR, why not apply the exact same approach and ditch the inner SHA completely?

### Outer AES

Again from Dropbox:

> Finally, the resulting bcrypt hash is encrypted with AES256 using a secret key (common to all hashes) that we refer to as a pepper. The pepper is a defense in depth measure. The pepper value is stored separately in a manner that makes it difficult to discover by an attacker (i.e. not in a database table). As a result, if only the password storage is compromised, the password hashes are encrypted and of no use to an attacker.

And also:

> Why is the global pepper used for encryption instead of hashing?
>
> Recall that the global pepper is a defense in depth measure and we store it separately. But storing it separately also means that we have to include the possibility of the pepper (and not the password hashes) being compromised. If we use the global pepper for hashing, we can’t easily rotate it. Instead, using it for encryption gives us similar security but with the added ability to rotate.

In general a pepper that is stored separately from the actual password hash is also a valid mechanism encouraged in the [OWASP cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#peppering):

> A pepper can be used in addition to salting to provide an additional layer of protection. The purpose of the pepper is to prevent an attacker from being able to crack any of the hashes if they only have access to the database, for example, if they have exploited a SQL injection vulnerability or obtained a backup of the database.

So essentially Dropbox approach to this is using encryption because it enables you to rotate the pepper, but without having to compute up to number of peppers hashes per password. Remember that password hash algorithms are computationally hard by purpose. Instead, encryption is used to allow fast iteration over all peppers. This is also implemented and can be nicely seen in this PR:

https://github.com/Wikia/kratos/blob/e4cc6f66585497e008bc3b6332ec9ba2740267d7/hash/hash_comparator.go#L113-L118

### Wrapping Up

To me it looks like this PR is in fact a valuable addition to Kratos and should be merged. I propose to just completely remove the inner most SHA and apply the same concept to the length constraint of bcrypt as we already do in the plain bcrypt implementation. This will also completely avoid the password shucking and all the discussion around that.

WDYT?

aeneasr:
As said before, I feel bad for rejecting this because it is great engineering work by @harnash . Unfortunately we can not rationalize away the very real limitations and implications of the implementation and my duty as the maintainer of this repo is to manage downside risk when it comes to sensitive functionality in modules such as password hashing.

Without inner SHA, the PR could be refactored so that global peppering is supported for password hashing in general, not just BCrypt. After recalling the troubles i had with implementing AES GCM for Hydra‘s JWK store I strongly discourage using an AEAD algorithm for this purpose. AEAD algorithms are twofold. They encrypt and verify messages - but they do so with trade offs such as a limited amount of IVs. If IVs (96bit int for AES GCM) are reused you risk absolute and irrecoverable compromise of the encryption key. If the IV space is exhausted an attacker with access to the whole DB can recover the original encryption key - completely nullifying the defense mechanism. AEAD / AES GCM is an algorithm well suited for signal encoding (e.g. TLS) but not for storage of large amounts of messages. The pepper is a mechanism which is global for all credentials (passwords/hashes) and thus, in my understanding, explicitly a non-use case for AES GCM.

With that in mind - whole the idea of global peppering is good and has my and Ory‘s support - I do not believe that AES GCM is the appropriate algorithm to use in this situation though.

So the question is, what do cryptographers recommend to use for global peppering? I believe I saw mentions of HMAC or DES but I definitively have to do some research here.

Sorry again @harnash, I hope you understand! Unfortunately this was one of the scenarios I feared when I first saw the PR which is why:

- I took a lot of time with the review as it requires focus and lots of research to make informed decisions
- We ask in our contribution guidelines to discuss larger changes as github issues with core maintainers prior to writing code changes

Again, I hope you understand!

We can still salvage ideas from this PR though with the global peppering IMO - but we need some sources of truth and validation for the approach to be taken by either popular OSS services (Kubernetes, Linux, …) or well known researchers.

aeneasr:
Limitations of AES GCM and nonces/IVs are documented for Ory Hydra here: https://www.ory.sh/hydra/docs/security-architecture/#aes-gcm

Unfortunately Dropbox does not specify what concrete AES family they are using - they just say AES256…

zepatrik:
> Without inner SHA, the PR could be refactored so that global peppering is supported for password hashing in general, not just BCrypt. [...]
>
> With that in mind - whole the idea of global peppering is good and has my and Ory‘s support - I do not believe that AES GCM is the appropriate algorithm to use in this situation though.
>
> So the question is, what do cryptographers recommend to use for global peppering? I believe I saw mentions of HMAC or DES but I definitively have to do some research here.

I fully support and agree on these points. Let's move this towards globally enabling peppering, regardless of the underlying hash. I am pretty sure that we can use this PR as a basis of that, it just needs a bit of refactoring. Thanks @harnash for all the work and sorry to make it more work for you.

harnash:
@aeneasr + @zepatrik I understand and I feel the same as you. I think we have learned quite a lot during the last days and I also started to have doubts about this being merged into the upstream. It is sad we had lost a lot of time on this (me putting the implementation together and time for everyone to read it through and understand it's implications) but I hope it was not all for nothing. I've managed to understand Kartos' hashing better, we did found research evidences that helped us understand risks involved in such implementation and there are plans to improve password hashing based on the research that was done here.

Again sorry for taking your time (that probably should be spent somewhere else and be more productive) :-( This PR started as a simple change and it grew quite a bit over the last days.
I am going to create GH issue to discuss few ideas that were brought up here (mainly the hashing algorithm detection improvements which should be helpful for us).

aeneasr:
Thank you for your understanding and great work :)

PR #1333, opened 2021-05-12
docs: Fix incorrect command

nickdbush:
Fixes doc typo in which files are suggested to be copied to `/ory/home` instead of `/home/ory`.

### DIFF
--- a/docs/versioned_docs/version-v0.6/guides/docker.md
+++ b/docs/versioned_docs/version-v0.6/guides/docker.md
@@ -82,7 +82,7 @@ below:
 
 ```dockerfile
 FROM oryd/kratos:latest
-COPY contrib/quickstart/kratos/email-password/kratos.yml /ory/home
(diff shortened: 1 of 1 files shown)


zepatrik (approved):
Awesome, thanks a lot :tada: 
The directory should indeed be `/home/ory` https://github.com/ory/kratos/blob/18518e93bad531fe57f0b073029aa8edd5822f60/.docker/Dockerfile#L13

nickdbush:
Also, I'm not sure if it's a bug, but the Docker command specified in that section of the docs doesn't seem to work, because kratos looks for the file `.kratos.yml` according to other sections of the docs.

Using the `serve -c kratos.yml` command causes an error saying that the file cannot be found (exec-ing into the container shows the files are present and have correct permissions. However, changing the entrypoing to `/bin/sh -c` _does_ work.

aeneasr:
I believe the correc tcommand should be:

```
CMD ["serve", "-c", "/home/ory/kratos.yml", "--watch-courier"]
```

PR #1352, opened 2021-05-17
feat: OIDC provider for Auth0.com Tenants.

glawler:
## Proposed changes

This PR adds a simple OIDC provider for auth0.com tenants. The code is modeled on the existing GitLab provider as Atuh0 has the same issue that gitlab does - the response does not include the proper claims. So this provider hits the `/userinfo` endpoint at Auth0 to fill out the claims. 

Note also that there is a known bug in the Auth0 OIDC implementation. This  bug causes the returned `imported_at` JSON field to be a string (in RFC 3339 format instead of `int64`. You can see the workaround for this in `selfservice/strategy/oidc/provider_auth0.go`. 

## Checklist

      the changes.
      works.


aeneasr:

```
$ make format
$ git push
```

Thank you!

glawler:
> $ make format
> $ git commit -a -m "styles: format code"
> $ git push

Done. I'm building on a golang.1.16 debian container. This may have caused more changes than you'd like.

glawler:
I did have to install `prettier` by hand and this is the error I see now.

selfservice/strategy/oidc/provider_auth0.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return nil, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("%s", err))
+	}
+	u.Path = path.Join(u.Path, "/userinfo")
+	req, err := http.NewRequest("GET", u.String(), nil)
+	if err != nil {
+		return nil, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("%s", err))
+	}
+
+	req.Header.Add("Authorization", "Bearer: "+exchange.AccessToken)
```
aeneasr, line 83: I believe this isn't necessary as o.Client should attach the bearer token automatically :)

selfservice/strategy/oidc/provider_auth0.go
```diff
@@ -0,0 +1,194 @@
...
+		if _, ok := v.(string); ok {
+			// This is the bug. We need to convert.
+			c, err := auth0Claims(b)
+			if err != nil {
+				return nil, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("%s", err))
+			}
+
+			return c, nil
+		}
+	}
```
aeneasr, line 120: I believe this could be simplified with tidwall/gjson along the lines of:


```suggestion
	updatedAt := 0
	// There is an updated_at field. Look at the type. If float64 (default for json numbers, then
	// we can just use the Claims. Otherwise we convert the value.
	if updatedAtField := gjson.GetBytes(b, "updated_at"); updatedAtField.Exists() && updatedAtField.IsString() {
			t, err := time.Parse(time.RFC3339, updatedAtField.String())
			updatedAt = t.Unix()
	}
	
	b, err := sjons.Delete(gjson.GetBytes(b, "updated_at")
	if err != nil {
		// ...
	}
```

This would remove the need for the `auth0Claims` function completely!

aeneasr (commented):
Awesome, thank you! Sorry for the late review. I have an idea how to reduce the amount of code :)

selfservice/strategy/oidc/provider_auth0.go
glawler: Yes, this is good. I figured there was a better way to do this and I'm glad you found it. Those changes look great.

selfservice/strategy/oidc/provider_auth0.go
glawler: In my testing, I found that I needed this. I'll take your word for it though.

aeneasr: Sorry for the long reply, currently on vacation! If you testing says it does not work without this then you're right, of course :) So, does this only work if it was added?

glawler:
Sorry for the delay in merging your suggestions. I was away. I've pulled ory:master updates and merged those into my fork. Then I merged in your suggested updates. I think everything is ready for another review. Thanks,

aeneasr:
Thank you! Unfortunately it looks like the build is failing due to static code errors, see: https://app.circleci.com/pipelines/github/ory/kratos/3833/workflows/b85db7aa-e8a8-4376-92b7-0e7ed1809de7/jobs/22179

Also formatting is needed :)

```
$ make format
$ git push
```

Thank you!

glawler:
I took a closer look at the changes. I don't think it is correct. It will check the type of `updatedAt` and convert it,  but then it just removes it as a field thus throwing it away. 

I'll update the formatting, fix this, and push.

glawler:
I've updated the code use generic json parse into a `map[string]interface{}` in the case of an incorrect type in the `updated_at` field.

I still hit a build/format error in CI though. `Prettier` is not installed. Not sure how to fix this CI to have this...

glawler:
Looks like CI passed but for that missing `Prettier`. I'm not sure I can do anything about that.

aeneasr:

```
$ make format
$ git push
```

Thank you!

glawler:
Merged with master again.

glawler:
The still existing CI problem is that `Prettier` is not getting installed for some reason.  I do not think this has anything t do with myu code. 

```Checking formatting...
[warn] test/e2e/cypress/helpers/index.js
[warn] Code style issues found in the above file(s). Forgot to run Prettier?

```

zepatrik (commented):
The way we install prettier in the CI is a bit hacky, locally you should use `npm i` instead to get the correct version.

glawler:
Sorry, I don't understand. My local builds do work. Why is `prettier` missing in your CI builds? Isn't that controlled by your CI scripts? I assume you do not want me changing those...

zepatrik:
It is working and running in the CI as well. You just have a different version locally than we have in the CI, that's why it errors in the CI and passes locally.
```
Checking formatting...
[warn] test/e2e/cypress/helpers/index.js
[warn] Code style issues found in the above file(s). Forgot to run Prettier?

```
This means there is something wrong with the format in the file `test/e2e/cypress/helpers/index.js`.

In the CI we don't want to run `npm i` because of time, so we instead use this hacky way of only installing prettier and the styles package:
```
mkdir -p node_modules/ory-prettier-styles
tar -xf "$(npm pack ory-prettier-styles)" -C node_modules/ory-prettier-styles --strip-components=1
sudo npm i -g "prettier@$(jq -r '.devDependencies.prettier' package.json)"
rm ory-prettier-styles-*.tgz
```
We don't install the latest prettier version, but instead the one fixed in the `package.json`. You probably just use the latest version which is not producing the same output as our version requires.

But as I said, locally you just want to run `npm i` instead and then `npm format` to auto format everything.

glawler:
I have run

```
npm i
make format
```

The only file that is changed in my repo is `go.sum`. 

I'm not sure what else I can do at this point.

glawler:
If you can just paste in that file in a format that makes your CI happy, I can push that.

package.json
```diff
@@ -17,7 +17,7 @@
     "cypress": "7.0.0",
     "dayjs": "^1.10.4",
     "ory-prettier-styles": "1.1.1",
-    "prettier": "2.2.1",
+    "prettier": "^2.2.1",
```
zepatrik, line 20: The problem is actually this change, in the CI it actually installed `2.3.1` instead of `2.2.1`. That's why the styles did not match up.

zepatrik:
Now we passed the format step but have an error from the linter :wink: 
```
selfservice/strategy/oidc/provider_auth0.go:110:4: SA4006: this value of `err` is never used (staticcheck)
			t, err := time.Parse(time.RFC3339, updatedAtField.String())
			^

```

aeneasr (approved):

aeneasr:
Wohooo! Thank you so much for your work! :)

glawler:
Very nice! Thanks for helping me with the submit process.

aeneasr:
👍

Issue #1328, opened 2021-05-11
Login only with verified address


dadrus:
**Describe the bug**

With kratos v0.5.x it was possible to disable login until the account is activated (until the email is verified). This does not work with the current version (v0.6.0-alpha.2) anymore.

**Reproducing the bug**

1. Register the user (since the registration flow does not have the password session hook configured, the user is not logged in). Kratos will send an email with an email verification link.
2. Don't open the link, instead try to login. Kratos will create a session (so the user is logged in)

Here's a stripped down kratos yaml with the most relevant settings, which worked with v0.5.x versions and also with v0.6.0-alpha.2 (but results in a different behavior)

```.yaml
....

selfservice:
  default_browser_return_url: http://127.0.0.1:9090/
  whitelisted_return_urls:
    - http://127.0.0.1:9090

  methods:
    link:
      enabled: true 
    password:
      enabled: true
    profile:
      enabled: true

  flows:
    login:
      ui_url: http://127.0.0.1:9090/identity/login
      lifespan: 10m

    registration:
      lifespan: 10m
      ui_url: http://127.0.0.1:9090/identity/register
      after:
        default_browser_return_url: http://127.0.0.1:9090/identity/activation_pending

    verification:
      enabled: true
      ui_url: http://127.0.0.1:9090/identity/activate_account
      after:
        default_browser_return_url: http://127.0.0.1:9090/identity/account_activated

....

```

**Expected behavior**

The v0.5.x functionality is preserved, so instead of logging in the user, an error is returned.

aeneasr:
This did not work in Ory Kratos 0.5 either - it's not a regression. Currently, it is not possible to require verification before login in Ory Kratos.

dadrus:
I would not file this issue, if it had not worked previously. Even the [documentation](https://www.ory.sh/kratos/docs/next/guides/account-activation-email-verification/) is talking about account activation. But maybe something else is meant by "account activation" and the behavior, I know from the past was an "unofficial" feature.

aeneasr:
I know that it did not work before - maybe it was something you implemented in your system - because we have tests that check verification. They explicitly check that you can sign in even if you do not have a verified email address. But it doesn't really matter - it's still a feature request :)

dadrus:
Ok. :) How relevant or important do you see this feature request for kratos?
I very appreciate if you could give me some hints where to look in the code as well to add the corresponding behavior. Thank you in advance.

aeneasr:
[...]

PR #1355, opened 2021-05-18
feat: require verified address

dadrus:
## Related issue

@aeneasr: This is a promised PR for #1328.

## Proposed changes

As discussed in #1328. The only difference is the configuration. Instead of making the hook available in `before` property, the `after` is used as the information about the identity is only available there.

```.yaml
flows:
    login:
      ui_url: http://127.0.0.1:9090/identity/login
      lifespan: 10m
      after:
        hooks:
          - hook: require_verified_address
```

## Checklist

      the changes.
      works.


dadrus:
@aeneasr: In addition to the documentation update (which will follow), I would like to clarify one thing: 

An identity can have multiple verifiable addresses. The current implementation of the PR expects all of these to be verified. This might be an issue. One could change the behavior to make it expect at least one verified address, but this might be problematic as well. In our case, we would like to have the eMail address used for the current login attempt verified. So maybe kind of a configuration with useful defaults is required. So something like this:
```.yaml
  hook: require_verified_address
  config:
    verify: all # or alternatively 'one', or 'current' 
```
If the config is omitted, `one` might be a useful default.

If we go this way, how can I get the information about the email address used for the login? But there are also other related topics. In our case, we also support logins with user names. In this case, the expectation is, that at least `one` email address is verified. This however would not be possible to configure given the above configuration approach.

Any ideas from your side?

Zageron:
What happens if someone's session expires immediately after changing their primary email, and they misspell it? Blocked from logging in because email is unverified and unable to verify because wrong email? Does the email change flow have a verification component in order to successfully apply?

radekg:
Interesting question the one above. I’d expect the app to not allow changing an email address if there is no another confirmed address on the account.

aeneasr:
Hello there, this is a great contribution! Sorry for the late review but heading off to holidays and had a ton of stuff on my plate :)

So generally I would suggest to avoid the one / all / current mechanism. If we add more authentication methods, this will get blurier. What for example about OpenID Connect logins that do not have an email address associated?

I think the "one" strategy is good enough. If the user has verified an address I think it is safe to assume that the user has passed that challenge. I'm not sure if having multiple addresses be verified would add an additional benefit - on the contrary! Users might be wondering "why do I need to verify several email addresses?" or "I verified my email address why can I not log in?".

From my side - the one strategy as the default would be absolutely good enough! :)

radekg:
On the other hand, if I signed up using an email and a password, I should be able to recover using one of the social logins. I actually did that once with meetup.com. I created an account using facebook sso but I deleted my facebook account and was unable to sign in. I was able to recover the account by signing up with google directly. This linked the account to the old one and all was good.

The second thought I have, isn't signing up with social login an implied verification? If I signed up with Google SSO, Google just told you that the identity is genuine. If someone else was able to sign up using my Google account, well, that was a game over because my Google account was already compromised. So a registration via SSO linking implies a verified address, IMO.

Enjoy your holidays.

dadrus:
@aeneasr: Yeah, simpler is better :) I'll update the code accordingly next week and will also add documentation.

Enjoy the long weekend!

dadrus:
@radekg: Maybe I miss something, but how do you see this PR is related to recovery?

> The second thought I have, isn't signing up with social login an implied verification?

radekg:
> @radekg: Maybe I miss something, but how do you see this PR is related to recovery?

Sorry, I think I’ve mixed stuff up with https://github.com/ory/kratos/pull/1323#issuecomment-846232227

dadrus:
@aeneasr: Please take a look. There are some questions (as comments), which I would like to clarify, before finalizing the implementation and based on that updating the documentation.

selfservice/flow/login/hook.go
```diff
@@ -63,6 +63,21 @@ func NewHookExecutor(d executorDependencies) *HookExecutor {
...
+	//
+	// f.Active = identity.CredentialsTypePassword // or identity.CredentialsTypeOidc
+	//
+	// if err = s.d.LoginFlowPersister().UpdateLoginFlow(r.Context(), f); err != nil {
+	// 	return nil, s.handleLoginError(w, r, f, &p, errors.WithStack(herodot.ErrInternalServerError.WithReason("Could not update flow").WithDebug(err.Error())))
+	// }
+	//
+	// in the right place. Otherwise the Flow.Active property is never set (only in tests).
+	// If done as described above, there wouldn't be a need to pass around the `ct` argument as well, like in this method.
+	// So, FMPOV, the following line is just a hack due to the missing implementation described above.
```
aeneasr, line 78: So `Active` refers to the method the user used to sign in / up / verify / ...

So once you e.g. click "sign in with username / password" this would be set to "password". If you use OIDC this would be "oidc". It's basically a hint for the UI implementor.

Therefore, I don't think we need to set it here!

aeneasr (commented):
This looks great! Only thing we need to add is some documentation to: https://www.ory.sh/kratos/docs/next/self-service/hooks/#after

It probably also makes sense to only allow this hook for password flows? For OIDC flows I don't think it makes sense?

selfservice/flow/login/hook.go
dadrus: This is exactly the point. `Active` is never set. If it would, I would not implemented the workaround after the comment

dadrus:
> It probably also makes sense to only allow this hook for password flows? For OIDC flows I don't think it makes sense?

Yes, I was thinking to adjust the schema for this purpose. Will come with the next commits

selfservice/flow/login/hook.go
dadrus: @aeneasr: Since `Active` is never set, I consider it as a bug and will fix it as part of the next commits.

dadrus:
@aeneasr: Please take a look. If I haven't forgotten anything, this PR should be complete.

aeneasr (approved):

dadrus:
@aeneasr: I need your help. This is first time I have to deal with JS and cypress. I tried to implement the test cases, but unfortunately there were issues running the registration scenario locally. Only after completely rewriting the `register` command in `commands.js` the registration was successful. This way I was able to finalize the negative test. But now I run into issues while trying to verify the email in a successful case. The mailslurper just doesn't receive any emails and I don't understand why. Could you please support here? Thank you in advance.

test/e2e/cypress/support/commands.js
```diff
@@ -92,6 +92,18 @@ Cypress.Commands.add('longRecoveryLifespan', ({} = {}) => {
...
+    delete config.selfservice.flows.registration.after['password']
+    return config
+  })
+})
+Cypress.Commands.add('enableLoginForVerifiedAddressOnly', ({} = {}) => {
+  updateConfigFile((config) => {
+    config.selfservice.flows.login['after'] = {password: {hooks: [{hook: 'require_verified_address'}]}}
+    return config
+  })
+})
```
aeneasr, line 106: This looks good! However, I think this might need the negative as well? So iterating over `login.['after'].password.hooks` and `filter` out any hooks that are `require_verified_address`

test/e2e/cypress/support/commands.js
```diff
@@ -92,6 +92,18 @@ Cypress.Commands.add('longRecoveryLifespan', ({} = {}) => {
...
 })
+Cypress.Commands.add('dontLoginUserAfterRegistration', ({} = {}) => {
+  updateConfigFile((config) => {
+    delete config.selfservice.flows.registration.after['password']
+    return config
+  })
+})
+Cypress.Commands.add('enableLoginForVerifiedAddressOnly', ({} = {}) => {
+  updateConfigFile((config) => {
+    config.selfservice.flows.login['after'] = {password: {hooks: [{hook: 'require_verified_address'}]}}
```
aeneasr, line 103: Instead of a replacement this should probably be adding the hook to the array, not replacing it :)

test/e2e/cypress/support/commands.js
```diff
@@ -92,6 +92,18 @@ Cypress.Commands.add('longRecoveryLifespan', ({} = {}) => {
     return config
   })
 })
+Cypress.Commands.add('dontLoginUserAfterRegistration', ({} = {}) => {
+  updateConfigFile((config) => {
+    delete config.selfservice.flows.registration.after['password']
```
aeneasr, line 97: Similar to my comment below this should probably have a filter. Also, we want the inverse (so add session back in)

test/e2e/cypress/integration/profiles/verification/login/success.spec.js
```diff
@@ -0,0 +1,37 @@
...
+    })
+
+    beforeEach(() => {
+      cy.dontLoginUserAfterRegistration()
+      cy.enableLoginForVerifiedAddressOnly()
+
+      identity = gen.identity()
+      cy.register(identity)
+    })
+
```
aeneasr, line 23: These tests need a `afterEach()` or `after()` which resets `cy.dontLoginUserAfterRegistration` and `cy.enableLoginForVerifiedAddressOnly` to the previous values or all subsequent tests will use this set up and thus fail.

test/e2e/cypress/integration/profiles/verification/login/errors.spec.js
```diff
@@ -0,0 +1,35 @@
...
+    })
+
+    beforeEach(() => {
+      cy.dontLoginUserAfterRegistration()
+      cy.enableLoginForVerifiedAddressOnly()
+
+      identity = gen.identity()
+      cy.register(identity)
+    })
+
```
aeneasr, line 23: See comment in `success.spec.js`

test/e2e/cypress/support/commands.js
```diff
@@ -108,48 +120,37 @@ Cypress.Commands.add(
...
-          url: form.action,
-          followRedirect: false
-        })
-      })
-      .then((res) => {
-        console.log('Registration sequence completed: ', { email, password })
-        expect(res.redirectedToUrl).to.not.contain(
-          APP_URL + '/auth/registration?flow='
-        )
+    cy.get('a[href*="/auth/registration"]').click()
```
aeneasr, line 135: Changing the logic here will break all other tests. Please revert any changes to this method! I don't believe you need any of those here.

dadrus:
The tests are not final, as there were two issues, I've written in https://github.com/ory/kratos/pull/1355#issuecomment-858553414. I appreciate, if you could help with those

Quasarman:
Hello,

First of all thanks for the efforts for implementing this behavior!
Are there any further updates when this will become available?

BR

aeneasr:

Thank you!

dadrus:
@Quasarman: I was on vacation until yesterday and before it, there were some urgent topics in a project I'm working for, which I had to address.
The actual implementation is actually complete. The things which are missing are the e2e tests, which I will address asap

Quasarman:
Hello @dadrus ,

no worries thanks a lot for your efforts!

aeneasr:
Nice, I hope you enjoyed your days off @dadrus :)

dadrus:
Definitely :) Recharging batteries feels always good ;)

dadrus:
@aeneasr: There are no much changes in the last commit - only some simplifications compared to what you've done during your call.

aeneasr:
@dadrus yes you are right, looks like we don't need the reset method at all :)

aeneasr:
Sorry for the breaking test - I had a fix locally but looks like you were faster 😅

dadrus:
@aeneasr: If you're not going to break any further tests in a review :), I would say I've done everything required for this feature. Last commits have a self explanatory commit comments :), so feel free to merge or ask for further updates ;)

aeneasr:
yeah sorry about that again 😓 also LGTM from my side

dadrus:
Don't worry ;)

aeneasr:
🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳

PR #1366, opened 2021-05-20
feat: Include Credentials Metadata in admin api Missing changes in ha…

martinei:
## Related issue
I am very sorry, but my recently merged PR was missing the essential changes in the handler.
https://github.com/ory/kratos/pull/1274

## Checklist

      the changes.
      works.

### DIFF
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -152,13 +152,13 @@ type getIdentityParameters struct {
 //       404: jsonError
 //       500: jsonError
 func (h *Handler) get(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
-	i, err := h.r.IdentityPool().GetIdentity(r.Context(), x.ParseUUID(ps.ByName("id")))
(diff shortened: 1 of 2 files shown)


aeneasr:
Could you add a test for this so that we catch any regressions here? Thank you! :)

martinei:
Hi @aeneasr 

 > Could you add a test for this so that we catch any regressions here? Thank you! :)

I already had a look in the past and would need some guidance here:
The existing test  (TestHandler in handler_test.go) is a kind of  integration test and creates identities using the API. For the new case I'd need an Identity with credentials to test it's retrieval. Creating an identity with credentials can not be done via the api of this handler.

The Option I see would be to create a new Test with a mocked PrivilegedIdentityPool to return deliver an Identity with Credentials. However this would introduce a different style of test.

What do you think?

aeneasr:
In [the test](https://github.com/ory/kratos/blob/master/identity/handler_test.go#L80) you could use `reg.PrivilegedIdentityPool()` directly in the store to create an identity which you later retrieve using HTTP :)

aeneasr (approved):
Thank you :)

PR #1376, opened 2021-05-24
fix: check filesystem before fallback to bundled templates

oliverpool:
## Proposed changes

I am interested in implementing #595 and I took a look at the email management. The way `filepath.Join` is used there does not seem cross-platform (`filepath.Join(..., "recovery/invalid/email.subject.gotmpl")` will not be converted to backward slashes).

This PR fixes that by refactoring CourierTemplatesRoot into CourierTemplatePath

## Checklist

      works.

### DIFF
--- a/courier/template/load_template.go
+++ b/courier/template/load_template.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"embed"
 	"io"
+	"io/fs"
 	"os"
+	"path"
 	"text/template"
 
 	_ "embed"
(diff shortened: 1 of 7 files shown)


aeneasr:
What problem are you facing with the previous code? Could you include an example and the error and the logs? I’m not sure if using windows slashes in go embed works, especially if the build is cross platform. Thank you! :)

zepatrik (changes requested):
The current code also does not work on windows, because
```go
filepath.Join(t.c.CourierTemplatesRoot(), "verification/invalid/email.body.gotmpl")
```
will result in `"courier/builtin/templates\verification/invalid/email.body.gotmpl"` on windows. IMO we should use `path.Join` instead, and in case it is not loaded from embed AND we are on windows, replace all `/` with `\`?

oliverpool:
@zepatrik you are totally right. I replaced the use of `filepath` with `path`.

I think that go is still able to open a file on windows when the path is split with forward slashed (but I don't have a windows computer to test this)

zepatrik:
I don't think that is the case, otherwise there would be no `filepath` package at all. You can also add a windows CI task instead, so we can check that.

driver/config/config.go
```diff
@@ -701,8 +702,16 @@ func (p *Config) CourierSMTPFromName() string {
...
+func (p *Config) CourierTemplatePath(subpath ...string) string {
+	var fullpath []string
+	if p.p.Koanf.Exists(ViperKeyCourierTemplatesPath) {
+		fullpath = []string{p.p.String(ViperKeyCourierTemplatesPath)}
+	} else {
+		fullpath = []string{"courier", "builtin", "templates"}
+	}
+	fullpath = append(fullpath, subpath...)
+
+	return path.Join(fullpath...)
```
aeneasr, line 714: I'm not sure if I understand what the previous error was - it looks like the code is doing the same thing (but is using `path` instead of `filepath`)?

Regarding `path` I think it would make sense to add a test for `CourierTemplatePath` which checks if the behaviour on windows is correct? So it is able to open files and also use the built in templates.

aeneasr (changes requested):
Sorry for the lengthy review time! I'm back from vacation :)

Then, in the go embed file I believe we should use the https://golang.org/pkg/path/filepath/#ToSlash method, and if it's not an embedded file use https://golang.org/pkg/path/filepath/#FromSlash to open it on the FS.

I'm actually not 100% sure if go embed respects filepath.PathSeparator when bundling...

oliverpool:
> I'm actually not 100% sure if go embed respects filepath.PathSeparator when bundling...

According to https://golang.org/pkg/io/fs/#ValidPath, it should always be a forward slash:
> Note that paths are slash-separated on all systems, even Windows. Paths containing other characters such as backslash and colon are accepted as valid, but those characters must never be interpreted by an FS implementation as path element separators. 

-----

Actually I am wondering if the following code is working as intended:
```go
	if file, err := templates.Open(path); err == nil {
		...
	} else {
		file, err := os.Open(path)
```
- since it first checks for the template in the embedded FS, there is no possibility for the user to override the template
- except if I specify a different `courier.template_override_path`, but then I have to duplicate all the templates (I can't override just one email)

I would suggest something like:
```go
func openTemplate(dir, name string) (fs.File, error) {
	if file, err := os.DirFS(dir).Open(name); err == nil {
		return file, nil
	}
	return templates.Open(path.Join("courier/builtin/templates", name))
}
```
This way:
- the filesystem is checked first (eventually at a custom location)
- if nothing was found, the embedded FS is checked (using the hard-coded location)

PR #1401, opened 2021-06-08
fix: check filesystem before fallback to bundled templates

oliverpool:
## Related issue

Follow up of #1376 (I made a new PR, this the proposed change is totally different)

## Proposed changes

I am wondering if the following code is working as intended:
```go
	if file, err := templates.Open(path); err == nil {
		...
	} else {
		file, err := os.Open(path)
```
- since it first checks for the template in the embedded FS, there is no possibility for the user to override the bundled template
- except if it specifies a different `courier.template_override_path`, but then I have to duplicate all the templates (I can't override just one email template)

This PR fixes this.

## Checklist


aeneasr:
Re-running CI...

aeneasr:
Looks like this needs to be fixed on master first (the CI)...will try to come up with something

aeneasr:
Once https://github.com/ory/kratos/pull/1411 is merged you can rebase on master and the CI should pass fine.

Issue #249, opened 2020-02-14
Add filters to admin api


gperdomor:
**Is your feature request related to a problem? Please describe.**

Hi, i want to build some kind of CRUD around kratos, a kind of kratos wrapper, but I don't see a way to apply filters, for example, i want to filter based on identity type (user/device/customer) or by the first name of a customer...

**Describe the solution you'd like**

Maybe we can use just query params for that, or some path param to handle the identity type and query params for the rest of fields, but the second approach we can't list for example users and devices, we are limited to one type 🤔

aeneasr:
Hey, this is definitely on the roadmap! We were planning to do something with GraphQL because building REST query languages sucks! :)

However, traits need to be indexed and stuff, so this is a bit further down the roadmap currently. We might even need something like Sphinx or Elastic.

UkonnRa:
1. If mentioned about queries on RESTful, you can check the discussion from JSON:API: https://github.com/json-api/json-api/issues/1445
2. Designing a query language IS a difficult problem, even using GraphQL

Right now I want to apply some complex queries when using Kratos (filter identities by some specific trait fields) and Cursor-based pagination. So I wonder:
1. **if I can do some queries on the SQL level?**
2. **are the schemas of SQL tables stable enough for customization?**

aeneasr:
> if I can do some queries on the SQL level?

You probably can, if your database supports JSON indexing!

> are the schemas of SQL tables stable enough for customization?

Could you explain what you mean with stable? We definitely will have migrations applied to the tables in the future!

g13013:
@aeneasr 

> However, traits need to be indexed and stuff, so this is a bit further down the roadmap currently. We might even need something like Sphinx or Elastic.

I suggest https://blevesearch.com/ for that to avoid 3rd party dependency, supporting basic searchability would be supported by kratos, and most advanced search capabilities could be the responsibility of the app provider

aeneasr:
Nice, that looks like a pretty cool tool and could work for small installations (similar to SQLite)! Not sure if we could use this in HA or larger envs?

g13013:
I think it can do more than that, for basic search capabilities I think that it's more than enough

here are some benchmarks https://blevesearch.com/bleve-bench/

fdelucchijr:
Probably related: #2448 

There’s any way we can do batching for now? N+1 Problem is a pretty serious thing with rest...
[...]

PR #1420, opened 2021-06-14
feat: Login method api

jess-sheneberger:
## Related issue
There is no related issue.

## Proposed changes

In my app I want to allow users to register via one of several methods: two OIDC providers or a password. When they return after sign-up, I want to provide a login experience that doesn't require them to remember how they registered. I want them to enter their username or email address and for the UI to start the correct OIDC provider flow or prompt for their password using the same method they registered with.

To make this possible, I've put together this login method API. It looks up the given email address and returns either the ID of the OIDC provider that should be used, or the "password" method. My app UI is able to interpret the result and either launch the correct OIDC flow or prompt for a password.

I also have a requirement that users who login via password need to have a username rather just than an email address. Login via email address is fine for users who login via OIDC.

I understand this is a potential account enumeration attack vector, so maybe it should be turned on with a configuration option and left off for most users.

I'm open to different/better approaches to accomplish this. I'm marking it WIP because I implemented what seemed to me to be the most straightforward way, but maybe there is a different approach that would work better.

## Checklist

      the changes.
      works.

## Further comments

I accidentally closed the old merge request here: https://github.com/ory/kratos/pull/1278

I've added tests to finish up my contribution.  I believe I've added an annotation to the documentation by mimicking the swagger auto-gen syntax but I don't know how to generate the documentation so I didn't test it. 

Please let me know what else I need to do to get this merged.


jess-sheneberger:
I've fixed the swagger auto-gen issue that occurred in the previous runs, but now tests appear to be failing for an unrelated reason in the "restoring cache" step.

jess-sheneberger:
@aeneasr I don't know how to mark this as ready, and I don't know how to tag the maintainers (other than you specifically)

It appears some of the CI jobs failed due to an issue not related to my changes.  Additionally I don't appear to have permission to re-start the jobs.  Can you help me figure out the right way to proceed?

aeneasr:
Thank you! Currently only I am a maintainer of the repo so it's ok to tag me :)

I've merged your branch with master so that should fix the CI issues. Let's see how it goes!

jess-sheneberger:
@aeneasr I merged latest again and now it appears tests are passing, can it be merged?

aeneasr (commented):
Thank you, this looks really good. I'll make some minor adjustments and push them directly to your branch. Basically just some naming conventions around the SDK methods but from my side this is good to be merged!

aeneasr:
I took a longer look at this PR and I think it goes in the right direction in terms of functionality :) One thing I had a hard time was to think about the API design - the proposal doesn't really follow how Ory's other APIs look like due to these reasons:

1. The search is a POST request - it is however considered good practice to have searches be a GET endpoint with query parameters
2. Depending on the credentials hint we search for the credentials identifier, OR the verified address (for oidc). But what if we want to find by recovery addresses? Or only by verified (with status verified) addresses? Or credential identifier, and verified address? I think this logic is not enough documented and flexible enough to be a generic endpoint included in Ory kRatos.

My idea for this PR is to implement the search / filter functionality in the `GET /identities` endpoint by using query parameters `credentials_identifier`, `verifiable_address`, `recovery_address`. This would allow to list identities like so:

```
GET /identities?verifiable_address=...

// Filters are an AND
GET /identities?verifiable_address=...&recovery_address=...
```

and would return a list of identities that match that. The list can include credentials so that we can build an endpoint which tells us what credential types are matching.

This would allow a much more flexible search and also the use case targeted in this PR.

Sorry if it took so long to look into this, I was really busy and reviewing conceptual PRs like this one (that introduce new APIs) takes a lot of focus and though.

jess-sheneberger:
This sounds fine to me, but it may be some time until I can implement it. For now, our needs are met by running the current version of this PR.

g13013:
@aeneasr it would be even better if we could add annotations in the identity schema to support other search fields, I know it's too much for this PR, but, every authentication service needs administration, and limiting the searchability to some keys do not give sense to use a third-party solution and then get obliged to search directly in the database to implement some solutions

here and example:

in our trait we have a field named `legacyId`, from an admin perspective I want to search by this field, a solution would be to add:

```jsonc
"legacyId": {
  "title": "Legacy ID",
  "type": "string",
  "ory.sh/kratos": {
    "search": {
       "enabled": true,
       "field": "leg-id" // name in the url instead of the field name ?leg-id="id"
    }
  }
}
```

This would allow to declare searchable fields beforehand and allow Kratos to do some database optimizations like `generated_columns` for mysql or jsonb indexes in posgresssql

zepatrik:
@g13013 general purpose filtering is tracked as #249, this is only about credentials identifiers as far as I can tell.

aeneasr:
Closing this but please do reopen it if you have some time to work on it! :)

PR #1385, opened 2021-05-31
feat: add instana as possible tracing provider

dadrus:
## Related PR

https://github.com/ory/x/pull/349

As soon as the referenced PR is merged, this PR can be merged as well.

## Proposed changes

Add "instana" as possible value for a tracing provider

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [x] I have added or changed [the documentation](docs/docs).

PR #1429, opened 2021-06-16
feat: add instana as possible tracing provider

aeneasr:
Supersedes #1385
Closes #1385


aeneasr:
thank you @dadrus ! :)

PR #1444, opened 2021-06-17
chore: remove CII best practices

vinckr:
The CII best practices actually only concern ory/hydra, so we should remove it from the Kratos README?
https://bestpractices.coreinfrastructure.org/projects/364


aeneasr:
Let's rather go through CII for kratos and link that. Could you do that?

vinckr:
Makes sense yes! I'll close this then.

PR #1445, opened 2021-06-18
feat: Disable HaveIBeenPwned validation when HaveIBeenPwnedEnabled is set to false

jertel:
## Related issue

Implements remaining request from #316 to support disabling the check for pwned passwords.

## Proposed changes

Environments disconnected from the Internet, and environments that enforce strict outbound network access restrictions should not attempt to access a remote URL to determine if a password has appeared in a prior breech. While the current implementation allows for custom host definition, environments that do not have a custom pwned password host currently have no method to skip this check. This PR provides that feature.

## Checklist

      the changes.
      works.

## Further comments

None


aeneasr:
You can already ignore network errors using `passwordPolicyConfig.IgnoreNetworkErrors`. Is there a particular reason for adding another control mechanism?

jertel:
Yes, with network errors ignored, the client will attempt several retries using an exponential backoff algorithm (see https://github.com/hashicorp/go-retryablehttp). This removes 15 seconds of unnecessary waiting from the password process, when it's known in advance that the remote host is unavailable.

aeneasr:
Right, that makes sense then! Would it be possible to introduce a new variable for this though? Something that makes it explicit that this is disabled.

jertel:
Yes, take a look a the new changes and let me know if it needs adjustments.

driver/config/config_test.go
```diff
@@ -144,7 +144,7 @@ func TestViperProvider(t *testing.T) {
 				config  string
 				enabled bool
 			}{
-				{id: "password", enabled: true, config: `{"haveibeenpwned_host":"api.pwnedpasswords.com","ignore_network_errors":true,"max_breaches":0}`},
+				{id: "password", enabled: true, config: `{"haveibeenpwned_host":"api.pwnedpasswords.com","haveibeenpwned_enabled":true,"ignore_network_errors":true,"max_breaches":0}`},
```
aeneasr, line 147: Add a test case with false here :)

aeneasr (commented):
Just one minor thing, then this is good to merge IMO!

driver/config/config_test.go
jertel: My understanding is that "group=methods" test case only validates config defaults; it does not allow for specifying inputs to new configs. I've added new unit test coverage for all of the haveibeenpwned settings.

docs/docs/guides/hosting-own-have-i-been-pwned-api.mdx
```diff
@@ -27,6 +27,17 @@ selfservice:
 An example of a minimal self-hosted, k-anonymity enabled HiBP API
 [can be found in this repository](https://github.com/radekg/hibp).
 
+Alternatively, this feature can be disabled altogether, when configured as 
+follows:
```
zepatrik, line 31: I would like to have the implications of this here as well. Something along the lines of:
```suggestion
Alternatively, this feature can be disabled altogether. Note however that by doing so
you will significantly decrease the security of the service and not follow [NIST guidelines](https://pages.nist.gov/800-63-3/sp800-63b.html):

> When processing requests to establish and change memorized secrets, verifiers SHALL compare the prospective secrets against a list that contains values known to be commonly-used, expected, or compromised. For example, the list MAY include, but is not limited to:
> 
> 1. Passwords obtained from previous breach corpuses.
> 2. Dictionary words.
> 3. Repetitive or sequential characters (e.g. ‘aaaaaa’, ‘1234abcd’).
> 4. Context-specific words, such as the name of the service, the username, and derivatives thereof.
> 
> If the chosen secret is found in the list, the CSP or verifier SHALL advise the subscriber that they need to select a different secret, SHALL provide the reason for rejection, and SHALL require the subscriber to choose a different value.

Disabling the feature will effectively remove checks for the points 1.-3.
The feature can be disabled by setting the following configuration values: 
```

aeneasr:

```
$ make format
$ git push
```

Thank you!

aeneasr (approved):

PR #1447, opened 2021-06-18
docs: correct CII badge

vinckr:
As discussed in https://github.com/ory/kratos/pull/1444, 
here is the completed CII badge for Kratos.


aeneasr (approved):

aeneasr:
Great, thank you! Can you do a follow up for keto and oathkeeper as well?

PR #1449, opened 2021-06-20
fix: use local schema URL for sorting UI nodes

### LINKED ISSUE #1448: fix: settings flow error handle should access schemas using configured "local" URLs instead of public URLs
### Discussed in https://github.com/ory/kratos/discussions/1373

<div type='discussions-op-text'>

<sup>Originally posted by **jertel** May 23, 2021</sup>
The issue I'm seeing involves Kratos, running in a container, attempting to access schemas via the public URL. `Persister.injectTraitsSchemaURL()` concatenates the schema ID (Ex: `default`) to the configured public self-service URL (specified in the configuration file). Then in error scenarios, such as attempting to change your password to an invalid value, the settings flow's `error.go` handler attempts to sort the traits for an improved UI experience. But that sorting also requires loading the schema, and it does so via `LoadURL()`, from within `GetKeysInOrder()` (in `schema.go`). Since the Kratos process now attempts to access this public self service URL with the concatenated schema ID, it can run into problems if the container does not have the right CAs or DNS or network routes defined to access the public URL (for example when it's running behind a reverse proxy).

Instead of the Kratos process itself trying to lookup schemas via the concatenated public URL, would it be better to have it load the URLs listed in the configuration, under the `identity` block?

```
identity:
  default_schema_url: file:///kratos-conf/schema.json
```
</div>

jertel:
## Related issue

Resolves #1448 

## Proposed changes

Use the schema URLs defined in the Kratos configuration yaml instead of using the public URLs. This ensures that containerized Kratos processes can successfully load the schema during the settings error flow handler, which is needed in environments that have network configurations that prohibit outward facing network traffic.

## Checklist

      the changes.
      works.

## Further comments

There does not appear to be test coverage for the `sortNodes()` method, so I'm open to suggestions on how to validate this via a unit test. Ideally I'd like to have the unit test check to see which URLs were attempted to be accessed, or something to that effect.

No doc changes are necessary for this fix.


aeneasr:
Unfortunately, for some reason, the CircleCI tests are not running. Do you maybe follow your/a fork of this repository on CircleCI? If so, you need to unsubscribe / unwatch from that CircleCI project. Then, make another push to your branch using:

```
$ git commit --amend --allow-empty
$ git push --force
```

That should get the CI running! Thank you :)

aeneasr (changes requested):
Thank you for the contribution! I think this is going into the right direction.

It would be awesome if you could add tests! :relieved:

Tests are important to open source projects because the allow reviewers to better judge if the changes are doing what was intended. They also ensure that your changes never regress / break when someone else works on code that is related to this.

If you need help with the tests feel free to push your work in progress and ping myself or another maintainer :raising_hand:

jertel:
Please see the comment about tests in the original PR description. It is difficult to add test coverage for PRs like this when the original code is missing test coverage. So yes, i am asking for advice in how you want to handle this particular test. 

Regarding CircleCI, i personally do not use it and do not have it linked to my GH account. Are you able to see specific reasons why CI had not launched?

aeneasr:
> There does not appear to be test coverage for the sortNodes() method, so I'm open to suggestions on how to validate this via a unit test. Ideally I'd like to have the unit test check to see which URLs were attempted to be accessed, or something to that effect.

> Please see the comment about tests in the original PR description. It is difficult to add test coverage for PRs like this when the original code is missing test coverage. So yes, i am asking for advice in how you want to handle this particular test.

RTFM @aeneasr 🤣 Sorry, I missed that...

So this is currently being tested here:

https://github.com/ory/kratos/blob/e74a9ce69f12b198d2509d19f1d2a7060adbe4f9/selfservice/flow/settings/error_test.go#L234-L245

The schema is coming from here:

https://github.com/ory/kratos/blob/e74a9ce69f12b198d2509d19f1d2a7060adbe4f9/selfservice/flow/settings/error_test.go#L60

and here:

https://github.com/ory/kratos/blob/e74a9ce69f12b198d2509d19f1d2a7060adbe4f9/selfservice/flow/settings/error_test.go#L40

I think that, if you change the `PublicURL` using `conf.Set(config.SomethingSomethingPublicURL)` to something random, the linked test above would fail without your changes.

So writing a sub-test which sets the public url to something random (e.g. google or ory.sh) should replicate the issue. You can use `t.Cleanup()` to reset the config value after the test finished so that other tests don't break :)

CONTRIBUTING.md
```diff
@@ -199,7 +199,10 @@ request, go through this checklist:
    commit history and helps structure the changelog.  
    Please refer to this
    [list of prefixes for Kratos](https://github.com/ory/kratos/blob/master/.github/semantic.yml).
-
+1. Sign-up with CircleCI so that it has access to your repository with the
```
aeneasr, line 202: Ahhh, that is the missing piece! @vinckr can you please transfer that into ory/meta?

aeneasr:
Looks like E2E failed with a flake, rerunning

aeneasr (approved):
Great job, thank you!

PR #1469, opened 2021-06-29
feat: add helper for starting kratos e2e

aeneasr:
## Checklist

- [ ] I am following the
      the changes.
      works.


cmd/identities/helpers_test.go
```diff
@@ -0,0 +1,73 @@
...
+	require.Len(t, stdErr, 0, stdOut)
+	return stdOut
+}
+
+func execErr(t *testing.T, cmd *cobra.Command, args ...string) string {
+	stdOut, stdErr, err := exec(cmd, nil, args...)
+	require.True(t, errors.Is(err, cmdx.ErrNoPrintButFail))
+	require.Len(t, stdOut, 0, stdErr)
+	return stdErr
+}
```
zepatrik, line 63: This part is now part of `cmdx`, maybe remove it in this go? I can also do that btw

aeneasr: Ah yeah makes sense, I first have to wrap everything in new* funcs to make it pass. We can then do another pass and remove these helpers and move to ory/x. Would be great if you could do that. I think it makes sense to do it once this PR is merged

zepatrik: They are already in `ory/x` 1:1, you can just remove them here and import instead. But we can also separate that.

aeneasr: The method signatures are a bit different so I'd separate it

Issue #979, opened 2021-01-12
Go get fails when building from source


zkrzyzanowski:
**Describe the bug**

Build from source based on doc instructions https://www.ory.sh/kratos/docs/install/#building-from-source fails with error

```
go get: github.com/ory/cli@v0.0.28-0.20200903094455-e203529bf12b updating to
	github.com/ory/cli@v0.0.37 requires
	github.com/ory/kratos@v0.5.5-alpha.1.0.20201219142622-e589c07bb081 requires
	github.com/ory/kratos-client-go@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

**Reproducing the bug**

1. `go get -d -u github.com/ory/kratos`

**Expected behavior**
Package installed to `$GOPATH/src/github.com/ory/kratos`

**Environment**

- MacOs Sierra 10.15.6
- Go 1.15

aeneasr:
Do not use "-u" with go modules.

PR #1473, opened 2021-06-29
docs: [install] fix building from source

spg:
The old instructions did not work (same result as described here: https://github.com/ory/kratos/issues/979)


aeneasr:
Thank you, this looks great! Could you please sign the CLA (just two clicks), then we can merge it right away!

aeneasr:
Awesome, thank you!

PR #1528, opened 2021-07-12
feat: Allow wildcard domains for redirect_to checks

### LINKED ISSUE #943: Support wildcard in whitelisted_return_urls

We are using multiple subdomains in our system (user1.mydomain.tld, user2.mydomain.tld, ...) with wildcard https certificate. When a user is not logged in, he is redirected to auth.mydomain.tld. I want to return him to the original url after login (for example user1.mydomain.tld).

Two solutions would fix our issue:
1. We could use glob or regex match here: 
https://github.com/ory/kratos/blob/ef1e3f77238d090112808e988b890ab7e8b58074/x/http_secure_redirect.go#L90

2. or specify only domain in `whitelisted_return_urls` and allow all its subdomain by default.

abador:
## Related issue
resolves #943
@aeneasr

## Proposed changes
Allow to whitelist a wildcard domain. Domains are filtered out using golang.org/x/net/publicsuffix

## Checklist

- [x] I am following the
      the changes.
      works.


driver/config/config.go
```diff
@@ -672,6 +674,17 @@ func (p *Config) SelfServiceBrowserWhitelistedReturnToDomains() (us []url.URL) {
 			p.l.WithError(err).Warnf("Ignoring URL \"%s\" from configuration key \"%s.%d\".", u, ViperKeyURLsWhitelistedReturnToDomains, k)
 			continue
 		}
+		if parsed.Host == "*" {
+			p.l.Warnf("Ignoring wildcard \"%s\" from configuration key \"%s.%d\".", u, ViperKeyURLsWhitelistedReturnToDomains, k)
+			continue
+		}
+		eTLD, icann := publicsuffix.PublicSuffix(parsed.Host)
+		if parsed.Host[:1] == "*" &&
```
aeneasr, line 682: This panics if `host` is an empty string. Also, what about `foo.*.bar`?

x/http_secure_redirect.go
```diff
@@ -55,6 +55,14 @@ func SecureRedirectOverrideDefaultReturnTo(defaultReturnTo *url.URL) SecureRedir
 	}
 }
 
+// SecureRedirectToIsWhitelisted validates if the redirect_to param is allowed for a given wildcard
+func SecureRedirectToIsWhiteListedHost(returnTo *url.URL, allowed url.URL) bool {
```
aeneasr, line 59: Could you please add tests for this?

aeneasr (changes requested):

x/http_secure_redirect.go
```diff
@@ -55,6 +55,14 @@ func SecureRedirectOverrideDefaultReturnTo(defaultReturnTo *url.URL) SecureRedir
...
+	if allowed.Host != "" && allowed.Host[:1] == "*" {
+		return strings.HasSuffix(returnTo.Host, allowed.Host[1:])
```
ttimonen, line 61: What's the plan for case sensitivity here?
- On documentation, the function docstring is ambiguous,
- when wildcards are used, the match is case-sensitive,
- when wildcards are not used, the match is case-insensitive.

To follow the principle of least surprise, you might want to do one of
- Do strings.HasSuffix(strings.ToLower(returnTo.Host), strings.ToLower(allowed.Host[1:]) here to be case-insensitive.
- .. or do return allowed.Host == returnTo.Host few lines later to be case-sensitive.
- .. or mention this semi-case-insensitivity behavior in the function comment if it is intentional and preferred.

aeneasr: Hostnames are case insensitive, so the correct approach would be to have the wildcard matches also case insensitive :)

aeneasr:

Thank you!

aeneasr (changes requested):
Great! Now, one last thing :) We need to document it! Could you please add some docs here: https://www.ory.sh/kratos/docs/concepts/browser-redirect-flow-completion

Thank you! 🙏

abador:
> Great! Now, one last thing :) We need to document it! Could you please add some docs here: https://www.ory.sh/kratos/docs/concepts/browser-redirect-flow-completion
> 
> Thank you! 🙏

I've added some more docs in my last commit.

aeneasr (approved):

aeneasr:
Great stuff!

PR #1545, opened 2021-07-14
refactor: Cleanup lint errors and add docstring to x

ttimonen:
The x package name is neither very descriptive nor has comment
describing what the package is. This addresses the latter.
Also fix two staticcheck warnings by doing minor restructuring of
two tests.


aeneasr (approved):

Issue #1563, opened 2021-07-17
/self-service/settings/api returning stacktrace


nanikjava:
**Describe the bug**

Accessing the `/self-service/settings/api` with an invalid `flow` parameter gets the following error

```
{
  "error": {
    "code": 403,
    "status": "Forbidden",
    "reason": "This endpoint can only be accessed with a valid session. Please log in and try again.",
    "debug": "rid=\nerror=request does not have a valid authentication session\nreason=No active session was found in this request.\ndetails=map[]\ndebug=\n\ngithub.com/ory/kratos/session.(*ManagerHTTP).FetchFromRequest\n\t/home/.....
....
....
runtime.goexit\n\t/home/Golang/go1.16/src/runtime/asm_amd64.s:1371",
    "message": "The requested action was forbidden"
  }
}
```

**Reproducing the bug**

Execute the following command:

`curl -s -X GET -H "Accept: application/json"   "http://127.0.0.1:4433/self-service/settings/api?flow=XXXX"`

**Expected behavior**

Not sure if this is the expected behaviour but surely sending stack trace to the API caller is not the right approach.

nanikjava:
The `FetchFromRequest` function inside `manager_http.go` is returning 

`return nil, errors.WithStack(ErrNoActiveSessionFound)`

changing to

`return nil, errors.Cause(ErrNoActiveSessionFound)`

got a more reasonable error

```

  "error": {
    "code": 403,
    "status": "Forbidden",
    "reason": "This endpoint can only be accessed with a valid session. Please log in and try again.",
    "debug": "rid=\nerror=request does not have a valid authentication session\nreason=No active session was found in this request.\ndetails=map[]\ndebug=\n",
    "message": "The requested action was forbidden"
  }
}
```

Thoughts ?

aeneasr:
Yeah, we should probably have a config switch to dis/enable traces ending up at the http response (I believer ory/herodot is already capable of it)

nanikjava:
> Yeah, we should probably have a config switch to dis/enable traces ending up at the http response (I believer ory/herodot is already capable of it)

@aeneasr can you kindly take a look a the WIP PR to check whether this is the `herodot` config that you are referring to ?

PR #1574, opened 2021-07-21
[WIP] fix: remove stacktrace from error

nanikjava:
Per the issue #1563 when a user submit to the `/api` endpoint with not a valid flow id it reports with stack trace.

This PR is to modify to ensure only relevant error message is reported (not including stacktrace)

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/session/manager_http.go
+++ b/session/manager_http.go
@@ -122,7 +122,7 @@ func (s *ManagerHTTP) extractToken(r *http.Request) string {
 func (s *ManagerHTTP) FetchFromRequest(ctx context.Context, r *http.Request) (*Session, error) {
 	token := s.extractToken(r)
 	if token == "" {
-		return nil, errors.WithStack(ErrNoActiveSessionFound)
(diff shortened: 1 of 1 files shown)


aeneasr (changes requested):
Thank you for the PR. Unfortunately it's not really what I imagined we need. So currently, stack traces are returned based on the `LOG_LEVEL` that is set. if the log level is e.g. `trace` or `debug`, these traces will be returned in the API. It is possible that this doesn't work here, though (I would have to check).

What we should do, is to have another setting which allows controlling wether or not we want stack traces to returned in errors (globally, not just for this one error).

These changes would need to be done here: https://github.com/ory/herodot

nanikjava:
From your explanation seems like this is a more broader logging configuration issue and not specifically for the `FetchFromRequest`

Will open a separate ticket specifically for the logging configuration issue and link the #1563 issue.

nanikjava:
I've created issue for the logging changes https://github.com/ory/kratos/issues/1589 

Closing this issue.

PR #1585, opened 2021-07-23
feat: implement blocking web hooks

### LINKED ISSUE #1724: Setting up a Criteria for self-service registration
The feature is meant to set custom criteria for the registration form, for example setting up a dob (date of birth) field, to prevent underage users to register to the product, because of the "COPPA".

I have considered just putting in the schema field type "number" with the name age and set the minimum, but it would also be nice to be able to apply cookie, but also instead of using the number filed to use the date

If someone has a good suggestion or solution, I would be happy to hear it out. Because the easiest thing now I am thinking about is to do the limitation in the front end, but I don't think that is enough

harnash:
This change allows webhooks to control the registration flow by either allowing process to continue or fail with an errors propagated to the UI (mostly validation errors at this point). This is not yet complete but the general idea is here to start any discussions and resolve concerns early. 

TODO:
* tests
* documentation update

Closes #1724

## Related issue(s)
https://github.com/ory/kratos/discussions/1483

## Checklist
- [x] I am following the
      the changes.
      works.


aeneasr:
Thank you for the PR! I will review it next week!

harnash:
> Thank you for the PR! I will review it next week!

No hurry there. I'm off next week so effectively I can resume work on this in two weeks more or less.

aeneasr:
Could you please enable [write access for maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)? 🤔 

Thank you! 🙏

aeneasr:
Alternatively, given that this is an org repo, you can also invite me as a collaborator to the fork

driver/config/.schema/config.schema.json
```diff
@@ -149,6 +149,11 @@
               "type": "string",
               "description": "Path to the jsonnet template used for payload generation. Only used for those HTTP methods, which support HTTP body payloads"
             },
+            "interrupt": {
```
aeneasr, line 152: Let's make this explicit:

```suggestion
            "can_interrupt": {
```

driver/config/.schema/config.schema.json
```diff
@@ -149,6 +149,11 @@
...
+              "type": "boolean",
+              "default": false,
+              "description": "Should hook be able to interrupt the flow. It only applies to a certain flows and require concrete response format"
```
aeneasr, line 155: Please also explain to which flows this applies:

```suggestion
              "description": "If enabled allows the web hook to interrupt / abort the self-service flow. It only applies to a certain flows (EXPLAIN WHICH ONES) and requires a valid response format."
```

schema/errors.go
```diff
@@ -173,3 +173,43 @@ func NewNoVerificationStrategyResponsible() error {
...
+		Messages: messages,
+	}
+}
+
+type ValidationListError struct {
+	Validations []*ValidationError
+}
+
+func (e ValidationListError) Error() string {
+	return fmt.Sprintf("%d validation errors occurred", len(e.Validations))
```
aeneasr, line 192: Shouldn't we also be able to see the concrete errors?

schema/errors.go
```diff
@@ -173,3 +173,43 @@ func NewNoVerificationStrategyResponsible() error {
...
+}
+
+func (e *ValidationListError) Add(v *ValidationError) {
+	e.Validations = append(e.Validations, v)
+}
+
+func (e ValidationListError) Empty() bool {
```
aeneasr, line 199: How about:

```suggestion
func (e ValidationListError) HasErrors() bool {
```

ui/container/container.go
```diff
@@ -189,6 +189,14 @@ func (c *Container) ParseError(group node.Group, err error) error {
 			}
 		}
 		return nil
+	} else if e := schema.NewValidationListError(); errors.As(err, &e) {
+		fmt.Println("list error validation")
```
aeneasr, line 193: Remove

ui/container/container.go
```diff
@@ -189,6 +189,14 @@ func (c *Container) ParseError(group node.Group, err error) error {
 			}
 		}
 		return nil
+	} else if e := schema.NewValidationListError(); errors.As(err, &e) {
```
aeneasr, line 192: Add a test for this

schema/errors.go
```diff
@@ -173,3 +173,43 @@ func NewNoVerificationStrategyResponsible() error {
...
+	return &ValidationError{
+		ValidationError: &jsonschema.ValidationError{
+			Message:     message,
+			InstancePtr: instancePtr,
+		},
+		Messages: messages,
+	}
+}
+
+type ValidationListError struct {
```
aeneasr, line 187: Add tests for this struct

selfservice/hook/web_hook.go
```diff
@@ -53,24 +59,43 @@ type (
...
-		RequestHeaders http.Header        `json:"request_headers"`
-		RequestMethod  string             `json:"request_method"`
-		RequestUrl     string             `json:"request_url"`
-		Identity       *identity.Identity `json:"identity"`
+		Flow           flow.Flow             `json:"flow"`
+		RequestHeaders http.Header           `json:"request_headers"`
+		RequestMethod  string                `json:"request_method"`
+		RequestUrl     string                `json:"request_url"`
+		Identity       *identity.Identity    `json:"identity,omitempty"`
+		Credentials    *identity.Credentials `json:"credentials,omitempty"`
```
aeneasr, line 75: I don't think we should include these. I believe they would be empty always anyways and it would not be a good idea to send e.g. hashes over the network. The risk of accidental logging by some component or other types of compromise is just too high! So I say we remove this:

```suggestion
```

selfservice/hook/web_hook.go
```diff
@@ -307,9 +350,58 @@ func doHttpCall(method string, url string, as AuthStrategy, body io.Reader) erro
 
 	if err != nil {
 		return err
-	} else if resp.StatusCode >= 400 {
+	} else if resp.StatusCode >= http.StatusBadRequest {
```
aeneasr, line 353: So we have two options here:

1. If the response is an error code we stop execution and try to understand the error, relaying it to the UI
2. If the response contains a special field (e.g. `abort_flow`) we interrupt the flow and display an error

Generally speaking I think calling the config `can_interrupt` is misleading, because an error will always interrupt the flow. What this feature is doing is to allow us to gracefully return validation errors to the user, as opposed to showing a 500 server error.

selfservice/hook/web_hook.go
```diff
@@ -307,9 +350,58 @@ func doHttpCall(method string, url string, as AuthStrategy, body io.Reader) erro
...
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return errors.Wrap(err, "could not read response body")
+	}
+	defer func(Body io.ReadCloser) {
+		if closeErr := Body.Close(); closeErr != nil {
+			err = closeErr
+		}
+	}(resp.Body)
```
aeneasr, line 378: Remove this - see comment above

```suggestion
```

selfservice/hook/web_hook.go
aeneasr, line 353: This actually has a connection leak. It should be written like so and also be using a resilient client:

```go
	client := httpx.NewResilientClient(/*...*/)
	resp, err := client.Do(req)
	if err != nil {
		return err
	}
	defer resp.Body.Close()

	if resp.StatusCode >= http.StatusBadRequest {
```

You can instantiate the resilient client once in the hook struct `New` method.

selfservice/hook/web_hook.go
```diff
@@ -307,9 +350,58 @@ func doHttpCall(method string, url string, as AuthStrategy, body io.Reader) erro
...
 
 	return nil
 }
+
+func parseResponse(resp *http.Response) (err error) {
+	if resp == nil {
+		return fmt.Errorf("empty response provided from the webhook")
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
aeneasr, line 370: Instead of reading the body into memory and then unmarshalling, it is much more effective to do:

```
if err := json.NewDecoder(resp.Body).Decode(); err != nil {
```

aeneasr (changes requested):
Thank you for the PR! This is going in the right direction :)

I have merged the conflicts (as we have changed something in the hooks) which I pushed to a separate branch and made a PR against your fork: https://github.com/Wikia/kratos/pull/14

High-level this looks good. To see it in action, we will need some tests. I am in particular interested to see how the error parsing will be working in combination with the UI messages / fields.

We also need to make some decisions on how Ory Kratos detects whether a hook wants to return validation errors or not. I also think we need to rename the feature to better match what it does.

* [ ] Define a better name for the feature in the config
* [ ] Come up with a definition how a web hook could tell Ory Kratos that we have a validation error going on
* [ ] Add functional tests for web hook
* [ ] Add unit tests for the new helper structs
* [ ] Add documentation on how to return validation errors for hooks
* [ ] Add e2e test that tests this feature from start to finish
* [ ] Remove credentials from hooks

schema/errors.go
harnash: Fix in 75a55945

schema/errors.go
harnash: Fix in 97d539aa

selfservice/hook/web_hook.go
harnash: Removed in d8511ebd

selfservice/hook/web_hook.go
harnash: I was thinking about it in a similar way and I've decided to go with HTTP status codes to be somewhat in sync with what the Kratos client will receive.

Option 1. Is a bit more logical from the hook perspective that 200s will mean success and will not interrupt the flow and any other status code will one way or the other cause errors.
Option 2. Is a bit cleaner on Kratos' side since it will explicitly indicate an event that is being triggered by the web hook reponse.

I can go with either one and do not have strong preferences. I guess it is your call then @aeneasr :-)

harnash: Done in 6b4c11081a322b71ff44231a9ead616580654442

selfservice/hook/web_hook.go
harnash: That is neat! Thanks for the tip!  Fix in: 81106d15.

ui/container/container.go
harnash: Damn. Fixed in 7177daf8

dadrus:
When I was implementing the web hooks, my extension idea was to let the response be verified in a yet another jsonet template to which, as with building of the web hook request, a couple of objects can be passed in, like the response and the actuall flow objekt. This way the protocol of the web hook is fully open and one can achieve much more compared with the implementation in this pr. What do you think?

harnash:
> When I was implementing the web hooks, my extension idea was to let the response be verified in a yet another jsonet template to which, as with building of the web hook request, a couple of objects can be passed in, like the response and the actuall flow objekt. This way the protocol of the web hook is fully open and one can achieve much more compared with the implementation in this pr. What do you think?

Yes I was thinking about this but I am not sure how we would properly propagate those responses up to the UI which requires some concrete structures/data to properly handle validation errors etc. (see: https://github.com/ory/kratos/blob/master/ui/container/container.go#L152). It would be great to come up with some flexible solution that would allow to validate response using Jsonnet and transform it into something Kratos can use. For the scenarion where  we have fixed data structures in Kratos and create custom payload for the webhooks Jsonnet works nicely providing the glue betwen Kratos and external services for the other way around it is a bit trickier. There must be some binding point. I will think about it more but right now I have no good solution.

dadrus:
I think this can remain as is implemented by this PR. I mean the implementation right now expects a message with some text in the web hook response. Why not just letting a jennet template call a function on the passed context object, like e.g. `ctx.abortFlow("some error description")` ."some error description" can then come from the payload received from the Web hook or be just set in the jsonet template based on some conditions which are up to the developer of that template. When the template execution is done the remaining code will have everything which is required. Or do I overlook something?

dadrus:
If you like I can contribute some lines of code 😉 so you can directly see if it fits.

aeneasr:
JsonNet is not really a programming language that executes things. It's for manipulating JSON! Adding functionality like `ctx.abortFlow` will not work because the control flow is not aborted - you still need to return some JSON. I also think it is not a good idea to have this (potentially untrusted) code be able to call into the internal API of Ory Kratos (as a general pattern, not particularly for this idea).

Keep in mind that regular code is much easier to program, test, and document. JsonNet is not that great when trying to test it. The less logic we have in JsonNet, the better!

aeneasr:

Thank you!

harnash:
OK. I had some interruptions lately and I wasn't able to devote much time to this but hopefully I can resume work on this soon.

aeneasr:
No rush! Unmerged PRs give me a bit of anxiety so I tend to try to keep the  backlog short 😅

aeneasr:
When this is good for review, mark it as ready!

harnash:
I've just added support for maintaining form/flow data when hook returns error. Only tests to go now I think.

harnash:
Sorry again for the delay. I was quite busy but I'm going to work on it this week. Will ping you when I'm done.

schema/errors.go
harnash: Done in d188b7c55

ui/container/container.go
harnash: Done in d188b7c55

harnash:
I think it is ready for review

aeneasr:
@harnash sorry I was really busy these past days. I will try to take a 👀 asap

harnash:
One question to answer is: how to handle errors when parsing responses from the blocking/interruptible web hooks? 

Currently when there is an issue when:
* response could not be unmarshaled
* could not call webhook
Kratos will return 5xx error resulting in selfservice `/error` redirection. From user perspective it might be better to return generic error (i.e. error calling webhook) and retain form and it's contents so it can be retried if needed. Right now it is not possible and user has to retype all data in the forms.

harnash:
@aeneasr did you had time to take a look?

selfservice/flow/login/hook.go
```diff
@@ -66,7 +69,30 @@ func (e *HookExecutor) requiresAAL2(r *http.Request, s *session.Session) (*sessi
 	return aalErr, errors.As(err, &aalErr)
 }
 
-func (e *HookExecutor) PostLoginHook(w http.ResponseWriter, r *http.Request, a *Flow, i *identity.Identity, s *session.Session) error {
+func (e *HookExecutor) handleLoginError(_ http.ResponseWriter, r *http.Request, g node.Group, f *Flow, i *identity.Identity, flowError error) error {
```
aeneasr, line 72: I think this is missing tests?

selfservice/flow/recovery/hook.go
```diff
@@ -60,28 +64,51 @@ func NewHookExecutor(d executorDependencies) *HookExecutor {
 	}
 }
 
+func (e *HookExecutor) handleRecoveryError(_ http.ResponseWriter, r *http.Request, f *Flow, i *identity.Identity, flowError error) error {
```
aeneasr, line 67: Missing tests?

selfservice/flow/registration/hook.go
```diff
@@ -82,6 +85,37 @@ func NewHookExecutor(d executorDependencies) *HookExecutor {
 	return &HookExecutor{d: d}
 }
 
+func (e *HookExecutor) handleRegistrationError(_ http.ResponseWriter, r *http.Request, ct identity.CredentialsType, f *Flow, i *identity.Identity, flowError error) error {
```
aeneasr, line 88: Missing tests?

selfservice/flow/settings/hook.go
```diff
@@ -94,6 +96,37 @@ func WithCallback(cb func(ctxUpdate *UpdateContext) error) func(o *postSettingsH
 	}
 }
 
+func (e *HookExecutor) handleSettingsError(_ http.ResponseWriter, r *http.Request, settingsType string, f *Flow, i *identity.Identity, flowError error) error {
```
aeneasr, line 99: Missing tests?

selfservice/flow/verification/hook.go
```diff
@@ -60,14 +64,37 @@ func NewHookExecutor(d executorDependencies) *HookExecutor {
 	}
 }
 
+func (e *HookExecutor) handleVerificationError(_ http.ResponseWriter, r *http.Request, f *Flow, i *identity.Identity, flowError error) error {
```
aeneasr, line 67: Missing tests?

ui/container/container.go
```diff
@@ -87,6 +87,17 @@ func NewFromJSON(action string, group node.Group, raw json.RawMessage, prefix st
 	return c
 }
 
+//NewFromStruct creates a UI Container based on serialized contents of the provided struct.
```
aeneasr, line 90:
```suggestion
// NewFromStruct creates a UI Container based on serialized contents of the provided struct.
```

selfservice/flow/registration/hook.go
```diff
@@ -82,6 +85,37 @@ func NewHookExecutor(d executorDependencies) *HookExecutor {
...
+	if f != nil {
+		if i != nil {
+			var group node.Group
+			switch ct {
+			case identity.CredentialsTypePassword:
+				group = node.PasswordGroup
+			case identity.CredentialsTypeOIDC:
+				group = node.OpenIDConnectGroup
```
aeneasr, line 96: Can you please extract this logic into a common helper function that converts from credential types to groups and vice versa? And then use this instead of having the correct group conversion applied in different files?

aeneasr:
@abador I have some changes locally which I would like to push to your branch. You invited me to the repo but I did not see it and the invitation has expired. Could you please invite me again? Sorry for the trouble :(

aeneasr:
Closed on accident

selfservice/hook/web_hook.go
aeneasr: The connection leak is still there, reopening this. I have a local commit which fixes this.

harnash:
@aeneasr I've just re-added you with `write` access

aeneasr:
Thank you! I pushed my changes!

aeneasr (commented):
Thank you, this looks great! I think we need a couple more things before we can merge this into master :)

1. [ ] It would be really good to have E2E tests for every flow that ensure that the web hook failure indeed leads to a UI message. Only that way can we be certain that it works as expected.
2. [ ] The documentation (as noted in the opening message) should be updated to reflect how hooks can intercept registration.
3. [ ] The new methods added to the hooks (e.g. `handleLoginError`, `handleRegistrationError`) are more or less duplicated and all do the same thing. Can we maybe extract them into a common helper function which is tested, and reuse the code?

Something else I noticed during code review is that you don't like the `json:""` tags :) For Ory's APIs we always use lower case and underscore naming for JSON. So `InstancePtr` becomes `instance_pointer` :) I updated the place accordingly wherever I found this to be missing!

aeneasr:
By the way, I am keen on getting this merged :) So will try to be as responsive as possible!

harnash:
Thanks. I will have time to work on this soon. I'll read through all comments and try to address them quickly!

harnash:
I've added the docs but I'm waiting for a fix to:
```json
Unable to build website for locale "en".
Error: Server responded to https://raw.githubusercontent.com/ory/docs/master/docs/docs/ecosystem/bugbounty.md with status code 404:
404: Not Found
``` which seems to be caused by https://github.com/ory/docs/pull/540

**Comment by vinckr** [comment]
Fixed now @harnash, my bad!

**Comment by lanphan** [comment]
Hello @aeneasr and @harnash ,

There is 1 special case here with registration flow: in case registration flow is configured with important web-hooks (such as hook to continue some business logic after creating new identity in kratos), is it possible to use this new feature, interuptable web-hook, to configure failed flow with some additional actions (such as, delete new created identity because web-hook failed)?

(In fact, I see that there is PostRegistrationPrePersistHooks in code, but I don't know how to configure it, seems it's not available)

**Comment by harnash** [comment]
@lanphan Right. So you could run some additional logic before identity is created (but not yet stored in the database) via `PostRegistrationPrePersistHook` and do complex business logic. That is exactly what we are doing right now. It is a bit tricky since there are timeouts (so the hook can't run for too long) but in most cases it should be OK.

**Comment by lanphan** [comment]
thank @harnash ,
looking forward to your PR :)

File: selfservice/flow/login/hook_test.go [hunk 19]
```diff
@@ -46,6 +48,14 @@ func TestLoginExecutor(t *testing.T) {
 					}
 				})
 
+				var group node.Group
+				switch strategy {
```
abador, line 52: shouldn't tests cover other types? https://github.com/Wikia/kratos/blob/upstream_blocking_hooks/identity/credentials.go#L46

harnash:
Alright! I think it is ready for re-review!

selfservice/flow/registration/hook.go
```diff
@@ -101,7 +103,26 @@ func (e *HookExecutor) PostRegistrationHook(w http.ResponseWriter, r *http.Reque
...
+			case identity.CredentialsTypeRecoveryLink:
+				group = node.RecoveryLinkGroup
+			case identity.CredentialsTypeLookup:
+				group = node.LookupGroup
+			case identity.CredentialsTypeTOTP:
+				group = node.TOTPGroup
+			case identity.CredentialsTypeWebAuthn:
+				group = node.WebAuthnGroup
+			}
```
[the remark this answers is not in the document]
harnash: Will do!

aeneasr:

Thank you!

harnash:
I should be done with the fixes next week. Right now I have to focus on other tasks. Sorry for the delay.

aeneasr:
No problem!

harnash:
Fix in 6a08bef8d8b4394fd834b70da6d7fc91df241c1d

harnash:
I believe I've addressed all comments.

hperl (changes requested):
Hi, thank you for your changes. I have taken over the review from @aeneasr and left some comments so that we can get this merged soon :muscle: . I am still somewhat new to the code base, so please forgive me if some comments are not applicable :wink: .

harnash:
@hperl thank you. I will be looking and addressing those comments today and tomorrow.

identity/credentials.go
```diff
@@ -42,6 +43,21 @@ func (c CredentialsType) String() string {
...
+	switch c {
+	case CredentialsTypePassword:
+		return node.PasswordGroup
+	case CredentialsTypeOIDC:
+		return node.OpenIDConnectGroup
+	case CredentialsTypeTOTP:
+		return node.TOTPGroup
+	case CredentialsTypeWebAuthn:
+		return node.WebAuthnGroup
+	default:
```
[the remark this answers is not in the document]
harnash: Fix in f7ef178ea

selfservice/flow/login/hook_test.go
```diff
@@ -26,11 +26,13 @@ import (
...
 
 func TestLoginExecutor(t *testing.T) {
-	for _, strategy := range []string{
-		identity.CredentialsTypePassword.String(),
-		identity.CredentialsTypeOIDC.String(),
+	for _, strategy := range []identity.CredentialsType{
+		identity.CredentialsTypePassword,
+		identity.CredentialsTypeOIDC,
+		identity.CredentialsTypeTOTP,
+		identity.CredentialsTypeWebAuthn,
```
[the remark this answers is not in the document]
harnash: Fix in f7ef178ea

schema/errors.go
```diff
@@ -247,6 +247,50 @@ func NewNoWebAuthnRegistered() error {
 	})
 }
 
+func NewHookValidationError(instancePtr, message string, messages text.Messages) *ValidationError {
+	return &ValidationError{
```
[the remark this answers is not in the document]
harnash: Right. This one is a bit special since it is wrapped by `ValidationListError`. I've added `errors.Wrap()` to `NewValidationListError()` but I don't think it will be good to wrap those additionally.

selfservice/flow/login/handler.go
```diff
@@ -171,7 +172,8 @@ preLoginHook:
 	}
 
 	if err := h.d.LoginHookExecutor().PreLoginHook(w, r, f); err != nil {
-		return nil, err
+		h.d.LoginFlowErrorHandler().WriteFlowError(w, r, f, s.NodeGroup(), err)
```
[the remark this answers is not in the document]
harnash: I was looking at `selfservice/flow/registration/handler.go` it is a bit similar but login flow might behave a bit differently than registration. I'll add a check that will prevent exception but I'm not sure what to choose there. Logic is a bit different and we do not stop after first successful strategy completed so I would not like to change this behaviour unless I'm pretty sure it is the right way to do it. We could use `node.DefaultGroup` but that would probably be problematic in some cases.

hperl:
Thanks, let me know when I can take a look :).

selfservice/flow/login/handler.go
```diff
@@ -606,7 +610,7 @@ continueLogin:
 		return
 	}
 
-	if err := h.d.LoginHookExecutor().PostLoginHook(w, r, f, i, sess); err != nil {
+	if err := h.d.LoginHookExecutor().PostLoginHook(w, r, group, f, i, sess); err != nil {
```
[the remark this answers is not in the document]
harnash: I think this one is in sync with registration. We use the first one that succeeded (last iteration). If there were errors than we deal with that in the loop above if none succeeded (or none are defined) then we do not have a valid identity in `i` and the process will halt before reaching this point.

selfservice/hook/web_hook.go
```diff
@@ -141,6 +174,11 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
 		}
 
 		if resp.StatusCode >= http.StatusBadRequest {
+			if gjson.GetBytes(e.conf, "can_interrupt").Bool() {
+				if err := parseWebhookResponse(resp); err != nil {
```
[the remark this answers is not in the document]
harnash: Partially out of habit :-) But this is an implementation assumption and caller might not be aware of this  and can cause issues when someone modifies `parseWebhookResponse()` and add a case where `nil` will be returned. Usually when function returns an `error` one might expect it to return `nil` in some cases (default Go contract) and keeping this logic would violate "least surprise" rule ;-)

selfservice/hook/web_hook.go
```diff
@@ -158,3 +196,39 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
...
+				Text:    detail.Text,
+				Type:    msgType,
+				Context: detail.Context,
+			})
+		}
+		validationErr.Add(schema.NewHookValidationError(msg.InstancePtr, msg.Message, messages))
+	}
+
+	if !validationErr.HasErrors() {
+		return errors.New("error while parsing hook response: got no validation errors")
```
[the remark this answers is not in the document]
harnash: Umm. So there is an case when hook is set to be "blocking" type. There is an contract/assumption that there webhook will either:
* succeed and flow will continue as normal (http status < 400)
* will process the request and decide to abort the flow (http status >= 400) and now we might have situation:
  * webhook returned response that Kratos could decode and could understand and propagates error to the UI
  * webhook returned outdated/corrupted response and Kratos could not parse it correctly
  * request was handled/interrupted somewhere between Kratos and webhook and response is also unrecognizable by Kratos (basically the same situation as above)
  * webhook returned an error but there are no errors inside a container (something wrong happened inside and probably needs attention, it looks like webhook decided to block the flow but couldn't add proper error) <- this is this case

It is probably not so convenient but we need to take into account that request may fail and Kratos should react somehow. It is just an extra check to make sure that webhook returns valid data so we don't break flow without specifying a reason for the user to react.

aeneasr:
To resolve the CI I think we need to resolve the merge conflicts :)

aeneasr:
There's still one failing test :) https://github.com/ory/kratos/runs/6671177484?check_suite_focus=true#step:16:20427

harnash:
@hperl I think I'm done addressing your comments. Let me know if there  is anything else.

hperl:
Thanks for the ping, I'll take a look tomorrow!

selfservice/hook/web_hook.go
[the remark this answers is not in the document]
hperl: One more thing, we only read from `errChan` once in https://github.com/ory/kratos/pull/1585/files#diff-6dbaf3f7f3f6f427e72de26f22beb17b0c56379f702ae5be81516c38765f8d13R220, but are writing to it in lines 201 and 204. The second send (in 204) will still succeed, since the channel is buffered, but it is still cleaner to return after the first send in 201, since that is what will really happen, right?

hperl (dismissed):

embedx/config.schema.json
```diff
@@ -208,6 +208,11 @@
                 }
               ]
             },
+            "can_interrupt": {
```
aeneasr, line 211: What happens if we set this value and `response.ignore` both to true? Should they be mutually exclusive?

aeneasr:
All in all this looks very, very, very good :) I will need to try it out end-to-end, but I don't think that there will be big topics outside of: https://github.com/ory/kratos/pull/1585#discussion_r891441712 :) :)

embedx/config.schema.json
harnash: Will take a look.

harnash: Not an JSON schema expert but I think I got this working: 97514c298

harnash:
I think I've addressed all comments @aeneasr + @hperl . We are going to try migrate to v0.10.x in few days and if you say it is in mergable state I will try to incorporate this into our branch and test it on DEV env to see if there are some unexpected issues.

aeneasr:
> I think I've addressed all comments @aeneasr + @hperl . We are going to try migrate to v0.10.x in few days and if you say it is in mergable state I will try to incorporate this into our branch and test it on DEV env to see if there are some unexpected issues.

Do you have an update on this? :)

harnash:
> > I think I've addressed all comments @aeneasr + @hperl . We are going to try migrate to v0.10.x in few days and if you say it is in mergable state I will try to incorporate this into our branch and test it on DEV env to see if there are some unexpected issues.
> 
> Do you have an update on this? :)

Sadly migration to v0.10 is being postponed a bit. Will let you know as soon as I know more. Currently we have this in backlog and planned for next sprint (~ two weeks).

aeneasr:
Ok, no problem :) I think the code looks good so I would be fine merging it after I do one final round of review

aeneasr (approved):
It's perfect. Thank you!

I've made two minor changes: I removed the `message` string and renamed `detailed_messages` to `messages`. The reason being that `message` did not actually have an effect and it was a bit confusing why we need it.

The second thing was that there was a connection leak in the webhook handler as `resp.Body.Close()` wasn't being called. This is fixed now :)

I've updated the docs also

Overall, great achievement!

harnash:
Thanks!
We are still working on upgrade to v0.10.1 but I think we could test this build soon enough.

PR #1606, opened 2021-07-29
feat: allow multiple webhook body sources

zepatrik:
## Related issue(s)

Currently the webhook body can only be loaded from a local file. This allows to use any source enabled in the JSONSchema loaders (file, base64, https, ...)

## Checklist

- [x] I am following the
      the changes.
      works.


selfservice/hook/web_hook_test.go
```diff
@@ -91,47 +91,78 @@ func TestApiKeyInCookieStrategy(t *testing.T) {
...
+					"Some-Header": []string{"Some-Value"},
+				},
+				RequestMethod: "POST",
+				RequestUrl:    "https://test.kratos.ory.sh/some-test-path",
+				Identity:      i,
+			},
+		},
+		{
+			desc:     "filepath without scheme",
+			template: "./stub/test_body.jsonnet",
```
aeneasr, line 121: Would it make sense to emit a warning if this is used, and at some point deprecate it (and create an issue tracking the deprecation)?

driver/config/.schema/config.schema.json
```diff
@@ -147,7 +147,13 @@
             },
             "body": {
               "type": "string",
-              "description": "Path to the jsonnet template used for payload generation. Only used for those HTTP methods, which support HTTP body payloads"
+              "description": "URI pointing to the jsonnet template used for payload generation. Only used for those HTTP methods, which support HTTP body payloads",
+              "examples": [
+                "file:///path/to/body.jsonnet",
+                "/path/to/body.jsonnet",
+                "./body.jsonnet",
+                "base64://ZnVuY3Rpb24oY3R4KSB7CiAgaWRlbnRpdHlfaWQ6IGlmIGN0eFsiaWRlbnRpdHkiXSAhPSBudWxsIHRoZW4gY3R4LmlkZW50aXR5LmlkLAp9="
```
aeneasr, line 155: Do we allow loading from `http` too?

selfservice/hook/web_hook_test.go
zepatrik: I mean we allow it for all other URIs like identity schema as well in the same way...

driver/config/.schema/config.schema.json
zepatrik: Yes, if the loader is registered.

selfservice/hook/web_hook_test.go
aeneasr: I am actually not sure if we support filepaths in the other things (e.g. jsonnet loader). If we don't I think we should deprecate, if we do, we can keep it

zepatrik: Ah right, missed that we have a different loader there. Will mark it as deprecated then.

aeneasr (approved):

PR #1624, opened 2021-08-05
feat: add mfa using totp

### LINKED ISSUE #26: Authentication refactoring and MFA capabilities: TOTP + Lookup Secrets
This is an umbrella issue for all things related to single and second factor authentication, the different factors, passwordless authentication, and more.

* [x] TOTP
* [x] Recovery Codes
* [x] WebAuthn
* [ ] Documentation
* [x] UI examples

Implementation of these features by the ORY Kratos maintainers has officially started: https://github.com/ory/kratos/issues/26#issuecomment-745464249

aeneasr:
*tbd*

Closes #26

## Checklist

- [x] I am following the
      the changes.
      works.

---

- [ ] TOTP issuer should be configurable (used in `totp.NewKey()`
- [ ] TOTP account name is currently the identity's UUID. Maybe we could somehow use the user's email address instead? Maybe as an annotation in JSON Schema?


aeneasr:
Implementing QR code and text handling was really a breeze with the changes introduced in 0.6:

Bildschirmfoto 2021-08-05 um 12 49 42

Unlinking is available also:

Bildschirmfoto 2021-08-05 um 13 54 58

Unlinking requires a privileged session right now.

aeneasr:
https://user-images.githubusercontent.com/3372410/128347384-3976918a-13af-4288-aa90-c1d4a0e93b45.mov

aeneasr:
So there are a couple of questions related to mfa login which we just discussed and I want to keep them for record here:

1. OIDC providers do not provide context about auth factor so we need to deal with them as if they were a first factor only
2. Sessions should have the auth factor in it
3. Login should always be first factor, with an optional hook to redirect to mfa login
4. App should be responsible for checking if first factor is enough or not / depending if mfa is enabled redirect to settings screen
5. Some type of configuration declaring what methods can be used for what factor

dadrus:
@aeneasr: Great stuff!!!!
I have however a "hint". If you could take it into account, that would be great.

When dealing with totp and/or hotp based authentication, one should always take a possible time/counter deviation on the client side into account. Even dealing with otp values generated on a smart phone might not always match the value generated on the server side. For that reason the provider of such tokens or services always work with time frames (if we ignore hotp for now) and not with a single time point. That basically means, the auth server generates multiple otp values and checks whether the received otp value matches one of those generated. If it doesn't match - the result is obvious. If it however matches, the server will adjust the deviation by moving the time window accordingly. So next time the client sends the otp, it can take that deviation into account and will build the otp verification window around that adjusted time point.

How big the time window should be is obviously a security concern. For that reason it should be configurable. Same is also true for the automatic deviation adjustment. The window configuration is don in steps. A step is just how many otps in the future and in the past from the given server side time point should be considered valid. So the value of 0 would mean only the value generated on the server for it current time. A value of 1 means, the window consist of the value generated for the current server time, as well as of one otp value in the future and one otp value in the past. And so on. How long that step is depends on the otp generation parameters. If I remember correctly OATH defines 30 seconds for it.

So if this would be the otp generation function: 
`ts=$(date +%s); echo $(($ts / 30)) |  openssl sha1 -hmac "password123" -r |  head -c 6`
with 
- `date` being obviously the current server time,
- `30` in the `echo $(($ts / 30))` part being the step length. Here 30 seconds
- `head` in `head -c 6` being a super cool but not compliant transformation function, which just shortens the result to 6 characters and
- `password123` being the shared secret

A window of the size 0 would only contain the otp value of `fc7e37` for the server side time `1612046280` +/- 30 sec, a window of 1 would contain the above otp, plus an otp generated for 1612046280 - 30s and and an otp generated for 1612046280 + 30s

aeneasr:
Yay, it is possible to replay webauthn with the same challenge - so we can actually test this stuff in the go code :)

aeneasr:
WebAuthn integration tests are done

aeneasr:
```

sql("UPDATE identity_credential_identifiers")

change_column("identity_credential_identifiers", "identity_credential_type_id", "uuid", {})
add_index("identity_credential_identifiers", ["nid", "identity_credential_type_id", "identifier"], {"unique": true, "name": "identity_credential_identifiers_identifier_nid_type_uq_idx"})

```

pradyuman:
I notice there hasn't been activity on this in the last few weeks, Just curious, what's blocking this? I might be able to help out.

aeneasr:
Vacation ;)

pradyuman:
Ah, no worries then. Hope you have a relaxing break!

Issue #1662, opened 2021-08-19
Successfull execution of a recovery flow does not update the Verified state of the used email address


dadrus:
**Describe the bug**

If kratos is configured to allow logins with verified addresses only (via the `require_verified_address` hook), a "login" is still possible via the recovery flow. The recovery flow does however not update the `Verified` state of the used email address to `true`, thus a regular login is still not possible.

**Reproducing the bug**

* Configure kratos self service flows to look like follows:
```.yaml
    login:
      ui_url: <url for login>
      after:
        password:
          hooks:
            - hook: require_verified_address

    registration:
      ui_url: <url for registration>
      after:
        default_browser_return_url: <url to inform the user, that email verification is pending>

    verification:
      enabled: true
      ui_url: <url for verification>
      after:
        default_browser_return_url: <url to inform the user, that email verification is done>

    recovery:
      enabled: true
      ui_url: <url for recovery>
```

When the user registers and does not click the link in the received email to verify the email address and tries to login, the login is rejected with an error message saying, the account is not yet active (which is expected). When the user now goes through the password recovery flow and clicks the recovery link in the received (password recovery) email, the user is able to change the password (as expected) and kratos issues a valid session, so the user is logged in (this is expected as well). If the user now logs out and tries to login regularly, he/she will again run into an error saying the account is not active (not expected), even the previously executed password recovery flow effectively did an email verification. 

**Expected behavior**

The email used for recovery flow is marked as verified, so the user can login.

**Additional notes**

@aeneasr: actually this is my fault, as I had not this part of kratos in my mind while working on the PR for https://github.com/ory/kratos/issues/1328. If https://github.com/ory/kratos/blob/1ba6c4ac8ba88ac4900c84a651f68807332d7c8e/selfservice/strategy/link/strategy_recovery.go#L258 is the proper function to fix it, a PR will follow soon. ;) Btw. where do I get the email used for recovery from?

dadrus:
My naive bug fix would look like this:

```.go
func (s *Strategy) recoveryIssueSession(w http.ResponseWriter, r *http.Request, f *recovery.Flow, recoveryAddress *identity.RecoveryAddress) error {
	recovered, err := s.d.IdentityPool().GetIdentity(r.Context(), recoveryAddress.IdentityID)
	if err != nil {
		return s.HandleRecoveryError(w, r, f, nil, err)
	}

	var address *identity.VerifiableAddress
	for _, va := range recovered.VerifiableAddresses {
[...]

PR #1665, opened 2021-08-19
feat: Mark recovery email address verified

dadrus:
## Related issue(s)

@aeneasr: This PR is for https://github.com/ory/kratos/issues/1662

## Checklist

- [x] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security. vulnerability, I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.

## Further Comments

The implementation idea is still the same as described in https://github.com/ory/kratos/issues/1662. This PR does however introduce a separate function for marking the address as verified instead of misusing the `recoveryIssueSession` method.


dadrus:
@aeneasr: if this is the way to go, I'll add the required tests and extend the documentation. Please take a look.

aeneasr:
Yeah that looks good! :)

dadrus:
@aeneasr : Functionality wise I consider this PR to be complete. So please take a look. After taking a closer look at already existing documentation, it already states, that the verification is initiated by the registration and verification flows. This way there is no need for documentation update.

selfservice/strategy/link/strategy_recovery.go
```diff
@@ -353,38 +360,73 @@ func (s *Strategy) retryRecoveryFlowWithMessage(w http.ResponseWriter, r *http.R
...
 		node.NewInputField("email", body.Email, node.RecoveryLinkGroup, node.InputAttributeTypeEmail, node.WithRequiredInputAttribute),
 	)
 
-	req.Active = sqlxx.NullString(s.RecoveryNodeGroup())
-	req.State = recovery.StateEmailSent
-	req.UI.Messages.Set(text.NewRecoveryEmailSent())
-	if err := s.d.RecoveryFlowPersister().UpdateRecoveryFlow(r.Context(), req); err != nil {
-		return s.HandleRecoveryError(w, r, req, body, err)
+	if address != nil && !address.Verified {
+		address.Status = identity.VerifiableAddressStatusSent
```
aeneasr, line 396: While I think it makes sense to mark the verified address as verified if we perform recovery with it, I think that it should stay untouched otherwise. For example, we are not marking the recovery email as status sent if we verify it. What do you think?

selfservice/strategy/link/strategy_recovery.go
```diff
@@ -353,38 +360,73 @@ func (s *Strategy) retryRecoveryFlowWithMessage(w http.ResponseWriter, r *http.R
...
+	if err := s.d.RecoveryFlowPersister().UpdateRecoveryFlow(r.Context(), f); err != nil {
+		return s.HandleRecoveryError(w, r, f, body, err)
+	}
+
+	return nil
+}
+
+func (s *Strategy) markRecoveryAddressVerified(w http.ResponseWriter, r *http.Request, f *recovery.Flow, id *identity.Identity, recoveryAddress *identity.RecoveryAddress) error {
+	var address *identity.VerifiableAddress
+	for _, va := range id.VerifiableAddresses {
```
aeneasr, line 415: Using the index here could help clean up the copy/reassignment stuff a bit

selfservice/strategy/link/strategy_recovery_test.go
```diff
@@ -159,18 +165,16 @@ func TestAdminStrategy(t *testing.T) {
 
 		require.Len(t, f.UI.Messages, 1)
 		assert.Equal(t, "You successfully recovered your account. Please change your password or set up an alternative login method (e.g. social sign in) within the next 60.00 minutes.", f.UI.Messages[0].Text)
+
+		addr, err := reg.IdentityPool().FindVerifiableAddressByValue(context.Background(), identity.VerifiableAddressTypeEmail, "recoverme@ory.sh")
+		assert.NoError(t, err)
+		assert.False(t, addr.Verified)
```
aeneasr, line 171: Shouldn't this be true now?

selfservice/strategy/link/strategy_recovery_test.go
```diff
@@ -181,8 +185,70 @@ func TestRecovery(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.False(t, addr.Verified)
+	assert.Equal(t, sqlxx.NullTime{}, addr.VerifiedAt)
+	assert.Equal(t, identity.VerifiableAddressStatusPending, addr.Status)
+
+	addr, err = reg.IdentityPool().FindVerifiableAddressByValue(context.Background(), identity.VerifiableAddressTypeEmail, recoveryEmail4)
+	assert.NoError(t, err)
+	assert.False(t, addr.Verified)
+	assert.Equal(t, sqlxx.NullTime{}, addr.VerifiedAt)
+	assert.Equal(t, identity.VerifiableAddressStatusPending, addr.Status)
```
aeneasr, line 251: Could you please wrap this in a function or a for loop to avoid the code duplication?

aeneasr (changes requested):
Nice, this looks pretty much mergeable! I’ve got some follow up questions and some ideas for improvement!

selfservice/strategy/link/strategy_recovery_test.go
dadrus: No. This test is about the creation of a recovery link by an admin. So it is not sent to the user's eMail. This way the email address is not verified.

selfservice/strategy/link/strategy_recovery.go
dadrus: The recovery address is actually a separate struct. As such it does even not have such fields. But if the recovery eMail address is the same, as the eMail address used for verification purposes. In such case, IMHO the verifiable address object should receive the same updates as it receives in a regular verification flow. In both cases an eMail verification is done by sending an email to the user and letting it click the corresponding link. This is also why there is a check for `address != nil`. So I would not change the current implementation on this PR.

selfservice/strategy/link/strategy_recovery_test.go
dadrus: done

selfservice/strategy/link/strategy_recovery.go
dadrus: done

selfservice/strategy/link/strategy_recovery_test.go
dadrus: done

dadrus:
@aeneasr: Could you please restart the test job? There is some issue with the db not related to the latest commits at all.

zepatrik:
Done :heavy_check_mark:

dadrus:
Looks much better now ;) Thank you. Please let me know if you would like to have something more changed/enhanced.

selfservice/strategy/link/strategy_recovery_test.go
aeneasr: Makes sense!

selfservice/strategy/link/strategy_recovery.go
dadrus: Are you not conviced with my answer? ;)

aeneasr: Sorry, the 🏝 got in the way of me reading it 😅

Half half, I still see them as separate things. I agree that the flows work the same way (OOB TOTP over email) implying that solving one challenge is line solving the other. But then again they have different behaviors, lifespans, timeouts, and might have different security guards in the future.

At the same time, the link strategy is on the way out due to it’s many issues so to me it’s not that important!

If I was a dev and would do a recovery flow, and then see „hhmm now verification was also impacted“, I would probably open a bug report about it 😉

dadrus: Everything is fine. I'm not pushing ;)

> But then again they have different behaviors, lifespans, timeouts, and might have different security guards in the future.
> f I was a dev and would do a recovery flow, and then see „hhmm now verification was also impacted“, I would probably open a bug report about it 

The current [account-activation-email-verification](https://www.ory.sh/kratos/docs/guides/account-activation-email-verification) documentation says something else:
> Email verification currently supports sending out a recovery link to an email address.

The [verify-email-account-activation](https://www.ory.sh/kratos/docs/self-service/flows/verify-email-account-activation) document does not differentiate recovery and verification flows for email verification as well.

Actually, I fully agree with you if the email address used for recovery purposes differs from the email for account identification purposes (thus for login). And this situation is taken into account by the given PR. If they are however the same, I personally would expect the same state changes. And this is IMHO not related to lifespans, timeouts etc. They can for sure differ. I can change the implementation as you're asking for (even I'm not convinced), but then there is a need for documentation update to make the differences obvious and clear.

aeneasr: > The verify-email-account-activation document does not differentiate recovery and verification flows effects for email verification as well.

That is definitely a copy/paste error :D

I agree with you here - if the address is verified the state should reflect that! And yes it’s definitely redundant. I thought it might be useful to have a simple bool flag reflecting the true/false status rather than the state, which is used internally.

By the way, I believe that unless you do not have the verify hook, the state here will almost always be „sent“! We use it internally to decide if a verification mail should be sent or not, for example when you update your email.

Regarding your recovery use case - it makes sense. It would be confusing to the user if they do recovery and still can‘t sign in because they should have done verification. I don’t really know how other sites are handling this. In general I think the problem is that we issue a session here, without doing any hook checks. This logic in general is still pretty akward IMO.

For the emails themselves I think there is a difference - you would e.g. want to send a double consent email (email verify) rather than mingle that with recovery.

In the end both approaches look acceptable to me here. I still feel that we are tricking the system here a bit - we are sending a recovery email and say that we sent the verification email too. Then again most people have this status set to sent already if the hook is used.

One thing is for sure though, if the status is verified already, this shouldn’t be touched!

aeneasr: ps: thank you for taking the input seriously even if the change is minimal 😅

dadrus:
@aeneasr: The https://github.com/ory/kratos/pull/1665/commits/41ce12c6c1e111133d88c4aa49964e443b6b0cab commit removes the status update as you've asked for. The other commits just ensure everything is working with the current master state.

aeneasr:
Awesome, thanks! Could you please add a note to the docs in the recovery part saying that we will mark the address as verified also if they match? Then this is mergable :)

PR #1671, opened 2021-08-22
feat: search

david972:
This feature is for adding the capability to use query parameter to filter identities result on `/identities` handler.

For now couple things to do :

- Review the paging
- Add identifier in the search field
- update the documentation 

## Related issue(s)

#1278 

## Checklist

- [x] I am following the
      the changes.
      works.


identity/identity_filter.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/tidwall/gjson"
+)
+
+func (i Identities) Filter(values url.Values) (Identities, error) {
```
aeneasr, line 12: This is a nice idea! Unfortunately I think it will become quite difficult once you hit ~10.000 identities or more as you will need to load everything into memory. Another alternative would be to support search for databases that have JSON indexing (so this would not work for SQLite but it does work for PostgreSQL, MySQL 8+, CockroachDB). For SQLite we could use the fallback defined here! Alternative would be to load a JSON extension for SQLite although I don't know how that works.

david972: Yes it what I did first by using scope from pop but effectively tests does not work due to SQLite incompatibility

david972: I’ll dive in it

david972: Just push my previous implementation
give me you thought

.build-cgo
```diff
@@ -0,0 +1 @@
+// #cgo CFLAGS: -DSQLITE_ENABLE_JSON1
```
aeneasr, line 1: Wow, how did you figure this out :D That's great!

aeneasr (commented):
Hey, this looks so good! Love it - definitely the right approach!

persistence/sql/persister_identity.go
```diff
@@ -472,3 +510,65 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+				}
+			}
+			return q
+		case strings.HasPrefix(dsn, "postgres"), strings.HasPrefix(dsn, "crdb"), strings.HasPrefix(dsn, "cockroach"):
+			field, innerField := extractFieldAndInnerFields(field)
+			for _, value := range values {
+				if innerField == "" {
+					q = q.Where(fmt.Sprintf(`%s @> '"%s"'`, field, value))
+				} else {
+					q = q.Where(fmt.Sprintf(`%s @> '{"%s":"%s"}'`, field, innerField, value))
```
zepatrik, line 534: Will this work for nested traits?
E.g. checking if `traits.name.first == "bob"`?

zepatrik: Also, is value in this case only allowed to be of type string? As the JSON type is always double quoted, it would always be a string in JSON.

persistence/sql/persister_identity.go
```diff
@@ -472,3 +510,65 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+	return func(q *pop.Query) *pop.Query {
+		for field, value := range values {
+			if IsStringInSlice([]string{"page","per_page"}, field) {
+				continue
+			}
+			if strings.HasPrefix(field, "traits") {
+				q = q.Scope(p.getJsonSearchQuery(field, value))
+				continue
+			}
+			q = q.Where(fmt.Sprintf("%s IN (?)", field), value)
```
zepatrik, line 553: Here also `field` could be used for SQL injection.

persistence/sql/persister_identity.go
```diff
@@ -281,6 +282,43 @@ func (p *Persister) ListIdentities(ctx context.Context, page, perPage int) ([]id
 	return is, nil
 }
 
+func (p *Persister) ListIdentitiesFiltered(ctx context.Context, values url.Values, page, perPage int) ([]identity.Identity, error) {
+	is := make([]identity.Identity, 0)
```
zepatrik, line 286: Guess you can add a capacity here already.
```suggestion
	is := make([]identity.Identity, 0, perPage)
```

persistence/sql/persister_identity.go
```diff
@@ -281,6 +282,43 @@ func (p *Persister) ListIdentities(ctx context.Context, page, perPage int) ([]id
...
+		Scope(p.buildScope(values)).
+		Paginate(page, perPage).Order("identities.id DESC").
+		All(&is)); err != nil {
+		return nil, err
+	}
+
+	schemaCache := map[string]string{}
+
+	for k := range is {
+		i := &is[k]
```
zepatrik, line 302: Taking a pointer of the slice element will not make a copy, but use the same underlying memory. Not sure what the intent was here :sweat_smile:

zepatrik (changes requested):
Nice, I think I found some things that need improvement. Please prove me wrong if I am mistaken with something :sweat_smile:

persistence/sql/persister_identity.go
david972: 1. Should work I forgot to add test with that
2. in this case as value come from the query it always a string. I assume that could be work for int too. I'll add a test for this like phone number

persistence/sql/persister_identity.go
david972: good point

.build-cgo
david972: It shouldn't have be there it was a memo. 
I think we do not need it for it to works

persistence/sql/persister_identity.go
```diff
@@ -472,3 +510,65 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+	return func(q *pop.Query) *pop.Query {
+		dsn := p.r.Config(context.Background()).DSN()
+		switch {
+		case strings.HasPrefix(dsn, "sqlite"), strings.HasPrefix(dsn, "mysql"):
+			field, innerField := extractFieldAndInnerFields(field)
+			for _, value := range values {
+				if innerField == "" {
+					q = q.Where(fmt.Sprintf(`json_extract(%s, '$') = "%v"`, field, value))
+				} else {
+					q = q.Where(fmt.Sprintf(`json_extract(%s, '$.%v') = "%v"`, field, innerField, value))
```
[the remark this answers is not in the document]
david972: Yes sure I agree about the SQL injection
I'll fix it by using sql argument for the value because it does not work for the path.

persistence/sql/persister_identity.go
david972: The point is I tried to use sql parameter here but it didn't work because of value that is an array

persistence/sql/persister_identity.go
david972: it's what I get from the original
https://github.com/ory/kratos/pull/1671/files/751a0e6d80a62eccea872c87e07c57738dfc36ec#diff-e6bdb8b4b6f8cb588c61fbd8410a5adfd4eea9f40eca74740bcac59d63f3f21bR265

persistence/sql/persister_identity.go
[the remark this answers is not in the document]
aeneasr: Thank you @zepatrik for catching that

david972:
finally I found an alternative for breaking sql injection
Need some documentation 
also need to do a PR for the go client

persistence/sql/persister_identity.go
```diff
@@ -472,3 +510,115 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+}
+
+func IsStringInSlice(slice []string, val string) bool {
+	for _, item := range slice {
+		if item == val {
+			return true
+		}
+	}
+	return false
+}
```
zepatrik, line 624: https://github.com/ory/x/blob/a065f5780b6040793599f5d9fd68a246322b56ab/stringslice/has.go :wink:

persistence/sql/persister_identity.go
```diff
@@ -472,3 +510,115 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+	}
+	parts := strings.Split(key, ".")
+
+	for i, part := range parts {
+		part = strings.Trim(part, `"`)
+		part = strings.Trim(part, "`")
+		part = strings.TrimSpace(part)
+
+		parts[i] = fmt.Sprintf(`%c%v%c`, c, part, c)
+	}
```
zepatrik, line 576: What happens if `key` is `foo"; DROP TABLES ALL; --` or something similar (assuming quote `"`)?

zepatrik: I would also change this func to do the quote detection itself, as it is already defined on the persister.
```suggestion
func (p *Persister) Quote(ctx context.Context, key string) string {
	n := p.Connection(ctx).Dialect.Name()
	c, ok := quoteChar[n]
	if !ok {
		// guess panic is OK here as the error is not fixable without a new release of Kratos
		panic("DSN is of unknown dialect " + n)
	}

	parts := strings.Split(key, ".")

	for i, part := range parts {
		part = strings.Trim(part, `"`)
		part = strings.Trim(part, "`")
		part = strings.TrimSpace(part)

		parts[i] = fmt.Sprintf(`%c%v%c`, c, part, c)
	}
```

persistence/sql/persister_identity.go
```diff
@@ -472,3 +510,115 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+
+	return strings.Join(parts, ".")
+}
+func (p *Persister) buildScope(queryValues url.Values) pop.ScopeFunc {
```
zepatrik, line 580:
```suggestion
func (p *Persister) buildScope(ctx context.Context, queryValues url.Values) pop.ScopeFunc {
```

persistence/sql/persister_identity.go
```diff
@@ -472,3 +510,115 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+			if strings.HasPrefix(field, "traits") {
+				q = q.Scope(p.getJsonSearchQuery(field, values))
+				continue
+			}
+			if strings.HasPrefix(field, "credentials") {
+				q = q.Scope(p.searchCredentialQuery(field, values))
+				continue
+			}
+			//q = q.Where("? = ?", field, values[0])
+			field = p.Quote(field, quoteChar[p.Connection(context.Background()).Dialect.Name()])
```
zepatrik, line 602:
```suggestion
			field = p.Quote(ctx, field)
```

persistence/sql/persister_identity.go
```diff
@@ -281,6 +282,43 @@ func (p *Persister) ListIdentities(ctx context.Context, page, perPage int) ([]id
...
 
+func (p *Persister) ListIdentitiesFiltered(ctx context.Context, values url.Values, page, perPage int) ([]identity.Identity, error) {
+	is := make([]identity.Identity, 0, perPage)
+
+	/* #nosec G201 TableName is static */
+	if err := sqlcon.HandleError(p.GetConnection(ctx).Where("identities.nid = ?", corp.ContextualizeNID(ctx, p.nid)).
+		LeftJoin("identity_verifiable_addresses verifiable_addresses", "verifiable_addresses.identity_id=identities.id").
+		LeftJoin("identity_recovery_addresses recovery_addresses", "recovery_addresses.identity_id=identities.id").
+		EagerPreload("VerifiableAddresses", "RecoveryAddresses").
+		Scope(p.buildScope(values)).
```
zepatrik, line 293:
```suggestion
		Scope(p.buildScope(ctx, values)).
```

zepatrik (commented):
Nice progress :+1:

zepatrik:
Do you mean the Kratos go client? That will be updated through CI automation :wink:

persistence/sql/persister_identity.go
david972: > Will this work for nested traits?
> E.g. checking if `traits.name.first == "bob"`?

it's work

persistence/sql/persister_identity.go
david972: > What happens if `key` is `foo"; DROP TABLES ALL; --` or something similar (assuming quote `"`)?

does not seem to work

persistence/sql/persister_identity.go
david972: ha ha ha ok I switch this

david972:
I find out a behavioural around persistent identity
If I do a first call (CURL) with the with_credentials flags all other call without the flag continuous to benefit to it.
example:
```
curl -X GET 'http://localhost:4434/identities?credentials.identifier=github:1423404' | jq '.'
{
  "error": {
    "code": 500,
    "status": "Internal Server Error",
    "message": "ERROR: no data source matches prefix: credential_identifiers (SQLSTATE 42P01)"
  }
}

curl -X GET 'http://localhost:4434/identities?with_credentials&credentials.identifier=github:1423404' | jq '.'
[
  {
    "id": "f972706a-0632-4d02-8c4c-44ea9c888ee6",
    "schema_id": "default",
    "schema_url": "http://127.0.0.1:4433/schemas/default",
    "state": "active",
    "state_changed_at": "2021-09-07T13:06:59.015672Z",
    "traits": {
      "email": "foo@bar.com",
      "name": {}
    },
    "verifiable_addresses": [
      {
        "id": "65f28f44-39e7-404f-9e2e-541bf654dc70",
        "value": "foo@bar.com",
        "verified": false,
        "via": "email",
        "status": "sent",
        "verified_at": null,
        "created_at": "2021-09-07T13:06:59.024327Z",
        "updated_at": "2021-09-07T13:06:59.024327Z"
      }
    ],
    "recovery_addresses": [
      {
        "id": "2f52fe89-9596-4555-b0ac-2da06dacd2ff",
        "value": "foo@bar.com",
        "via": "email",
        "created_at": "2021-09-07T13:06:59.030167Z",
        "updated_at": "2021-09-07T13:06:59.030167Z"
      }
    ],
    "created_at": "2021-09-07T13:06:59.018277Z",
    "updated_at": "2021-09-07T13:06:59.018277Z"
  },
  {
    "id": "f972706a-0632-4d02-8c4c-44ea9c888ee6",
    "schema_id": "default",
    "schema_url": "http://127.0.0.1:4433/schemas/default",
    "state": "active",
    "state_changed_at": "2021-09-07T13:06:59.015672Z",
    "traits": {
      "email": "foo@bar.com",
      "name": {}
    },
    "verifiable_addresses": [
      {
        "id": "65f28f44-39e7-404f-9e2e-541bf654dc70",
        "value": "foo@bar.com",
        "verified": false,
        "via": "email",
        "status": "sent",
        "verified_at": null,
        "created_at": "2021-09-07T13:06:59.024327Z",
        "updated_at": "2021-09-07T13:06:59.024327Z"
      }
    ],
    "recovery_addresses": [
      {
        "id": "2f52fe89-9596-4555-b0ac-2da06dacd2ff",
        "value": "foo@bar.com",
        "via": "email",
        "created_at": "2021-09-07T13:06:59.030167Z",
        "updated_at": "2021-09-07T13:06:59.030167Z"
      }
    ],
    "created_at": "2021-09-07T13:06:59.018277Z",
    "updated_at": "2021-09-07T13:06:59.018277Z"
  }
]


curl -X GET 'http://localhost:4434/identities?credentials.identifier=github:1423404' | jq '.'
[
  {
    "id": "f972706a-0632-4d02-8c4c-44ea9c888ee6",
    "schema_id": "default",
    "schema_url": "http://127.0.0.1:4433/schemas/default",
    "state": "active",
    "state_changed_at": "2021-09-07T13:06:59.015672Z",
    "traits": {
      "email": "foo@bar.com",
      "name": {}
    },
    "verifiable_addresses": [
      {
        "id": "65f28f44-39e7-404f-9e2e-541bf654dc70",
        "value": "foo@bar.com",
        "verified": false,
        "via": "email",
        "status": "sent",
        "verified_at": null,
        "created_at": "2021-09-07T13:06:59.024327Z",
        "updated_at": "2021-09-07T13:06:59.024327Z"
      }
    ],
    "recovery_addresses": [
      {
        "id": "2f52fe89-9596-4555-b0ac-2da06dacd2ff",
        "value": "foo@bar.com",
        "via": "email",
        "created_at": "2021-09-07T13:06:59.030167Z",
        "updated_at": "2021-09-07T13:06:59.030167Z"
      }
    ],
    "created_at": "2021-09-07T13:06:59.018277Z",
    "updated_at": "2021-09-07T13:06:59.018277Z"
  },
  {
    "id": "f972706a-0632-4d02-8c4c-44ea9c888ee6",
    "schema_id": "default",
    "schema_url": "http://127.0.0.1:4433/schemas/default",
    "state": "active",
    "state_changed_at": "2021-09-07T13:06:59.015672Z",
    "traits": {
      "email": "foo@bar.com",
      "name": {}
    },
    "verifiable_addresses": [
      {
        "id": "65f28f44-39e7-404f-9e2e-541bf654dc70",
        "value": "foo@bar.com",
        "verified": false,
        "via": "email",
        "status": "sent",
        "verified_at": null,
        "created_at": "2021-09-07T13:06:59.024327Z",
        "updated_at": "2021-09-07T13:06:59.024327Z"
      }
    ],
    "recovery_addresses": [
      {
        "id": "2f52fe89-9596-4555-b0ac-2da06dacd2ff",
        "value": "foo@bar.com",
        "via": "email",
        "created_at": "2021-09-07T13:06:59.030167Z",
        "updated_at": "2021-09-07T13:06:59.030167Z"
      }
    ],
    "created_at": "2021-09-07T13:06:59.018277Z",
    "updated_at": "2021-09-07T13:06:59.018277Z"
  }
]
```

Certainly due to the pointer
We can let like this or I can search a way to reinit the pointer
what do you think ?

docs/docs/contributing.md
```diff
@@ -145,13 +145,12 @@ request, go through this checklist:
 1. Create a feature branch off of `master` so that changes do not get mixed up.
 1. [Rebase](http://git-scm.com/book/en/Git-Branching-Rebasing) your local
    changes against the `master` branch.
-1. Run the full project test suite with the `go test -tags sqlite ./...` (or
-   equivalent) command and confirm that it passes.
+1. Run the full project test suite with the `go test -tags sqlite,json1 ./...`
```
aeneasr, line 148: Can you please update the instructions in the README as well? :)

identity/identity.go
```diff
@@ -21,9 +21,7 @@ import (
 )
 
 // An Identity's State
-//
 // The state can either be `active` or `inactive`.
-//
```
aeneasr, line 26: Please re-add these newlines, they are important for doc generation :)

persistence/sql/persister_identity.go
```diff
@@ -472,3 +512,110 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
 	i.SchemaURL = s.SchemaURL(p.r.Config(ctx).SelfPublicURL(nil)).String()
 	return nil
 }
+
+func (p *Persister) getJsonSearchQuery(field string, values []string) pop.ScopeFunc {
+	return func(q *pop.Query) *pop.Query {
+		switch p.Connection(context.Background()).Dialect.Name() {
```
aeneasr, line 518: Please pass the context from the outer struct - never use context.Background outside of tests :)

persistence/sql/persister_identity.go
```diff
@@ -472,3 +512,110 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+		case "sqlite3", "mysql", "mariadb":
+			field, innerField := extractFieldAndInnerFields(field)
+			for _, value := range values {
+				if innerField == "" {
+					q = q.Where(fmt.Sprintf(`json_extract(%s, '$') = ?`, field), value)
```
aeneasr, line 523: This is vulnerable to SQL Injection. I saw that there is a sanitize function further down but I do not trust it and in the future developers might mess up the use. Best would be to allow only regex `[a-zA-Z0-9\.]+` here where \. is the delimiter and directly sanitize the field in the Sprintf statement.

persistence/sql/persister_identity.go
```diff
@@ -472,3 +512,110 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+		}
+		return q
+	}
+}
+
+func (p *Persister) searchCredentialQuery(field string, values []string) pop.ScopeFunc {
+	var innerField string
+	_, innerField = extractFieldAndInnerFields(field)
+	switch innerField {
+	case "type":
```
aeneasr, line 549: Prefix with „credential_“

persistence/sql/persister_identity.go
```diff
@@ -472,3 +512,110 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+		return func(q *pop.Query) *pop.Query { return q.Where("credential_types.name IN (?)", values) }
+	case "identifier":
```
aeneasr, line 551: Prefix with „credential_“

persistence/sql/persister_identity.go
```diff
@@ -472,3 +512,110 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+var quoteChar = map[string]uint8{
+	"cockroach": '"',
+	"mariadb":   '`',
+	"mysql":     '`',
+	"postgres":  '"',
+	"sqlite3":   '"',
+}
+
+func (p *Persister) Quote(ctx context.Context, key string) string {
+	n := p.Connection(ctx).Dialect.Name()
```
aeneasr, line 567: This is not good enough - there are so many ways to achieve SQL injection. It would also skip any single quote. Please use a strict AlphaNumeric + escaped dot regex here!

aeneasr (changes requested):
Looking pretty good but we need a bit more work to get where we want it to be :)

persistence/sql/persister_identity.go
david972: ok
I'll do it

persistence/sql/persister_identity.go
david972: and `_`

persistence/sql/persister_identity.go
aeneasr: nice :)

persistence/sql/persister_identity.go
david972: I was thinking about this.
credentials is already part of the request

persistence/sql/persister_identity.go
david972: done

aeneasr:
@david972 sorry I missed looking into this - is this good for 👀 ?

identity/pool.go
```diff
@@ -13,7 +11,7 @@ type (
 		ListIdentities(ctx context.Context, page, itemsPerPage int) ([]Identity, error)
 
 		// ListIdentitiesFiltered lists all identities in the store given the page and itemsPerPage.
-		ListIdentitiesFiltered(ctx context.Context, values url.Values, page, itemsPerPage int) ([]Identity, error)
+		ListIdentitiesFiltered(ctx context.Context, filters map[string][]string, page, itemsPerPage int) ([]Identity, error)
```
aeneasr, line 14: Filters should be explicit in my opinion

persistence/sql/persister_identity.go
```diff
@@ -253,73 +250,65 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
...
-	/* #nosec G201 TableName is static */
-	if err := sqlcon.HandleError(p.GetConnection(ctx).Where("nid = ?", corp.ContextualizeNID(ctx, p.nid)).
-		EagerPreload("VerifiableAddresses", "RecoveryAddresses").
-		Paginate(page, perPage).Order("id DESC").
-		All(&is)); err != nil {
-		return nil, err
-	}
-
-	schemaCache := map[string]string{}
+func (p *Persister) setIdentitySchemas(ctx context.Context, is []identity.Identity) error {
```
aeneasr, line 253: I extracted this into a common function because it was copy&pasted in the filter code

persistence/sql/persister_identity.go
```diff
@@ -253,73 +250,65 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
...
 
-func (p *Persister) ListIdentitiesFiltered(ctx context.Context, values url.Values, page, perPage int) ([]identity.Identity, error) {
-	is := make([]identity.Identity, 0, perPage)
+func (p *Persister) ListIdentities(ctx context.Context, page, perPage int) ([]identity.Identity, error) {
+	is := []identity.Identity{}
 
-//`[a-zA-Z0-9\.]+`
 	/* #nosec G201 TableName is static */
-	if err := sqlcon.HandleError(p.GetConnection(ctx).Where("identities.nid = ?", corp.ContextualizeNID(ctx, p.nid)).
-		LeftJoin("identity_verifiable_addresses verifiable_addresses", "verifiable_addresses.identity_id=identities.id").
```
aeneasr: `EagerPreload` should take care of these joins automatically.

persistence/sql/persister_identity.go
```diff
@@ -253,73 +250,65 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
...
-			schemaCache[i.SchemaID] = i.SchemaURL
-		}
+func (p *Persister) ListIdentitiesFiltered(ctx context.Context, filters map[string][]string, page, perPage int) ([]identity.Identity, error) {
+	is := []identity.Identity{}
 
-		is[k] = *i
+	if err := sqlcon.HandleError(p.GetConnection(ctx).Where("identities.nid = ?", corp.ContextualizeNID(ctx, p.nid)).
+		LeftJoin("identity_credentials ic", "identities.id=ic.identity_id").
+		LeftJoin("identity_credential_types credential_types", "credential_types.id=ic.identity_credential_type_id").
+		LeftJoin("identity_credential_identifiers credential_identifiers", "credential_identifiers.identity_credential_id=ic.id").
```
aeneasr, line 301: IMO we can always join these

persistence/sql/persister_identity.go
```diff
@@ -517,31 +507,36 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
 	return nil
 }
 
-func (p *Persister) getJsonSearchQuery(ctx context.Context, field string, values []string) pop.ScopeFunc {
+func (p *Persister) getJSONSearchQuery(ctx context.Context, field string, values []string) pop.ScopeFunc {
```
aeneasr, line 510: This code still had unfortunately several SQL injections.

Can you please write an extensive test suite for this? Without it we can not go forward and merge it. It needs to include fuzzing and SQL injection vectors such as `--` and `"` and `'` and other SQL injection entrypoints.

persistence/sql/persister_identity.go
```diff
@@ -559,66 +554,28 @@ func (p *Persister) searchCredentialQuery(field string, values []string) pop.Sco
...
-	for i, part := range parts {
-		part = strings.Trim(part, `"`)
-		part = strings.Trim(part, "`")
-		part = strings.TrimSpace(part)
-
-		parts[i] = fmt.Sprintf(`%c%v%c`, c, part, c)
-	}
-
-	return strings.Join(parts, ".")
-}
```
aeneasr: I wasn't sure where this code comes from. Looking at it, I am not very confident that it actually prevents SQL injection. What about other escape characters such as `'` or `--` or other unicode chars?

persistence/sql/persister_identity.go
```diff
@@ -474,3 +506,103 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+	case "type":
+		return func(q *pop.Query) *pop.Query { return q.Where("credential_types.name IN (?)", values) }
+	case "identifier":
+		return func(q *pop.Query) *pop.Query { return q.Where("credential_identifiers.identifier IN (?)", values) }
+	default:
+		return func(q *pop.Query) *pop.Query { return q }
+	}
+}
+
+func (p *Persister) buildIdentitySearchScope(ctx context.Context, filters map[string][]string) pop.ScopeFunc {
```
aeneasr, line 557: Please add tests and verify robustness against SQL injections

persistence/sql/persister_identity.go
```diff
@@ -474,3 +506,103 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+				continue
+			}
+
+			q = q.Where(fmt.Sprintf("%s IN (?)", field), values)
+		}
+		return q
+	}
+}
+
+func extractFieldAndInnerFields(field string) (string, string) {
```
aeneasr, line 585: Please add tests

persistence/sql/persister_identity.go
```diff
@@ -474,3 +506,103 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+var validValues = regexp.MustCompile(`[%]+`)
+
+func (p *Persister) validateValues(values []string) bool {
+	for _, value := range values {
+		if validValues.MatchString(value) {
+			return false
+		}
+	}
+	return true
+}
```
aeneasr, line 608: What was your intention with this function?

persistence/sql/persister_identity.go
```diff
@@ -249,37 +250,67 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
...
+	}
+
+	if err := p.setIdentitySchemas(ctx, is); err != nil {
+		return nil, err
+	}
+
+	return is, nil
+}
+
+func (p *Persister) ListIdentitiesFiltered(ctx context.Context, filters map[string][]string, page, perPage int) ([]identity.Identity, error) {
```
aeneasr, line 295: This function needs tests ensure that filters work and no SQL injections can happen.

aeneasr (changes requested):
Thank you again! I looked at the code. Unfortunately, there were still security issues with SQL injections. The rule is: Never, ever, use untrusted user input in raw SQL. The only time this *would* be fine is with a very strict and clear character set (e.g. a-zA-Z0-9) that can never possibly yield SQL injections.

We also need many more tests. Right now, the code only tests if SQLite works - but we also need to test MySQL and PostgreSQL, CockroachDB!

We also need to add tests which explicitly try to do SQL injections, and we need to prove that these queries are rejected.

Lastly, I think that we should add fuzzing to these SQL generator functions to ensure that no random input ever can cause SQL injections.

Thank you :)

aleksdikanski:
I would like to contribute to this branch as I have a use case where this would be pretty useful. Any way this would be possible?

david972:
> I would like to contribute to this branch as I have a use case where this would be pretty useful. Any way this would be possible?

I can give access to it

aeneasr:
> I would like to contribute to this branch as I have a use case where this would be pretty useful. Any way this would be possible?

Awesome, welcome to the ory contributor community :) To make sure that we're all pulling on the same rope, could you share a bit what you want to introduce? That will help us progress and merge faster :)

aleksdikanski:
> > I would like to contribute to this branch as I have a use case where this would be pretty useful. Any way this would be possible?
> 
> Awesome, welcome to the ory contributor community :) To make sure that we're all pulling on the same rope, could you share a bit what you want to introduce? That will help us progress and merge faster :)

Sure thing, I'm eager to test this search feature as I'm evaluating kratos for a use case at my current company. As the core search is already implemented I guess I would like to add some more tests for now. 
There are things like to be able to select only specific attributes of the identity (and / or it's traits) that might be good to have in the future, but I assume this goes more in the direction of having a query language for Rest APIs. I saw this is also discussed in other issues/MR so I would also move this to another discussion.

Besides this MR I'm just getting started with kratos (ory in general) and could also improve other tests and / or documentation etc.

david972:
> > > I would like to contribute to this branch as I have a use case where this would be pretty useful. Any way this would be possible?
> > 
> > 
> > Awesome, welcome to the ory contributor community :) To make sure that we're all pulling on the same rope, could you share a bit what you want to introduce? That will help us progress and merge faster :)
> 
> Sure thing, I'm eager to test this search feature as I'm evaluating kratos for a use case at my current company. As the core search is already implemented I guess I would like to add some more tests for now. There are things like to be able to select only specific attributes of the identity (and / or it's traits) that might be good to have in the future, but I assume this goes more in the direction of having a query language for Rest APIs. I saw this is also discussed in other issues/MR so I would also move this to another discussion.
> 
> Besides this MR I'm just getting started with kratos (ory in general) and could also improve other tests and / or documentation etc.

yes help is welcoming.

N.B. with this PR we can search in traits too

aeneasr:
Awesome, we can definitely need help with tests! One major area will be to ensure that no input can cause SQL injection. @david972 has done a really good job already with validation - there are many places where SQL statement generation based on user input happens, and we need to seal it air tight. :)

identity/handler.go
```diff
@@ -116,9 +103,12 @@ type adminListIdentities struct {
...
 //
-// Filter can be done by adding URL parameters
-// Filter can be done by any field from identityList
-// To filter by credential this flag `with_credential` as to be set
+// This endpoint can filter identities using URL Query parameters. You can filter by traits and by credentials, for example:
+//
+// - `?traits.email=bar@example.org&with_traits.name=foo`: Returns only identities with the email `bar@example.org` and name `foo`.
+// - `?traits.consent=true`: Returns identities whose `consent` property is `true`.
+// - `?credentials.type=oidc`: Returns identities who have an `oidc` credential attached.
+// - `?credentials.type=password&credentials.identifier=bar@example.org`: Returns identities who use `bar@example.org` to sign in.
```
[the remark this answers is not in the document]
aeneasr: Ping :) I updated the comment - does it reflect what you implemented?

identity/handler.go
```diff
@@ -118,7 +126,15 @@ type adminListIdentities struct {
 //       500: jsonError
 func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	page, itemsPerPage := x.ParsePagination(r)
-	is, err := h.r.IdentityPool().ListIdentities(r.Context(), page, itemsPerPage)
+
+	filters := map[string][]string{}
```
aeneasr, line 130: I think it would be better if we had a proper type for this! For example:

```
type Filter struct {
  Traits map[string]string
  // ...
}
```

identity/pool.go
aeneasr: With that I mean it should be a struct (see my other comment)

persistence/sql/persister_identity.go
aeneasr: @david972 this function needs several tests (you can write table tests)! The tests for this are located here: https://github.com/ory/kratos/blob/master/identity/test/pool.go

persistence/sql/persister_identity.go
aeneasr: Can you add tests with table tests that cover a variety of inputs (potentially using go fuzzing?) to ensure this never ends up in SQL injections?

persistence/sql/persister_identity.go
aeneasr: @david972 see above

aeneasr (changes requested):
I took another look :) There's still a few things we need to do:

1. [ ] Tests for the persister ensuring filter works properly
2. [ ] Tests for the helpers ensuring they escape SQL properly
3. [ ] Tests for the handler ensuring results are properly paginated & returned
4. [ ] Some documentation explaining how to search for identities
5. [ ] Make filters an explicit type - for example a struct, instead of a map

Once these are addressed, we can merge this!

Would you be up for these changes @david972 ? If not, I'll mark this PR as draft and maybe someone else can take over. Thank you!

david972:
> I took another look :) There's still a few things we need to do:
> 
> 1. [ ]  Tests for the persister ensuring filter works properly
> 2. [ ]  Tests for the helpers ensuring they escape SQL properly
> 3. [ ]  Tests for the handler ensuring results are properly paginated & returned
> 4. [ ]  Some documentation explaining how to search for identities
> 5. [ ]  Make filters an explicit type - for example a struct, instead of a map
> 
> Once these are addressed, we can merge this!
> 
> Would you be up for these changes @david972 ? If not, I'll mark this PR as draft and maybe someone else can take over. Thank you!

yes. I will
I posted an issue
I used this PR maybe something I did wrong here

david972:
Sorry
So much work on my side. 
I’ll try to get some time free on next week to finish both PR

aeneasr:
A new version of SQLite is out which now supports JSON syntax like Postgres. This will make this PR much easier!

david972:
> A new version of SQLite is out which now supports JSON syntax like Postgres. This will make this PR much easier!

Yes
BTW I changed the way and have rewritten the update token PR

identity/handler.go
```diff
@@ -118,7 +126,15 @@ type adminListIdentities struct {
...
+	for k, v := range r.URL.Query() {
+		if strings.HasPrefix(k, "traits.") || strings.HasPrefix(k,"credentials.") {
+			filters[k] = v
```
ianmobbs, line 133: If you have multiple values for the same parameter, this will overwrite the most recently found parameter. Since `filters` is of type `map[string][]string`, this should be something like:
```
if _, ok := filters[k]; !ok {
  filters[k] = []string{}
}

filters[k] = append(filters[k], v)
```

david972: 👍🏽

aeneasr:
See #2442

Issue #1419, opened 2021-06-14
Allow account recovery for identities without email address


andrewbanchich:
For instances where identities don't require an email address, it would be useful to still allow account recovery. Since it's just a link, an admin could create it and send it to the user themselves.

aeneasr:
https://www.ory.sh/kratos/docs/admin/managing-users-identities#invite-a-user

andrewbanchich:
@aeneasr I may be misunderstanding something, but the docs you linked to show how to recover and account *with* an email address. The docs also say [that is currently the only supported method](https://www.ory.sh/kratos/docs/self-service/flows/account-recovery#methods).

aeneasr:
Ah, my bad

aeneasr:
However, I believe it should work also if you don't have an email. It generates a link:

```
$ curl --request POST -sL \
    --header "Content-Type: application/json" \
    --request POST \
    --data '{
  "expires_in": "12h",
  "identity_id": "954f7f59-16a5-4152-8ce7-ad7c73bb124a"
}' \
    http://127.0.0.1:4434/recovery/link

{
  "expires_at": "2020-07-27T10:47:45.806Z",
  "recovery_link": "http://127.0.0.1:4433/self-service/browser/flows/recovery/link?request=8b6fd3e4-1de2-49bf-aa88-1a26634bf062\u0026token=b1tGmHf64cYDeHB9wKiuCF1FfycMJEyf"
}
```

Have you tried if this works?

andrewbanchich:
`POST http://localhost:4434/identities`
 
```json
{
    "schema_id": "default",
    "traits": {
        "username": "test"
    }
}
```
`GET http://localhost:4434/identities`

```json
[
    {
        "id": "36fe1677-83b6-4b44-87e3-c6dee4c2d160",
        "schema_id": "default",
        "schema_url": "https://mydomain.com/.ory/kratos/public/schemas/default",
        "traits": {
            "username": "test"
        }
    }
]
```

`POST http://localhost:4434/recovery/link`

```json
{
  "expires_in": "12h",
  "identity_id": "36fe1677-83b6-4b44-87e3-c6dee4c2d160"
}
```

returns: 

```json
{
    "error": {
        "code": 400,
        "status": "Bad Request",
        "reason": "The identity does not have any recovery addresses set.",
        "message": "The request was malformed or contained invalid parameters"
    }
}
```

aeneasr:
Ok, tracked as a feature :)

[...]

PR #1685, opened 2021-08-25
feat: #1419 - make value nullable for Recovery address

dibyajyotibehera:
## Checklist

- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/migrations/templates/20210825184606_recovery_addresses.down.fizz
+++ b/persistence/sql/migrations/templates/20210825184606_recovery_addresses.down.fizz
@@ -1,0 +1,1 @@
+change_column("identity_recovery_addresses", "value", "string", {"size": 400,"null": false})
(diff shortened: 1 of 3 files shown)


persistence/sql/migrations/templates/20210825184606_recovery_addresses.down.fizz
```diff
@@ -0,0 +1 @@
+change_column("identity_recovery_addresses", "value", "string", {"size": 400,"null": false})
```
aeneasr, line 1: Fizz has a bug where `null: false` doesn't work, see https://github.com/gobuffalo/fizz/issues/62

```suggestion
change_column("identity_recovery_addresses", "value", "string", {"size": 400})
```

aeneasr (commented):
just fyi :)

aeneasr:
As per discussion on slack, I believe this is not the best approach to solve the recovery without email issue. Should we keep this open or do you want to close it?

dibyajyotibehera:
Yes i will close it now

Issue #1699, opened 2021-08-30
Endpoint for fetching all schema IDs or all schemas


jld3103:
**Is your feature request related to a problem? Please describe.**

I want to create a user dynamically without hardcoding which schemas the kratos instance has available.

**Describe the solution you'd like**

An endpoint to either fetch all schema IDs or all schemas directly.

aeneasr:
Makes sense!

PR #1703, opened 2021-08-30
feat: Add all json schemas endpoint

jld3103:
This PR implements an endpoint to list all Json schemas

## Related issue(s)
https://github.com/ory/kratos/issues/1699

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/schema/handler.go
+++ b/schema/handler.go
@@ -4,10 +4,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 
 	"github.com/ory/kratos/driver/config"
+	"github.com/ory/x/urlx"
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/pkg/errors"
@@ -41,7 +43,8 @@ const SchemasPath string = "schemas"
 
 func (h *Handler) RegisterPublicRoutes(public *x.RouterPublic) {
 	h.r.CSRFHandler().IgnoreGlobs(fmt.Sprintf("/%s/*", SchemasPath))
-	public.GET(fmt.Sprintf("/%s/:id", SchemasPath), h.get)
+	public.GET(fmt.Sprintf("/%s/:id", SchemasPath), h.getByID)
+	public.GET(fmt.Sprintf("/%s", SchemasPath), h.getAll)
 }
 
 func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
@@ -77,34 +80,134 @@ type getJsonSchema struct {
 //       200: jsonSchema
 //       404: jsonError
 //       500: jsonError
-func (h *Handler) get(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+func (h *Handler) getByID(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	s, err := h.r.IdentityTraitsSchemas(r.Context()).GetByID(ps.ByName("id"))
 	if err != nil {
 		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrNotFound.WithDebugf("%+v", err)))
 		return
 	}
-	var src io.ReadCloser
 
-	if s.URL.Scheme == "file" {
-		src, err = os.Open(s.URL.Host + s.URL.Path)
-		if err != nil {
-			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
-			return
-		}
-		defer src.Close()
-	} else {
-		resp, err := http.Get(s.URL.String())
-		if err != nil {
-			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
-			return
-		}
-		defer resp.Body.Close()
-		src = resp.Body
+	src, err := ReadSchema(s)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+		return
 	}
+	defer src.Close()
 
 	w.Header().Add("Content-Type", "application/json")
 	if _, err := io.Copy(w, src); err != nil {
 		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
 		return
 	}
+}
+
+// Raw JSON Schema list
+//
+// swagger:model JSONSchemas
+type JSONSchemas []identitySchema
+
+type identitySchema struct {
+	ID     string          `json:"id"`
+	Schema json.RawMessage `json:"schema"`
+}
+
+// nolint:deadcode,unused
+// swagger:parameters getJsonSchemas
+type getJsonSchemas struct {
+	// Items per Page
+	//
+	// This is the number of items per page.
+	//
+	// required: false
+	// in: query
+	// default: 100
+	// min: 1
+	// max: 500
+	PerPage int `json:"per_page"`
+
+	// Pagination Page
+	//
+	// required: false
+	// in: query
+	// default: 0
+	// min: 0
+	Page int `json:"page"`
+}
+
+// swagger:route GET /schemas v0alpha1 getJsonSchemas
+//
+// Get all JSON Schemas
+//
+//     Produces:
+//     - application/json
+//
+//     Schemes: http, https
+//
+//     Responses:
+//       200: JSONSchemas
+//       500: jsonError
+func (h *Handler) getAll(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	page, itemsPerPage := x.ParsePagination(r)
+
+	schemas := h.r.IdentityTraitsSchemas(r.Context()).List(page, itemsPerPage)
+	total := h.r.IdentityTraitsSchemas(r.Context()).Total()
+
+	var ss JSONSchemas
+
+	for _, schema := range schemas {
+		s, err := h.r.IdentityTraitsSchemas(r.Context()).GetByID(schema.ID)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrNotFound.WithDebugf("%+v", err)))
+			return
+		}
+
+		src, err := ReadSchema(s)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+			return
+		}
+		defer src.Close()
+
+		raw, err := ioutil.ReadAll(src)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+			return
+		}
+
+		var p json.RawMessage
+		err = json.Unmarshal(raw, &p)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+			return
+		}
+
+		ss = append(ss, identitySchema{
+			ID:     s.ID,
+			Schema: p,
+		})
+	}
+
+	x.PaginationHeader(w, urlx.AppendPaths(h.r.Config(r.Context()).SelfPublicURL(r), fmt.Sprintf("/%s", SchemasPath)), int64(total), page, itemsPerPage)
+
+	w.Header().Add("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(ss); err != nil {
+		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+		return
+	}
+}
+
+func ReadSchema(schema *Schema) (src io.ReadCloser, err error) {
+	if schema.URL.Scheme == "file" {
+		src, err = os.Open(schema.URL.Host + schema.URL.Path)
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+	} else {
+		resp, err := http.Get(schema.URL.String())
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		src = resp.Body
+	}
+	return src, nil
 }
--- a/schema/handler.go
+++ b/schema/handler.go
@@ -101,19 +101,22 @@ func (h *Handler) getByID(w http.ResponseWriter, r *http.Request, ps httprouter.
 	}
 }
 
-// Raw JSON Schema list
+// Raw identity Schema list
 //
-// swagger:model JSONSchemas
-type JSONSchemas []identitySchema
+// swagger:model IdentitySchemas
+type IdentitySchemas []identitySchema
 
+// swagger:model identitySchema
 type identitySchema struct {
-	ID     string          `json:"id"`
+	// The ID of the Identity JSON Schema
+	ID string `json:"id"`
+	// The actual Identity JSON Schema
 	Schema json.RawMessage `json:"schema"`
 }
 
 // nolint:deadcode,unused
-// swagger:parameters getJsonSchemas
-type getJsonSchemas struct {
+// swagger:parameters listIdentitySchemas
+type listIdentitySchemas struct {
 	// Items per Page
 	//
 	// This is the number of items per page.
@@ -134,9 +137,9 @@ type getJsonSchemas struct {
 	Page int `json:"page"`
 }
 
-// swagger:route GET /schemas v0alpha1 getJsonSchemas
+// swagger:route GET /schemas v0alpha2 listIdentitySchemas
 //
-// Get all JSON Schemas
+// Get all Identity Schemas
 //
 //     Produces:
 //     - application/json
@@ -144,7 +147,7 @@ type getJsonSchemas struct {
 //     Schemes: http, https
 //
 //     Responses:
-//       200: JSONSchemas
+//       200: IdentitySchemas
 //       500: jsonError
 func (h *Handler) getAll(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	page, itemsPerPage := x.ParsePagination(r)
@@ -152,7 +155,7 @@ func (h *Handler) getAll(w http.ResponseWriter, r *http.Request, ps httprouter.P
 	schemas := h.r.IdentityTraitsSchemas(r.Context()).List(page, itemsPerPage)
 	total := h.r.IdentityTraitsSchemas(r.Context()).Total()
 
-	var ss JSONSchemas
+	var ss IdentitySchemas
 
 	for _, schema := range schemas {
 		s, err := h.r.IdentityTraitsSchemas(r.Context()).GetByID(schema.ID)
--- a/schema/handler.go
+++ b/schema/handler.go
@@ -103,7 +103,7 @@ func (h *Handler) getByID(w http.ResponseWriter, r *http.Request, ps httprouter.
 
 // Raw identity Schema list
 //
-// swagger:model IdentitySchemas
+// swagger:model identitySchemas
 type IdentitySchemas []identitySchema
 
 // swagger:model identitySchema
@@ -147,7 +147,7 @@ type listIdentitySchemas struct {
 //     Schemes: http, https
 //
 //     Responses:
-//       200: IdentitySchemas
+//       200: identitySchemas
 //       500: jsonError
 func (h *Handler) getAll(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	page, itemsPerPage := x.ParsePagination(r)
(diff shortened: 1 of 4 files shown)


jld3103:
I'm not sure which docs I'm supposed to update and also if I should commit the generated code.

schema/handler.go
```diff
@@ -108,3 +110,39 @@ func (h *Handler) get(w http.ResponseWriter, r *http.Request, ps httprouter.Para
...
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrNotFound.WithDebugf("%+v", err)))
+		return
+	}
+
+	w.Header().Add("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(ids); err != nil {
+		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+		return
+	}
```
aeneasr, line 147: Please use `h.r.Writer().Write` instead :)

schema/handler.go
```diff
@@ -108,3 +110,39 @@ func (h *Handler) get(w http.ResponseWriter, r *http.Request, ps httprouter.Para
...
+//     Produces:
+//     - application/json
+//
+//     Schemes: http, https
+//
+//     Responses:
+//       200: jsonSchemaIDs
+//       500: jsonError
+func (h *Handler) getIDs(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	ids, err := h.r.IdentityTraitsSchemas(r.Context()).GetIDs()
```
aeneasr, line 137: I think it would make more sense to just return the complete schemas instead, and add pagination?

aeneasr (changes requested):

schema/handler.go
jld3103: Yes, but that approach is much more effort and I don't know enough about the architecture of kratos to do it. I also just need a solution that works and not something fancy

aeneasr: Ok, I understand - but Ory Kratos is used by thousands and we need solutions that work in general and not just for individuals. As it is right now, I feat that we can not accept it. Adding the required changes however shouldn't be too hard. We have helpers for pagination:

https://github.com/ory/kratos/blob/c2c5a588a4f0e6a1876ee7cfec772e2499120ebe/identity/handler.go#L111-L112

If you don't want to work on the suggestions I fear we need to close this PR

jld3103: I can try doing it. It will just take me a bit longer than someone who knows kratos better.

jld3103:
@aeneasr i hope this is to your liking

aeneasr:
Thank you! It looks like the tests fail with a panic right now:

https://app.circleci.com/pipelines/github/ory/kratos/4621/workflows/b18b066e-0dab-4771-8311-79846e0692fe/jobs/25049

Could you please take a look? :)

jld3103:
Uh the endpoint returns the same schema multiple times. I need to fix that.

jld3103:
I think the implementation should be fixed now. I also improved the test, because the bug slipped past it. Sorry for the many force pushes :(

jld3103:
The test fail seems completely unrelated to this PR

schema/schema.go
```diff
@@ -38,6 +38,16 @@ func (s Schemas) GetByID(id string) (*Schema, error) {
...
 
+func (s Schemas) GetIDs() ([]string, error) {
+	ids := []string{}
+
+	for _, ss := range s {
+		ids = append(ids, ss.ID)
+	}
+
+	return ids, nil
+}
```
zepatrik, line 49: Instead of building a list of all schemas here and then slicing it later on, I think it makes more sense to paginate here directly and return what is needed.
```suggestion
func (s Schemas) List(page, perPage int) Schemas {
	ss := make(Schemas, 0, perPage)

	// TODO add guard checks
	copy(s[page*perPage, (page+1)*perPage], ss)

	return ss
}
```

schema/handler.go
```diff
@@ -77,34 +80,140 @@ type getJsonSchema struct {
...
+
+	enc := json.NewEncoder(w)
+	enc.SetIndent("", "  ")
+	if err := enc.Encode(ss); err != nil {
 		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
 		return
 	}
 }
+
+func readSchema(schema *Schema) (json.RawMessage, error) {
```
zepatrik, line 193: Maybe instead of JSON decoding and encoding, you can instead return the `io.ReadCloser` here and then use `io.Copy` as before?

jld3103:
@aeneasr i think there is a but in the admin list identities endpoint. requesting page 0 and page 1 both return the same results. In this endpoint pages start at 0 and the bug is not present. this should be consistent though and the spec says pages start at 0, so the admin list identities endpoint needs to be fixed.

schema/handler.go
```diff
@@ -77,34 +80,140 @@ type getJsonSchema struct {
...
+		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+		return
+	}
+}
+
+// Raw JSON Schema map
+//
+// swagger:model jsonSchemas
+// nolint:deadcode,unused
+type jsonSchemas map[string]json.RawMessage
```
zepatrik, line 109: It makes sense from a data structure perspective to return a map of schema ID to schema, but that will mean that requesting the same page will include the same schemas but in a different order. I know that JSON objects have no order, but I would expect a list API to return something ordered. Maybe make this an array instead? Not sure though, just a thought.

zepatrik (commented):
Thanks for updating the PR :+1:
I have some ideas how to improve the code even further :wink:

schema/handler.go
jld3103: I see your point, but i chose a map because the schema json doesn't include the id of the schema itself. I don't know if there is a solution to this problem

zepatrik: You could also return
```json
[{"id": "default", "schema": {...}}, {"id": "other", "schema": {...}}]
```

jld3103: Yeah I'll do it that way

schema/schema.go
jld3103: with the guard checks you mean checking that it doesn't exceed the boundaries of the slice that is copied from?

jld3103:
@zepatrik i applied all your suggestions

schema/schema.go
zepatrik: Yes, exactly what you did already :wink:

schema/schema_test.go
```diff
@@ -70,6 +70,37 @@ func TestSchemas_GetByID(t *testing.T) {
...
+			ids_orig = append(ids_orig, s.ID)
+		}
+		ids_list := []string{}
+		for _, s := range []Schema(ss.List(0, 4)) {
+			ids_list = append(ids_list, s.ID)
+		}
+		for _, id := range ids_orig {
+			require.Contains(t, ids_list, id)
+		}
+	})
```
zepatrik, line 101: This test looks a bit weird to me, probably because you previously only returned the IDs?
What do you think about:

```suggestion
	t.Run("case=get all schemas", func(t *testing.T) {
		p0 := ss.List(0, 4)
		assert.Equal(t, ss, p0)
	})
	
	t.Run("case=smaller pages", func(t *testing.T) {
		p0, p1 := ss.List(0, 2), ss.List(2, 2)
		assert.Equal(t, ss, append(p0, p1...))
	}
	
	t.Run("case=indexes out of range", func(t *testing.T) {
		p0 := ss.List(-1, 10)
		assert.Equal(t, ss, p0)
	}
```

and maybe there is some other case I missed?

schema/handler_test.go
```diff
@@ -126,3 +127,97 @@ func TestHandler(t *testing.T) {
...
+		return body
+	}
+
+	getFromFS := func(id string) string {
+		f, err := os.Open(strings.TrimPrefix(getSchemaById(id).RawURL, "file://"))
+		require.NoError(t, err)
+		raw, err := ioutil.ReadAll(f)
+		require.NoError(t, err)
+		require.NoError(t, f.Close())
+		return string(raw)
```
zepatrik, line 174:
```suggestion
		raw, err := os.ReadFile(strings.TrimPrefix(getSchemaById(id).RawURL, "file://"))
		require.NoError(t, err)
		return string(raw)
```

schema/handler.go
```diff
@@ -77,34 +81,140 @@ type getJsonSchema struct {
...
+}
+
+// Raw JSON Schema list
+//
+// swagger:model jsonSchemas
+// nolint:deadcode,unused
+type jsonSchemas []struct {
+	ID     string          `json:"id"`
+	Schema json.RawMessage `json:"schema"`
+}
```
zepatrik, line 112: Can you define the struct from this type as it's own type? You copy the type definition multiple times, so it would be easier to make this it's own type.
```suggestion
type JSONSchemas []jsonSchema

type jsonSchema struct {
	ID     string          `json:"id"`
	Schema json.RawMessage `json:"schema"`
}
```

schema/handler.go
```diff
@@ -77,34 +81,140 @@ type getJsonSchema struct {
...
+		err = json.Unmarshal(raw, &p)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+			return
+		}
+
+		ss = append(ss, struct {
+			ID     string          `json:"id"`
+			Schema json.RawMessage `json:"schema"`
+		}{
```
zepatrik, line 187:
```suggestion
		ss = append(ss, jsonSchema{
```

schema/handler_test.go
```diff
@@ -126,3 +127,97 @@ func TestHandler(t *testing.T) {
...
+	conf.MustSet(config.ViperKeyDefaultIdentitySchemaURL, getSchemaById(config.DefaultIdentityTraitsSchemaID).RawURL)
+	conf.MustSet(config.ViperKeyIdentitySchemas, schemasConfig)
+
+	t.Run("case=get all schemas", func(t *testing.T) {
+		body := getFromTS(http.StatusOK)
+
+		var result []struct {
+			ID     string          `json:"id"`
+			Schema json.RawMessage `json:"schema"`
+		}
```
zepatrik, line 197:
```suggestion
		var result schema.JSONSchemas
```

schema/handler.go
```diff
@@ -77,34 +81,140 @@ type getJsonSchema struct {
...
+//     Responses:
+//       200: jsonSchemas
+//       500: jsonError
+func (h *Handler) getAll(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	page, itemsPerPage := x.ParsePagination(r)
+
+	schemas := h.r.IdentityTraitsSchemas(r.Context()).List(page, itemsPerPage)
+	total := h.r.IdentityTraitsSchemas(r.Context()).Total()
+
+	var ss jsonSchemas
```
zepatrik, line 155:
```suggestion
	var ss JSONSchemas
```

schema/handler.go
```diff
@@ -77,34 +81,140 @@ type getJsonSchema struct {
...
+	if _, err := io.Copy(w, src); err != nil {
+		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+		return
+	}
+}
+
+// Raw JSON Schema list
+//
+// swagger:model jsonSchemas
+// nolint:deadcode,unused
```
zepatrik, line 108:
```suggestion
```

schema/handler_test.go
```diff
@@ -126,3 +127,97 @@ func TestHandler(t *testing.T) {
...
+		for _, id := range ids_orig {
+			require.Contains(t, ids_list, id)
+		}
+
+		for _, s := range schemas {
+			for _, r := range result {
+				if r.ID == s.ID {
+					p, err := json.MarshalIndent(r.Schema, "", "  ")
+					require.NoError(t, err)
+					require.EqualValues(t, getFromFS(s.ID), string(p)+"\n")
```
zepatrik, line 217: https://pkg.go.dev/github.com/stretchr/testify@v1.7.0/assert#JSONEq
```suggestion
					assert.JSONEq(t, getFromFS(s.ID), string(r.Schema))
```

schema/handler_test.go
```diff
@@ -126,3 +127,97 @@ func TestHandler(t *testing.T) {
...
+				}
+			}
+
```
zepatrik, line 220: I would like to also have a case with pagination. No edge cases, just general case that ensures pagination is actually applied.

zepatrik (changes requested):
Nice, this is going in a very good direction :wink: 
Just some final details, then it should be good to merge :+1:

schema/handler.go
jld3103: The problem I see is that `jsonSchema` is already defined as `json.RawMessage`. Changing this would be a breaking change, but I think it is better then.

jld3103: Sounds good

jld3103:
I hope this is all good now

schema/handler_test.go
```diff
@@ -98,31 +97,138 @@ func TestHandler(t *testing.T) {
...
+		require.EqualValues(t, expectCode, res.StatusCode, "%s", body)
+		return body
+	}
+
+	getFromFS := func(id string) []byte {
+		raw, err := os.ReadFile(strings.TrimPrefix(getSchemaById(id).RawURL, "file://"))
+		require.NoError(t, err)
+		return raw
+	}
```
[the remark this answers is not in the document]
zepatrik: Actually, I would prefer using `t.Run` so you can even share the registry and stuff, makes it a bit faster.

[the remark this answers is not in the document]
zepatrik: Which reminds me: in case one of the schemas is not found, should the list API omit it or return 404? Not sure what would be the best approach here.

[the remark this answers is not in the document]
jld3103: Yeah the problem was the api only returns the schemas that actually exist, so the test can't compare it to the list where also non existent schemas are listed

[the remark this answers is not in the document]
zepatrik: But IRL it might happen that because of misconfiguration some old, unused schema is still around that is not resolvable. That's why I am bringing up the question how the API should behave :wink: 
You can also set config values mutliple times btw, it is hot reloaded. So you can set different schemas for each subtest.

schema/handler.go
```diff
@@ -77,34 +81,134 @@ type getJsonSchema struct {
...
+			return
+		}
+
+		ss = append(ss, identitySchema{
+			ID:     s.ID,
+			Schema: p,
+		})
 	}
 
+	x.PaginationHeader(w, urlx.AppendPaths(h.r.Config(r.Context()).SelfAdminURL(), fmt.Sprintf("/%s", SchemasPath)), int64(total), page, itemsPerPage)
```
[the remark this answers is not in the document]
aeneasr: I think it might be easier to expose it on the public port only as there is really no need to have redundancy.

jld3103:
@zepatrik  i don't know how the hot-reloading of the config should work, can you give me a hint?

zepatrik:
What you did should work already, isn't it?

jld3103:
yes it works, but it's a bit hacky

jld3103:
circleci test fails because of something unrelated with goreleaser

zepatrik:
You have to merge/rebase to current master :wink: https://github.com/ory/kratos/pull/1730
Btw, you don't have to force push all the time, we squash the commits in the end anyway. You can push later changes as individual commits.

jld3103:
Ok :) I wasn't sure what the merge strategy is, so I didn't want to spam the git log

schema/handler_test.go
```diff
@@ -61,68 +63,177 @@ func TestHandler(t *testing.T) {
...
+		for _, s := range schemas {
+			if s.ID != config.DefaultIdentityTraitsSchemaID {
+				schemasConfig = append(schemasConfig, config.Schema{
+					ID:  s.ID,
+					URL: s.RawURL,
+				})
+			}
 		}
+		return schemasConfig
 	}
```
zepatrik, line 102: Maybe make it less hacky by
```suggestion
	setSchemas := func(newSchemas schema.Schemas) () {
		schemas = newSchemas
		schemasConfig := make([]config.Schema)
		for _, s := range schemas {
			if s.ID != config.DefaultIdentityTraitsSchemaID {
				schemasConfig = append(schemasConfig, config.Schema{
					ID:  s.ID,
					URL: s.RawURL,
				})
			}
		}
		conf.MustSet(config.ViperKeyIdentitySchemas, schemasConfig)
	}
```

jld3103:
the failed test to me seems unrelated to these changes

jld3103:
@zepatrik i fixed the coverage as you suggested. the remaining uncovered lines are all error handling. i don't know if that is something that should be covered, because it wasn't covered before too.

zepatrik:
That is fine, only errors that are expected to occur during normal usage have to be covered. These are not expected errors.

aeneasr:
@zepatrik green light?

schema/handler.go
```diff
@@ -77,34 +80,134 @@ type getJsonSchema struct {
...
+	w.Header().Add("Content-Type", "application/json")
+	if _, err := io.Copy(w, src); err != nil {
+		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+		return
+	}
+}
+
+// Raw JSON Schema list
+//
+// swagger:model JSONSchemas
```
aeneasr, line 106:
```suggestion
// swagger:model identitySchemas
```

schema/handler.go
```diff
@@ -77,34 +80,134 @@ type getJsonSchema struct {
...
+type JSONSchemas []identitySchema
+
+type identitySchema struct {
+	ID     string          `json:"id"`
+	Schema json.RawMessage `json:"schema"`
+}
```
aeneasr, line 112:
```suggestion

// swagger:model identitySchema
type identitySchema struct {
	// The ID of the Identity JSON Schema
	ID     string          `json:"id"`
	// The actual Identity JSON Schema
	Schema json.RawMessage `json:"schema"`
}
```

schema/handler.go
```diff
@@ -77,34 +80,134 @@ type getJsonSchema struct {
...
+
+// nolint:deadcode,unused
+// swagger:parameters getJsonSchemas
```
aeneasr, line 115:
```suggestion
// swagger:parameters listIdentitySchemas
```

schema/handler.go
```diff
@@ -77,34 +80,134 @@ type getJsonSchema struct {
...
+	// Pagination Page
+	//
+	// required: false
+	// in: query
+	// default: 0
+	// min: 0
+	Page int `json:"page"`
+}
+
+// swagger:route GET /schemas v0alpha1 getJsonSchemas
```
aeneasr, line 137:
```suggestion
// swagger:route GET /schemas v0alpha2 listIdentitySchemas
```

schema/handler.go
```diff
@@ -77,34 +80,134 @@ type getJsonSchema struct {
...
+//
+// Get all JSON Schemas
+//
+//     Produces:
+//     - application/json
+//
+//     Schemes: http, https
+//
+//     Responses:
+//       200: JSONSchemas
```
aeneasr, line 147:
```suggestion
//       200: identitySchemas
```

aeneasr (changes requested):
Just some swagger stuff :)

zepatrik (approved):
Nice, looks good to me :wink: Thanks for all the hard work :+1:

jld3103:
Thanks for all your nice comments :) It was a pleasure to work with so friendly developers

jld3103:
When will this be merged/released?

zepatrik:
I am waiting for @aeneasr to approve.

schema/handler.go
```diff
@@ -77,34 +80,137 @@ type getJsonSchema struct {
...
+	w.Header().Add("Content-Type", "application/json")
+	if _, err := io.Copy(w, src); err != nil {
+		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+		return
+	}
+}
+
+// Raw identity Schema list
+//
+// swagger:model IdentitySchemas
```
aeneasr, line 106:
```suggestion
// swagger:model identitySchemas
```

schema/handler.go
```diff
@@ -77,34 +80,137 @@ type getJsonSchema struct {
...
+//
+// Get all Identity Schemas
+//
+//     Produces:
+//     - application/json
+//
+//     Schemes: http, https
+//
+//     Responses:
+//       200: IdentitySchemas
```
aeneasr, line 150:
```suggestion
//       200: identitySchemas
```

schema/handler.go
```diff
@@ -77,34 +80,137 @@ type getJsonSchema struct {
...
+			ID:     s.ID,
+			Schema: p,
+		})
 	}
 
+	x.PaginationHeader(w, urlx.AppendPaths(h.r.Config(r.Context()).SelfPublicURL(r), fmt.Sprintf("/%s", SchemasPath)), int64(total), page, itemsPerPage)
+
 	w.Header().Add("Content-Type", "application/json")
-	if _, err := io.Copy(w, src); err != nil {
+	if err := json.NewEncoder(w).Encode(ss); err != nil {
```
aeneasr, line 196: Please use `h.d.Writer().Write` so that the correct content type etc is used :)

schema/handler.go
```diff
@@ -77,34 +80,137 @@ type getJsonSchema struct {
...
 		}
-		defer resp.Body.Close()
-		src = resp.Body
+
+		var p json.RawMessage
+		err = json.Unmarshal(raw, &p)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The file for this JSON Schema ID could not be found or opened. This is a configuration issue.").WithDebugf("%+v", err)))
+			return
+		}
```
aeneasr, line 185: You don’t need to unmarshall. `json.RawMessage` is just `[]byte` so no need to decode here, just assign `Schema: raw` :)

schema/schema.go
```diff
@@ -38,6 +38,25 @@ func (s Schemas) GetByID(id string) (*Schema, error) {
 	return nil, errors.WithStack(herodot.ErrBadRequest.WithReasonf("Unable to find JSON Schema ID: %s", id))
 }
 
+func (s Schemas) Total() int {
+	return len(s)
+}
+
+func (s Schemas) List(page, perPage int) Schemas {
+	if page < 0 {
```
aeneasr, line 46: The pagination module has a helper for this :)

https://pkg.go.dev/github.com/ory/x/pagination#Index

aeneasr (changes requested):
Thank you ! Looking great and also awesome tests 😎 You’re almost there!

schema/schema.go
jld3103: Can you give me a hint on how to do it? I'm a bit confused by the function/how I should use it.

jld3103:
@aeneasr anything missing for the merge?

aeneasr:
Just got back from holidays. Will review as soon as possible!

aeneasr (approved):

aeneasr (approved):

PR #1736, opened 2021-09-07
fix: omitempty for VerifiedAt and StateChangedAt

AntiSC2:
Currently the Python SDK is unusable because the spec doesn't match what Kratos actually returns. OpenAPI expects that Kratos only returns date-time strings for `state_changed_at` and `verified_at`. I think the original idea was that if `verified_at` was null it would be omitted entirely. The problem is that `omitempty` doesn't work for structs in Go, only pointers to structs. This means that `verified_at` is returned as null anyways from the API.

This PR changes VerifiedAt and StateChangedAt into pointers so that omitempty works as expected.

## Related issue(s)

https://github.com/ory/sdk/issues/95

## Checklist

- [X] I am following the
      the changes.
      works.

### DIFF
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -220,7 +220,8 @@ func (h *Handler) create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
 		return
 	}
 
-	i := &Identity{SchemaID: cr.SchemaID, Traits: []byte(cr.Traits), State: StateActive, StateChangedAt: sqlxx.NullTime(time.Now())}
+	stateChangedAt := sqlxx.NullTime(time.Now())
+	i := &Identity{SchemaID: cr.SchemaID, Traits: []byte(cr.Traits), State: StateActive, StateChangedAt: &stateChangedAt}
 	if err := h.r.IdentityManager().Create(r.Context(), i); err != nil {
 		h.r.Writer().WriteError(w, r, err)
 		return
@@ -318,8 +319,10 @@ func (h *Handler) update(w http.ResponseWriter, r *http.Request, ps httprouter.P
 			h.r.Writer().WriteError(w, r, herodot.ErrBadRequest.WithReasonf("%s", err).WithWrap(err))
 		}
 
+		stateChangedAt := sqlxx.NullTime(time.Now())
+
 		identity.State = ur.State
-		identity.StateChangedAt = sqlxx.NullTime(time.Now())
+		identity.StateChangedAt = &stateChangedAt
 	}
 
 	identity.Traits = []byte(ur.Traits)
--- a/identity/handler_test.go
+++ b/identity/handler_test.go
@@ -164,10 +164,11 @@ func TestHandler(t *testing.T) {
 			for name, ts := range map[string]*httptest.Server{"public": publicTS, "admin": adminTS} {
 				t.Run("endpoint="+name, func(t *testing.T) {
 					res := send(t, ts, "POST", "/identities", http.StatusCreated, json.RawMessage(`{"traits": {"bar":"baz"}}`))
+					stateChangedAt := sqlxx.NullTime(res.Get("state_changed_at").Time())
 
 					i.Traits = []byte(res.Get("traits").Raw)
 					i.ID = x.ParseUUID(res.Get("id").String())
-					i.StateChangedAt = sqlxx.NullTime(res.Get("state_changed_at").Time())
+					i.StateChangedAt = &stateChangedAt
 					assert.NotEmpty(t, res.Get("id").String())
 
 					assert.EqualValues(t, "baz", res.Get("traits.bar").String(), "%s", res.Raw)
--- a/identity/identity.go
+++ b/identity/identity.go
@@ -77,7 +77,7 @@ type Identity struct {
 	State State `json:"state" faker:"-" db:"state"`
 
 	// StateChangedAt contains the last time when the identity's state changed.
-	StateChangedAt sqlxx.NullTime `json:"state_changed_at" faker:"-" db:"state_changed_at"`
+	StateChangedAt *sqlxx.NullTime `json:"state_changed_at,omitempty" faker:"-" db:"state_changed_at"`
 
 	// Traits represent an identity's traits. The identity is able to create, modify, and delete traits
 	// in a self-service manner. The input will always be validated against the JSON Schema defined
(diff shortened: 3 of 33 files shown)


identity/identity_verification.go
```diff
@@ -68,7 +68,7 @@ type VerifiableAddress struct {
 	//
 	// example: 2014-01-01T23:28:56.782Z
 	// required: false
-	VerifiedAt sqlxx.NullTime `json:"verified_at,omitempty" faker:"-" db:"verified_at"`
+	VerifiedAt sqlxx.NullTime `json:"verified_at" faker:"-" db:"verified_at"`
```
aeneasr, line 71: We could also try making this a pointer instead?

internal/httpclient/api/openapi.yaml
```diff
@@ -2893,6 +2893,7 @@ components:
       format: date-time
       title: NullTime implements sql.NullTime functionality.
       type: string
+      nullable: true
```
aeneasr, line 2896: You will need to do this somewhere else as this is autogenerated. In particular, it will probably require a patch along the lines of this one:

https://github.com/ory/kratos/blob/4f05d64ea8e04feb2791cf83b2296e825e3bd8ac/.schema/openapi/patches/schema.yaml#L15-L17

aeneasr (changes requested):
Thank you! I think a better approach would be to try and return `nil` `isZero` is true here:

https://github.com/ory/x/blob/a065f5780b6040793599f5d9fd68a246322b56ab/sqlxx/types.go#L104

If a test is added to ensure it behaves as expected, and it works, it would mean that this is fixed for all fields immediately. I recommend using the "gjson" library in the test.

AntiSC2:
> Thank you! I think a better approach would be to try and return `nil` `isZero` is true here:
> 
> https://github.com/ory/x/blob/a065f5780b6040793599f5d9fd68a246322b56ab/sqlxx/types.go#L104
> 
> If a test is added to ensure it behaves as expected, and it works, it would mean that this is fixed for all fields immediately. I recommend using the "gjson" library in the test.

Maybe I'm not understanding it correctly but doesn't the code already return nil there (more specifically, `json.Marshal(nil)`). As far as I understand the omitempty option only works for things that have a default zero value which structs don't. So we could change VerifiedAt to be a pointer but we also have to do the same change for StateChangedAt.

Would that approach work, changing both fields to pointers and having omitempty on both of them?

aeneasr:
You are right, I thought that we could do this

```go
func (ns NullTime) MarshalJSON() ([]byte, error) {
	t := time.Time(ns)
	if t.IsZero() {
		return nil, nil
	}
	return json.Marshal(&t)
}
```

but apparently this causes an error in the JSON compiler so it won't work. You are right thus, these fields must indeed be pointers for `omitempty` to work!

> Would that approach work, changing both fields to pointers and having omitempty on both of them?

Yes, that should work and it would be awesome if you could contribute towards this. Please also add a test or two covering this use case so we do not regress this at some point! :)

AntiSC2:
I think I have manged it now. There are already a lot of existing tests (both in the regular tests and the e2e-tests) checking if the verified_at/state_changed_at field was null so I changed it so they instead check that the field doesn't exist. So I haven't added any new tests, I'm not super familiar with Go so that's also a reason why I haven't done it.

identity/handler.go
```diff
@@ -220,7 +220,8 @@ func (h *Handler) create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
 		return
 	}
 
-	i := &Identity{SchemaID: cr.SchemaID, Traits: []byte(cr.Traits), State: StateActive, StateChangedAt: sqlxx.NullTime(time.Now())}
+	var time = sqlxx.NullTime(time.Now())
```
aeneasr, line 223: In go it is considered bad practice to have variable names overshadow package names, so it’s best to rename this:

```suggestion
	stateChangedAt := sqlxx.NullTime(time.Now())
```

identity/handler.go
```diff
@@ -318,8 +319,10 @@ func (h *Handler) update(w http.ResponseWriter, r *http.Request, ps httprouter.P
 			h.r.Writer().WriteError(w, r, herodot.ErrBadRequest.WithReasonf("%s", err).WithWrap(err))
 		}
 
+		var time = sqlxx.NullTime(time.Now())
```
aeneasr, line 322:
```suggestion
		stateChangedAt := sqlxx.NullTime(time.Now())
```

identity/handler_test.go
```diff
@@ -164,10 +164,11 @@ func TestHandler(t *testing.T) {
 			for name, ts := range map[string]*httptest.Server{"public": publicTS, "admin": adminTS} {
 				t.Run("endpoint="+name, func(t *testing.T) {
 					res := send(t, ts, "POST", "/identities", http.StatusCreated, json.RawMessage(`{"traits": {"bar":"baz"}}`))
+					var time = sqlxx.NullTime(res.Get("state_changed_at").Time())
```
aeneasr, line 167: Same here :)

identity/identity.go
```diff
@@ -208,14 +208,15 @@ func NewIdentity(traitsSchemaID string) *Identity {
 		traitsSchemaID = config.DefaultIdentityTraitsSchemaID
 	}
 
+	var time = sqlxx.NullTime(time.Now())
```
aeneasr, line 211: And here :)

selfservice/hook/verification_test.go
```diff
@@ -59,9 +59,11 @@ func TestVerifier(t *testing.T) {
 			actual, err = reg.IdentityPool().FindVerifiableAddressByValue(context.Background(), identity.VerifiableAddressTypeEmail, "baz@ory.sh")
 			require.NoError(t, err)
 			assert.EqualValues(t, "baz@ory.sh", actual.Value)
+
+			var time = sqlxx.NullTime(time.Now())
```
aeneasr, line 63: And here :)

selfservice/strategy/link/strategy_verification.go
```diff
@@ -238,7 +238,8 @@ func (s *Strategy) verificationUseToken(w http.ResponseWriter, r *http.Request,
 
 	address := token.VerifiableAddress
 	address.Verified = true
-	address.VerifiedAt = sqlxx.NullTime(time.Now().UTC())
+	var time = sqlxx.NullTime(time.Now().UTC())
+	address.VerifiedAt = &time
```
aeneasr, line 242: And here

test/e2e/cypress/helpers/index.js
```diff
@@ -13,7 +13,7 @@ const assertVerifiableAddress = ({ isVerified, email }) => ({ identity }) => {
   if (isVerified) {
     expect(address.verified_at).to.not.be.null
   } else {
-    expect(address.verified_at).to.be.null
+    expect(address).to.not.have.property('verified_at')
```
aeneasr, line 16: Nice :)

aeneasr (changes requested):

aeneasr (approved):

Issue #655, opened 2020-08-18
Admin/Selfservice session management


zepatrik:
**Is your feature request related to a problem? Please describe.**

There are a few things we need for session management. See also #615 

**Describe the solution you'd like**

Selfservice:
- [ ] List all my sessions
- [ ] View one of my sessions
- [ ] Revoke one of my sessions

Admin:
- [ ] List an identity's sessions
- [ ] View an identity's sessions
- [ ] Delete one or all identity's sessions

aeneasr:
**Is your feature request related to a problem? Please describe.**

We need a couple of endpoints to allow us to manage an identity's session.

**Describe the solution you'd like**

* [ ] we need a global sessions API that allows us to list all sessions (with pagination!) `GET /sessions` at the admin endpoint.
* [ ] we must be able to get a specific session `GET /sessions/<id>` and also delete/revoke a session `DELETE /sessions/<id>`
* [ ] We should be able to see an identity's sessions (again at the admin endpoint) `GET /identities/<id>/sessions`

**Additional context**

These endpoints are needed to display and manage user sessions in the Ory Console.

danthegoodman1:
Would love to see this soon. Once this is in I think I will feel comfortable switching to Ory!

danthegoodman1:
> **Is your feature request related to a problem? Please describe.**
> 
> We need a couple of endpoints to allow us to manage an identity's session.
> 
> **Describe the solution you'd like**
> 
> * [ ]  we need a global sessions API that allows us to list all sessions (with pagination!) `GET /sessions` at the admin endpoint.
> * [ ]  we must be able to get a specific session `GET /sessions/<id>` and also delete/revoke a session `DELETE /sessions/<id>`
> * [ ]  We should be able to see an identity's sessions (again at the admin endpoint) `GET /identities/<id>/sessions`
> 
> **Additional context**
> 
> These endpoints are needed to display and manage user sessions in the Ory Console.

It would also be nice if there was an endpoint to handle deleting all sessions for a single identity, rather than having to list them all then delete one by one, to serve as a "log out everywhere". Maybe `DELETE /identities/<id>/sessions`?

@aeneasr @Benehiko

zepatrik:
### Self-service API Design Proposal

[...]

PR #1740, opened 2021-09-08
feat: Implement endpoint for invalidating all sessions for a given identity

harnash:
This PR introduces endpoint to destroy all sessions for a given identity which effectively logouts user from all devices/sessions. This is useful when for some security concern we want to make sure there are no "old" sessions active or other "staff" related actions (such as force logout after password change etc.).

Contributes to #655 

## Checklist

- [x] I am following the
      the changes.
      works.


session/handler.go
```diff
@@ -153,6 +160,42 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+//
+//     Security:
+//       oryAccessToken:
+//
+//     Responses:
+//       202: emptyResponse
+//       401: jsonError
+//       500: jsonError
+func (h *Handler) logout(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	if err := h.r.SessionPersister().DeleteSessionsByIdentity(r.Context(), x.ParseUUID(ps.ByName("id"))); err != nil {
```
zepatrik, line 191: `x.ParseUUID` will panic (I think) on error. As the parameter comes from the outside, we cannot assume that it is valid, so instead parse the UUID with error handling:
```suggestion
	iID, err := uuid.FromString(ps.ByName("id"))
	if err != nil {
		h.r.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(err.Error())).WithDebug("...something a bit useful")
	}
	if err := h.r.SessionPersister().DeleteSessionsByIdentity(r.Context(), iID); err != nil {
```

session/handler.go
```diff
@@ -153,6 +160,42 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+// - To forcefully logout Identity from all devices and sessions
+//
+//     Schemes: http, https
+//
+//     Security:
+//       oryAccessToken:
+//
+//     Responses:
+//       202: emptyResponse
+//       401: jsonError
```
zepatrik, line 188:
```suggestion
//       400: jsonError
//       401: jsonError
```

session/handler.go
zepatrik: Also, what does the persister return in case of unknown identity or no session? I think it should still return `http.StatusAccepted` in those cases because the user is logged out, so the desired outcome is reached.

session/handler.go
```diff
@@ -153,6 +160,42 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+	// ID is the identity's ID.
+	//
+	// required: true
+	// in: path
+	ID string `json:"id"`
+}
+
+// swagger:route DELETE /sessions/identity/{id} v0alpha1 adminLogoutIdentity
+//
+// Calling this endpoint irrecoverably and permanently Invalidates all sessions tha belongs to a given Identity.
```
zepatrik, line 175:
```suggestion
// Calling this endpoint irrecoverably and permanently Invalidates all sessions that belong to a given Identity.
```

zepatrik (changes requested):
Have you seen #655? It covers your use case as well. Just FYI and for linking this PR to that issue.
Not too happy about the path of the endpoint, but don't have a better idea right now :sweat_smile:

internal/httpclient/docs/V0alpha1Api.md
```diff
@@ -377,6 +378,74 @@ Name | Type | Description  | Notes
 [[Back to README]](../README.md)
 
 
+## AdminLogoutIdentity
+
+> AdminLogoutIdentity(ctx, id).Execute()
+
+Calling this endpoint irrecoverably and permanently Invalidates all sessions tha belongs to a given Identity.
```
danthegoodman1, line 385: typo: `tha belongs` -> `that belong` :)

harnash:
> Have you seen #655? It covers your use case as well. Just FYI and for linking this PR to that issue.

Oh. I've missed that. Thanks!

> Not too happy about the path of the endpoint, but don't have a better idea right now 😅

Yeah. I would choose something more appropriate but moving it to `/identites` (I think I've tried that once) causes circular dependencies issues.

internal/httpclient/docs/V0alpha1Api.md
harnash: Done 😄

session/handler.go
```diff
@@ -153,6 +161,48 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+//     Responses:
+//       202: emptyResponse
+//       400: jsonError
+//       401: jsonError
+//       404: jsonError
+//       500: jsonError
+func (h *Handler) logout(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	iID, err := uuid.FromString(ps.ByName("id"))
+	if err != nil {
+		h.r.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(err.Error()).WithDebug("could not parse UUID"))
```
aeneasr, line 196: Missing return

session/handler.go
```diff
@@ -153,6 +161,48 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+	}
+	if err := h.r.SessionPersister().DeleteSessionsByIdentity(r.Context(), iID); err != nil {
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusAccepted)
```
aeneasr, line 203: HTTP 202 means:

> The HyperText Transfer Protocol (HTTP) 202 Accepted response status code indicates that the request has been accepted for processing, but the processing has not been completed; in fact, processing may not have started yet.

Instead, return StatusNoContent here:

```suggestion
	w.WriteHeader(http.StatusNoContent)
```

session/handler_test.go
```diff
@@ -213,3 +219,62 @@ func TestIsAuthenticated(t *testing.T) {
 		})
 	}
 }
+
+func TestSessionLogout(t *testing.T) {
```
aeneasr, line 223:
```suggestion
func TestHandlerDeleteSessionByIdentityID(t *testing.T) {
```

:)

aeneasr (changes requested):

session/handler.go
```diff
@@ -153,6 +161,48 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+// nolint:deadcode,unused
+type adminLogoutIdentity struct {
+	// ID is the identity's ID.
+	//
+	// required: true
+	// in: path
+	ID string `json:"id"`
+}
+
+// swagger:route DELETE /sessions/identity/{id} v0alpha1 adminLogoutIdentity
```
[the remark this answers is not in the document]
harnash: Right. That was my initial idea but doing so would introduce import cycle:
```
	imports github.com/ory/kratos/cmd
	imports github.com/ory/kratos/cmd/courier
	imports github.com/ory/kratos/driver
	imports github.com/ory/kratos/continuity
	imports github.com/ory/kratos/identity
	imports github.com/ory/kratos/session
	imports github.com/ory/kratos/identity: import cycle not allowed
```
since we need an access to `session.PersistenceProvider`.  Unless we keep it in `session/handler.go` but change the route path only.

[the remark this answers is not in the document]
zepatrik: Keeping it in the session handler but having an `/identity` path sounds good to me :+1:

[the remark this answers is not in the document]
aeneasr: What @zepatrik said :) Routes do not necessarily need to match packages :)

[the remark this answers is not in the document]
harnash: OK. Will fix that then. Its functionality fits `/identity` endpoint better.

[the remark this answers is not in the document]
harnash: Done in 713447ee

zepatrik (approved):

session/handler.go
```diff
@@ -53,17 +57,21 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
...
+
+	admin.DELETE(RouteDeleteSession, h.deleteIdentitySessions)
 }
 
 func (h *Handler) RegisterPublicRoutes(public *x.RouterPublic) {
-	// We need to completely ignore the whoami path so that we do not accidentally set
+	// We need to completely ignore the whoami/logout path so that we do not accidentally set
 	// some cookie.
 	h.r.CSRFHandler().IgnorePath(RouteWhoami)
+	h.r.CSRFHandler().IgnoreGlob(RouteIdentity + "/*")
```
aeneasr, line 68: Please only allow the specific URL added!

aeneasr (changes requested):
Great! Just one minor thing :)

session/handler.go
harnash: Done in cf2b30e

PR #1750, opened 2021-09-14
feat: Make admin recovery to work without emails #1419

abador:
## Related issue(s)
#1419 

## Checklist

- [x] I am following the
      the changes.
      works.

## Further Comments
It's possible now to initiate and finish the whole recovery flow without a valid recovery email


persistence/sql/migrations/templates/20210913095309_identity_recovery_tokens.up.fizz
```diff
@@ -0,0 +1,2 @@
+add_column("identity_recovery_tokens", "identity_id", "uuid", {"size": 36})
```
aeneasr, line 1: This should have a foreign key :)

persistence/sql/migrations/templates/20210913095309_identity_recovery_tokens.down.fizz
```diff
@@ -0,0 +1,2 @@
+change_column("identity_recovery_tokens", "identity_recovery_address_id", "uuid", {"size": 36})
```
aeneasr, line 1: Without setting all NULL columns to some valid value, or deleting their rows, this will fail.

persistence/sql/persister_recovery.go
```diff
@@ -75,12 +74,10 @@ func (p *Persister) UseRecoveryToken(ctx context.Context, token string) (*link.R
 		}
 
 		var ra identity.RecoveryAddress
-		if err := tx.Where("id = ? AND nid = ?", rt.RecoveryAddressID, nid).First(&ra); err != nil {
-			return sqlcon.HandleError(err)
+		if err := tx.Where("id = ? AND nid = ?", rt.RecoveryAddressID, nid).First(&ra); err == nil {
```
aeneasr, line 77: Any error would be ignored here. Please return them appropriately. You can use `errors.Is(sql.ErrNoRows)` to detect if the row could not be found

aeneasr (changes requested):
Thank you! This looks much more like what we need! I haven’t vetted the tests yet but the logic looks 👌. The SQL statements need a bit more work but generally it appears to be good to go. Could you please also update the docs a bit that talk about this feature to reflect the new behavior? :)

abador:
@aeneasr I pushed the fixes and docs. Still the tests fail on running goimports. I'm not sure why this fails since I've got it configured locally and I see no errors :/

persistence/sql/migrations/templates/20210913095309_identity_recovery_tokens.down.fizz
abador: Fixed in https://github.com/ory/kratos/pull/1750/commits/c6f3d98a92cdbb928e105fa1ae653e5796c6a155

persistence/sql/migrations/templates/20210913095309_identity_recovery_tokens.up.fizz
abador: Fixed in https://github.com/ory/kratos/pull/1750/commits/c6f3d98a92cdbb928e105fa1ae653e5796c6a155

harnash (approved):

persistence/sql/persister_recovery.go
```diff
@@ -6,16 +6,13 @@ import (
...
 
-	"github.com/ory/kratos/corp"
-	"github.com/ory/kratos/identity"
-
 	"github.com/gobuffalo/pop/v5"
 	"github.com/gofrs/uuid"
-
-	"github.com/ory/x/sqlcon"
-
+	"github.com/ory/kratos/corp"
```
zepatrik, line 11:
```suggestion

	"github.com/ory/kratos/corp"
```

Because we use `goimports -local github.com/ory` instead of only `goimports`. The local flag results in "local" imports to be grouped on the bottom of the imports, so you have three sections of imports: stdlib, third party, and first party.

zepatrik: To run it locally, just use `make format`.

persistence/sql/migrations/templates/20210913095309_identity_recovery_tokens.up.fizz
```diff
@@ -0,0 +1,9 @@
...
+{{ if not .IsSQLite }}
+  add_foreign_key("identity_recovery_tokens", "identity_id", {"identities": ["id"]}, {
+      "name": "identity_recovery_tokens_identity_id_fk_idx",
+      "on_delete": "CASCADE",
+      "on_update": "RESTRICT",
+  })
+{{ end }}
```
aeneasr, line 8: You can add foreign keys in sqlite like so:

https://github.com/ory/kratos/blob/4bec9ef50f14f22342a311f09ba1b59cde47befc/persistence/sql/migrations/templates/20210410175418_network.up.fizz#L4

aeneasr (changes requested):
Almost :)

persistence/sql/migrations/templates/20210913095309_identity_recovery_tokens.up.fizz
abador: Sorry missed that comment. I'll try to fix it today

abador: @aeneasr if I add a column this way the migrations don't generate properly. The identity_id column is added i one of the first migrations and then is being removed after some temp table is generated and data is copied there. This causes tests to fail since the identity_id column doesn't exist
`"sqlite create: table identity_recovery_tokens has no column named identity_id"`

aeneasr: Could you share the full up/down migration?

abador: @aeneasr https://github.com/ory/kratos/pull/1750/commits/45ec59bd08509c775520bc2db59a12626c1c64b5

aeneasr: did you run `migrations-render-replace`? Another good practice is to clear out all the migrations in the sql repository related to your change. if you change migration templates the output might be a file that should have been changed or deleted for example

aeneasr: Did it help? :)

abador: sorry @aeneasr  no it didn't. I still get "sqlite create: table identity_recovery_tokens has no column named identity_id" . The only change that wasn't in the pr is in this file: 20210410175418000063_network.sqlite3.down.sql, but it doesn't matter.

aeneasr: Ok, I'll take a look

aeneasr: I just checked out your branch, ran `rm -rf persistence/sql/migrations/sql/* && make migrations-render-replace` and it worked. Not sure what's going on on your end?

Assuming tests pass, is this PR complete in your opinion?

abador: > I just checked out your branch, ran `rm -rf persistence/sql/migrations/sql/* && make migrations-render-replace` and it worked. Not sure what's going on on your end?
> 
> Assuming tests pass, is this PR complete in your opinion?

`make migrations-render-replace` works, but tests fail after that the same way. After regenerating migrations  the only difference between my pr and local changes is the file I mentioned in the previous comment.

As for the PR I would say it's finished.

Issue #1446, opened 2021-06-18
Relative path in ui_url or default_browser_return_url cause runtime crash


zhming0:
**Describe the bug**

When `selfservice.default_browser_return_url` or `selfservice.flows.XXX.ui_url`  is set to be relative path (`/foo/bar`), Kratos will crash whenever it access those setting.

**Reproducing the bug**

1. Setup the demo kratos server using demo config.
2. Change `selfservice.flows.login.ui_url` to `/login`
3. Try to access login page
4. See Kratos crashes with following log

<img width="957" alt="Screen Shot 2021-06-18 at 5 39 39 pm" src="https://user-images.githubusercontent.com/1054703/122525252-667d7a00-d05c-11eb-90cc-2f546e3c6440.png">

_Server configuration_

Just like the demo config: 

```yaml
serve:
  public:
    base_url: http://127.0.0.1:3001/kratos/
  admin:
    base_url: http://kratos:4434/

selfservice:
  default_browser_return_url: http://127.0.0.1:3001/
  whitelisted_return_urls:
    - /

  methods:
    password:
      enabled: true

  flows:
    login:
      ui_url: /login
```

**Expected behavior**

It works just like v0.5 or it throws a proper error instead of crashing

**Environment**

- Version: v0.6.3
- Environment: Docker

**Additional context**

* It appears this new behavior was introduced in v0.6. 
* The same problem applies to the default_browser_return_url setting

zepatrik:
I am not sure that ever worked, are you sure that it worked in v0.5 to have a relative URL for `selfservice.flows.login.ui_url`? The history shows it always used the same parse function: https://github.com/ory/kratos/blame/2e6f87ba7f3d9b4df8851e204554464ecc199c15/driver/config/config.go#L565
Same goes for `default_browser_return_url`.
However, at least the error should be handled correctly.

aeneasr:
We should probably change this in the schema config and only allow URL-URIs to be defined

zhming0:
@zepatrik I am sure it worked in v0.5. I have been running it on prod using such configuration. I thought the `serve.public.base_url` was designed to support this feature? 

For `default_browser_return_url`, [the doc](https://www.ory.sh/kratos/docs/reference/configuration) mentions this (`/dashboard`) as an example.

zhming0:
[...]

PR #1754, opened 2021-09-15
fix: Allow use of relative URLs in config

chlasch:
This PR reverts to previous behavior and allows relative uri-references for self-service URLS. This functionality matches the current documentation and addresses what seems like a bug that was introduced.

Our specific use case is an on-prem application that does not know it's hostname at deployment, and needs to be accessible by arbitrary DNS names or IP hostnames.

## Related issue(s)

#1446 

## Checklist

- [x] I am following the
      the changes.
      works.

## Further Comments

I don't believe a documentation change is needed, as this should be working already according to the config examples and schema. eg. `/Dashboard`


driver/config/config_test.go
```diff
@@ -61,20 +61,18 @@ func TestViperProvider(t *testing.T) {
 					config.ViperKeySelfServiceLoginUI:        "http://test.kratos.ory.sh/#/login",
 					config.ViperKeySelfServiceSettingsURL:    "http://test.kratos.ory.sh/#/settings",
 					config.ViperKeySelfServiceRegistrationUI: "http://test.kratos.ory.sh/#/register",
-					config.ViperKeySelfServiceErrorUI:        "http://test.kratos.ory.sh/#/error",
+					config.ViperKeySelfServiceErrorUI:        "/error",
```
zepatrik, line 64: Please add new tests and cases instead of modifying existing ones that should still pass.

zepatrik (changes requested):
To prevent regression of this bug (again :sweat_smile:), it would be good to add a case to one of the handlers that ensures it actually does a relative redirect when configured to do so.
Ideally even in all flow handlers like https://github.com/ory/kratos/blob/43c315022f05206f1ab0db82b3c98b3fd333921d/selfservice/flow/login/handler_test.go
https://github.com/ory/kratos/blob/43c315022f05206f1ab0db82b3c98b3fd333921d/selfservice/flow/registration/handler_test.go
and the others.

driver/config/config.go
```diff
@@ -734,10 +734,6 @@ func (p *Config) ParseURIOrFail(key string) *url.URL {
 		p.l.WithError(errors.WithStack(err)).
 			Fatalf("Configuration value from key %s is not a valid URL: %s", key, p.p.String(key))
 	}
-	if parsed.Scheme == "" {
```
aeneasr, line 737: Please add a new method that is called for example `ParseAbsoluteOrRelativeURIOrFail` and use it in the places where you want to use it :)

aeneasr (changes requested):

driver/config/config.go
chlasch: Done.

chlasch:
> To prevent regression of this bug (again 😅), it would be good to add a case to one of the handlers that ensures it actually does a relative redirect when configured to do so.
> Ideally even in all flow handlers like https://github.com/ory/kratos/blob/43c315022f05206f1ab0db82b3c98b3fd333921d/selfservice/flow/login/handler_test.go
> https://github.com/ory/kratos/blob/43c315022f05206f1ab0db82b3c98b3fd333921d/selfservice/flow/registration/handler_test.go
> and the others.

I added an explicit redirect check to the 3 handler tests that already had `TestInitFLow` functions (login, recovery, registration). I opted for a testhelper function to check the redirect, instead of duplicating the test function in each handler_test, but I'm not sure if this is how you want it done.

internal/testhelpers/selfservice.go
```diff
@@ -197,3 +197,17 @@ func SelfServiceMakeHookRequest(t *testing.T, ts *httptest.Server, suffix string
 	require.NoError(t, err)
 	return res, string(body)
 }
+
+func GetSelfServiceRedirectLocation(t *testing.T, url string) string {
```
zepatrik, line 201: Perfect :+1:

zepatrik (approved):
Thanks a lot, this looks very good now :tada:

aeneasr (commented):
Great work! :)

Could you add tests for the remaining (verification, settings, error) flows also please? :)

That ensures that this never unintentionally breaks - so it’s like an insurance for you :)

chlasch:
> Great work! :)
> 
> Could you add tests for the remaining (verification, settings, error) flows also please? :)
> 
> That ensures that this never unintentionally breaks - so it’s like an insurance for you :)

I can give it a shot, but since those flows don't have the `TestInitFlow` functions, I got a little stuck in understanding where it should fall within the test.

aeneasr:
Oh I see! If you have difficulties figuring it out feel free to ask here :)

aeneasr:

Thank you!

PR #1757, opened 2021-09-18
Attempt to fix flaky tests when calling StartE2EServer()

harnash:
It seems that tests are failing on the `master`. I don't have statistics since I do not (and can't) have access to CircleCI tests results but it seems that it boils down to `waitToComeAlive()` not being able to to succeed. I've increased the timeout for the health checks requests and it seems to help... we will see when this PR gets checked by CircleCI. Could be that the increased number of migrations is causing some issues but increasing the number of retries did not seem to help.

## Related issue(s)
None.

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/internal/testhelpers/e2e_server.go
+++ b/internal/testhelpers/e2e_server.go
@@ -113,7 +113,7 @@ func waitToComeAlive(t *testing.T, publicUrl, adminUrl string) {
 		}
 		return nil
 	},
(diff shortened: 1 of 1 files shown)


harnash:
I guess my assumptions were wrong. It seems to fail despite the fix.

aeneasr:
Yeah, I think the cause is something else

Issue #1659, opened 2021-08-18
Add PBKDF2 as password hashing algorithm


sawadashota:
**Is your feature request related to a problem? Please describe.**

I want to choose PBKDF2 as password hashing algorithm.
I'm planning to migrate ID system to kratos but a password hashing algorithm in my system is PBKDF2 (This is Django Default).

According to kratos overview, we can choose algorithms but currently we can choose only argon2 or bcrypt.
https://www.ory.sh/kratos/docs/concepts/index/#roll-your-own
https://www.ory.sh/kratos/docs/reference/configuration/

If kratos adds PBKDF2 as password hashing algorithm, I can migrate ID system with less user pain.

**Describe the solution you'd like**

Add PBKDF2 hasher
https://github.com/ory/kratos/tree/master/hash

**Describe alternatives you've considered**

**Additional context**

aeneasr:
Adding PKBDF2 as a n password import option is definitely an option!

PR #1774, opened 2021-09-22
feat(hash): Add PBKDF2 as password hashing algorithm

sawadashota:
## Related issue(s)

#1659

## Checklist

- [x] I am following the
      the changes.
      works.


driver/config/.schema/config.schema.json
```diff
@@ -1732,6 +1733,41 @@
...
+                "sha512"
+              ]
+            },
+            "iterations": {
+              "type": "integer",
+              "minimum": 1000,
+              "default": 100000
+            },
+            "salt_length": {
+              "type": "integer",
```
zepatrik, line 1760:
```suggestion
              "type": "integer",
              "description": "The length of the salt in bytes.",
```

driver/config/.schema/config.schema.json
```diff
@@ -1732,6 +1733,41 @@
...
+              "minimum": 16,
+              "default": 32
+            },
+            "key_length": {
+              "type": "integer",
```
zepatrik, line 1765:
```suggestion
              "type": "integer",
              "description": "The length of the key in bytes.",
```

zepatrik (commented):
This looks pretty good already. I am not sure though if this is the intended way discussed in the issue. This does not allow to update imported hashes but just adds PBKDF2 as an additional hasher. The password upgrade process might actually be implemented on top, so we might actually need it as is.
We could have something along the lines of
```yaml
hashers:
  bcrypt:
    cost: 10
  pbkdf2:
    upgrade_to: bcrypt
```
in the config, which would mean that all pbkdf2 hashes will be upgraded once checked.

sawadashota:
@zepatrik Thank you for your super fast review!
I have fixed it to use PBKDF2 as just compare hash for existing credential and upgrade hashing algorithm after authenticated.

And I remove configuration for PBKDF2 because we can upgrade verified and not primary hashing algorithm automatically.

sawadashota:
Added tests.

selfservice/strategy/password/login.go
```diff
@@ -68,6 +71,12 @@ func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow)
 		return nil, s.handleLoginError(w, r, f, &p, errors.WithStack(schema.NewInvalidCredentialsError()))
 	}
 
+	if !s.d.Hasher().IsSameAlgorithm([]byte(o.HashedPassword)) {
```
zepatrik, line 74: ~This check does not make sense here. The clear-text password does not contain the algorithm.~ - you fixed that already :wink: 

In the long term, we have to make it configurable which hash alg should be upgraded to which one. And some, like PBKDF2, will just be upgraded by default, as they are only intended for import.
For this PR it would be sufficient to just auto-upgrade PBKDF2 without a config option. So adding a check here that returns whether a hash is auto-upgrade (currently only PBKDF2), should be the way to go IMO.

zepatrik (commented):
Looking pretty good now :+1: 
One big thing missing now is documentation. I am also not :100: sure how that should look like, maybe you can start and we discuss it then?

hash/hash_comparator.go
```diff
@@ -105,3 +134,38 @@ func decodeArgon2idHash(encodedHash string) (p *config.Argon2, salt, hash []byte
 
 	return p, salt, hash, nil
 }
+
+// decodePbkdf2Hash decodes PBKDF2 encoded password hash.
+// format: $pbkdf2_<algorithm>$c=<iteration>$<salt>$<hash>
+func decodePbkdf2Hash(encodedHash string) (p *Pbkdf2, salt, hash []byte, err error) {
```
aeneasr, line 140: Where is this function body coming from and what standard does it rely on? :) Is there maybe some well-tested library we can use for this?

selfservice/strategy/password/login.go
sawadashota: Made it configurable!

> One big thing missing now is documentation

Hmm... currently, Kratos doesn't have ability to import user credentials now.
So we have 2 choices to document currently.

* Document "Wait to resolve #605 " 323fdc3
* Implement for #605 in thi PR

Which do you think better?

hash/hash_comparator.go
sawadashota: I could not found any well-tested Go libraries 😢 

This encoding structure is coming from [Django](https://docs.djangoproject.com/en/3.2/topics/auth/passwords/) and [passlib](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.pbkdf2_digest.html) and implementation is from [Django](https://github.com/django/django/blob/8036b53de61f16098a2f22c50621987182bdfaed/django/contrib/auth/hashers.py#L265-L317).
In former commit, iteration part ( `c=<iteration>` ) refer to argon2 in kratos but I changed it Django style at 75b8fcd.

aeneasr: Hm, what about other languages and frameworks? Are all of them using this syntax?

sawadashota: I found more PBKDF2 hasher in 2 framework but all are different syntax.

In case of Symfony(PHP), salt is stored another column.
https://github.com/symfony/symfony/blob/88f3ef62d6ab870128352c8686c7889562698faf/src/Symfony/Component/Security/Core/Encoder/UserPasswordEncoder.php#L37

In case of Spring Boot(Java), syntax is `<salt><hash>` without a separator.
https://github.com/spring-projects/spring-security/blob/main/crypto/src/main/java/org/springframework/security/crypto/password/Pbkdf2PasswordEncoder.java#L201

zepatrik: Hm that would then be part of the documentation/endpoint for importing those hashes I guess. The django format makes most sense IMO, as it includes all parameters.

aeneasr: Ok I see - so we would need to support multiple formats in the end to be able to import user passwords. I guess we can live with that!

sawadashota: Does it mean "Store single format per algorithm, support multiple formats in the end"?

I wrote documentation about format at the commit 060cbdb.

sawadashota:
Sounds good!
Then configuration goes to like...

```yaml
hashers:
  algorithm: bcrypt
  migrate: true
  bcrypt:
    cost: 10
```

@aeneasr @zepatrik Do you have any concerns?

zepatrik:
Sounds reasonable to me, only thing would be that you might want to specify a per-alg migrate flag.
So e.g.
```yaml
hashers:
  algorithm: bcrypt
  bcrypt:
    cost: 10
  pbkdf2:
    migrate: true
  argon2:
    migrate: false
```
which would keep argon2 hashes, but migrate pbkdf2. Not sure if this is actually a use case, but I can imagine you could safe some resources with that strategy.

hash/hash_comparator.go
aeneasr: Yeah, I think we probably need to do some translation at some point, but for now the approach you took is fine :)

aeneasr:
I would enforce migration. We don't want to stick with old PKBDF2 hashes in the DB forever but slowly migrate everyone over :)

seremenko-wish:
I also think we have to force passwords migration, maybe even do not provide an option not to migrate.

As I separate topic, I wanted to discuss a format of hash. Because we already use the PHC standard for password hashes, we have to stick with it https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md. So hash for this algorithm should have a format: 
```
$pbkdf2-<digest>$i=<iterations>$<salt>$<hash>
```

Example here https://www.npmjs.com/package/@phc/pbkdf2#phc-string-format

aeneasr:
Ok, that makes a ton of sense @seremenko-wish - thank you for the input

sawadashota:
For now, I want to clarify what is configurable to enforce migration.
Is it from only pbkdf2 or all non-primary algorithms?
I think the later is simple if it is enough.

```yaml
hashers:
  algorithm: bcrypt
  bcrypt:
    cost: 10
  # Cannot configure for pbkdf2. Auto-upgrade to primary algorithm.
  #pbkdf2:
  #  migrate: true
  argon2:
    migrate: false # Is this configurable like this, leave it or auto-upgrade to primary algorithm?
```

aeneasr:
I would remove the migrate option. We always migrate to the globally configured hasher (`hashers.algorithm`). So:

```
hashers:
  algorithm: bcrypt
  bcrypt:
    cost: 10
  pbkdf2:
    # not sure if this needs any configuration?
  argon2:
    memory: ...
    # ....
```

sawadashota:
I fixed to always migrate to the globally configured hasher.

sawadashota:
Resolved conflict.

aeneasr:
@sawadashota what do you think about [this comment](https://github.com/ory/kratos/pull/1774#issuecomment-931528674) from @seremenko-wish ? 

> I also think we have to force passwords migration, maybe even do not provide an option not to migrate.
>
> As I separate topic, I wanted to discuss a format of hash. Because we already use the PHC standard for password hashes, we have to stick with it https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md. So hash for this algorithm should have a format: 
```
$pbkdf2-<digest>$i=<iterations>$<salt>$<hash>
```
>
> Example here https://www.npmjs.com/package/@phc/pbkdf2#phc-string-format

I think it would make sense to follow the PHC format here.

aeneasr:
It's part of 0.8.0 now so due to be released before the [ory summit](https://events.hubilo.com/ory-summit/register?mtm_campaign=ory-summit-2021&mtm_kwd=banner-landingpage)

aeneasr:
Failing e2e tests are probably my fault

Issue #1775, opened 2021-09-22
request logger middleware is being run after the request was already handled


abador:
**Describe the bug**

When making a request to kratos, the reqlog.Middleware is being run after the request is finished. Logs don't shop proper times of the request. This impacts other middlewares as well

**Reproducing the bug**

1. Make API Request to any public endpoint(or admin)
2. Verify the order of logs

_Server logs_

"started handling request"
"completed handling request"

**Additional context**

This was checked with a very slow mysql replica, so the problem was easier to find.

aeneasr:
Please provide more detailed logs

PR #1776, opened 2021-09-22
fix: Add middleware earlier #1775

abador:
## Related issue(s)
#1775

## Checklist

- [x] I am following the
      the changes.
      works.

## Further Comments
No docs and tests, itonly changes the order of middlewares

### DIFF
--- a/cmd/daemon/serve.go
+++ b/cmd/daemon/serve.go
@@ -79,6 +79,9 @@ func ServePublic(r driver.Registry, wg *sync.WaitGroup, cmd *cobra.Command, args
 	c := r.Config(cmd.Context())
 	l := r.Logger()
 	n := negroni.New()
+	n.Use(reqlog.NewMiddlewareFromLogger(l, "public#"+c.SelfPublicURL(nil).String()))
+	n.Use(sqa(ctx, cmd, r))
+	n.Use(r.PrometheusManager())
 	for _, mw := range modifiers.mwf {
 		n.UseFunc(mw)
 	}
@@ -92,9 +95,6 @@ func ServePublic(r driver.Registry, wg *sync.WaitGroup, cmd *cobra.Command, args
 
 	r.RegisterPublicRoutes(ctx, router)
 	r.PrometheusManager().RegisterRouter(router.Router)
-	n.Use(reqlog.NewMiddlewareFromLogger(l, "public#"+c.SelfPublicURL(nil).String()))
-	n.Use(sqa(ctx, cmd, r))
-	n.Use(r.PrometheusManager())
 
 	if tracer := r.Tracer(ctx); tracer.IsLoaded() {
 		n.Use(tracer)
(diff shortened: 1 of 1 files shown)


cmd/daemon/serve.go
```diff
@@ -79,6 +79,9 @@ func ServePublic(r driver.Registry, wg *sync.WaitGroup, cmd *cobra.Command, args
 	c := r.Config(cmd.Context())
 	l := r.Logger()
 	n := negroni.New()
+	n.Use(reqlog.NewMiddlewareFromLogger(l, "public#"+c.SelfPublicURL(nil).String()))
+	n.Use(sqa(ctx, cmd, r))
+	n.Use(r.PrometheusManager())
```
aeneasr, line 84: Please move this after the for loop below

cmd/daemon/serve.go
```diff
@@ -138,16 +138,16 @@ func ServeAdmin(r driver.Registry, wg *sync.WaitGroup, cmd *cobra.Command, args
 	c := r.Config(ctx)
 	l := r.Logger()
 	n := negroni.New()
+	n.Use(reqlog.NewMiddlewareFromLogger(l, "admin#"+c.SelfPublicURL(nil).String()))
+	n.Use(sqa(ctx, cmd, r))
+	n.Use(r.PrometheusManager())
```
aeneasr, line 143: Please move this after the for loop below

aeneasr (changes requested):

aeneasr (changes requested):
Actually, I don't think that this is really solving the issue you described. The only thing this does is that the negroni middleware is executed before cleanpath and the csrf middleware!

abador:
> Actually, I don't think that this is really solving the issue you described. The only thing this does is that the negroni middleware is executed before cleanpath and the csrf middleware!

Yeah this also looked a bit weird to me. This shouldn't make much difference, but it does, since the CSRFHandler is created based on the router

aeneasr:
Oh I see, that makes sense! :)

cmd/daemon/serve.go
```diff
@@ -79,9 +79,12 @@ func ServePublic(r driver.Registry, wg *sync.WaitGroup, cmd *cobra.Command, args
 	c := r.Config(cmd.Context())
 	l := r.Logger()
 	n := negroni.New()
+	n.Use(reqlog.NewMiddlewareFromLogger(l, "public#"+c.SelfPublicURL(nil).String()))
+	n.Use(sqa(ctx, cmd, r))
 	for _, mw := range modifiers.mwf {
 		n.UseFunc(mw)
 	}
+	n.Use(r.PrometheusManager())
```
aeneasr, line 87: I meant to say to move all n.Use after the modifiers :) The modifiers are used only for testing, and it is important for them to execute before all other middleware. Thanks!

abador: My bad :) Done

aeneasr (approved):

PR #1780, opened 2021-09-23
Update the field updatedAt of the verifiable address when resending email

mgurevin:
There is a way to abuse verification e-mails sent by kratos-courier. A client can request frequently re-sending verification e-mail and change e-mail address frequently. In these cases, kratos send an e-mail for every client request. This behaviour leads us to become a spam source for the SMTP provider, also overcharge is possible.

I couldn't found a way to prevent this abuse by configuring Kratos, so I tried to solve this problem on my App side. But to do this, I need the last e-mail sent date for a verifiable address and I thought the updateAt field can be used for this since the status changed after the e-mail is sent. When I look the courier code I noticed the updateAt filed doesn't update by the Pop library if the status value doesn't change(for example re-sending a verification mail). So in this patch, we force update this field.

IMHO, this is not a way to prevent the above case. Besides I can not cover recovery flow with this method due to there is no session in this case. But it is a way to me prevent some abuse cases until we found a proper way to handle this situation.

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/link/sender.go
+++ b/selfservice/strategy/link/sender.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


aeneasr:
Thank you for the PR. This is indeed an issue, in fact, the issue touches several parts:

1. Creating mass new accounts if the verify hook is enabled
2. Requesting recovery / verification for random email addresses
3. Creating an account and continuously updating the verification email so that the verify hook triggers

Adding an indicator of last delivery is IMO not a good approach though:

1. Some users might re-request sending the email a couple of times if it lands in the e.g. spam folder
2. Emails will be delivered for on-existing accounts also (so you have no previous delivery time)

I'd love to hear ideas though on how to solve it. One possibility would be adding captchas to these submission forms.

PR #1782, opened 2021-09-23
feat: sign in with apple

themech:
Sign in with Apple isn't working with the generic OIDC provider. I'm adding a dedicated provider here.
The main changes are:

- Apple does not have a regular `client_secret`. Instead it uses team id and a private key to generate a temporary JWT token that acts as a secret
- When requesting user's email address, `form_post` parameter has to be used. This means Apple will send a POST request to the callback and this request will be missing Kratos CSRF and session cookies. In strategy I'm adding a new route that redirects a browser to a GET endpoint. After the redirect browser will attach Kratos cookies.
- Contrary to other providers, Apple sends the `email_verified` field as a string. I'm adding a new type so both booleans and strings can be unmarshalled.

## Checklist

- [x] I am following the
      the changes.
      works.

## Further Comments

I'm not particularly happy about the way I disabled the CSRF checks. First this works only when the provider id is `apple`. Second - it is effective even if Apple provider is not enabled. And third - the CSRF exempt affects both GET and POST requests while it is only needed for the POST one.
Also the POST->GET redirect seems like a workaround but I honestly don't have a better idea how to solve the missing cookies issues.
I'll appreciate any suggestions to improve this change.


driver/config/.schema/config.schema.json
```diff
@@ -403,34 +428,101 @@
         "id",
         "provider",
         "client_id",
-        "client_secret",
```
themech, line 406: `client_secret` is not required for Apple. I'm adding is as required in the condition below: https://github.com/ory/kratos/pull/1782/files#diff-c0162691c6563183b7fde70a705098b8460397b751f4176f671733ed40d3b1aeR489

docs/docs/guides/sign-in-with-github-google-facebook-linkedin.mdx
```diff
@@ -1001,6 +1001,82 @@ selfservice:
 Next, open the login endpoint of the SecureApp and you should see the Yandex
 Login option!
 
+## Apple
+
+To set up "Sign in with Apple" you must
+[create an app, a service and a private key key](https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple#create-an-app-id).
```
zepatrik, line 1007:
```suggestion
[create an app, a service and a private key](https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple#create-an-app-id).
```

driver/config/.schema/config.schema.json
zepatrik: This gets a bit messy now :sweat_smile: I will restructure it at some point.... tracked as #1783

themech: Oh, that's very much true. Especially with both Apple and Microsoft having custom fields and `else` sections in the condition.
I'm wondering how strict we want to be when validating especially with fields like `tenant` or `private_key` being forbidden for providers that do not use of them. It makes the schema validation more complicated. And using `tenant` with Github is quite similar to using `issuer_url` with Facebook or Google as those providers override it anyway.

docs/docs/guides/sign-in-with-github-google-facebook-linkedin.mdx
themech: fixed in https://github.com/ory/kratos/pull/1782/commits/23bde7c7e9f07635ddf1f9e63019bced48662eee

aeneasr:
Looks like there are some compile issues :/

themech:
@aeneasr Sorry about that. I'm pretty sure Kratos was compiling fine as I tested the Apple flow like 20 times using our fork: https://github.com/Wikia/kratos/pull/27. But our PR checks and IDE settings were slightly different, I'm fixing it now.

x/json_bool.go
```diff
@@ -0,0 +1,20 @@
...
+
+func (bit *ConvertibleBoolean) UnmarshalJSON(data []byte) error {
+	asString := string(data)
+	if asString == "true" || asString == `"true"` {
+		*bit = true
+	} else if asString == "false" || asString == `"false"` {
+		*bit = false
+	} else {
+		return fmt.Errorf("boolean unmarshal error: invalid input %s", asString)
+	}
```
zepatrik, line 18:
```suggestion
	switch string(data) {
	case "true", `"true"`:
		*bit = true
	case "false", `"false"`:
		*bit = false
	default:
		return fmt.Errorf("boolean unmarshal error: invalid input %s", asString)
	}
```
I would probably write it like that, but that's just code style, so keep what you like more :wink:

themech: Thank you for your suggestion. I changed the code to use the `switch` statement and added some tests in https://github.com/ory/kratos/pull/1782/commits/70b1932b64d98d24afc4647ea8e988c8f58c195a

themech:
I addressed your initial comments and added tests for the new code in strategy.go and i x/json_bool.go.
Existing Kratos tests seem to be a little flaky, my last commit removed an empty line I missed previously and tests went from green to red.

aeneasr (approved):
Nice work!

PR #1806, opened 2021-10-04
fix: corrected typo

dgoosens:
`selfservice.flows.settings.ui_url` should be `selfservice.flows.settings.ui.url`

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/docs/docs/configuring.md
+++ b/docs/docs/configuring.md
@@ -23,7 +23,7 @@ described in the following section.
 
 Environmental variables take precedence over config file values. Nested paths
 get mapped to config values by putting an underscore `_` between every level,
-e.g. `selfservice.flows.settings.ui_url` becomes
+e.g. `selfservice.flows.settings.ui.url` becomes
 `SELFSERVICE_FLOWS_SETTINGS_UI_URL=<value>`.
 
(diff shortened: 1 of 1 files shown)


vinckr:
Thanks for catching this! 
Could you make the same change in the versioned docs, please:
https://github.com/ory/kratos/blob/master/docs/versioned_docs/version-v0.7/configuring.md
🙏

aeneasr (changes requested):
The current value is correct. The key is `ui_url`, not `ui.url`

dgoosens:
Not sure...  
How would a parser be able to make a distinction between those two?

aeneasr:
We have added the appropriate conversions and tests for it!

dgoosens:
> We have added the appropriate conversions and tests for it!

well... yeah....  
if you have a parser that converts all `.` into `_`  
both `selfservice.flows.settings.ui_url` and `selfservice.flows.settings.ui.url` will be converted to `selfservice_flows_settings_ui_url`  
so yeah, there is no error, but this still is not correct

aeneasr:
Using the proposed format would be misleading, because the configuration key is not correct. I will therefore close this, but thank you for elaborating!

Issue #1794, opened 2021-09-29
Panic on recovery for deactivated user


LeonidChistov:
**Describe the bug**

When deactivated user opens recovery link from the recovery email, following error message appears in Kratos log and `500` HTTP error code is reported for call to `/self-service/recovery?flow=<..>." 

```
kratos_1            | 2021/09/29 13:47:36 http: panic serving 172.28.0.4:58654: runtime error: invalid memory address or nil pointer dereference
kratos_1            | goroutine 64686 [running]:
kratos_1            | net/http.(*conn).serve.func1(0xc0008459a0)
kratos_1            | 	/usr/local/go/src/net/http/server.go:1824 +0x153
kratos_1            | panic(0x1560460, 0x2317060)
kratos_1            | 	/usr/local/go/src/runtime/panic.go:971 +0x499
kratos_1            | github.com/ory/kratos/persistence/sql.(*Persister).CreateSession(0xc00067c060, 0x19eb518, 0xc001309e60, 0x0, 0xc000cdafd8, 0x4977c6)
kratos_1            | 	/home/ory/persistence/sql/persister_session.go:39 +0x8e
kratos_1            | github.com/ory/kratos/session.(*ManagerHTTP).CreateAndIssueCookie(0xc0003ef3f8, 0x19eb518, 0xc001309e60, 0x7f5d65a7d6e0, 0xc0000a3540, 0xc0008a5600, 0x0, 0x19c4860, 0xc00041c080)
kratos_1            | 	/home/ory/session/manager_http.go:43 +0x6d
kratos_1            | github.com/ory/kratos/selfservice/strategy/link.(*Strategy).recoveryIssueSession(0xc0000ee3d8, 0x7f5d65a7d6e0, 0xc0000a3540, 0xc0008a5600, 0xc000265a40, 0xd5410070fe24f9fb, 0xaa1e735a5e5c96bd, 0x0, 0x596)
kratos_1            | 	/home/ory/selfservice/strategy/link/strategy_recovery.go:276 +0x42d
kratos_1            | github.com/ory/kratos/selfservice/strategy/link.(*Strategy).recoveryUseToken(0xc0000ee3d8, 0x7f5d65a7d6e0, 0xc0000a3540, 0xc0008a5600, 0xc000c51860, 0x4, 0x7f5d65a7c2a0)
kratos_1            | 	/home/ory/selfservice/strategy/link/strategy_recovery.go:330 +0x419
[...]

PR #1826, opened 2021-10-09
feat: fix 1794

lanphan:
Fix issue #1794 

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/link/strategy_recovery_test.go
+++ b/selfservice/strategy/link/strategy_recovery_test.go
@@ -12,40 +12,29 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
-
 	"github.com/gofrs/uuid"
-
 	"github.com/pkg/errors"
-
-	"github.com/ory/kratos/selfservice/flow"
-	"github.com/ory/kratos/selfservice/strategy/link"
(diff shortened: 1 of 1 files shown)


lanphan:
@aeneasr ,

Please help to review, flow will end up with error page like below when identity is inactive after sending recovery email
Screen Shot 2021-10-10 at 01 54 52
:

lanphan:
@aeneasr ,

I did write some testcases for this, but seems I cannot reach to my code: I found the main flow is completely different if I use 

* delete identity command:
```
reg.Persister().DeleteIdentity(context.Background(), id.ID)
```

* inactivate identity command:
```
id.State = identity.StateInactive
reg.Persister().UpdateIdentity(context.Background(), id)
```

with ```id``` is a new created identity returned by ```createIdentityToRecover()``` method.

because all commands I use will delete (or delete then creating the new one) identity recovery address which will cause ```s.d.RecoveryTokenPersister().UseRecoveryToken``` return ```sqlcon.ErrNoRows``` --> cannot reach to my code.

In order to reach to my code, I think I must setup some twist to update directly db (for ex., update directly state of identity), not using above commands, what do you think?

PR #1948, opened 2021-11-11
build: Speed up docker build in some setups

abador:
This change adds builder cache in dockerfile. This speeds up the build in some cases.
## Checklist

- [x ] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [x ] I am following the
      the changes.
      works.
- [x ] I have added or changed [the documentation](docs/docs).

### DIFF
--- a/.docker/Dockerfile-build
+++ b/.docker/Dockerfile-build
@@ -1,4 +1,5 @@
-FROM golang:1.16-alpine AS builder
+# syntax = docker/dockerfile:1-experimental
+FROM golang:1.16-alpine AS base
 
 RUN apk -U --no-cache add build-base git gcc bash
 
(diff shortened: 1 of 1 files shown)


aeneasr:
Could you please link some docs explaining how this works? Thank you :)

abador:
This change is basically a way to add caching to our docker builds on a local machine that runs that build. If you run it on a different machine there is no difference. I also split this into a multi-stage build (this also might speed some things up). Some example doc about this: https://www.docker.com/blog/containerize-your-go-developer-environment-part-2/

.docker/Dockerfile-build
```diff
@@ -1,4 +1,5 @@
-FROM golang:1.16-alpine AS builder
+# syntax = docker/dockerfile:1-experimental
```
aeneasr, line 1: Is this needed?

.docker/Dockerfile-build
```diff
@@ -13,13 +14,15 @@ ENV CGO_ENABLED 1
 
 RUN go mod download
 
+FROM base as builder
```
aeneasr, line 17: We are using the same image as in the previous step, do we need to or can we leave this out?

abador: it will still work porperly without this

.docker/Dockerfile-build
abador: As far as I know this changes the way dockerfiles are parsed and it's still needed in the current version

PR #1964, opened 2021-11-19
refactor: identity.default_schema_url is now `identity.default_schema_id`

aeneasr:
BREAKING CHANGES: A major issue has been lingering in the configuration for a while. What happens to your identities when you update a schema? The answer was, it depends on the change. If the change is incompatible, some things might break!

To resolve this problem we changed the way you define schemas. Instead of having a global `default_schema_url` which developers used to update their schema, you now need to define the `default_schema_id` which must reference schema ID in your config. To update your existing configuration, check out the patch example below:

```patch
identity:
-  default_schema_url: file://stub/identity.schema.json
+  default_schema_id: default
+  schemas:
+  - id: default
+    url: file://stub/identity.schema.json
```

Ideally, you would version your schema and update the `default_schema_id` with every change to the new version:

```yaml
identity:
  default_schema_id: user_v1
  schemas:
    - id: user_v0
      url: file://path/to/user_v0.json
    - id: user_v1
      url: file://path/to/user_v1.json
```


driver/config/config.go
```diff
@@ -256,7 +256,7 @@ func (s Schemas) FindSchemaByID(id string) (*Schema, error) {
 		}
 	}
 
-	return nil, errors.Errorf("could not find schema with id \"%s\"", id)
+	return nil, errors.Errorf("unable to find identity schema with id: %s", id)
```
zepatrik, line 259: Example for https://github.com/ory/fosite/issues/578
This error will have a stack trace of the `init` func. Instead, we should always use `fmt.Errorf` or `import stderrs "error"`. The stack trace should instead be added once the error is actually returned somewhere.

zepatrik: Oh sorry, thought this is a package level var init.

aeneasr: Not sure I follow? The code style guide for Ory at the moment is to add the stack trace as "deep" as possible in the stack trace in order to get a reliable path to where the error actually happened. This works really well when tracing errors, at least in my experience. I haven't looked at ory/fosite#578

`"errors"` does to this day not include stack traces, which is why we use `pkg/errors` almost everywhere.

Technohacker:
I was having issues with the quickstart configuration on v0.8.2-alpha.1 due to this PR, didn't realize this change was very recent. Could I know which release version this will land in?

aeneasr:
v0.9.0!

PR #1984, opened 2021-11-24
Fix migrations for AWS Aurora Serverless MySQL

splaunov:
AWS Aurora Serverless MySQL blocks setting column as NOT NULL if there is a foreign key constraint with the column.

This PR reorders the DDL statements to avoid this issue.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/migrations/sql/20210817181232000002_unique_credentials.cockroach.up.sql
+++ b/persistence/sql/migrations/sql/20210817181232000002_unique_credentials.cockroach.up.sql
@@ -1,1 +1,1 @@
-ALTER TABLE "identity_credential_identifiers" ADD CONSTRAINT "identity_credential_identifiers_type_id_fk_idx" FOREIGN KEY ("identity_credential_type_id") REFERENCES "identity_credential_types" ("id") ON UPDATE RESTRICT ON DELETE CASCADE;
+UPDATE identity_credential_identifiers SET identity_credential_type_id = (SELECT  ict.id FROM identity_credential_types as ict JOIN identity_credentials AS ic ON (ic.identity_credential_type_id = ict.id) WHERE ic.id = identity_credential_id);
--- a/persistence/sql/migrations/sql/20210817181232000002_unique_credentials.postgres.up.sql
+++ b/persistence/sql/migrations/sql/20210817181232000002_unique_credentials.postgres.up.sql
@@ -1,1 +1,1 @@
-ALTER TABLE "identity_credential_identifiers" ADD CONSTRAINT "identity_credential_identifiers_type_id_fk_idx" FOREIGN KEY ("identity_credential_type_id") REFERENCES "identity_credential_types" ("id") ON UPDATE RESTRICT ON DELETE CASCADE;
+UPDATE identity_credential_identifiers SET identity_credential_type_id = (SELECT  ict.id FROM identity_credential_types as ict JOIN identity_credentials AS ic ON (ic.identity_credential_type_id = ict.id) WHERE ic.id = identity_credential_id);
--- a/persistence/sql/migrations/sql/20210817181232000003_unique_credentials.cockroach.up.sql
+++ b/persistence/sql/migrations/sql/20210817181232000003_unique_credentials.cockroach.up.sql
@@ -1,1 +1,1 @@
-UPDATE identity_credential_identifiers SET identity_credential_type_id = (SELECT  ict.id FROM identity_credential_types as ict JOIN identity_credentials AS ic ON (ic.identity_credential_type_id = ict.id) WHERE ic.id = identity_credential_id);
+ALTER TABLE "identity_credential_identifiers" RENAME COLUMN "identity_credential_type_id" TO "_identity_credential_type_id_tmp";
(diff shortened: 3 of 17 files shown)


aeneasr (changes requested):
Thank you for the PR! Unfortunately, it is not possible to retrospectively change migrations, as that would lead to inconsistent states between different systems. Especially adding or removing existing migrations will mess with the system and leave people without a way to run migrations.

Is there another way to address this, for example with documentation on what needs to be done before running migrations?

splaunov:
> Is there another way to address this, for example with documentation on what needs to be done before running migrations?

I don't see how we could resolve this in the documentation only. There is no other way to set up a new db schema without running the migration scripts. But the migration scripts just don't work with Aurora currently.

A workaround could be just to keep these changes in a separate branch and state in the documentation that those who target Aurora MySQL should not use the ory published kratos images but build their own from the branch instead.

zepatrik:
Is there maybe a way to temporarily deactivate FK checks?

splaunov:
If could use

SET FOREIGN_KEY_CHECKS = 0;

but the constraints will not be revalidated after we set it back to 1 unfortunately.

aeneasr:
> We could use
>
> SET FOREIGN_KEY_CHECKS = 0;
>
> but the constraints will not be revalidated after we set it back to 1 unfortunately.

That could be something we can communicate in the documentation!

splaunov:
> That could be something we can communicate in the documentation!

We need to extend the migration tool then. To provide an option for disabling foreign key checks.
Otherwise, the only way would be is to disable them globally. And this is unacceptable because will influence all sessions targeting other schemas on the same db instance.

Can we add the option to the migrator?

aeneasr:
> We need to extend the migration tool then. To provide an option for disabling foreign key checks.
Otherwise, the only way would be is to disable them globally. And this is unacceptable because will influence all sessions targeting other schemas on the same db instance.

Oh man :( Is it not possible to disable the checks only for one schema (the one kratos uses)?

splaunov:
Never tested it, but according to the docs we have only three options:
- temporary for single SQL statement (SET_VAR hint)
- local to the connection
- global (for all new connections)

https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html#foreign-key-checks
https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_foreign_key_checks

aeneasr:
Maybe it is already possible to set this config using the DSN connection string? I'm not sure if the MySQL Go Driver or MySQL itself support this?

Also, is this issue strictly for AWS AuroraDB or does it affect MySQL 8+ in general?

splaunov:
>Maybe it is already possible to set this config using the DSN connection string? I'm not sure if the MySQL Go Driver or MySQL itself support this?

No, it's not possible. There are only three ways to set MySQL system variables:
- on the server startup through command line parameters
- in the options file
- with the SET statement

``` System variables can be set at server startup using options on the command line or in an option file. Most of them can be changed dynamically while the server is running by means of the SET statement, which enables you to modify operation of the server without having to stop and restart it. You can also use system variable values in expressions.```

https://dev.mysql.com/doc/refman/8.0/en/using-system-variables.html

splaunov:
>Also, is this issue strictly for AWS AuroraDB or does it affect MySQL 8+ in general?

Might be that this problem is specific to Aurora only.
Tested it with Aurora Serverless MySQL 5.7 and Aurora Provisioned MySQL 5.7. Both failed on Kratos migration.
But on my local MySQL 5.7 in docker, the migration runs well.

zepatrik:
So we would need to add `SET` statements in the migrator if configured to do so? I propose we add the possibility to set system variables to the DNS String and then in the migrator execute `SET ...` for every variable before running the migrations. Sounds good?

splaunov:
Do you mean hardcoding the "set_" prefix in the DSN params for MySQL system variables?
Sounds good to me. Until there are no other parameters starting from "set_".

aeneasr:
We could potentially add one extra migration file for aurora which just deals with this particular problem. However, I don't think we can even know right now wether we're dealing with aurora or not?

Is there no other way to deal with this on aurora's side, given that this is valid MySQL?

aeneasr:
Is it ok if we move this to an issue (e.g. design document) to scope out how to solve it? I think we can not merge the PR as is because of the reasons state above, and it would declutter our review backlog a bit :)

splaunov:
> Is it ok if we move this to an issue (e.g. design document) to scope out how to solve it? I think we can not merge the PR as is because of the reasons state above, and it would declutter our review backlog a bit :)

Yes, no problem. Still, we will need to modify the code in some way as it seems to be no other workaround for the issue.

aeneasr:
Great, I've opened up an issue. One thing I'd like to note is that, since this works in plain MySQL, maybe you could ask at AWS why it is built that way and if it's a bug or if there's a workaround.

https://github.com/ory/kratos/issues/2044

seremenko-wish:
FYI, check my findings related to this issue here https://github.com/ory/kratos/issues/2167#issuecomment-1023556058

PR #2015, opened 2021-12-01
docs: Add content-type to allowed headers

DenisPalnitsky:
It's a documentation update. Cors fails in case when Kratos public is not on the same subdomain as client. 
I could be fixed by adding `Content-Type` to allow headers. 
Refer to https://github.com/ory/kratos/discussions/1993

### DIFF
--- a/docs/docs/guides/setting-up-cors.mdx
+++ b/docs/docs/guides/setting-up-cors.mdx
@@ -26,6 +26,7 @@ serve:
       allowed_headers:
         - Authorization
         - Cookie
+        - Content-Type
       exposed_headers:
         - Content-Type
         - Set-Cookie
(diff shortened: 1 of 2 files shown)


aeneasr:
Thank you, this looks great! Could you please sign the CLA (just two clicks), then we can merge it right away!

---

Thank you very much for your contribution to the documentation! Due to the way [Docusaurus](https://docusaurus.io) versioning works, you need to make the changes in two locations:

- `docs/versioned_docs/version-v{version}/path/to/file`
- `docs/docs/path/to/file`

Without making them in both locations, the changes will most likely be lost when releasing another version. Thank you! :)

DenisPalnitsky:
> Thank you, this looks great! Could you please sign the CLA (just two clicks), then we can merge it right away!

Done. 
Also added the same fix to unversioned docs.

zepatrik (approved):

Issue #1970, opened 2021-11-21
feat: implement include_credential for oidc tokens in cmd


AlexandreBrg:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

The new feature implemented in #1818 adds a new query param into route `GET /admin/identities/:id`. Cmd doesn't implement this query param, so we can't fetch oidc tokens. 

### Describe your ideal solution

Add a cmd flag so for command `kratos identities get ${ID}` we can fetch tokens. 

I suggest adding a flag named `--oidc` to solve the problem. Also, I don't know how to name the alias of `--oidc` as `-o` can be used in the future to define an output file ?  

### Workarounds or alternatives

A workaround for now is to keep using the API to fetch identities

### Version

from v0.8.0-alpha.3

aeneasr:
I'd say it should be `--include-credential oidc` as that allows us to expand it later on: `--include-credential oidc --include-credential webauthn`.

PRs welcomed!

PR #2017, opened 2021-12-01
feat(cmd): add oidc credential include 

AlexandreBrg:
## Related issue(s)
Implement this https://github.com/ory/kratos/issues/1970

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments
This is my first contrib :partying_face:

### DIFF
--- a/cmd/identities/get.go
+++ b/cmd/identities/get.go
@@ -2,6 +2,7 @@ package identities
 
 import (
 	"fmt"
+
 	kratos "github.com/ory/kratos-client-go"
 	"github.com/ory/kratos/x"
 	"github.com/ory/x/cmdx"
--- a/cmd/identities/get.go
+++ b/cmd/identities/get.go
@@ -2,10 +2,10 @@ package identities
 
 import (
 	"fmt"
-
 	kratos "github.com/ory/kratos-client-go"
 	"github.com/ory/kratos/x"
 	"github.com/ory/x/cmdx"
+	"github.com/ory/x/stringsx"
 
 	"github.com/ory/kratos/internal/clihelpers"
 
@@ -14,8 +14,16 @@ import (
 	"github.com/ory/kratos/cmd/cliclient"
 )
 
+const (
+	FlagIncludeCreds = "include-credentials"
+)
+
 func NewGetCmd() *cobra.Command {
-	return &cobra.Command{
+	var (
+		includeCreds []string
+	)
+
+	cmd := &cobra.Command{
 		Use:   "get <id-0 [id-1 ...]>",
 		Short: "Get one or more identities by ID",
 		Long: fmt.Sprintf(`This command gets all the details about an identity. To get an identity by some selector, e.g. the recovery email address, use the list command in combination with jq.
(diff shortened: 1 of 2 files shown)


cmd/identities/get.go
```diff
@@ -29,10 +38,25 @@ func NewGetCmd() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			c := cliclient.NewClient(cmd)
 
+			// we check includeCreds argument is valid
+			for _, opt := range includeCreds {
+				switch opt {
```
zepatrik, line 43: Here you can use https://github.com/ory/x/blob/727727f69ff97272f0a0e2ccd6bd6e3d19c3bc6d/stringsx/switch_case_test.go#L59-L66 to have an automatically proper default error.

cmd/identities/get.go
```diff
@@ -29,10 +38,25 @@ func NewGetCmd() *cobra.Command {
...
+				case "oidc":
+					break
+				default:
+					_, _ = fmt.Fprintln(os.Stderr, `You have to put a valid value of credentials type to be included, try --help for details.`)
+					os.Exit(1)
```
zepatrik, line 48: Don't print to `os.Std...` but instead use `cmd.ErrOrStderr`. In tests you can then use a buffer and assert on the output.
Also, never `os.Exit` directly in the command, as it makes it impossible to test this. Always just return the error.

zepatrik (changes requested):
Just some minor improvements :+1:

cmd/identities/get.go
AlexandreBrg: Oh ok ty, as shown in https://github.com/ory/x/blob/727727f69ff97272f0a0e2ccd6bd6e3d19c3bc6d/stringsx/switch_case_test.go#L37, I moved out from a switch case into a single condition. As we only got one case

cmd/identities/get.go
AlexandreBrg: Understood for `os.Exit`. 

I updated with `cmd.PrintErrln` which does both the print & `cmd.ErrOrStderr`

zepatrik: Ah did not know `cmd.PrintErrln`, nice :+1:

cmd/identities/get.go
```diff
@@ -29,10 +37,23 @@ func NewGetCmd() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			c := cliclient.NewClient(cmd)
 
+			// we check includeCreds argument is valid
+			for _, opt := range includeCreds {
+				e := stringsx.SwitchExact(opt)
+				if !e.AddCase("oidc") {
+					cmd.PrintErrln(`You have to put a valid value of credentials type to be included, try --help for details.`)
+					return nil
```
zepatrik, line 45: Returning `nil` here will result in the exit code being 0.
Instead, if you use https://github.com/ory/x/blob/c4721212a9d61b94046379934c1f99bb2f917792/cmdx/helper.go#L31-L37
```suggestion
					return cmdx.FailSilently(cmd)
```
it will not print the error, but fail with non-zero exit code.

cmd/identities/get.go
```diff
@@ -29,10 +37,23 @@ func NewGetCmd() *cobra.Command {
...
+				}
+			}
+
 			identities := make([]kratos.Identity, 0, len(args))
 			failed := make(map[string]error)
 			for _, id := range args {
-				identity, _, err := c.V0alpha2Api.AdminGetIdentity(cmd.Context(), id).Execute()
+				identity, _, err := c.V0alpha2Api.
+					AdminGetIdentity(cmd.Context(), id).
+					IncludeCredential(includeCreds).
```
zepatrik, line 54: Will this also work if there is no credential? Not sure how the SDK works :sweat_smile:

zepatrik: But I guess it works as the current test cases seem to pass.

zepatrik (changes requested):
Thanks for coming back to this :+1:

zepatrik:
Thanks again :+1:

cmd/identities/get.go
AlexandreBrg: I wasn't sure about it, now got it, ty!

cmd/identities/get.go
AlexandreBrg: Well, as in specified in https://github.com/ory/kratos/blob/31a3c19034456dac1810d3c29c87ab49ad7cfd94/identity/identity.go#L324 if no OIDC, then it will process but just do nothing

PR #2018, opened 2021-12-02
chore: fix all issues reported by the CI

zepatrik:
This PR aims to fix all CI issues that make master fail currently.


zepatrik (commented):
I mean, it still times out and flakes, but seems to pass in general.

PR #2025, opened 2021-12-06
feat: buildkit with multi stage build

aeneasr:
Closes #1948

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/.docker/Dockerfile-build
+++ b/.docker/Dockerfile-build
@@ -1,5 +1,4 @@
-# syntax = docker/dockerfile:1-experimental
-FROM golang:1.16-alpine AS base
+FROM golang:1.17-alpine AS base
 
 RUN apk -U --no-cache add build-base git gcc bash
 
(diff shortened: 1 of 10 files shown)


.docker/Dockerfile-build
```diff
@@ -1,4 +1,5 @@
-FROM golang:1.16.10-alpine3.14 AS builder
+FROM golang:1.17-alpine3.15 AS base
+# syntax = docker/dockerfile:1-experimental
```
mitar, line 2: This should be the first line in the file.

mitar: Personally I do not like to use it and instead rely on Docker 20+ already supporting those additional features. The reason why I do not like it is because if you provide it it makes Docker always pull the `docker/dockerfile` image first, before doing anything else, which delays the build process.

martinei: Looks like this has been merged with the "comment" in the wrong line. This way it does nothing and is very missleading.

Issue #2063, opened 2021-12-15
Accept header checking is broken


jld3103:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

I'm sending a request to `/self-service/login/browser` with the `Accept` header set to `application/json; charset=utf-8`. As a response I get a redirect and not the login flow.  
I looked through the code and found the problem:
https://github.com/ory/kratos/blob/8b0d1ee66f1ee2b9f37cd178ac2bcbd8980d6f1d/x/http.go#L110-L113
This method check how the header is set, but if you also have the charset in the header, it defaults to `text/html` which is an incorrect behaviour. Setting the header to `application/json` is just fine.

I think this is an upstream bug, but the repo is archived and the last update to it (the sub directory) was 4 years ago. I'd say the code should no longer be used as it's not maintained and broken.  
I'm open to making an PR to fix this, but I feel like this should be backported (I'm currently using v0.7.6-alpha.1).

### Reproducing the bug

```log
[work@archlinux ~]$ curl http://localhost:4455/api/public/self-service/login/browser -H "Accept: application/json"
[...]

PR #2071, opened 2021-12-16
fix: Accept header checking

jld3103:
This fixes the checking of the accept header for specific mime types. It also removes the usage of `github.com/golang/gddo/httputil` which hasn't been updated in 4 years and is responsible for the breakage (described in the linked issue).

## Related issue(s)
https://github.com/ory/kratos/issues/2063

## Checklist

      introduces a new feature.
      the changes.
      works.

## Further Comments

I found that the `AcceptToRedirectOrJSON` and `ContentNegotiationRedirection` were duplicated, so i removed the former and replaced all occurences with the later.

### DIFF
--- a/x/http_test.go
+++ b/x/http_test.go
@@ -28,27 +28,48 @@ func TestRequestURL(t *testing.T) {
 	}).String(), "http://foobar/foo")
 }
 
-func TestAcceptToRedirectOrJSON(t *testing.T) {
-	wr := herodot.NewJSONWriter(logrusx.New("", ""))
+func TestAcceptsContentType(t *testing.T) {
+	assert.True(t, AcceptsContentType(&http.Request{
+		Header: map[string][]string{"Accept": {"application/json"}},
+	}, "application/json"))
+	assert.True(t, AcceptsContentType(&http.Request{
+		Header: map[string][]string{"Accept": {"application/json; charset=utf-8"}},
+	}, "application/json"))
+	assert.False(t, AcceptsContentType(&http.Request{
+		Header: map[string][]string{"Accept": {"application/json"}},
+	}, "text/html"))
+	assert.False(t, AcceptsContentType(&http.Request{
+		Header: map[string][]string{"Accept": {"application/json; charset=utf-8"}},
+	}, "text/html"))
+	assert.True(t, AcceptsContentType(&http.Request{
+		Header: map[string][]string{"Accept": {"text/html"}},
+	}, "text/html"))
+	assert.True(t, AcceptsContentType(&http.Request{
+		Header: map[string][]string{"Accept": {"text/html, application/json;q=0.9, */*;q=0.8"}},
+	}, "application/json"))
+}
 
-	t.Run("case=browser", func(t *testing.T) {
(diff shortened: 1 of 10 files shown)


x/http_test.go
```diff
@@ -28,27 +28,48 @@ func TestRequestURL(t *testing.T) {
...
+	}, "text/html"))
+	assert.False(t, AcceptsContentType(&http.Request{
+		Header: map[string][]string{"Accept": {"application/json; charset=utf-8"}},
+	}, "text/html"))
+	assert.True(t, AcceptsContentType(&http.Request{
+		Header: map[string][]string{"Accept": {"text/html"}},
+	}, "text/html"))
+	assert.True(t, AcceptsContentType(&http.Request{
+		Header: map[string][]string{"Accept": {"text/html, application/json;q=0.9, */*;q=0.8"}},
+	}, "application/json"))
```
aeneasr, line 49: Please add a test for:

```
assert.True(t, AcceptsContentType(&http.Request{
  Header: map[string][]string{"Accept": {"application/json;q=0.9, */*;q=0.8"}},
}, "text/html"))
assert.True(t, AcceptsContentType(&http.Request{
  Header: map[string][]string{"Accept": {"application/json;q=0.9, text/html;q=0.8, */*;q=0.8"}},
}, "text/html"))
```

And also different spacing and numeric variations:

```
 application/json;q=0.9
 application/json; q=0.9
 application/json; q=0.9,*/*;q=1
 application/json; q=0.9, */*;q=1
 application/json; q=0.9 ,*/*;q=1
```

aeneasr (commented):
After looking into this a bit more, we actually behave correctly here already because `application/json; charset=utf-8` is not a valid accept header. In correct HTTP, this would be:

```
Accept: application/json
```

That is because you can not force a certain encoding at the server. You just negotiate the content type, not the content charset. 

See also:

* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept#directives
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset
* https://stackoverflow.com/questions/7055849/accept-and-accept-charset-which-is-superior

As SO points out:

> Altough you can set media type in Accept header, the charset parameter definition for that media type is not defined anywhere in RFC 2616 (but it is not forbidden, though).

However, there appears to be a more recent RFC covering additional parameters:

https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2

In any case, it makes sense IMO to handle this gracefully :)

aeneasr:
Maybe we could fork the gddo implementation, because the parsing itself works very well, and just add that it ignores additional parameters like `charset`?

aeneasr:

Thank you!

jld3103:
Sure! I'm currently in my vacation and will be back in the beginning of January.

aeneasr:
Awesome :) Enjoy your time off! :)

aeneasr:
Closing this as it is a non-conformity issue and there has been no progress on the PR, but thank you for the work done! :) Feel free to pick it up again if you find the time

Issue #654, opened 2020-08-18
Throttling repeated login requests


aeneasr:
**Describe the solution you'd like**

We should throttle repeated login requests by adding an exponential backoff with a max cap to prevent DoS attacks against certain accounts.

Throttling should be IP-based and/or identity-based.

Donatzsky:
In most cases IP-based is best, since it prevents an attacker from locking an account. But there are legitimate reasons for identity-based as well, so it should be an option.

Don't know if this has been discussed elsewhere, but having a threshold that triggers some sort of account verification would also be useful.

aeneasr:
You mean something like a captcha?

Donatzsky:
Captcha, 2FA, side-channel (email, sms, ...), support etc.

Perhaps two (or more) thresholds: First would be a captcha, just to annoy the low-effort bots, and second would be 2FA or side-channel.

A pseudo-config could look like this:

```
1:
  - attempts: 3-6 # Every attempt from 3 to 6 shows a captcha
  - method: captcha
2:
  - attempts: 7
  - method: sms # SMS verification
3:
  - attempts: 10
  - method: support-lockout # Need to contact support
```

This is somewhat orthogonal to throttling, so perhaps it should be a separate issue.

aeneasr:
So the "best practice" way of throttling this is to simply increase the amount of time it takes to log in with a maximum cap (e.g. 10 seconds). This significantly reduces the ability to run brute force attacks or other type of dictionary attacks. The maximum wait is set up because this would be a DoS attack otherwise where you could lock people out of their accounts.

This would reduce the amount of work we have to put in to support different methods significantly while having similar protective effects. What do you think?

We could still enable a full account lockout though which could be used in e.g. banks!

robbietree8:
Looking forward, will this be addressed in 0.6.x ?

aeneasr:
Not sure, we are quite busy with other features at the moment. Contributions are however always welcomed!

robbietree8:
Good to know, not quite familiar with Golang, just keep watching... : )

Robert-Bosse:
There should be a throttling of failed login attempts on identity basis.
After first failed login 5 seconds cooldown. After second 10 seconds. After third 30 seconds. After fourth and higher 1 minute. 
Do prevent credentials stuffing even with the minute long cooldown a ratio good logins / bad logins on tenant basis that triggers a tenant based captcha worked well for us till now.
[...]

PR #2092, opened 2021-12-31
feat: rate limit failed password attempts

djcrabhat:
Adding a per Identifier+requesting IP [Limiter](https://pkg.go.dev/golang.org/x/time/rate#Limiter), per @aeneasr 's [lovely explanation](https://github.com/ory/kratos/issues/654#issuecomment-728958470).  (hope that suffices for a design doc :))

Starts by making a repeated failed attempt for a user from the same IP wait for 1s.  We double this wait time for every subsequent failed attempt, up to the configurable `selfservice.methods.password.config.rate_limit_max_duration` (Default 30s).  Resets the limiter on successful login.  Adds a background goroutine to forget any tracked attempts older than 10 minutes.

## Related issue(s)

#654 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments
I _think_ an in-memory tracking of failed attempts is good enough here.  I don't think it's worth persisting to the DB for, far too chatty and a drag on performance to be worth the coordinated effort across nodes.

### DIFF
--- a/cmd/clidoc/main.go
+++ b/cmd/clidoc/main.go
@@ -87,6 +87,7 @@ func init() {
 		"NewErrorValidationAddressNotVerified":                    text.NewErrorValidationAddressNotVerified(),
 		"NewErrorValidationNoTOTPDevice":                          text.NewErrorValidationNoTOTPDevice(),
 		"NewErrorValidationNoLookup":                              text.NewErrorValidationNoLookup(),
+		"NewErrorPasswordRateLimit":                               text.NewErrorPasswordRateLimit(),
 		"NewErrorValidationNoWebAuthnDevice":                      text.NewErrorValidationNoWebAuthnDevice(),
 		"NewInfoLoginReAuth":                                      text.NewInfoLoginReAuth(),
 		"NewInfoLoginMFA":                                         text.NewInfoLoginMFA(),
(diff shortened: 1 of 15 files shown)


aeneasr (commented):
This is awesome! Thank you :)

One problem I see right now (we do not have to address it immediately) is that as we scale Ory Kratos (e.g. run 3 pods of it), the rate limit will be per node! So I think it would be good to have a rate limit backend (in memory is a good first step) which can be configured. I think we could have, in the future:

1. in memory / per process
2. in SQL
3. in a high performance system (e.g. redis)

Maybe we could lay the foundation for such a design in this PR and implement in memory, and follow up in another PR with SQL?

aeneasr:
By the way, maybe it makes sense to use an existing rate limiting algo? E.g. https://github.com/uber-go/ratelimit

djcrabhat:
Certainly @aeneasr , would be happy to.  Was hoping this would be a start of the conversation on this feature, propose _something_ with some code, and would be happy to iterate on it.  And will definitely look at uber's library there, I just took go's super simple library as a starter./

Thanks for your support and helpful suggestions.  Will get on those today.

aeneasr:
Absolutely! And thank you so much for the contribution:) And happy new year!! :)

djcrabhat:
Sorry, time got away from me over January.  Still want to clean this up!

aeneasr:
Would be awesome! I think this is a great feature!

PR #2100, opened 2022-01-04
fix: ory cloud mentions + spelling

vinckr:
Fixes a few minor issues in the documentation, most notably the outdated feature descriptions of Ory Cloud. 

Sidenote: In general we should not advertise future/coming soon-features or anything like that, see also [Google style guide](https://developers.google.com/style/future)


aeneasr:
>Sidenote: In general we should not advertise future/coming soon-features or anything like that, see also Google style guide

👍

Issue #2116, opened 2022-01-06
Add Twitter as an identity provider


mfzl:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Context and scope

Current `oidc.Provider` interface works well for providers that doesn't deviate too much from `oauth2` specification.  However Twitter is using OAuth1 for general authentication and identification purposes. This makes it tricky to add as a provider within the current OIDC strategy as it assumes `oauth2` in main `oidc.Provider` interface and handlers utilizing it.

### Goals and non-goals

- Make the Provider interface general enough to accommodate non-oauth2 providers.
- Add Twitter as a provider

### The design

Although OAuth1.1 is different in specification, overall OAuth "dance" is similar enough that it could be accommodated within the current OIDC strategy implementation. Current implementation is general enough to make few adjustments to main interface and how it's utilized without modifying existing Provider implementations. 

Current `Provider` interface is tied to `oauth2` library

```go
type Provider interface {
	Config() *Configuration
	OAuth2(ctx context.Context) (*oauth2.Config, error)
	Claims(ctx context.Context, exchange *oauth2.Token) (*Claims, error)
	AuthCodeURLOptions(r ider) []oauth2.AuthCodeOption
}
```

As a result `oauth2.Token` and `oauth2.Config` is assumed in few other places of OIDC strategy implementation.

This proposes a new general interface for `Provider` with an additional interface for `Token`

```go
// Token returned by providers
type Token interface {
	Claims(ctx context.Context) (*Claims, error)
	CredentialsConfig(ctx context.Context, c cipher.Cipher) (*ProviderCredentialsConfig, error)
}

// Provider for implementations of identity providers
type Provider interface {
	Config() *Configuration
	Token(ctx context.Context, req *http.Request) (Token, error)
[...]

PR #2117, opened 2022-01-06
feat: Provider Twitter

mfzl:
Add twitter as an identity provider and required changes to accommodate it.

## Related issue(s)
#2116

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


aeneasr (commented):
Sorry for the late review! This looks very good :) Could you please add a guide to the docs explaining how to add Twitter auth? That would be awesome! :)

mfzl:
Any update on this?

Benehiko:
Hi @mfzl 

I looked into this PR a bit and there are a lot of changes to the other providers with the implementation of OAuth1.0 flows.
Why do we need to use OAuth1.0 here?

Can we not use Authorization Code Flow?
https://developer.twitter.com/en/docs/authentication/oauth-2-0/user-access-token

mfzl:
Hey @Benehiko 

Thank you for checking into this. 

It's a bit confusing what you can do with Twitter OAuth 1.0a and OAuth 2.0. As of now there's no way to retrieve user profile information with email without falling back to OAuth 1.0a AFAIK. That's the main reason. 

The closest endpoint we have is this https://developer.twitter.com/en/docs/twitter-api/users/lookup/api-reference/get-users-me#Optional

Here's the endpoint that allows retrieving user's email, which requires OAuth 1.0a

https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials

And the changes are necessary to make the `Provider` interface generic as I've briefly highlighted in the linked issue.

Benehiko:
Hi @mfzl 

Maybe we should rather rely on the Twitter v2 endpoint which uses OAuth2.0 flows instead. This would then be a cleaner implementation. We can then request the user to add their `email` (if required by the Identity Schema) through the registration flow after they have logged in through twitter - but this is up to the person implementing Kratos.

https://www.ory.sh/docs/kratos/self-service/flows/user-registration#registration-with-google-facebook-github--openid-connect--oauth-20-1

@aeneasr what do you think?

mfzl:
That would break user expectations given how Login with Twitter is implemented and used in the wild. Which as I understand (as limited as it may be) is a big deal when it comes to user experience.

It would also make it less useful to people migrating over to Kratos using Twitter authentication currently. 

It's also one of the reasons why V2 is not being used a lot in the wild for authentication. They'll get there eventually (hopefully). 
From my readings though, they don't have any plans to deprecate the current API, meaning it's here to stay.

Some references:

https://github.com/supabase/gotrue/blob/6de5ec1e65f9a904207d4ae054db4d677da4158b/api/provider/twitter.go#L23

https://www.passportjs.org/concepts/authentication/twitter/

https://auth0.com/rules/get-twitter-email

Benehiko:
Hey @mfzl 

Thank you for explaining. I understand the reasoning and use case now. Maybe we can implement the `v2` APIs for now in this PR which can be merged relatively quickly, since it will just reuse what the other OAuth2.0 providers use.  

Then after that we can look into implementing `v1` for `email` support. But for this we would need to look into using official libraries or come up with a list of alternatives.

Benehiko:
> @Benehiko how about we send a new PR for v2. You could close this one if you feel that's the best way forward.

Alright, let's do that then! :)

mfzl:
Unfortunately I don't think it's possible because of this issue https://github.com/golang/oauth2/issues/534

aeneasr:
try using go-oidc :)

mfzl:
>

humb1t:
Hi guys! Also curious about Twitter support in Kratos.

aeneasr:
twitter / x is now supported

humb1t:
@aeneasr that's awesome!

Issue #2121, opened 2022-01-08
Duplicate continuity cookies cause requests to fail


aeneasr:
As experienced multiple times by @amorevino and just captured by @sashatalalasha

Google callback request:

```
curl 'https://project.console.ory.sh/api/kratos/public/self-service/methods/oidc/callback/google?....' \
  -H 'authority: project.console.ory.sh' \
  -H 'sec-ch-ua: " Not A;Brand";v="99", "Chromium";v="96", "Google Chrome";v="96"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'upgrade-insecure-requests: 1' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.110 Safari/537.36' \
  -H 'accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9' \
  -H 'sec-fetch-site: cross-site' \
  -H 'sec-fetch-mode: navigate' \
  -H 'sec-fetch-user: ?1' \
  -H 'sec-fetch-dest: document' \
  -H 'referer: https://accounts.google.com/' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'cookie: ory_kratos_continuity=1; __cfruid=eed8572e5941e594b11253131c2656bb86fe5533-1640599019; csrf_token_e9d6d5f4d737a26e1f7090a79427dd512f6=1=; __cflb=; sticky=; ory_kratos_continuity=2' \
  --compressed
```

Google callback response headers:

```
    :authority: project.console.ory.sh
    :method: GET
    :path: /api/kratos/public/self-service/methods/oidc/callback/google?state=...
    :scheme: https
    accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
    accept-encoding: gzip, deflate, br
    accept-language: en-US,en;q=0.9
    cookie: ory_kratos_continuity=1; __cfruid=...-1640599019; ...=...=; __cflb=...; sticky=...; ory_kratos_continuity=2--VTVmGZlj1Iuziui62qXHY_p
    referer: https://accounts.google.com/
[...]

PR #2122, opened 2022-01-08
fix: choose correct CSRF cookie when multiple are set

aeneasr:
Resolves an issue where, when multiple CSRF cookies are set, a random one would be used to verify the CSRF token. Now, regardless of how many conflicting CSRF cookies exist, if one of them is valid, the request will pass and clean up the cookie store.

See ory/kratos#2121
See ory-corp/cloud#1786

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


Imtinan1996:
Hi i am facing this issue on Ory Kratos CLI with local deployment. Are the fixes not pushed as part of the package?

As you can see i am facing the same issue with multiple cookies where one is definitely correct and hence have to clear the cookies before every login flow is created to be able to log in again

image

PR #2124, opened 2022-01-08
fix: resolve magic link issues by introducing another prompt

### LINKED ISSUE #1202: Email verification link is automatically opened by mail provider
This is the same problem as described in #1142 but with verification links.

When registering a new user the email verification link is automatically clicked by the email provider. This leads to an automatic verification of the users email. This could be potentially exploited.

This was reproduced with Office 365.

Solutions could be:
- user interaction before verifying the email (clicking a button).
- ~~user needs to login before the email is verified.~~

### LINKED ISSUE #1451: Replace magic links with TOTP OOB codes

As we've seen in several issues #1202 #1142 the current link strategy is flawed because email clients with antivirus open the link and invalidate the token.

Instead of relying on links we should use recovery / verification codes (TOTP) which have enough entropy to [avoid distributed brute force attacks](https://thezerohack.com/apple-vulnerability-bug-bounty) meaning at least a 8 char alphanumeric password.

This strategy would deprecate and eventually fully replace the current link strategy.

Resolving this issue in the link strategy is difficult because we need workarounds like UA detection. Antiviruses probably spoof this UA to avoid detection, making it a cat-and-mouse game.

### LINKED ISSUE #1765: I need custom path in recovery email link
**Describe the bug**

I need to use a custom URL for recovery and I'm using this env:

`SELFSERVICE_FLOWS_RECOVERY_UI_URL="http://my.domain/customPath/recovery"`

but if I use link method for recovery with the standard email template it sends an email with this link:

`http://my.domain/self-service/recovery?flow=...&token=...`

As you can see it creates a link with `self-service` in it.

_Server configuration_

```yaml
selfservice:
  methods:
    password:
      enabled: true
      config:
        haveibeenpwned_enabled: false

    link:
      enabled: true

  flows:
    recovery:
      enabled: true

log:
  format: text
```

**Expected behavior**

It should create the link in mail (`.recoveryURL` var template) with `customPath` not `self-service`.

Is this my fault? Some config to override?

**Environment**

- Version: oryd/kratos:v0.7.6-alpha.1
- Environment: Docker, Linux, Windows

aeneasr:
This patch addresses issues related to magic links which were in use for verification and recovery flows. With this change, the UI will now display a code input field once the email is sent out. The user then copies the code she/he received in the email into the code input field. This flow will circumvent any magic link related issues on many mobile platforms and with a variety of email providers.

To solve edge cases where the user might not have initiated the flow and instead e.g. an administrator, a link is still included in the email which will open the same page with the code input field pre-filled.

Closes #1765
Closes #1202
Closes #1451

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/link/token_recovery.go
+++ b/selfservice/strategy/link/token_recovery.go
@@ -67,7 +67,7 @@ func NewSelfServiceRecoveryToken(address *identity.RecoveryAddress, f *recovery.
 	}
 	return &RecoveryToken{
 		ID:                x.NewUUID(),
-		Token:             randx.MustString(32, randx.AlphaNum),
+		Token:             randx.MustString(8, randx.Numeric),
 		RecoveryAddress:   address,
 		ExpiresAt:         now.Add(expiresIn),
 		IssuedAt:          now,
--- a/selfservice/strategy/link/token_verification.go
+++ b/selfservice/strategy/link/token_verification.go
@@ -60,7 +60,7 @@ func NewSelfServiceVerificationToken(address *identity.VerifiableAddress, f *ver
 	now := time.Now().UTC()
 	return &VerificationToken{
 		ID:                x.NewUUID(),
-		Token:             randx.MustString(32, randx.AlphaNum),
+		Token:             randx.MustString(8, randx.Numeric),
 		VerifiableAddress: address,
 		ExpiresAt:         now.Add(expiresIn),
 		IssuedAt:          now,


lanphan:
Hi @aeneasr ,

Any update on this, @aeneasr ? I think we can consider this as a big limitation, due to some most of enterprise users cannot use due to their IT policy (Outlook, firewall check link ...)

aeneasr:
Contributions are always welcomed, I unfortunately won't have time to work in this in the next 1 1/2 months :/

PR #2129, opened 2022-01-10
fix: minor fixes from faq update

vinckr:
Follow up PR as requested in https://github.com/ory/kratos/pull/2101#pullrequestreview-846823931.

Next step will be to add the rest of items to documentation 🥳


vinckr:
Let me reopen this properly, my bad.

PR #2154, opened 2022-01-17
feat: Trace web hook calls

martinei:
This Pull request attempt to improve tracing in two places:

- Tracing Middleware was not register in the correct order in daemon/serve.go. That led to many DB calls being orphaned because their parent span did not exist. **Now most DB calls are under a correct parent span**. (I noticed further orphaned DB call spans, but the other root cause is in pop: https://github.com/gobuffalo/pop/issues/685)
- The http client for webhooks is now instrumented as well,  so that **webhooks are traced**

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

## Further Comments

This PR still creates the Http client in webhooks. I am not sure if it would be a good Idea to move it to the registry.
Further the client has no timeouts configured. Making the configurable could be  a task for further improvements.

I had some trouble because the the config and thus the tracer are "contextualized" - at least that is what the function signatures say. In reality I believe the context param is never used and I think it makes the overall thingh to complex.


martinei:
I am not sure about the failing tests. As far as I can see those tests are also red on master

aeneasr:
Those are flaky :) Should be fine!

selfservice/hook/web_hook.go
```diff
@@ -173,12 +180,63 @@ func newWebHookConfig(r json.RawMessage) (*webHookConfig, error) {
...
+// functions as RoundTrippers. If f is a function with the appropriate
+// signature, RountTripperFunc(f) is a RoundTripper that calls f.
+type RoundTripperFunc func(req *http.Request) (*http.Response, error)
+
+// RoundTrip implements the RoundTripper interface.
+func (rt RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	return rt(r)
+}
+
+func createHttpClient(tracer *tracing.Tracer) *http.Client {
```
aeneasr, line 199: Is there some way to write a test for this functionality? It looks fine but we need to ensure:

1. It works
2. It doesn't break when someone changes something

selfservice/hook/web_hook.go
```diff
@@ -173,12 +180,63 @@ func newWebHookConfig(r json.RawMessage) (*webHookConfig, error) {
...
+	if tracer == nil || tracer.IsLoaded() == false {
+		return http.DefaultClient
+	}
+
+	delegateClient := http.DefaultTransport
+
+	var roundTripper RoundTripperFunc = func(req *http.Request) (*http.Response, error) {
```
aeneasr, line 206: Could you move this to `ory/x/httpx`? It could be an option like

https://github.com/ory/x/blob/e38bb388d3f8982c170d256247a3b4f42a8ce589/httpx/resilient_client.go#L40

so for example:

```
httpx.ResilientClientWithTracer()
```

that would allow this to be used across the ecosystem :)

martinei: I will have a look, also at the test

aeneasr:
ory/x is merged :)

Probably makes sense to redo this PR completely from master and just add tracing here:

https://github.com/ory/kratos/blob/d9c82173a1926e815dbb3528cf0c1595d976e84c/driver/registry_default.go#L687-L698

martinei:
Ups. Updating ory/x did not go well. There appear to be breaking changes in the upgraded jsonschema/v3

aeneasr:
Oh yeah, if you make your changes against https://github.com/ory/kratos/pull/2166 then you won't have any problems upgrading ory/x :)

Benehiko (commented):
Looks good! I just think it is missing some tests to verify that the changes do fix the problem. Other than that it's good :)

aeneasr:
Thank you for the review @Benehiko :) @martinei added tests in the PR in ory/x. Given that we only do some bootstrapping here it will be difficult to test this e2e with Jaeger, so I think we can take a leap of faith on this one!

There are however some CI failures, I'm not sure if they're flakes or not (rerunning CI now) but could you (@martinei ) please take a look if they fail?

martinei:
Thanks @aeneasr This is based on the v0.8 branch which also has failing tests. I was hopping to first see green test there, so I can sort out if this PR brakes anything new.

aeneasr:
Ah, got it, fair point! Thank you :)

Issue #2054, opened 2021-12-13
Specify email templates as separate configuration keys


aeneasr:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Currently, you can override the whole directory of where email templates are mounted from using:

```
courier:
  template_override_path: /my/path
```

The problem with this is that we cannot use any remote locations, and it's difficult to specify precedence. 

### Describe your ideal solution

Instead, it should be possible to specify for each template the location:

```
courier:
  templates:
    recovery:
      invalid:
        email:
          plaintext: # base64://, https://, file://, ...
          html: # base64://, https://, file://, ...
        # In the future maybe:
        # sms: ...
    # recovery ...
```

### Workarounds or alternatives

None so far, if you have better ideas please LMK :)

### Version

master

### Additional Context

* [ ] Write migration guide
* [ ] Write breaking changes
* [ ] Document configuration with a guide / examples
* [ ] Document how to do translations ( https://github.com/ory-corp/cloud/issues/1695 )

PR #2156, opened 2022-01-18
feat: courier template configs

Benehiko:
## Related issue(s)

Implements: https://github.com/ory/kratos/issues/2054

## Checklist

      introduces a new feature.
- [X] I am following the
      the changes.
      works.


courier/template/load_template.go
```diff
@@ -65,6 +85,32 @@ func loadBuiltInTemplate(filesytem fs.FS, name string, html bool) (Template, err
 	return tpl, nil
 }
 
+func loadRemoteTemplate(url string, name string, html bool, root string) (Template, error) {
+	if t, found := cache.Get(name); found {
+		return t.(Template), nil
+	}
+
+	f := fetcher.NewFetcher()
```
aeneasr, line 93: This is vulnerable to SSRF, please use the global HTTP client from the registry :) Here's an example:

```
fetch := fetcher.NewFetcher(fetcher.WithClient(s.d.HTTPClient(r.Context())))
```

courier/template/load_template_test.go
```diff
@@ -61,4 +66,83 @@ func TestLoadTextTemplate(t *testing.T) {
...
+				m := map[string]interface{}{"Body": "something"}
+				tp, err := LoadTextTemplate(nil, "", "", m,
+					"file://courier/builtin/templates/test_stub/email.body.plaintext.gotmpl",
+					"base")
+				require.NoError(t, err)
+				assert.Contains(t, tp, "stub email body something")
+			})
+		})
+
+		t.Run("case=http resource", func(t *testing.T) {
```
aeneasr, line 118: Can you please add a test where the config is set so that `clients.http.disallow_private_ip_ranges=true` and then use a localhost URL and see whether it fails?

courier/template/verification_invalid_test.go
```diff
@@ -1,24 +1,77 @@
...
-	require.NoError(t, err)
-	assert.NotEmpty(t, rendered)
+	t.Run("test=with courier templates directory", func(t *testing.T) {
+		conf, _ := internal.NewFastRegistryWithMocks(t)
+		tpl := template.NewVerificationInvalid(conf, &template.VerificationInvalidModel{})
+
+		testhelpers.TestRendered(t, tpl)
+	})
+
+	t.Run("test=with remote resources", func(t *testing.T) {
```
aeneasr, line 26: This test structure is present 4 times, can we DRY it up into a common helper function or is there a good reason to have it duplicated?

driver/config/config.go
```diff
@@ -822,6 +834,54 @@ func (p *Config) CourierTemplatesRoot() string {
 	return p.p.StringF(ViperKeyCourierTemplatesPath, "courier/builtin/templates")
 }
 
+func (p *Config) courierTemplatesHelper(key string) *CourierEmailTemplate {
```
aeneasr, line 837: Could you add a test for this?

aeneasr (changes requested):
Thank you! This looks really solid. A few follow ups:

* [ ] Could you please add documentation to the courier section which explains how to customize individual templates?
* [ ] Is it possible to load only one template from a custom source, and then all other templates are used from the default source?
* [ ] Could you customize the recovery or verification e2e test and set one custom template and then, in the respective e2e test, ensure that the custom template is being used? You can add one or two custom templates to [here](https://github.com/ory/kratos/blob/bb1e8668561e9b6ee939bf9688efa906ed597d5b/test/e2e/profiles/recovery/.kratos.yml#L1) and then assert the contents around [here](https://github.com/ory/kratos/blob/bb1e8668561e9b6ee939bf9688efa906ed597d5b/test/e2e/cypress/integration/profiles/recovery/recovery/success.spec.ts#L90). Alternatively you could also update the config on demand in the e2e test and write an individual test just for the custom template as e2e. To set a custom config value within your test, see [here](https://github.com/ory/kratos/blob/bb1e8668561e9b6ee939bf9688efa906ed597d5b/test/e2e/cypress/integration/profiles/recovery/recovery/success.spec.ts#L83) and subsuquently [here](https://github.com/ory/kratos/blob/bb1e8668561e9b6ee939bf9688efa906ed597d5b/test/e2e/cypress/support/commands.ts#L182). I recommend using base64 based template loading to make your life easier.
* [ ] Can you please add a [schema test](https://github.com/ory/kratos/tree/master/test/schema) to verify that your JSON Schema is correct?

courier/template/load_template.go
Benehiko: So looking into this a bit. When using the `registryDefault.HTTPClient(ctx)` it returns a `*retryablehttp.Client` which is not supported by fetcher. I would have to add an optional function in ory/x for this

```go
func WithResilientClient(rc *retryablehttp.Client) func(*opts) {
	return func(o *opts) {
		o.hc = rc
	}
}
```

Does that make sense? Also, it seems doing so would break a lot of things in Kratos as there are many changes to ory/x

Benehiko: When just returning an `HTTPClient` from the `retryablehttp.Client` it removes a lot of the setup such as max retries etc.

```go
f := fetcher.NewFetcher(fetcher.WithResilientClient(d.HTTPClient(ctx).HTTPClient))
```

Benehiko: based on branch v0.8 solved this

courier/template/verification_invalid_test.go
Benehiko: fixed :)

aeneasr:
@Benehiko are you waiting for a review on this one?

Benehiko:
I'm still writing e2e tests, it's almost ready

courier/template/load_template.go
```diff
@@ -61,12 +71,39 @@ func loadBuiltInTemplate(filesytem fs.FS, name string, html bool) (Template, err
 		tpl = t
 	}
 
-	_ = cache.Add(name, tpl)
+	_ = Cache.Add(name, tpl)
 	return tpl, nil
 }
 
+func loadRemoteTemplate(ctx context.Context, d templateDependencies, url string, name string, html bool, root string) (Template, error) {
+	if t, found := Cache.Get(name); found {
```
aeneasr, line 79: The `name` is a fixed string, it is always `recovery/invalid/email.body.plaintext.gotmpl` for example. If you update your configuration, or if you have a system where you have multiple templates for the same name, this cache will break that as the cache will always return the first template it found for `recovery/invalid/email.body.plaintext.gotmpl`.

Please first add a test verifying that this is an issue (changing the template remote source does in fact not have an effect), and then remove this cache.

courier/template/load_template.go
```diff
@@ -61,12 +71,39 @@ func loadBuiltInTemplate(filesytem fs.FS, name string, html bool) (Template, err
...
+		return t.(Template), nil
+	}
+
+	f := fetcher.NewFetcher(fetcher.WithClient(d.HTTPClient(ctx)))
+
+	bb, err := f.Fetch(url)
```
aeneasr, line 85: In contrast, we can definitely cache this request, so a cache with key `url` that returns `bb`

docs/docs/concepts/email-sms.md
```diff
@@ -92,12 +92,70 @@ courier:
...
+  #
+  # Supported templates are:
+  # - verification
+  #   - valid
+  #   - invalid
+  # - recovery
+  #   - valid
+  #   - invalid
+  #
+  # Each template supports the following layout. A singular key can be specified under `email` to override the defaults.
```
aeneasr, line 108: SO this means specify e.g. `subject` does not require us to set `html`?

embedx/config.schema.json
```diff
@@ -783,6 +783,80 @@
...
+              "examples": [
+                "file://path/to/body.html.gotmpl",
+                "https://foo.bar.com/path/to/body.html.gotmpl",
+                "base64://e3sgZGVmaW5lIGFmLVpBIH19CkhhbGxvLAoKSGVyc3RlbCBqb3UgcmVrZW5pbmcgZGV1ciBoaWVyZGllIHNrYWtlbCB0ZSB2b2xnOgp7ey0gZW5kIC19fQoKe3sgZGVmaW5lIGVuLVVTIH19CkhpLAoKcGxlYXNlIHJlY292ZXIgYWNjZXNzIHRvIHlvdXIgYWNjb3VudCBieSBjbGlja2luZyB0aGUgZm9sbG93aW5nIGxpbms6Cnt7LSBlbmQgLX19Cgp7ey0gaWYgZXEgLmxhbmcgImFmLVpBIiAtfX0KCnt7IHRlbXBsYXRlICJhZi1aQSIgLiB9fQoKe3stIGVsc2UgLX19Cgp7eyB0ZW1wbGF0ZSAiZW4tVVMiIH19Cgp7ey0gZW5kIC19fQo8YSBocmVmPSJ7eyAuUmVjb3ZlcnlVUkwgfX0iPnt7IC5SZWNvdmVyeVVSTCB9fTwvYT4"
+              ]
+            }
+          },
+          "required": [
+            "plaintext",
+            "html"
```
aeneasr, line 847: Are they really required? Wouldn't these fall back to the default if undefined?

docs/docs/concepts/email-sms.md
```diff
@@ -201,6 +261,45 @@ the following pattern: `email.body*`. You can also see that the `Identity` of
...
+### Nested templates with remote templates
+
+When remote templates are used in Kratos, the dynamics of loading nested
+templates change. The templates cannot reference templates outside itself as
+with templates loaded from a singular directory.
+
+The template will need to contain the nested templates in the same file. See
+below for an example. In this template we even wrap the main content in a define
+block called "base". We can then call the "base" define block by using the
+`template_root` configuration.
```
aeneasr, line 273: So `template_root` defines the primary template "id" that is used to render the email? If so, maybe we could change it to `use_id` or `use`? Alternatively we could also just expect this to be e.g. `body` or `subject` which would match the template target and omit this config key?

aeneasr (changes requested):
Looks really good and I have some ideas for improvement :)

courier/template/load_template.go
Benehiko: So instead of caching and retrieving the template we cache the fetch `bytes.Buffer`?

docs/docs/concepts/email-sms.md
Benehiko: yes, they are independent

embedx/config.schema.json
Benehiko: They would, but they are only required when an `email` key was specified. This forces the user to implement html and plaintext templates, instead of e.g. just html

docs/docs/concepts/email-sms.md
Benehiko: Yeah, I agree that we can just always expect a `body` key defined and a `subject`. But this would mean you will always bundle everything in a singular template with the translations. Is that okay? Would we then also expect plaintext and html to be inside the same gotmpl?

e.g.

```
{{ define "body.html" }}
 <h1>an html body</h1>
{{- end -}}

{{ define "subject" }}

{{- end -}}

{{ define "body.plaintext" }}
a plaintext body
{{- end -}}

{{ define "en.US" }}
language specific template
{{- end -}}
```

Benehiko: Then in the config we only have

```yaml

courier:
  templates:
    recovery:
      invalid: "https://remote-template/gotmpl"
...
```

Benehiko:
Should be good to go on green CI

README.md
```diff
@@ -504,6 +504,17 @@ For more details, run:
 ./test/e2e/run.sh
 </pre>
 
+**Run only a singular test**
+Edit the `cypress.json` file located in the `test/e2e/` folder.
```
aeneasr, line 508: This would be easier with using [`only`](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Excluding-and-Including-Tests) :)

docs/versioned_docs/version-v0.8/concepts/email-sms.md
```diff
@@ -3,8 +3,8 @@ id: email-sms
 title: Out-of-band communication via E-Mail and SMS
```
aeneasr, line 3: Please move this to docs/docs or it will be overwritten by the next automation step :)

test/e2e/run.sh
```diff
@@ -108,7 +108,7 @@ prepare() {
 
   (
     cd "$rn_ui_dir"
-    npm i -g expo-cli
+    npm i expo-cli
```
aeneasr, line 111: Is this confirmed to work? I believe the expo-cli must be installed globally: https://docs.expo.dev/get-started/installation/#installing-expo-cli

test/e2e/cypress/integration/profiles/recovery/recovery/errors.spec.ts
```diff
@@ -122,6 +122,17 @@ context('Account Recovery Errors', () => {
           cy.noSession()
         })
       })
+
+      it('invalid remote recovery email template', () => {
+        const identity = gen.identityWithWebsite()
```
aeneasr, line 127: We wanted to add a test which writes to the config file that then changes the schema, to ensure the hot reloading works.

embedx/config.schema.json
aeneasr: Ok got it, you can probably use `if` statements to check whether one of them is set and then require both :)

aeneasr (changes requested):

courier/template/load_template.go
```diff
@@ -61,12 +71,45 @@ func loadBuiltInTemplate(filesytem fs.FS, name string, html bool) (Template, err
...
 	return tpl, nil
 }
 
+func loadRemoteTemplate(ctx context.Context, d templateDependencies, url string, html bool) (Template, error) {
+	var bb *bytes.Buffer
+	var err error
+
+	// instead of creating a new request always we always cache the bytes.Buffer using the url as the key
+	if t, found := Cache.Get(url); found {
+		bb = t.(*bytes.Buffer)
```
aeneasr, line 84: Caching buffers is pretty difficultl. In HTTP for example, the buffer can only be read once because the server sends the body only once. Please use `[]byte` for caching instead

courier/template/load_template_test.go
```diff
@@ -61,4 +78,117 @@ func TestLoadTextTemplate(t *testing.T) {
...
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), "is in the")
+
+			_, err = template.LoadTextTemplate(ctx, reg, nil, "", "", map[string]interface{}{}, "http://localhost:8080/1234")
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), "is in the")
+
+		})
+
+		t.Run("method=cache works", func(t *testing.T) {
```
aeneasr, line 183: Nice :)

aeneasr (changes requested):

README.md
Benehiko: Maybe I can keep this here and just say - "if you want to run a subset of tests for debugging" and then add "only" on running a singular test?

courier/template/load_template.go
Benehiko: Ah okay, makes sense!

test/e2e/run.sh
aeneasr: ok, if it passes CI i guess it’s ok :)

embedx/config.schema.json
Benehiko: https://github.com/Benehiko/kratos/blob/ea6a585d9d4a91b53acf971cca74ad7b45bc3e19/embedx/config.schema.json#L817-L844

This is using a definition that does not require the `body`. It only activates the required fields when setting `body`.

docs/versioned_docs/version-v0.8/concepts/email-sms.md
aeneasr: Ah ok, I think the two documents did not match fully so I wasn't sure whether you might have copied something into this doc that should have been in the other doc (as this will be overwritten). If the docs/concepts/email-sms guide is up to date then you can ignore this comment :)

Benehiko: Yeah there's no diff :)

test/e2e/cypress/integration/profiles/recovery/recovery/errors.spec.ts
Benehiko: So we want to update the config on the fly here inside this test? So instead of having the courier configs added to the already generated kratos configs you want it added here?

aeneasr: Exactly, that would catch resolve any future regressions :) If you need help with the code LMK!

test/e2e/cypress/support/commands.ts
```diff
@@ -462,6 +462,38 @@ Cypress.Commands.add('browserReturnUrlOry', ({} = {}) => {
   })
 })
 
+Cypress.Commands.add('remoteCourierRecoveryTemplates', ({} = {}) => {
```
aeneasr, line 465: 👍 👍 👍 👍

aeneasr (approved):

PR #2160, opened 2022-01-18
fix: Ignore whitespace around identifier with password strategy

### LINKED ISSUE #2158: Email address with leading/trailing whitespace treated as different identifier
### Preflight checklist

### Describe the bug

Suppose an account exists for `"foo@example.com"`, then it is not possible to log in using `"foo@example.com "` (same string but with trailing whitespace, which often happens on mobile phones with auto-completion) or `" foo@example.com"` (leading whitespace). In both cases, the error message reads `"The provided credentials are invalid, check for spelling mistakes in your password or username, email address, or phone number."`. I think whitespace around the identifier should be ignored.

Even worse, entirely new accounts can be registered for all of the following:

```json
" foo@example.com"
"  foo@example.com"
"   foo@example.com"
```

etc.

Interestingly, trying to register an account for `"foo@example.com "` (trailing space instead of leading) will fail, as it should. The error message there reads `"\"foo@example.com \" is not valid \"email\""`.

### Reproducing the bug

Create an account for `"foo@example.com"` and verify that logging into it works, then log out again.

At this point the following things will fail, although they should succeed:

* Logging in with `"foo@example.com "`
* Logging in with `" foo@example.com"`

The following will work, although it should fail:

* Creating an account for `" foo@example.com"`

The following will correctly fail:

* Creating an account for `"foo@example.com "`

(Note that the quotation marks are not part of user input. I had to add them to deal with Markdown formatting issues.)

### Relevant log output

_No response_

### Relevant configuration

```yml
{
  "$id": "https://schemas.ory.sh/presets/kratos/quickstart/email-password/identity.schema.json",
  "title": "Person",
  "type": "object",
  "properties": {
    "traits": {
      "type": "object",
      "properties": {
        "email": {
          "type": "string",
          "format": "email",
          "title": "E-Mail",
          "minLength": 3,
          "ory.sh/kratos": {
            "credentials": {
              "password": {
                "identifier": true
              }
            },
            "verification": {
              "via": "email"
            }
          }
        },
        "name": {
          "type": "string",
          "title": "Full name"
        }
      },
      "required": [
        "email",
        "name"
      ],
      "additionalProperties": false
    }
  }
}
```

### Version

v0.8.0-alpha.3

Linux

Kubernetes with Helm

### Additional Context

_No response_

meyfa:
All identifiers are trimmed before they are passed to the database.

## Related issue(s)

closes #2158

## Checklist

      introduces a new feature.
- [X] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/password/login_test.go
+++ b/selfservice/strategy/password/login_test.go
@@ -677,6 +677,24 @@ func TestCompleteLogin(t *testing.T) {
 		assert.Equal(t, identifier, gjson.Get(body2, "identity.traits.subject").String(), "%s", body2)
 	})
 
+	t.Run("should login same identity regardless of leading or trailing whitespace", func(t *testing.T) {
+		identifier, pwd := x.NewUUID().String(), "password"
+		createIdentity(identifier, pwd)
+
+		browserClient := testhelpers.NewClientWithCookies(t)
+		f := testhelpers.InitializeLoginFlowViaBrowser(t, browserClient, publicTS, false, false)
+
+		values := url.Values{"method": {"password"}, "password_identifier": {"  " + identifier + "  "}, "password": {pwd}, "csrf_token": {x.FakeCSRFToken}}.Encode()
+
+		_, res := testhelpers.LoginMakeRequest(t, false, false, f, browserClient, values)
+		assert.EqualValues(t, http.StatusOK, res.StatusCode)
+
+		f = testhelpers.InitializeLoginFlowViaBrowser(t, browserClient, publicTS, true, false)
+		body2, res := testhelpers.LoginMakeRequest(t, false, false, f, browserClient, values)
+
+		assert.Equal(t, identifier, gjson.Get(body2, "identity.traits.subject").String(), "%s", body2)
+	})
+
 	t.Run("should fail as email is not yet verified", func(t *testing.T) {
 		conf.MustSet(config.ViperKeySelfServiceLoginAfter+".password.hooks", []map[string]interface{}{
 			{"hook": "require_verified_address"},
--- a/selfservice/strategy/password/login_test.go
+++ b/selfservice/strategy/password/login_test.go
@@ -668,13 +668,10 @@ func TestCompleteLogin(t *testing.T) {
 
 		values := url.Values{"method": {"password"}, "password_identifier": {strings.ToUpper(identifier)}, "password": {pwd}, "csrf_token": {x.FakeCSRFToken}}.Encode()
 
-		_, res := testhelpers.LoginMakeRequest(t, false, false, f, browserClient, values)
+		body, res := testhelpers.LoginMakeRequest(t, false, false, f, browserClient, values)
+
 		assert.EqualValues(t, http.StatusOK, res.StatusCode)
-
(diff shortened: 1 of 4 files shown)


selfservice/strategy/password/login_test.go
```diff
@@ -677,6 +677,24 @@ func TestCompleteLogin(t *testing.T) {
...
+		browserClient := testhelpers.NewClientWithCookies(t)
+		f := testhelpers.InitializeLoginFlowViaBrowser(t, browserClient, publicTS, false, false)
+
+		values := url.Values{"method": {"password"}, "password_identifier": {"  " + identifier + "  "}, "password": {pwd}, "csrf_token": {x.FakeCSRFToken}}.Encode()
+
+		_, res := testhelpers.LoginMakeRequest(t, false, false, f, browserClient, values)
+		assert.EqualValues(t, http.StatusOK, res.StatusCode)
+
+		f = testhelpers.InitializeLoginFlowViaBrowser(t, browserClient, publicTS, true, false)
+		body2, res := testhelpers.LoginMakeRequest(t, false, false, f, browserClient, values)
```
zepatrik, line 693: So you are making 2 requests here, but don't really use the first one? I mean, the values are the same in both requests.

selfservice/strategy/password/registration_test.go
```diff
@@ -587,43 +587,77 @@ func TestRegistration(t *testing.T) {
...
+			t.Run("case=leading whitespace", func(t *testing.T) {
+				testWithTransform(t, "leading", func(v url.Values) {
+					v.Set("traits.username", "  "+v.Get("traits.username"))
+				})
+			})
+
+			t.Run("case=trailing whitespace", func(t *testing.T) {
+				testWithTransform(t, "trailing", func(v url.Values) {
+					v.Set("traits.username", v.Get("traits.username")+"  ")
+				})
```
zepatrik, line 660: So you added the two transforms that add whitespace in the beginning and end, but there is no more test without adding whitespace? Can you also add a noop transform to keep the current test behavior?

zepatrik (changes requested):
Thanks, this looks very good already. Just some minor test improvements and we should be ready to go :wink:

selfservice/strategy/password/registration_test.go
meyfa: The no-op case still exists of course, it can be found starting at line 639. I apologize if this setup is too confusing, it was the best I could come up with to not duplicate lots and lots of code...

selfservice/strategy/password/login_test.go
meyfa: You are right, that is a bit weird. Honestly I just copy-pasted the capitalization test case from above and changed the identifier transformation. I think the capitalization test case might originate from `"should be a new session with refresh flag"` which would explain why there are two requests.

I can push a simplified version of both my new test case and the one above (as it suffers from the same problem), if you like?

zepatrik: Yeah, makes sense.

selfservice/strategy/password/registration_test.go
zepatrik: I see, sorry. It is a good approach, the diff is just hard to read.

zepatrik (approved):

aeneasr (approved):

Issue #1777, opened 2021-09-22
Add support for transient properties


dadrus:
**Is your feature request related to a problem? Please describe.**

The domain addressed by Kratos is identity management. Identity Management as "feature set" alone is however not enough and there are other domains involved to have all facets of an Identity Management System covered. Things like the management of privacy policies and/or terms and conditions documents (and the corresponding versions) as well as the management of the corresponding consents given by the users are required, but FMPOV clearly outside of scope for Kratos. As of today the integration of such systems with Kratos require the definition of corresponding properties as traits in the identity schema so that one can hook into a specific flow, e.g. registration, via web hooks and pass down the information to the corresponding system. This approach has a huge drawback - such properties are not managed by Kratos, but are required to be used in each and every request e.g. to update the identity object.

**Describe the solution you'd like**

To overcome the above said limitations and to enable better integration with other systems, there is a need for "transient" properties. That is properties, which are not stored by Kratos, but are just used for passing them down to other systems via web hooks. One approach would be to extend the web hook implementation to support passing the body object of the request to the jsonet template. This way the front end application for Kratos could include whatever is required into the rendered form and the web hook could make use of it upon submission to Kratos, without Kratos even knowing about these additional properties. Care must however be taken as some such requests (e.g. registration and login) have also the password send in the body. These shall not be exposed to the jsonet template.

**Describe alternatives you've considered**

Another approach could be to allow patching of identities in Kratos. So the properties required by other systems would not be required to be passed around in requests related to identity object updates. It has however the downside, that Kratos would still know about these and still not be responsible for managing these.

What do you think? If you agree with extending the web hooks implementation, as written above, I'll volunteer for the corresponding PR :)

mitar:
I agree something should be done about this. I do not mind the alternative though, storing those additional properties in Kratos as traits. The only issues I see is that:
* Currently hooks cannot modify which traits are stored.
* All traits can be further changed by end-users. We need internal traits or something.

dadrus:
I already had a chat about it with @aeneasr and started a branch to implement the idea described above. There is however a dependency to ory-x, on which I started working as well. Because of some time constraints, it was however not yet possible for me to finalize it on time.

aeneasr:
See #3102 fyi @CaptainStandby

PR #2161, opened 2022-01-19
feat: Support for transient properties

dadrus:
## Related issue(s)

@aeneasr: This PR is for #1777

## Checklist

- [x] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security. vulnerability, I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
      works.

## Further Comments

Unlike written in the linked issue, this PR does not require any modification of the ory-x lib. Instead it requires a specific json schema to define which properties are expected in the payload from the downstream service. So to make use of the functionality to access the downstream payload body while creating the upstream payload body in the jsonnet template, the web-hook config has been extended to require a new `request_body_schema` property:
```.yaml
- hook: web_hook
  config:
    url: http://some-upstream-endpoint
    method: POST
    body: file:///etc/config/kratos/template.jsonnet
    request_body_schema: file:///etc/config/kratos/downstream-payload-schema.json
```
Without this property, respectively the referenced schema, the body of the downstream request cannot be accessed, respectively an empty payload is provided to the jsonnet template. In the jsonnet template the downstream request body is accessible via the `request_body` property of the well-known `ctx` object like this:
```.yaml
function(ctx) {
    transient_property: ctx.request_body.some_downstream_request_property
}
```
For the above example the aforesaid `downstream-payload-schema.json` could look like this:
```.json
{
    "$id": "https://example.com/downstream-payload-schema.json",
    "type": "object",
    "properties": {
        "some_downstream_request_property": {
            "type": "number"
        }
    },
    "required": [
        "some_downstream_request_property"
    ],
    "additionalProperties": true
}

```
If `additionalProperties` is used, it must be set to `true` as there are kratos specific request attributes present. If the schema is going to "whitelist" password or/and "csrf_token", it will stay without effect, as these two are removed from the payload passed to the template. So these cannot be accessed anyway.

There are a couple of tests already available, but if the approach is ok for you, these will be extended to provide better coverage (if required). Same applies for the documentation.

The only thing I'm struggling a bit with is the name of the `request_body_schema` property. It is not obvious from it, what it means and enables.

What do you think?


aeneasr:
Thank you for the PR! I have a quick question about how this additional schema would behave during Ory Kratos' validation and UI generation phase? Is it being respsected?

aeneasr:
And with regards to your question, I think an alternative to this could be to have available in the webhook context

```
{
  request: {
    form: {
       // Data from form decoding or empty if not www-url-form
    },
    json: {
       // JSON decoded data if available
    },
    headers: {
       // Here we should only include "safe" headers like `Accept`, `Content-Type`, `Content-Length and not things like `X-*` or `Cookie` or `Authorization`, ...
    }
  }
}
```

that way Kratos just forwards the decoded request and the web hook receiver can do whatever with it. This does however make validation a bit tricky, but we can probably have a rule that validation errors propagated to the UI should have a specific error format & error code.

dadrus:
> I have a quick question about how this additional schema would behave during Ory Kratos' validation and UI generation phase? Is it being respsected?
For the validation purpose, the config schema has been extended accordingly.
The UI is respected in the same way, any error raised from a web hook is respected.

dadrus:
Regarding 

> And with regards to your question, I think an alternative to this could be to have available in the webhook context
> 
> ```
> {
>   request: {
>     form: {
>        // Data from form decoding or empty if not www-url-form
>     },
>     json: {
>        // JSON decoded data if available
>     },
>     headers: {
>        // Here we should only include "safe" headers like `Accept`, `Content-Type`, `Content-Length and not things like `X-*` or `Cookie` or `Authorization`, ...
>     }
>   }
> }
> ```
> 
> that way Kratos just forwards the decoded request and the web hook receiver can do whatever with it. This does however make validation a bit tricky, but we can probably have a rule that validation errors propagated to the UI should have a specific error format & error code.

The inner structure for the web hook content is

```.go
templateContext struct {
  Flow flow.Flow `json:"flow"`
  RequestHeaders http.Header `json:"request_headers"`
  RequestMethod string `json:"request_method"`
  RequestUrl string `json:"request_url"`
  RequestBody interface{} `json:"request_body,omitempty"`
  Identity *identity.Identity `json:"identity"`
}
```
The change introduced by this PR is the addition of the `RequestBody` property. So headers were actually already present. 
Regarding the `form` or `json` properties, which you suggest: We cannot just forward the request to the web hook. we have to filter out specific properties, like the csrf token and the password value first.. Since parsing of the request payload is done according to a flow specific json schema, the web hook implementation would have to know all these schemas (of each and every flow). That would a) make the implementation dependent on the implementation of each and every flow and b) require an update to ory-x to not filter out those elements, which are not defined in the schema file. This dependency is imho an explicit no-goal. So the only clear way out of that situation is to let the kratos integrator define its own schema for those properties, which can be included into the request to a kratos flow (but which the kratos is unaware about).

aeneasr:
Your points are very good, I need a bit of thinking time to weight any possible alternatives or whether this is without alternative. Nudge me next week if I don't reply :)

aeneasr:
After sleeping several days 😅 I like your idea as JSON Schema is just cool and it provides great control over validation. At the same time I think this is not the right component to use it for due to primarily developer experience. Thus, I would like to propose the following:

We use the plain bodies but filter out certain values. The filtering works by calling a central function e.g. `ListConfidentialFields()` which can be implemented by every strategy (e.g. `password`, ...) and provides a list of keys which will be omitted from body when forwarding it to the hooks. Since all confidential keys are at the top-level, it should not be too difficult to filter them. So basically something like:

```
body := fromJsonOrForm(req.Body)
fields := registry.ListConfidentialFields()
filter(body, fields)
renderWebHookJsonnet(...)
```

The reason why I think that the JsonSchema is difficult is because developers need to worry about two configuration items when they work with WebHooks now. They must ensure that the key in the JsonNet is correctly typed and available in the JSON Schema. Any mistake here will be very difficult to debug because we have four moving components:

1. The UI form
2. The JSON Schema
3. The JsonNet
4. The web hook target

For example, when developers want to create a new field or change a field, they have to update 4 things. With the suggested approach they might only need to modify the JsonNet, or, if they have the JsonNet forward everything, only the web hook target and the UI.

Another reason is that I believe that we should not push "internal challenges" (such as who knows what fields to hide) to third parties if that makes it difficult to use the feature. For example, the JSON Schema for identities influences what Ory Kratos suggests the UI to render. In the case of web hooks, it would only be used for validation / filtering.

I hope this makes sense to you!

dadrus:
I fully agree with you. That makes sense. I'll check what it means in sense of the effort and then see how to better proceed with this PR.

aeneasr:

Thank you!

aeneasr:
@dadrus are you still up for this PR?

dadrus:
Hi. Please excuse me for the silence. Because of my project (ypu're aware about ;) ) I had other priorities. I'll come back to this PR however as soon as possible.

PR #2164, opened 2022-01-20
feat: courier templates fs support

Benehiko:
Adds `embed.FS` support to the templates.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/courier/template/load_template.go
+++ b/courier/template/load_template.go
@@ -8,6 +8,7 @@ import (
 	"io/fs"
 	"path"
 	"path/filepath"
+	"strings"
 	"text/template"
 
 	"github.com/Masterminds/sprig/v3"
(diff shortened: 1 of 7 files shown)


courier/template/load_template.go
```diff
@@ -105,23 +129,41 @@ func loadTemplate(osdir, name, pattern string, html bool) (Template, error) {
...
 	var b bytes.Buffer
 	if err := t.Execute(&b, model); err != nil {
 		return "", err
 	}
 	return b.String(), nil
 }
 
-func LoadHTMLTemplate(osdir, name, pattern string, model interface{}) (string, error) {
-	t, err := loadTemplate(osdir, name, pattern, true)
+func LoadHTMLTemplate(osdir, name, pattern string, model interface{}, fs ...embed.FS) (string, error) {
```
aeneasr, line 153: It looks like we will only use the first embedFS argument. So maybe just make this a regular arg?

aeneasr (changes requested):

courier/template/load_template.go
Benehiko: but then it won't be optional anymore. we will then require the current functions to pass a nil i guess

Benehiko: I see, okay, I will implement option 2.

aeneasr: Thank you @Benehiko  :)

Benehiko:
Just to make sure I'm not adding any regressions. What is the purpose of the `pattern` parameter in this case? To me, it seems strange since if `name` doesn't match, `pattern` doesn't matter. So we know the file exists through its `name`. But then we use the `pattern` (if it exists) to call up the file instead. 

https://github.com/ory/kratos/blob/master/courier/template/load_template.go#L80-L87

zepatrik (approved):

aeneasr:
Great stuff, thank you @Benehiko !

PR #2033, opened 2021-12-07
feat: Passwordless SMS code authentication

splaunov:
Implements new "SMS" login strategy. This strategy requires two login flow submissions:

- the first one is to send a phone number to the server
- and the second - to send the received SMS code for validation

## Related issue(s)

#1570

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [x] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](docs/docs).

PR #2184, opened 2022-02-02
OTP recovery and verification strategy 

oleksiireshetnik:
Changes:
- introduced a new `otp` strategy (similar to link) with recovery and verification methods implementation
- added sender, which uses SMS sending functionality from courier to send OTP
- moved logic for persisting recovery/verification tokens to separate package (for reusing between otp and link strategies)

## Related issue(s)
Needs #1941 to be merged
Implements #1451 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


alexey-reshetnik:
@aeneasr will add a small guide about how to configure Kratos with Twilio and use phone numbers as identifiers, plus recovery and verification process

alexey-reshetnik:
@splaunov just for your information - this PR implements recovery and verification by using OTP codes. Maybe you will find it useful

aeneasr:
@alexey-reshetnik could you give a brief flow chart of how you except the flow to work? You can use mermaid to create such a chart :) I think we have a couple of states we need to take care of:

- New flow
- OTP sent
- OTP entered but wrong, expired, ...
- OTP valid

and probably a few other error flows :)

Also, it would be awesome if the OTP strategy could work with links as well, so for example:

```
/verification?code=123456&flow=...
```

which would pre-fill the form with the OTP code. This could be useful if we want to use this strategy to send magic links and would basically render the link strategy completely obsolete.

Also, we need to talk about entropy. Either, we make the OTP entropy high enough (e.g. `[a-zA-Z0-9]{8}`) or we introduce something which marks the flow as failed after the OTP has been entered incorrectly for e.g. 3 times.

WDYT?

splaunov:
Hi!
I've tried to review the otp strategy draft and did some modifications to make all existing tests pass:
https://github.com/monta-app/kratos/commit/613a599085b4605c139265a77b68ca8951d1f540

But after the review it seems to me that there is a good reason to merge it with the 'code' registration/login strategy.
All the `code` processing logic should be the same in #2033 and in #2184

What do you think?

aeneasr:
fyi I am currently working on passwordless auth using webauthn. I suggest that once it is merged we copy that pattern for SMS auth: https://github.com/ory/kratos/pull/2260

langovoi:
@alexey-reshetnik any progress? Maybe need help?

Issue #1077, opened 2021-02-17
Recovery link doesn't log out existing users


andrewbanchich:
To reproduce:

1. Create an identity using `foo@test.com`.
2. Generate recovery link for that account.
3. Set password using the recovery link.
4. While logged in on the same computer, create a new identity `bar@test.com`.
5. Generate a recovery link for `bar@test.com`.
6. Clicking on the link opens the UI, but still logged in as `foo@test.com`.

aeneasr:
Thank you! Added to the 0.6 milestone.

mattbonnell:
So, when we [use the recovery token](https://github.com/ory/kratos/blob/83aedcb885acb96c5deb39fff675d5f0528af32d/selfservice/strategy/link/strategy_recovery.go#L277), we [issue a new session](https://github.com/ory/kratos/blob/83aedcb885acb96c5deb39fff675d5f0528af32d/selfservice/strategy/link/strategy_recovery.go#L384) by [calling]((https://github.com/ory/kratos/blob/83aedcb885acb96c5deb39fff675d5f0528af32d/selfservice/strategy/link/strategy_recovery.go#L327)) CreateAndIssueCookie. Looking at the code for CreateAndIssueCookie here, it would seem that we should be overwriting any old session cookie: https://github.com/ory/kratos/blob/83aedcb885acb96c5deb39fff675d5f0528af32d/session/manager_http.go#L63-L96

We could explicitly purge the cookie from the request [here](https://github.com/ory/kratos/blob/83aedcb885acb96c5deb39fff675d5f0528af32d/selfservice/strategy/link/strategy_recovery.go#L326), but it strikes me as strange that this would be necessary. 

Can you share your configuration file, @andrewbanchich?

aeneasr:
I think another good approach would be to try and recreate the issue and see if we encounter it - from there the debugging can start :)

andrewbanchich:
```
version: v0.5.5-alpha.1

identity:
  default_schema_url: file:///etc/config/kratos/identity.traits.schema.json

selfservice:
  default_browser_return_url: https://myapp.com/
  whitelisted_return_urls:
    - https://myapp.com/
  flows:
    settings:
      ui_url: https://myapp.com/settings
      lifespan: 1h
      privileged_session_max_age: 1h
      after:
        default_browser_return_url: https://myapp.com/
    logout:
      after:
        default_browser_return_url: https://myapp.com/
    registration:
      ui_url: https://myapp.com/auth/registration
[...]

PR #2195, opened 2022-02-05
fix: accept recovery link from authenticated users (#1077)

kszafran:
When a recovery link is opened while the user already has a session cookie (possibly for another account), the endpoint will now correctly complete the recovery process and issue new cookies.

```
BREAKING CHANGES: Calling /self-service/recovery without flow ID or with an invalid flow ID while authenticated will now respond with an error instead of redirecting to the default page.
```

## Related issue(s)

Closes https://github.com/ory-corp/cloud/issues/2173

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

I don't like my solution too much, it feels like a hack. Since I'm not too familiar with the codebase, I didn't want to make too many changes. I'm open to suggestions. I was thinking that perhaps instead of doing this in `submitFlow` (pseudocode):
```
fetch flow by ID

for each strategy:
    strategy.TryToHandle(...)
```
we could do:
```
for each strategy:
    selected = strategy.TryToAccept(...)

if some specific error:
    redirect authenticated

fetch flow by ID

selected.Handle(...)
```

This would also help avoid that minor breaking change.


selfservice/strategy/link/strategy_recovery.go
```diff
@@ -250,7 +255,7 @@ func (s *Strategy) Recover(w http.ResponseWriter, r *http.Request, f *recovery.F
 func (s *Strategy) recoveryIssueSession(w http.ResponseWriter, r *http.Request, f *recovery.Flow, id *identity.Identity) error {
 	f.UI.Messages.Clear()
 	f.State = recovery.StatePassedChallenge
-	f.SetCSRFToken(flow.GetCSRFToken(s.d, w, r, f.Type))
+	f.SetCSRFToken(s.d.CSRFHandler().RegenerateToken(w, r))
```
kszafran, line 258: I needed to change this, otherwise the flow didn't work for authenticated calls. The sequence of events was:
1. CSRF token is added to the ctx in nosurf handler
2. `flow.GetCSRFToken` gets that token and sets it for the flow (but does not send `Set-Cookie`)
3. later on, when we get to `IssueCookie`, a new CSRF token is generated and sent, because the identity has changed (https://github.com/ory/kratos/blob/894a2cc39671fbc9d2c13b1fc1b45b217da5145d/session/manager_http.go#L87)
4. we end up with one CSRF token in the flow and another in `Set-Cookie`

By always regenerating the token here, we avoid this issue. Previously, this endpoint was inaccessible to authenticated users, so `flow.GetCSRFToken` would always regenerate the token. Since we want to overwrite user cookies, there's no need to try and grab an existing CSRF token.

kszafran:
The e2e tests failed apparently due to some random timeout. I'll ignore that for now, since I'll probably need to push some changes and rerun tests after code review.

kszafran:
And now it failed on `make test-coverage` with:
```
=== RUN   TestError
    e2e_server.go:83: Starting migrations...
Successfully applied SQL migrations!
    e2e_server.go:85: Migration done
    e2e_server.go:87: Starting server...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x15afdc6]

goroutine 4418 [running]:
github.com/ory/x/prometheusx.(*MetricsManager).RegisterRouter(...)
	/home/circleci/go/pkg/mod/github.com/ory/x@v0.0.334/prometheusx/middleware.go:34
github.com/ory/kratos/cmd/daemon.ServeAdmin(0x1ec6098, 0xc0004d0d80, 0xc0019d2460, 0xc0000f6780, 0xc00195adb0, 0x0, 0x3, 0x0, 0x0, 0x0)
	/home/circleci/project/cmd/daemon/serve.go:170 +0x7e6
created by github.com/ory/kratos/cmd/daemon.ServeAll.func1
	/home/circleci/project/cmd/daemon/serve.go:283 +0x199
FAIL	github.com/ory/kratos/examples/go/selfservice/error	1.903s
FAIL
```
It does not look related to my changes 🤔.

selfservice/strategy/link/strategy_recovery.go
aeneasr: I think that makes sense. Do you think that there are implications on users who do not have a session yet, where this could cause trouble?

aeneasr (commented):
Thank you! This looks great! Could you please also add an end-to-end test, for example to this file ( https://github.com/ory/kratos/blob/master/test/e2e/cypress/integration/profiles/recovery/recovery/success.spec.ts#L72 ) to ensure these changes work as expected end-to-end? :)

selfservice/strategy/link/strategy_recovery.go
```diff
@@ -221,6 +221,11 @@ func (s *Strategy) Recover(w http.ResponseWriter, r *http.Request, f *recovery.F
 		return s.recoveryUseToken(w, r, body)
 	}
 
+	if _, err := s.d.SessionManager().FetchFromRequest(r.Context(), r); err == nil {
+		session.RedirectOnAuthenticated(s.d)(w, r, nil)
```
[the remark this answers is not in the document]
kszafran: I took this straight from https://github.com/ory/kratos/blob/2a55f3f90805df4e43cec7a3d009a39dd328aa61/selfservice/flow/recovery/handler.go#L68. I think it redirects to a default page?

kszafran:
Are you OK with this approach in general, or would you prefer that I redo it like described in the "Further Comments" section (or solve this some other way)?

selfservice/strategy/link/strategy_recovery.go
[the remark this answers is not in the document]
aeneasr: Ok, that works - is there a test that ensures we are really redirecting to there? :)

aeneasr:
Yes, the approach is good! :)

selfservice/strategy/link/strategy_recovery.go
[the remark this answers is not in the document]
kszafran: Check the test in `handler_test.go`. It initiates a new recovery flow, navigates to `ui.action`, and ends up redirected.

kszafran:
I added an e2e test.

I aldo added a simple test for `GetCSRFToken` to get the coverage back up (seems that the recovery flow tests were going through a branch that verification tests don't). I'm not a fan of such mocked, fine-grained tests personally. Let me know if you'd prefer some other test, or just ignore that minimal drop in coverage and delete that test.

selfservice/strategy/link/strategy_recovery.go
kszafran: Hm... But the user is logged in as another identity, so any flows they have open are for it. Maybe a very edge case would be that we regenerate the token (sending `Set-Cookie`), but then fail with 4xx. I've never tried that, but according to this [SO answer](https://stackoverflow.com/a/21554371/4107551), error response can also set cookies.

kszafran: One way of addressing it would be to extend `retryRecoveryFlowWithError` (called when there's an error after we've regenerated the token) so that it sends `Set-Cookie: ory_kratos_session; Max-Age=-1; ...` and explicitly removes current session.

But I guess it would be even better if we can just keep using the existing CSRF token 🤔. I'll see if I can make it happen.

kszafran: @aeneasr What are your thoughts on this?

aeneasr: Thank you for the thoughts, I think this should be correct as it is right now! We need to rotate the CSRF token when the identity changes, as this is a so-called principal change :)

aeneasr:
Hey, I need more time to review this. Somehow in my head it does not make sense that the introduced change has the effect it should, so I need to go through the changed code line-by-line, but I don't really have time for it this week :/

kszafran:
> Hey, I need more time to review this. Somehow in my head it does not make sense that the introduced change has the effect it should, so I need to go through the changed code line-by-line, but I don't really have time for it this week :/

No worries. I described the problem here: https://github.com/ory/kratos/issues/1077#issuecomment-1028879862. The TL;DR is that if you're authenticated, then `/self-service/recovery` _always_ redirects to the default page, whether you wanted to submit a recovery form or you clicked on a recovery link (it's the same endpoint). The fix is to not redirect immediately on the route level (in `recovery.RegisterPublicRoutes`), but do it later on when we know that what we're handling is not a recovery link.

kszafran:
@aeneasr Any chance you'll look into this PR this week?

kszafran:
> @kszafran can you please take a look at CI failures?

This issue is supposedly fixed in master already: https://github.com/ory/kratos/issues/1077 I haven't closed this PR yet, because had trouble testing the solution in master locally (see my latest comment in the linked issue). I'm hoping I can test it without issues once v0.8.3 is released.

selfservice/strategy/link/strategy_recovery.go
aeneasr: I checked, this is OK

kszafran:
@piotrmsc I tested `revoke_active_sessions` locally today and it doesn't solve the issue. I will try to rebase this PR onto current master.

aeneasr (approved):

aeneasr:
Thank you, I looked into the PR and resolved a couple of issues I found

kszafran:
@aeneasr In this case, is there still anything you need from me?

aeneasr:
I think we're good :)

vinckr:
Hello @kszafran 

Issue #1969, opened 2021-11-20
Documentation says APIs return 302 but actually return 303


hi117:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

API docs say APIs (ie /self-service/login/browser) return 302, actually return 303.

### Reproducing the bug

Curl `/self-service/login/browser`

### Relevant log output

```shell
curl -v 'http://127.0.0.1:4455/.ory/kratos/public/self-service/login/browser'                                     
*   Trying 127.0.0.1:4455...
* Connected to 127.0.0.1 (127.0.0.1) port 4455 (#0)
> GET /.ory/kratos/public/self-service/login/browser HTTP/1.1
> Host: 127.0.0.1:4455
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 303 See Other
< Cache-Control: private, no-cache, no-store, must-revalidate
< Content-Length: 101
< Content-Type: text/html; charset=utf-8
< Date: Sat, 20 Nov 2021 08:12:03 GMT
< Location: http://127.0.0.1:4455/auth/login?flow=6f6ff6bb-4df7-435a-ae53-55c3ec346447
< Set-Cookie: csrf_token_2d785ea2f6536900108362e9bfff73eef3a32be3921b3abc74b0a990093487c9=uN7r1lyekt86LsMfe8T9uMdSLNIaEs7ws4Ne/p6YU+I=; Path=/; Max-Age=31536000; HttpOnly; SameSite=Lax
< Vary: Origin
< Vary: Cookie
< 
<a href="http://127.0.0.1:4455/auth/login?flow=6f6ff6bb-4df7-435a-ae53-55c3ec346447">See Other</a>.

[...]

PR #2206, opened 2022-02-10
fix: move to new post-release steps

aeneasr:
## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ ] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [ ] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [ ] I have read the [security policy](../security/policy).
- [ ] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](docs/docs).

PR #2215, opened 2022-02-11
fix: use HTTP 303 instead of 302 for selfservice redirects

maurice-freitag:
The issue mentions a discrepancy between the documented response code and the actual response code sent by Kratos during selfservice flow redirects. Upon further inspection I noticed that some handlers actually did return 302 which I have changed to 303 where appropriate. Fixed docs & rebuilt sdk as well. 

## Related issue(s)

#1969

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
- [x] I have ~~added~~ adjusted tests that prove my fix is effective or that my feature
      works.


maurice-freitag:
The end-to-end tests failing seems to be an issue with the current version of yq. Running `yq merge ../test/e2e/profiles/kratos.base.yml ../test/e2e/profiles/email/.kratos.yml` with a local installation (v3.4.1) works fine and gives the expected output. Running the same command with the version build via `make .bin/yq` (v4.19.1) produces the error visible in the logs here:
> Error: Parsing expression: Lexer error: could not match text starting at 1:1 failing at 1:3.
	unmatched text: "me"

:/

zepatrik (commented):
`yq` seems to be broken since #2206, so blame @aeneasr :wink: 
This looks very good, now all redirects seem consistent. Is it expected that anyone depends on a `302` code and we need to mark this as a breaking change? I would assume not...
It would be nice if you could add some tests (or probably just extend existing ones) that ensure no regression will happen.

maurice-freitag:
> `yq` seems to be broken since #2206, so blame @aeneasr wink This looks very good, now all redirects seem consistent. Is it expected that anyone depends on a `302` code and we need to mark this as a breaking change? I would assume not... It would be nice if you could add some tests (or probably just extend existing ones) that ensure no regression will happen.

Handling the 302/303 redirect is the user agent's responsibility. According to MDN, both are supported by all browsers but [302](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302) handling seems to be inconsistently implemented, whereas with [303](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303) the subsequent request method is always GET, which is probably what we want when redirecting to the application UI. So if anything there is a very slim chance this could fix otherwise broken redirects if some obscure browser decides to handle 302 weirdly. In any case it makes user agent behavior more predictable I guess. I'll look into writing tests, I guess this would be a case for the e2e tests, right?

Benehiko:
@maurice-freitag I agree, the response code should be 303. 

I believe you can write some tests in each flows `handler_test.go` file. I did take a quick look and it seems we are mocking some of the calls which don't test the redirect (just something I saw when I did a quick look at the codebase this morning). In the e2e tests, we can also ensure that the response code is 303 and that the behaviour is correct.

I can take a look today to direct you properly on this :)

zepatrik (commented):
I'd also say ensuring the status code in the unit tests is enough. We can't test all weird browsers anyways, so no point in adding specific cases for this issue.

Benehiko:
So after looking at this a bit, we don't have any tests currently covering if the redirect takes place within the flows' `handler_test.go`.

So I wrote a quick sample for you to add and which you can reuse in the other flows.

under `selfservice/flow/login/handler_test.go`

```go 

t.Run("lifecycle=init", func(t *testing.T) {
...
  t.Run("flow=browser", func(t *testing.T) {
  ...
    t.Run("case=redirects with 303", func(t *testing.T) {
	    c := ts.Client()
	    // prevent the redirect
	    c.CheckRedirect = func(req *http.Request, via []*http.Request) error {
		    return http.ErrUseLastResponse
	    }
	    req, err := http.NewRequest("GET", ts.URL+login.RouteInitBrowserFlow, nil)
	    require.NoError(t, err)
    
	    res, err := c.Do(req)
	    require.NoError(t, err)
	    // here we check that the redirect status is 303
	    require.Equal(t, http.StatusSeeOther, res.StatusCode)
	    defer res.Body.Close()
    })
```

Benehiko:
@maurice-freitag I have opened a PR against your branch implementing the `login` and `logout` handler tests. 
You can merge it and continue with the rest of the tests :) 
It should give you enough context to continue, if you get stuck don't be afraid to ping me or @zepatrik 
https://github.com/maurice-freitag/ory-kratos/pull/1

maurice-freitag:
> ests :)
> It should give you enough context to continue, if you get stuck don't be afraid to ping me or @zepatrik
> [maurice-freitag#1](https://github.com/maurice-freitag/ory-kratos/pull/1)

ty! I merged your changes and will get going with the other tests asap :)

maurice-freitag:
Seems the docs have moved 🙈 I'll look into it when I get the chance

Benehiko:
Ah yes @maurice-freitag you can add the docs to https://github.com/ory/docs

Sorry about that, we've been doing some major changes to the structure of our documentation.

selfservice/flow/settings/handler_test.go
```diff
@@ -159,6 +159,24 @@ func TestHandler(t *testing.T) {
...
+				req, err := http.NewRequest("GET", publicTS.URL+settings.RouteInitBrowserFlow, nil)
+				require.NoError(t, err)
+
+				res, err := c.Do(req)
+				require.NoError(t, err)
+				// here we check that the redirect status is 303
+				require.Equal(t, http.StatusSeeOther, res.StatusCode)
+				defer res.Body.Close()
+			})
 		})
```
Benehiko, line 180: Not sure if this will work, since we need a session first before we initialize a settings flow

https://www.ory.sh/kratos/docs/self-service/flows/user-settings#user-and-profile-settings-for-server-side-browser-clients

Also shown in one of the tests above:

```go
t.Run("description=success", func(t *testing.T) {
	user1 := testhelpers.NewHTTPClientWithArbitrarySessionCookie(t, reg)
	res, body := initFlow(t, user1, false)
	assert.Contains(t, res.Request.URL.String(), reg.Config(context.Background()).SelfServiceFlowSettingsUI().String())
    assertion(t, body, false)
})
```

Benehiko (changes requested):
This looks quite good :)
I think it's almost done, just a small change and the docs and this should be good to go!

selfservice/flow/settings/handler_test.go
maurice-freitag: The test ran through, I switched the default client with the one including the session cookie anyways since all other tests here use it as well

maurice-freitag: I also instantiated a new http client instead of using `http.DefaultClient` because some subsequent tests where failing due to the changed redirect behavior on the default instance.

maurice-freitag:
The PR for the docs is now open as well: https://github.com/ory/docs/pull/598

aeneasr (approved):

aeneasr:
Thank you! :)

vinckr:
Hello @maurice-freitag 

PR #2221, opened 2022-02-14
fix: incorrect rule in makefile

icyphox:
Removes missing make target `docs/package.json`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ docs/cli:
 node_modules: package.json Makefile
 		npm ci
 
-docs/node_modules: docs/package.json
+docs/node_modules:
 		cd docs; npm ci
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -46,7 +46,7 @@ docs/cli:
 node_modules: package.json Makefile
 		npm ci
 
-docs/node_modules: docs/package.json
+docs/node_modules:
```
aeneasr, line 49: This can be removed completely I suppose

Benehiko:
should this not be merged to v0.8? Or will we just update it on v0.8 from master?

aeneasr:
You're right, it should be against 0.8

icyphox:
Okay well in that case I'll close and re-open against v8.

PR #2222, opened 2022-02-14
fix: incorrect makefile rule

icyphox:


Makefile
```diff
@@ -130,7 +125,7 @@ quickstart-dev:
 
 # Formats the code
 .PHONY: format
-format: .bin/goimports docs/node_modules node_modules
+format: .bin/goimports node_modules
 		goimports -w -local github.com/ory .
 		cd docs; npm run format
```
aeneasr, line 130: What about this?

icyphox: Wops. Fixed.

PR #2232, opened 2022-02-16
fix: pass docs output dir as a separate argument

icyphox:
Fixes Kratos' docs generation. Ref: https://github.com/ory/ci/commit/0851b01317a6b63b334ed38e06e648080663bf06


icyphox:
@aeneasr can you confirm that
- the paths are correct
- that it produces the desired output in the docs repo

PR #2249, opened 2022-02-22
Fix Apple's mess

JiggyDown:
Apple POSTs user first name, last name, and email directly to the callback URL, as opposed to embed them in the ID token like any sane company would do.


JiggyDown:
Sorry wrong organization. We plan to test this first before upstreaming the changes.

Issue #2049, opened 2021-12-10
Recovery link answers with JSON payload for API flows


aeneasr:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

When initializing a recovery or verification link using an API flow, and then clicking on the link, the browser shows a JSON instead of the UI

### Reproducing the bug

Not sure yet

### Version

v0.8.0-alpha.4

aeneasr:
Reported by @sashatalalasha

Demonsthere:
Context: issue related to link expiration.

Demonsthere:
Judging from the JSON response sample, the request_url was missing the return_to parameter: `"request_url":"http://foobar.com/self-service/registration/api?return_to=`

PR #2250, opened 2022-02-23
fix: don't inherit flow type in recovery and verification flows

Demonsthere:
- Set flow type to Browser in recovery and verification flows
- Add tests for both
- Add e2e tests for expired validation flow

## Related issue(s)
- Fixes https://github.com/ory/kratos/issues/2049

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


selfservice/flow/recovery/flow.go
```diff
@@ -130,7 +130,12 @@ func NewFlow(conf *config.Config, exp time.Duration, csrf string, r *http.Reques
 }
 
 func FromOldFlow(conf *config.Config, exp time.Duration, csrf string, r *http.Request, strategies Strategies, of Flow) (*Flow, error) {
-	nf, err := NewFlow(conf, exp, csrf, r, strategies, of.Type)
+	f := of.Type
```
aeneasr, line 133: Is this method only called for expired flows, or could it break some other behavior?

selfservice/flow/verification/flow.go
```diff
@@ -133,7 +133,12 @@ func NewFlow(conf *config.Config, exp time.Duration, csrf string, r *http.Reques
 }
 
 func FromOldFlow(conf *config.Config, exp time.Duration, csrf string, r *http.Request, strategies Strategies, of *Flow) (*Flow, error) {
-	nf, err := NewFlow(conf, exp, csrf, r, strategies, of.Type)
+	f := of.Type
```
aeneasr, line 136: Is this method only called for expired flows, or could it break some other behavior?

aeneasr (commented):
For web-based flows we have a test that recovers / verifies an account with an expired link here:

https://github.com/ory/kratos/blob/c5a04b5f174e06faca99ebc7461c8ebe8e1f694d/test/e2e/cypress/integration/profiles/recovery/recovery/errors.spec.ts#L69-L82

I am under the impression that this test is actually recovering the account using an API flow, and then testing for this bug. So it appears that this bug is actually not happening on account recovery, only on verification?

If you (more or less) copy the test code of recovery to verification as well:

https://github.com/ory/kratos/blob/c5a04b5f174e06faca99ebc7461c8ebe8e1f694d/test/e2e/cypress/integration/profiles/verification/verify/errors.spec.ts#L50-L63

we would have a full e2e test covering this use case. Could you please add that?

It is a bit mysterious to me though why recovery seems to pass without our changes? But I do remember that I at one point introduced a fix for this...

selfservice/flow/recovery/flow.go
Demonsthere: If i understand [this code](https://github.com/ory/kratos/blob/master/selfservice/flow/verification/error.go#L70) correctly, then yes, only if we get an `ExpiredError`

aeneasr: It appears that the code is a bit messy, for some of the flows we set the flow type to browser:

https://github.com/ory/kratos/blob/c5a04b5f174e06faca99ebc7461c8ebe8e1f694d/selfservice/strategy/link/strategy_recovery.go#L310

for others, we seem to reuse the flow type:

https://github.com/ory/kratos/blob/c5a04b5f174e06faca99ebc7461c8ebe8e1f694d/selfservice/strategy/link/strategy_recovery.go#L244

so we need to make sure that this PR here does not add any unwanted regressions :/

Demonsthere: If this exists only on verification, i can remove the added logic from recovery flow, that should ensure we do not introduce any regressions there

test/e2e/cypress/support/commands.ts
```diff
@@ -402,6 +402,51 @@ Cypress.Commands.add('recoverApi', ({ email, returnTo }) => {
     })
 })
 
+Cypress.Commands.add('verificationApiExpired', ({ email, returnTo }) => {
```
zepatrik, line 405: This is not creating an expired flow, but creates one and checks if it is expired right? Do you first have to set `cy.shortVerificationLifespan()` for this to work? Could we make it part of this command then?

selfservice/flow/recovery/flow.go
zepatrik: So I checked, this is only called on expired flows.

selfservice/flow/verification/flow.go
zepatrik: Same here, only expired.

zepatrik (commented):
Looks solid IMO.

test/e2e/cypress/support/commands.ts
Demonsthere: Good point, I've set the short lifespan in the test before calling this, might be a better approach to include directly in the call

test/e2e/cypress/support/commands.ts
```diff
@@ -402,6 +402,52 @@ Cypress.Commands.add('recoverApi', ({ email, returnTo }) => {
...
+  cy.shortVerificationLifespan()
```
zepatrik, line 406: This will affect following tests right? Can you maybe also after creating the expired flow, reset the config?

Demonsthere: Sure, I could, but we do reset it for each test case [here](https://github.com/ory/kratos/blob/master/test/e2e/cypress/integration/profiles/verification/verify/errors.spec.ts#L40)

zepatrik: I see, then this is fine :+1:

zepatrik (approved):

aeneasr:
Thank you! I wanted to add a little test to recovery as well and while doing so I realized that the bug is still there. You can find my fixes here: https://github.com/ory/kratos/pull/2250/commits/8d191d59324622786392a4ebd5470cf06d118019

The problem was basically the redirection logic as it was still using the old flow type, hence redirecting to a JSON endpoint!

aeneasr (approved):

selfservice/flow/recovery/error.go
```diff
@@ -83,7 +83,10 @@ func (s *ErrorHandler) WriteFlowError(
 			return
 		}
 
-		if f.Type == flow.TypeAPI || x.IsJSONRequest(r) {
+		// We need to use the new flow, as that flow will be a browser flow. Bug fix for:
+		//
+		// https://github.com/ory/kratos/issues/2049
```
zepatrik, line 88:
```suggestion
		// https://github.com/ory/kratos/issues/2049!!
```

https://github.com/ory/closed-reference-notifier/commit/5d419e0124280696469e6e862a4ab1cfb39f9d84

selfservice/flow/verification/error.go
```diff
@@ -81,7 +81,10 @@ func (s *ErrorHandler) WriteFlowError(
 			return
 		}
 
-		if f.Type == flow.TypeAPI || x.IsJSONRequest(r) {
+		// We need to use the new flow, as that flow will be a browser flow. Bug fix for:
+		//
+		// https://github.com/ory/kratos/issues/2049
```
zepatrik, line 86:
```suggestion
		// https://github.com/ory/kratos/issues/2049!!
```

PR #2316, opened 2022-03-17
feat: TLS certs auto-reload

StarAurryon:
Kubernetes is able to auto-renew certificates with cert-manager and update files in the container. I would be cool if Ory Kratos could support auto certificate update on file change.

## Related issue(s)
None

## Checklist

      introduces a new feature.
- [X] I am following the
      the changes.
      works.

## Further Comments
Adds dependencies to fsnotify to support Mac/Windows/Linux.

This is an example of a working PoC. I need feedback, 95% of the changes are similar as proposed for hydra ;-)

### DIFF
--- a/cmd/daemon/cert_refresh.go
+++ b/cmd/daemon/cert_refresh.go
@@ -1,0 +1,150 @@
+package daemon
+
+import (
+
+	// #nosec G505 - This is required for certificate chains alongside sha256
+	"crypto/tls"
+	"crypto/x509"
+	"path"
+	"runtime"
+	"sync"
+	"time"
+
+	"github.com/ory/kratos/driver"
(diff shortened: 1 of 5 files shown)


cmd/daemon/cert_refresh.go
```diff
@@ -0,0 +1,150 @@
+package daemon
```
aeneasr, line 1: Would it be possible to add some tests for tis file? In case you wish to re-use it in Ory Hydra, it might also make sense to move this into ory/x so that it can be used across different repos :)

aeneasr (changes requested):

aeneasr:
Closing due to inactivity

PR #2318, opened 2022-03-18
feat: switch to opentelemetry tracing

icyphox:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


aeneasr (changes requested):
Thank you! Similar to the PR in Keto, could you please update the examples around tracing here too - I think we have examples in the docker compose files as well as the documentation!

embedx/config.schema.json
```diff
@@ -1827,160 +1827,7 @@
...
-              "examples": [
-                {
-                  "server_url": "http://localhost:9411/api/v2/spans"
-                }
-              ]
-            }
-          }
-        }
-      }
+      "$ref": "ory://tracing-config"
```
zepatrik, line 1830: This will work internally, but not for others consuming our schema. You will have to add a small script that renders the config schema as we do in Keto: https://github.com/ory/keto/pull/740
We can make that a separate PR, but it will block release. Maybe it would also be better to write a small go utility instead of bash scripts, but that can also be done later.

.github/workflows/ci.yaml
```diff
@@ -259,7 +259,7 @@ jobs:
       - uses: ory/ci/releaser/render-version-schema@master
         with:
           token: ${{ secrets.ORY_BOT_PAT }}
-          schema-path: embedx/config.schema.json
+          schema-path: .schema/config.schema.json
```
aeneasr, line 262: Can you please change this back and remove the file `.schema/config.schema.json`?

cmd/courier/watch.go
```diff
@@ -47,16 +48,21 @@ func ServeMetrics(ctx cx.Context, r driver.Registry) {
...
-		n.Use(tracer)
-	}
-
 	n.UseHandler(router)
 
-	server := graceful.WithDefaults(&http.Server{
-		Addr:    c.MetricsListenOn(),
-		Handler: n,
-	})
+	var server *http.Server
```
aeneasr, line 53: It would probably be easier to have `var h http.Handler` and wrap it with otelx if tracing is enabled, instead of instantiating the server in the if statement:

```
var n http.Handler = negroni.New()

// ...

if tracer.isLoaded() {
  n = otelx.NewHandler...
}

server := ...

File: cmd/daemon/serve.go [hunk 4]
```diff
@@ -122,10 +121,19 @@ func ServePublic(r driver.Registry, wg *sync.WaitGroup, cmd *cobra.Command, args
 	}
 
 	certs := c.GetTSLCertificatesForPublic()
-	server := graceful.WithDefaults(&http.Server{
-		Handler:   handler,
-		TLSConfig: &tls.Config{Certificates: certs, MinVersion: tls.VersionTLS12},
-	})
+	var server *http.Server
+	if tracer := r.Tracer(ctx); tracer.IsLoaded() {
```
aeneasr, line 125: Similar here. Also, is there a reason why the handler can't be a function that we can pass to `n.Use()`? I think that would be less code and easier to handle. All you have to do is to implement a function that follows the interface of the argument to `n.Use`

script/render-schemas.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
aeneasr, line 1: Would it be possible to move this to go code? One option is to use go code generation. Another is to use `embed` in `otelx` and expose the schema as a global variable, that you then import here (see @zepatrik suggestion)

aeneasr (changes requested):

script/render-schemas.sh
icyphox: Bear with me because I'm new to this, but aren't we doing so already? See:
- https://github.com/ory/x/blob/999103ce50c0c88f193202c0371437eaeeb43fec/otelx/config.go#L28
- https://github.com/ory/x/blob/999103ce50c0c88f193202c0371437eaeeb43fec/configx/schema.go#L31

Hence, the otelx schema is already being loaded.
As I understand it, Patrik's suggested script is to simply replace the "internal" schema ref (`ory://tracing-config`) to one that is accessible "globally" (i.e. using the absolute URL).

aeneasr: Ah yes, perfect :) So we can remove this file then?

zepatrik: We need this for schemas to be consumed externally, like through schemastore.

aeneasr: Ah, I see. In that case a bit of documentation what this script's purpose is and how to use it would be appreciated. It's also a bit confusing now, because we have two identical config schemas and it's not clear what's what, especially since the newly generated schema lives now under the location where the old schema was stored previously. Is there a better way to publish this for schemastore?

zepatrik: We can put it anywhere we want to. The new schema is now at `.schema/config.schema.json`, while the internal one is still under `embedx`. Any consuming go service would still use the one from `embedx`, but it also has to register the dependency schemas. That's why there should be a helper function that does exactly that, basically the same as https://github.com/ory/keto/blob/273b3eada1b376d4b09a0f286d87a01e2d619ef8/embedx/schemas.go#L28-L40.

I'll write something better in go later, but did not have time yet. This was only a hack to get it working.

zepatrik: We could then also organize the various openapi specs properly. Not sure what the best naming would be for both of those assets.

icyphox: Folks, so what's the resolution on this? Shall I just plop it into `.schemastore/`?

aeneasr: Yes? :)

aeneasr (commented):
Once the rendered schema is moved to e.g. `.schemastore/` along the script, updated CI definition, and a small README this is good to merge :)

aeneasr:
The updated schema store location will also need PRs against: https://github.com/SchemaStore/schemastore/blob/cc6104c5127ccb681600728c1bd7f116aebca51a/src/api/json/catalog.json#L1933

Can you open those too please @icyphox ?

sidharthramesh:
The documentation still says Open Telemetry is not supported - https://www.ory.sh/docs/kratos/guides/tracing

Can Kratos propagate open telemetry contexts?

PR #2313, opened 2022-03-17
fix: Add the expired verification flows return_to to the error object

MilesNorton:
## Related issue(s)
Closes #2279 

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [x] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [x] I have added or changed [the documentation](docs/docs).

Issue #2285, opened 2022-03-04
The `return_to` on recovery is not being returned with the settings flow after the recovery email has been clicked


MilesNorton:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

When I initialise a recovery browser flow with a `return_to` param and then submit the recovery form and then click on the link in my email which redirects me to the settings page, I'm not able to grab the `return_to` in the settings browser flow return object.

### Reproducing the bug

1. Initiate a recover browser flow with a return_to param
2. Go to email and click on recovery link

### Relevant log output

Request_url in db:
http://localhost:4433/self-service/recovery?flow=29fb8969-276d-45cf-a63a-955dae408d04&token=kyEgHVr8Snjs9CKDsG1TVuCECXaW6Qql

### Version

image: oryd/kratos:v0.8.2-alpha.1

### In which environment are you deploying?

Kubernetes with Helm

MilesNorton:
Similar discussion/issue to:
#2279

jacoblehr:
Looks like this one has been resolved in the [0.8.3-alpha.1.pre.0] (2022-01-21) release; see [https://github.com/ory/kratos/commit/b925d351dd0ce48cb6aed046dcf2698796453751](https://github.com/ory/kratos/commit/b925d351dd0ce48cb6aed046dcf2698796453751)

MilesNorton:
I've tested this locally and it isn't working still. This only occurs if it has been clicked before and therefore expired

Issue #2279, opened 2022-03-03
Clicking on an old verification link redirects you to the default return url and not the return url specified with the initial verification flow


MilesNorton:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

I signed up with a new account using the following queries: `return_to` and `after_verification_return_to`, which are both different to the `default_return_to` url. Once I receive the email to verify my account and clicked on the link, I get redirected to the `after_verification_return_to` url which is expected. 
Now, when I go back to my email and click on the verification email again it redirects me to the `default_return_to` url.

 
Please note I have also tested this from the normal verification service flow and it also produces the same bug which means it isn't specific to the `after_verification_return_to` url.

### Reproducing the bug

1. Set up second domain that will be used for return to url's
2. Navigate to the registration page and make sure the second domain is set as the  `return_to` and `after_verification_return_to` query params.
3. Sign up
4. Click on verification email link 
5. Click on verification email link again

### Relevant configuration

```yml
serve:
  public:
    base_url: http://localhost:4433/
  admin:
    base_url: http://localhost:4434/

selfservice:
  default_browser_return_url: http://localhost:8100
  whitelisted_return_urls:
    - http://localhost:8101/
  flows:
    registration:
      ui_url: http://localhost:8100/registration
      lifespan: 1h
      before: {}
      after:
        default_browser_return_url: http://localhost:8100
        password:
          hooks:
            - hook: session
    verification:
      ui_url: http://localhost:8100/verification
      after:
        default_browser_return_url: http://localhost:8100
      lifespan: 1h
      enabled: true
```

### Version

v0.8.2-alpha.1

aeneasr:
> I get redirected to the after_verification_return_to url which is expected.
[...]

Issue #2275, opened 2022-03-02
Pass back the 'return_to' url when errors are thrown so that we can redirect with the return_to to get a new flow


MilesNorton:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

When using Nextjs (react) to get or submit self service browser flows and you receive an error, it would be nice for the error message to return the `redirect_to` with the original `return_to` in the url. 

So initially when I called the page I passed in the return_to and then I reload the page with the Flow Id which renders everything. But if that flow id were to expire say, it would spit out an error and we would then need to call `'http://localhost:4433/self-service/login/browse` to initialize a new flow, at this point we don't have any notion of the return_to that was originally used.

Note, we use the return_to to render the page differently for different applications.

Here is an example of the error message return with the `self_service_flow_expired` error:
```
{
  error: {
    id: 'self_service_flow_expired',
    code: 410,
    status: 'Gone',
    reason: 'The login flow has expired. Redirect the user to the login flow init endpoint to initialize a new login flow.',
    details: { redirect_to: 'http://localhost:4433/self-service/login/browser' },
    message: 'The requested resource is no longer available because it has expired or is otherwise invalid.'
  }
}
```

Below is an example of the login screen, notice where I've commented `SOLVE:` which is what I'm having an issue with;

```tsx
type LoginProps = {
	flow: SelfServiceLoginFlow;
};

const Login: React.FC<LoginProps> = ({ flow }) => {
	const submitFlow = async (flowId: string, values: SubmitSelfServiceLoginFlowBody) => {
		await ory.submitSelfServiceLoginFlow(flowId, undefined, values)
            .catch(handleRedirectFlowError({
			    redirectCallback: clientRedirectHelper,
			    route: "login",
			    return_to: flow.return_to
            }))
[...]

PR #2333, opened 2022-03-22
fix: load return_to and append to errors

jacoblehr:
This merge request fixes the way that return_to urls are set on the self service flow objects; previously, this was only being set when marshalling the flow to JSON; I have added a call to set this when the flow is retrieved from the database. This means that the return_to can be appended to each of the error responses in the event that the flow is expired. There is an existing pull request that fixes this for the verification flow (I'm actually a colleague of the original submitter MilesNorton) but I think it's better in this case to fix all of these at once.

## Related pull requests
- https://github.com/ory/kratos/pull/2313

## Related issue(s)

- https://github.com/ory/kratos/issues/2275
- https://github.com/ory/kratos/issues/2279
- https://github.com/ory/kratos/issues/2285

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

N/A


selfservice/flow/login/handler.go
```diff
@@ -422,9 +422,17 @@ func (h *Handler) fetchFlow(w http.ResponseWriter, r *http.Request, _ httprouter
 
 	if ar.ExpiresAt.Before(time.Now()) {
 		if ar.Type == flow.TypeBrowser {
+			redirectURL := urlx.AppendPaths(h.d.Config(r.Context()).SelfPublicURL(), RouteInitBrowserFlow)
+			if ar.ReturnTo != "" {
+				query := redirectURL.Query()
```
aeneasr, line 427: Use `ory/x/urlx.CopyWithQuery`

selfservice/flow/verification/handler.go
```diff
@@ -248,9 +248,16 @@ func (h *Handler) fetch(w http.ResponseWriter, r *http.Request, _ httprouter.Par
...
 		if req.Type == flow.TypeBrowser {
+			redirectURL := urlx.AppendPaths(h.d.Config(r.Context()).SelfPublicURL(), RouteInitBrowserFlow)
+			if req.ReturnTo != "" {
+				query := redirectURL.Query()
+				query.Set("return_to", req.ReturnTo)
+				redirectURL.RawQuery = query.Encode()
+			}
+
 			h.d.Writer().WriteError(w, r, errors.WithStack(x.ErrGone.
-				WithReason("The verification flow has expired. Redirect the user to the verification flow init endpoint to initialize a new verification flow.").
```
aeneasr, line 252: Please re-add the reason

persistence/sql/persister_verification.go
```diff
@@ -33,6 +33,8 @@ func (p Persister) GetVerificationFlow(ctx context.Context, id uuid.UUID) (*veri
 		return nil, sqlcon.HandleError(err)
 	}
 
+	r.SetReturnTo()
```
aeneasr, line 36: Implement `db.Scan` instead which sets this, then this field will always be set when loaded from the DB

selfservice/flow/settings/handler.go
```diff
@@ -377,9 +377,16 @@ func (h *Handler) fetchFlow(w http.ResponseWriter, r *http.Request) error {
 
 	if pr.ExpiresAt.Before(time.Now().UTC()) {
 		if pr.Type == flow.TypeBrowser {
+			redirectURL := urlx.AppendPaths(h.d.Config(r.Context()).SelfPublicURL(), RouteInitBrowserFlow)
+			if pr.ReturnTo != "" {
+				query := redirectURL.Query()
+				query.Set("return_to", pr.ReturnTo)
+				redirectURL.RawQuery = query.Encode()
+			}
```
aeneasr, line 385: This is repetitive, you could instead create a function which generates the detaiols to pass to the error below. That way you can just reuse this. Or, alternative, create an error creation function which you use and pass the config and returnTo as args

aeneasr (changes requested):

jacoblehr:
Thanks for your feedback @aeneasr - good suggestions all round 😄 

I'll work on the suggested improvements and update the PR

selfservice/flow/login/handler.go
jacoblehr: Updated

selfservice/flow/verification/handler.go
jacoblehr: Updated

aeneasr:
Could you please resolve the CI issues? :) 

https://github.com/ory/kratos/runs/5676532411?check_suite_focus=true#step:13:41

persistence/sql/persister_verification.go
jacoblehr: It looks like this one may be a little tricky - I've tried making ReturnTo a new type that just wraps string, and implemented the Scan method on the new type (along with setting the db tag to 'request_url' and the rw tag to 'r'), but it doesn't seem to be picking this up when I re-run the tests with this change (presumably as that column is already referenced?)

Is this what you were thinking @aeneasr or did you have another approach in mind?

selfservice/flow/settings/handler.go
jacoblehr: Updated

persistence/sql/persister_verification.go
aeneasr: Sorry yeah, it's indeed not possible to use a custom `db.Scan` because we this won't work on the outer struct. I thought sql uses the same concept as json but apparently not so!

selfservice/flow/login/handler_test.go
```diff
@@ -553,18 +554,24 @@ func TestGetFlow(t *testing.T) {
 	})
 
 	t.Run("case=expired with return_to", func(t *testing.T) {
```
aeneasr, line 556: I ran these tests without your modification and they passed, can you please revert this test and add tests that would fail without the modifications? Preferrably:

1. tests that reside in `/selfservice/flow/<the-flow-type>/test` where you create a flow and then get it, and expect the return url to be set.
2. At least one test that would otherwise fail, either in handler_test or in the e2e tests

aeneasr (changes requested):
I've updated the PR using pop's callbacks which is a bit more save as they will always run :)

As noted in one of the comments, could you pelase add tests for these? A good indicator of whether tests are effective is to see if they pass when you comment your fixes. If they do pass there as well, they are not effective (as they do not test your changes but something else).

selfservice/flow/login/handler_test.go
jacoblehr: Nice work :D I haven't used Soda before but I was going to suggest using post-load hooks if they were available so great to see it working!

Hmm... on reverting back to my commit, commenting out the r.SetReturnTo in persister_login.go, I get the following:

Screen Shot 2022-03-28 at 5 45 01 pm

I am running the tests with `go test -p 1 -tags sqlite -count=1 -failfast ./selfservice/flow/login`

aeneasr: You're right, not sure why it passed for me! Thank you for validating :)

vinckr:
Hello  @jacoblehr 

Issue #707, opened 2020-09-28
Support OAuth2 and OpenID Connect in API-based flows


aeneasr:
**Is your feature request related to a problem? Please describe.**

It's currently not possible to use Google to sign in through ORY Kratos' Admin APIs which would be necessary in mobile environment (e.g. React Native).

**Describe the solution you'd like**

This should be implemented. Keep in mind that Facebook does not allow login through a browser, but instead requires us to use their SDK.

splaunov:
Can we just add auth code and access/refresh token as parameters to the submit method of the oidc flow?
Would it be the right solution?
On mobile platforms it's not hard to get auth code or access token directly from OAuth provider with numerous plugins available. Only the last step is missing: we need to exchange OAuth code/token for Kratos session token.

aeneasr:
The access token would not be a good option because that would open a very specific vulnerability of OAuth2 (as it does not prove authentication but authorization). There is a good resource on that [here](https://oauth.net/articles/authentication/).

The most secure and interoperable option would be an OpenID Connect ID Token. Here, the proposed submit flow could work! I am not sure if all social sign in providers support that but for those that do, this could be an option. For those that do not (like facebook) we probably have to implement some custom stuff anyways!

splaunov:
Yes, I've just misused the terminology. It definitely should be an ID Token.
And auth code should also work fine. If one prefers to exchange it for ID token server-side. Right?

I would try to implement this if the concept seems right to you.

splaunov:
My first step would be to separate 'browser' and 'api' tests in `selfservice/strategy/oidc/strategy_test.go`

aeneasr:
That's a good first step :)

splaunov:
Remove flow type check in Strategy.PopulateLoginMethod

aeneasr:
👍

splaunov:
This is what I've done so far as an experiment:
- Deleted throwing of `ErrAPIFlowNotSupported` in `oidc.Strategy.validateFlow()`.
- Deleted API flow blocking in `oidc.Strategy.PopulateLoginMethod()`.
- Changes in `oidc.Strategy.Login()`:
    - Added validation check if `id_token` parameter present for API flow.
    - Added id token claims extraction.
    - Added search of identity in db by `claims.Subject`.
    - Added code to create and persist new identity if db search failed with `ErrNoRows`.
      Duplicated this code from `oidc.Strategy.processRegistration()`.

[...]

PR #2346, opened 2022-03-28
feat: support OAuth2 and OpenID Connect in API-based flows

splaunov:
## Related issue(s)

#707 

## Checklist

- [ x] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [ x] I am following the
      the changes.
      works.


aeneasr:
Bruh are you for real? Awesome!

Could you please provide a high-level overview of how this feature worked? Preferably with an example for e.g. the Facebook SDK?

embedx/config.schema.json
```diff
@@ -435,6 +435,12 @@
         },
         "requested_claims": {
           "$ref": "#/definitions/OIDCClaims"
+        },
+        "skip_audience_check": {
```
aeneasr, line 439: Why do we need this?

selfservice/strategy/oidc/provider.go
```diff
@@ -12,6 +12,8 @@ type Provider interface {
 	Config() *Configuration
 	OAuth2(ctx context.Context) (*oauth2.Config, error)
 	Claims(ctx context.Context, exchange *oauth2.Token) (*Claims, error)
+	ClaimsFromIdToken(ctx context.Context, rawIdToken string) (*Claims, error)
+	ClaimsFromAccessToken(ctx context.Context, accessToken string) (*Claims, error)
```
aeneasr, line 16: Instead of adding these two to the base interface, you could create a new interface which can then be used in type assertion to check whether it implements thes functions :)


```
type APIFlowProvider struct {
 	ClaimsFromIdToken(ctx context.Context, rawIdToken string) (*Claims, error)
 	ClaimsFromAccessToken(ctx context.Context, accessToken string) (*Claims, error)
}

// ...
if p, ok := provider.(APIFlowProvider); ok {
  // do something
}
```

selfservice/strategy/oidc/provider_generic_oidc.go
```diff
@@ -103,10 +108,43 @@ func (g *ProviderGenericOIDC) Claims(ctx context.Context, exchange *oauth2.Token
...
+func (g *ProviderGenericOIDC) ClaimsFromIdToken(ctx context.Context, rawIdToken string) (*Claims, error) {
+	p, err := g.provider(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return g.verifyAndDecodeClaimsWithProvider(ctx, p, rawIdToken)
+}
+
+func (g *ProviderGenericOIDC) ClaimsFromAccessToken(ctx context.Context, accessToken string) (*Claims, error) {
```
aeneasr, line 123: Using an access token to authenticate a user is extremely dangerous. The problem is that an evil user eve can create a legitimate OAuth2 App, trick good user alice to use that OAuth2 app, and then use the access token eve received from that app at Ory Kratos, even though the alice never authenticated in Ory Kratos.

Therefore, we need to axe this flow. It can only ever be possible to use `ClaimsFromIDToken()`

selfservice/strategy/oidc/provider_generic_oidc.go
```diff
@@ -83,8 +83,13 @@ func (g *ProviderGenericOIDC) AuthCodeURLOptions(r ider) []oauth2.AuthCodeOption
 	return options
 }
 
-func (g *ProviderGenericOIDC) verifyAndDecodeClaimsWithProvider(ctx context.Context, provider *gooidc.Provider, raw string) (*Claims, error) {
-	token, err := provider.Verifier(&gooidc.Config{ClientID: g.config.ClientID}).Verify(ctx, raw)
+func (g *ProviderGenericOIDC) verifyAndDecodeClaimsWithProvider(ctx context.Context, provider *gooidc.Provider, rawIdToken string) (*Claims, error) {
+	token, err := provider.
+		Verifier(&gooidc.Config{
+			ClientID:          g.config.ClientID,
+			SkipClientIDCheck: g.config.SkipAudienceCheck,
```
aeneasr, line 90: This is dangerous and we need to remove this feature for the same reason we need to remove the `ClaimsFromAccessToken` one.

```suggestion
```

aeneasr (changes requested):

splaunov:
Just submit login flow with 

```
{
  "method": "oidc",
  "porvider": "name it",
  "id_token": "your token"
}
```

or

```
{
  "method": "oidc",
  "porvider": "name it",
  "access_token": "your token"
}
```

in the body.
As simple as that.
Have tested it with google and apple. Not sure about Facebook.

embedx/config.schema.json
splaunov: This was just a workaround to the following problem.
go-oidc simply compares aud to client id. But Apple puts different values to the audience field for iOS and web.
They use AppId for iOS and ServiceId (ClientId) for web.

So, the best solution would be to extend go-oidc to check list of allowed aud values instead of single ClientId value.

For now I'm removing `skip_audience_check` parameter from this PR. But be aware that iOS clients will not work until go-oidc is extended.

selfservice/strategy/oidc/provider_generic_oidc.go
splaunov: Yes, we have discussed this earlier.
Will remove `access_token` option.

embedx/config.schema.json
aeneasr: > go-oidc simply compares aud to client id. But Apple puts different values to the audience field for iOS and web.
They use AppId for iOS and ServiceId (ClientId) for web.

I see, we then definitely need to add the ability to use a custom audience instead. Do you have an example you could show so I understand the problem? Unfortunately I don't have any mobile programming experience with this. 

Is it not possible to register to apple providers, so one for web and one for iOS in Ory Kratos?

selfservice/strategy/oidc/provider_generic_oidc.go
aeneasr: Nice, thanks :)

embedx/config.schema.json
splaunov: > Do you have an example you could show so I understand the problem?


Apple recommends these identifiers to be in a reverse domain name notation. But you can also use any unique string.

> Is it not possible to register to apple providers, so one for web and one for iOS in Ory Kratos?


Should be possible. Not sure what consequences it will have but this will make app management more complex.
And this might not be an option for those who migrates existing apps to Kratos.

I have also created an issue to go-oidc project: https://github.com/coreos/go-oidc/issues/334

splaunov: I'm adding `audiences` list into provider config. If the list is empty, then the audience will be checked against client id, as it's implemented now. So, this will not be a breaking change.
@aeneasr please let me know if you see any disadvantages in such solution.

aeneasr:
@splaunov feel free to re-request a review any time you're ready :)

selfservice/strategy/oidc/provider.go
splaunov: done

selfservice/strategy/oidc/provider_generic_oidc.go
splaunov: done

selfservice/strategy/oidc/provider_generic_oidc.go
splaunov: done

aeneasr:
I tried pushing some changes required for merging the PR to your fork & branch, but it appears that I am not allowed to do so 😕 

```
% git push ...
ERROR: Permission to push denied to aeneasr.
fatal: could not read from the remote repository.

Please make sure that you have the correct access rights
and the repository exists.
```

But the good news is, giving access is easy! :relaxed: All you need to do is [enable write access for maintainers](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). Thank you! 😄

If the repository belongs to an organization, please add me for the [project as a collaborator](https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/adding-outside-collaborators-to-repositories-in-your-organization)!

aeneasr:

```
$ make format
$ git push
```

Thank you!

embedx/config.schema.json
```diff
@@ -448,6 +448,21 @@
         },
         "requested_claims": {
           "$ref": "#/definitions/OIDCClaims"
+        },
+        "audiences": {
```
aeneasr, line 452: Maybe rename this to `allowed_audiences`?

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -69,6 +71,11 @@ type SubmitSelfServiceLoginFlowWithOidcMethodBody struct {
 
 	// The identity traits. This is a placeholder for the registration flow.
 	Traits json.RawMessage `json:"traits"`
+
+	// Only used in API-type flows, when an id token has been received by mobile app directly from oidc provider.
+	//
+	// required: false
+	IdToken string `json:"id_token"`
```
aeneasr, line 78: In Golang, it is common to use upper-case for abbreviations, so

```suggestion
	IDToken string `json:"id_token"`
```

selfservice/strategy/oidc/provider.go
```diff
@@ -15,6 +15,10 @@ type Provider interface {
 	AuthCodeURLOptions(r ider) []oauth2.AuthCodeOption
 }
 
+type APIFlowProvider interface {
+	ClaimsFromIdToken(ctx context.Context, rawIdToken string) (*Claims, error)
```
aeneasr, line 19: In Go it is common to name abbreviations with upper case:

```suggestion
	ClaimsFromIDToken(ctx context.Context, rawIDToken string) (*Claims, error)
```

selfservice/strategy/oidc/provider_config.go
```diff
@@ -94,6 +94,10 @@ type Configuration struct {
 	//
 	// More information: https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter
 	RequestedClaims json.RawMessage `json:"requested_claims"`
+
+	// List of values to check audience field of ID Token.
+	// The audience field of ID Token should be equal to one of the items in this list.
+	Audiences []string `json:"audiences"`
```
aeneasr, line 100: To make the use of this config clear I suggest renaming this to:

```suggestion
	AllowedAudiences []string `json:"allowed_audiences"`
```

selfservice/strategy/oidc/provider_generic_oidc.go
```diff
@@ -97,16 +109,31 @@ func (g *ProviderGenericOIDC) verifyAndDecodeClaimsWithProvider(ctx context.Cont
...
 
+func verifyAudience(received []string, expected []string) error {
+	for _, r := range received {
+		for _, e := range expected {
+			if r == e {
+				return nil
+			}
+		}
+	}
+	return fmt.Errorf("oidc: audience not valid: %v", received)
```
aeneasr, line 120: To add a stack trace we use `pkg/errors`:

```suggestion
	return errors.Errorf("oidc: audience not valid: %v", received)
```

selfservice/strategy/oidc/provider_generic_oidc.go
```diff
@@ -97,16 +109,31 @@ func (g *ProviderGenericOIDC) verifyAndDecodeClaimsWithProvider(ctx context.Cont
 	return &claims, nil
 }
 
+func verifyAudience(received []string, expected []string) error {
```
aeneasr, line 112: While the code is simple enough, it would be good to still have a test so that noone can introduce a regression here ever.

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -162,14 +188,133 @@ func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow,
...
+
+			s.d.Logger().
+				WithRequest(r).
+				WithField("oidc_provider", provider.Config().ID).
+				WithSensitiveField("oidc_claims", claims).
+				WithField("mapper_jsonnet_output", evaluated).
+				WithField("mapper_jsonnet_url", provider.Config().Mapper).
+				Debug("OpenID Connect Jsonnet mapper completed.")
+
+			//option, err := decoderRegistration(s.d.Config(r.Context()).DefaultIdentityTraitsSchemaURL().String())
```
aeneasr, line 264: Do we still need these comments?

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -162,14 +188,133 @@ func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow,
...
+			i.SetCredentials(s.ID(), *creds)
+
+			if err := s.d.IdentityManager().Create(r.Context(), i); err != nil {
+				if errors.Is(err, sqlcon.ErrUniqueViolation) {
+					return nil, schema.NewDuplicateCredentialsError()
+				}
+				return nil, errors.WithStack(err)
+			}
+
+			//i.Traits = identity.Traits(fmt.Sprintf("{\"phone\": \"%s\"}", p.Phone))
```
aeneasr, line 298: Consider removing these comments

aeneasr (changes requested):
Thank you for these great changes, I think we are now on track to get this shipped. I haven't had a lot of time to really look into the details of the implementation yet (currently on vacation) but from a high level it looks very good.

One question I have is how we deal with validation errors? In the browser flow, if a validation error occurrs (e.g. some required field is missing), we show the registration form again and ask the user to fill out their details. How does that work for API flows?

Another question is, how do we deal with detecting whether we need to execute login or registration? How is that information relayed to the application? In the browser flow, we just redirect to the appropriate login (user exists) or registration (user does not exist yet) endpoint.

Another thing we will need to add to this PR is an e2e test for this new feature. We already have e2e tests for browser OIDC flows which you can find here:

- https://github.com/ory/kratos/blob/master/test/e2e/cypress/integration/profiles/oidc/registration
- https://github.com/ory/kratos/tree/master/test/e2e/cypress/integration/profiles/oidc/login
- https://github.com/ory/kratos/tree/master/test/e2e/cypress/integration/profiles/oidc/settings

The profile for mobile apps (API flows) you can find here: https://github.com/ory/kratos/tree/master/test/e2e/cypress/integration/profiles/mobile

It will probably not be trivial to add OIDC tests to the mobile app as the flow works differently (we need a valid ID token) but it is a requirement from our end to get this merged into master, as it is the only way to verify that everything works as expected from end-to-end. The tests will need to cover:

1. Login with an ID Token and a user which exists
2. Login with an ID Token and a user which does not exist -> we end up registering a new user
3. Registration with an ID Token and a user which exists -> we end up signed in
4. Registration with an ID Token and a user which does not exist, and validation errors (e.g. website missing)
5. Successful registration with an ID Token
7. Linking and unlinking OIDC providers in the settings flow
6. Error cases for each flow (login, registration, settings)
    1. Audience mismatch
    2. Signature key mismatch
    3. ID Token generally invalid
    4. Issuer mismatch

splaunov:
> I tried pushing some changes required for merging the PR to your fork & branch, but it appears that I am not allowed to do so 😕

That's strange. I added you as a collaborator some time ago and you've already pushed some changes to our fork, but to another branch.

image

splaunov:
> One question I have is how we deal with validation errors? In the browser flow, if a validation error occurrs (e.g. some required field is missing), we show the registration form again and ask the user to fill out their details. How does that work for API flows?
> Another question is, how do we deal with detecting whether we need to execute login or registration? How is that information relayed to the application? In the browser flow, we just redirect to the appropriate login (user exists) or registration (user does not exist yet) endpoint.

As for now, there is no registration API flow for oidc. New users are silently provisioned in login flow.
This works fine if identity has no required traits.

You are right in that we need to support registration flow as well.
To do this, I can move the provisioning code from login to registration flow. So, login flow will return error to mobile app if credentials not found and new registration flow will need to be started by the mobile app in this case.

selfservice/strategy/oidc/provider_generic_oidc.go
splaunov: Have added test to `selfservice/strategy/oidc/strategy_test.go`

splaunov:
>Another thing we will need to add to this PR is an e2e test for this new feature. We already have e2e tests for browser OIDC flows which you can find here:

This assumes extending the React Native app. It should get ID Token from Hydra.
I feel like I can't spend enough time for this in nearest future. If someone can help with RN, I would be just happy to work on cypress part.

aeneasr:
Thank you! I will for sure be able to help when I'm back from vacation in May. I think we need to iron out some of the big questions (registration vs login) before we can finalize it and merge it to upstream. But your work so far has been great! If you want to work on this in the meanwhile please do let me know if there's anything I can help with (e.g. pointers) :))

klueken:
Hi everyone, is there any planned work to come back and finish this PR? This is a critically important feature that is blocking my project.

I'm not well-versed in Golang, but could try to help with documentation or testing if that helps move this forward.

aeneasr:
Hey there, I think this is a great PR and feature, and it's on our roadmap and backlog! For us unfortauntely it has not as high of a priority as a couple of other very urgent topics.

Probotect0r:
@aeneasr Are there any alternative solutions for mobile based OAuth flows while this PR is still in progress?

hernangonzalez:
Would be very nice if this could be prioritized :)

splaunov:
Have added support for apps using WebView for oidc logins

dmakarenko:
@splaunov thank you for pushing this feature forward 🙏 

I have a question though. I am not sure how the proposed WebView flow would work for native apps. AFAIK the final redirect often relies on a custom URL scheme (at least on iOS) which passes the control to the application and the only information available to the native app is the URL so any extra details (like credentials) should be passed via query params. Usually mobile apps don't have access to the WebView context (for security and privacy reasons) especially if something like [Apple's Authentication Services framework](https://developer.apple.com/documentation/authenticationservices) is used

There is an open discussion on that topic here: https://github.com/ory/kratos/discussions/2434

It could be that I am missing something here. If so, my apologies in advance.

David-Wobrock:
Hi all 👋 
First of all, thanks a lot for working on this. It seems to me that it will play an important role in the adoption of Kratos :)

Two questions:

1/ About this:

> > One question I have is how we deal with validation errors? In the browser flow, if a validation error occurrs (e.g. some required field is missing), we show the registration form again and ask the user to fill out their details. How does that work for API flows?
> > Another question is, how do we deal with detecting whether we need to execute login or registration? How is that information relayed to the application? In the browser flow, we just redirect to the appropriate login (user exists) or registration (user does not exist yet) endpoint.
> 
> As for now, there is no registration API flow for oidc. New users are silently provisioned in login flow. This works fine only if identity has no required traits.
> 
> You are right in that we need to support registration flow as well. To do this, I can move the provisioning code from login to registration flow. So, login flow will return error to mobile app if credentials not found and new registration flow will need to be started by the mobile app in this case.

As @aeneasr stated, in the current social browser flow, the decision between login and register is made when the callback is called whether the user (or more precisely the OIDC credentials) exist or not.
I'm wondering whether it would make sense to have a similar logic for the API flow.
Basically, somewhere here https://github.com/ory/kratos/pull/2346/files#diff-a847b9b099ed3a4c0c0eed58467442371554463a07f865dcaa3d8f034759f875R233 when we try to log the user in, but it doesn't exist, switch over to the registering logic.

So that we can avoid having the mobile app starting a registration flow when a new user connects through a social provider.

2/ About the method used.
I completely understand that passing the `id_token` from the mobile app to the server is enough for the server to connect/create a user, as the JWT contains enough information.
Would it also work to take as input the authorization code, so that Kratos uses its client secret to exchange the code for the oauth2 tokens? (and therefore it could re-use the existing social callback logic).
A bit like it is described in [this documentation from Google](https://developers.google.com/identity/sign-in/android/offline-access).

sayoun:
Thanks for this PR, we really need this feature and want to help, can someone list what is missing here so it could be merged ? @aeneasr or @splaunov ?

splaunov:
Here is what missed to my understanding:

1. We need an end-to-end test. To make it happen, React Native app should be extended first to register/login with pure api and WebView flows using Hydra for OIDC. And then these flows should be covered by cypress tests.
2. A review of my recently added WebView related code needed. @aeneasr

splaunov:
> @splaunov thank you for pushing this feature forward 🙏
> 
> I have a question though. I am not sure how the proposed WebView flow would work for native apps. AFAIK the final redirect often relies on a custom URL scheme (at least on iOS) which passes the control to the application and the only information available to the native app is the URL so any extra details (like credentials) should be passed via query params. Usually mobile apps don't have access to the WebView context (for security and privacy reasons) especially if something like [Apple's Authentication Services framework](https://developer.apple.com/documentation/authenticationservices) is used
> 
> There is an open discussion on that topic here: #2434
> 
> It could be that I am missing something here. If so, my apologies in advance.

@dmakarenko yes, you're right. I have added the token to query params of the final redirect.
Thanks for pointing it out!

peturgeorgievv:
Can we merge this, as we really need this functionality in our app?

chancezeus:
Hi @splaunov (and @aeneasr),

Since I'm interested in having this feature in native apps as well I did some checking on the code changes that are made. Most of them look OK but I have to put some question marks at the "webview" flow, the reason mainly being that it would be all to easy for the dev to not open it in an actual webview but to delegate it to the system browser and then redirecting back using a "custom url scheme". This on it's own would not be a problem if those custom schemes would actually require some form of registration at a central provider but unfortunately (or luckily depending on how you look at it) they don't and thus it'd be possible for any app to hijack the redirect and have access to the session and/or session_token. I would suggest making the flow similar to the OAuth2 with PKCE flow where the app needs to generate a challenge + verifier, first it sends the verifier together with the flow request, opens the browser, get's back a code and then exchanges the code together with the challenge for the actual session and session_token. This makes it more flexible, secure and actually easier (since we don't have to decode the redirect url query string) at the cost of a single extra request.

FYI: I'm not a Go dev but I've been doing web and app development for over a decade and been coding for even longer and in that time I've written a lot of code in a lot of different (programming) languages (and read even more in even more (programming) languages) so although I don't know the specifics of Go, I do have a quite broad "general programming" skill-set and can read it well enough to understand what's happening.

splaunov:
Hi @chancezeus!
Thank you for the feedback.
I agree in that WebView method is known to have some weeknesses.
But it's handy to have it in the tool chest when the task is to migrate legacy apps to kratos. When app's installed base is huge and we can't upgrade it in one go.
So, I would keep it even if there are more advanced/secure methods to be added in the future.

chancezeus:
> Hi @chancezeus! Thank you for the feedback. I agree in that WebView method is known to have some weeknesses. But it's handy to have it in the tool chest when the task is to migrate legacy apps to kratos. When app's installed base is huge and we can't upgrade it in one go. So, I would keep it even if there are more advanced/secure methods to be added in the future.

I'm not saying we should remove it (especially since we'd need it for some flows where a native SDK wouldn't be available or provide an ID token (like Facebook and/or potentially some generic provider), I'm just suggesting we should use a code exchange like flow with PKCE similar to OAuth2 to make it more secure so that a malicious party is less likely to hijack the session token.

chancezeus:
_Note for previous remark:_ you wouldn't even need to put anything in the query, since we could use the flows id as the "code" and the challenge/verifiers are generated (and thus known) and send by the frontend anyway, so you could simply say:
1. If WebView is needed -> app MUST send a PKCE verifier together with the submit login/registration request
2. Kratos saves the PKCE verifier in the flow
3. Kratos returns URL
4. App opens URL (either in actual WebView OR in system browser)
5. WebView flow is run and in the end Kratos redirects to the given webview return url without any details
6. Application detects the return url in webview OR gets (re)opened by system browser
7. Application retrieves flow id/PKCE challenge from (secure) storage or memory
8. Application submits the flow again but this time it includes the "challenge" instead of the verifier
9. Kratos checks if the flow is in the correct state, i.e. steps 1 to 5 have completed correctly, if not return the correct error (login cancelled, etc) -> app must restart flow from at least step 1
10. Kratos checks the challenge against the verifier
11. If challenge is incorrect -> invalidate flow (session hijack) return error
12. Return session + session_token

=> You could/should even generalize this together with the regular browser flow so it wouldn't matter if you're using browser OR api, but that would be a (major) breaking change...

iakrevetkho:
Hi!
It it very common feature for us, because we are using `kratos` to authenticate users. But we can't add authentication via `Social networks` in our mobile app without it.
May we help you with it? Or what is blocking current review process?

jwmay2012:
> Does this PR enable me to use [Apple Authentication](https://github.com/invertase/react-native-apple-authentication) and [Google Sign In](https://github.com/react-native-google-signin/google-signin) libraries to get sign in tokens from the mobile device directly and then pass that token to Kratos over API in exchange for a Kratos auth token? Without webview?

To answer my own question, yes it does, and it works great!

And thanks @splaunov for the recent update from master. (I had done my own master -> feature/oidc-api update a few weeks ago and was going to offer it up this week, but you beat me to it 😄 )

@aeneasr @ory-team This PR passes all unit tests, builds, and functions well for us. We're running our development stacks from this build. Mobile API social logins are a requirement for us to go live with our switch to Ory, which we have operational with this PR. 

Are there any remaining barriers to getting this merged to master? 
If `webview` flow isn't considered complete until PKCE, could we split it out from the rest to get API Social flows merged? 

We were aiming to make the switch to Ory/Kratos end of January. Would be preferable to use official images instead of our own Kratos build pipeline.

go.mod
```diff
@@ -27,7 +27,7 @@ require (
 	github.com/bwmarrin/discordgo v0.23.0
 	github.com/bxcodec/faker/v3 v3.3.1
 	github.com/cenkalti/backoff v2.2.1+incompatible
-	github.com/coreos/go-oidc v2.2.1+incompatible
+	github.com/coreos/go-oidc/v3 v3.1.0
```
kszafran, line 30: BTW, looks like there's a newer version now:
```suggestion
	github.com/coreos/go-oidc/v3 v3.5.0
```

kmherrmann:
We are definitely highly interested to make this part of Kratos out of the box - we're just time constrained on our end. We'll give this a review soon!

septian-machtwatch:
> 

hi @jwmay2012 could you please explain the steps to exchange OIDC id token for kratos session using API? I also would like to implement the flow in our mobile app development

septian-machtwatch:
> > Does this PR enable me to use [Apple Authentication](https://github.com/invertase/react-native-apple-authentication) and [Google Sign In](https://github.com/react-native-google-signin/google-signin) libraries to get sign in tokens from the mobile device directly and then pass that token to Kratos over API in exchange for a Kratos auth token? Without webview?
> 
> To answer my own question, yes it does, and it works great!
> 
> And thanks @splaunov for the recent update from master. (I had done my own master -> feature/oidc-api update a few weeks ago and was going to offer it up this week, but you beat me to it 😄 )
> 
> @aeneasr @ory-team This PR passes all unit tests, builds, and functions well for us. We're running our development stacks from this build. Mobile API social logins are a requirement for us to go live with our switch to Ory, which we have operational with this PR.
> 
> Are there any remaining barriers to getting this merged to master? If `webview` flow isn't considered complete until PKCE, could we split it out from the rest to get API Social flows merged?
> 
> We were aiming to make the switch to Ory/Kratos end of January. Would be preferable to use official images instead of our own Kratos build pipeline.

Hi @jwmay2012 could you please explain the steps to exhcanges OIDC id token to Kratos session token using API? I also would like to implement the flow in our mobile app development.

MortalKastor:
Hi!

We're using the code from this PR on our backend, and I'm trying to do a Facebook login on our React Native app.

The issue I have is that the Facebook SDK doesn't give me an ID Token, but an Access Token and I can't get the API to accept it (sending it as `access_token` fails with `Authentication failed because no id_token was returned. Please accept the "openid" permission and try again.` (I verified `openid` was part of the scope passed to the SDK)).

Is there a way to do API Facebook login that's known to work, by any chance?

rivten:
Just for info, we implemented @MortalKastor's feature with the following patch. I thought we should share it with you all :) 

```
diff --git a/selfservice/strategy/oidc/provider_facebook.go b/selfservice/strategy/oidc/provider_facebook.go
index f875e592..fd9f3fe7 100644
--- a/selfservice/strategy/oidc/provider_facebook.go
+++ b/selfservice/strategy/oidc/provider_facebook.go
@@ -126,3 +126,12 @@ func (g *ProviderFacebook) Claims(ctx context.Context, exchange *oauth2.Token, q
 		Birthdate:         user.BirthDay,
 	}, nil
 }
+
+func (g *ProviderFacebook) ClaimsFromIDToken(ctx context.Context, rawIDToken string) (*Claims, error) {
+	claims, err := g.ProviderGenericOIDC.verifyAndDecodeClaimsWithProvider(ctx, g.ProviderGenericOIDC.p, rawIDToken)
+	if err != nil {
+		// some facebook SDK do not provide idToken after oauth, only an access_token so we need to use it instead
+		return g.Claims(ctx, &oauth2.Token{AccessToken: rawIDToken}, nil)
+	}
+	return claims, err
+}

```

hperl:
Hi everybody!

First of all a big thank you to @splaunov for this PR. Based on this work, we revisited the flow to get OIDC working on mobile and ultimately settled on an approach implemented here: https://github.com/ory/kratos/pull/3216.

The main difference between that PR and this is that we continue to let Kratos fetch the identity from the OIDC provider, which also means that we can continue using the more secure authorization code grant.

The gets the session token as follows:
* When initializing a flow, you can request a session token exchange code (only available for app flows)
* After the user logged in, you can exchange the code for a session token

If you are interested, please take a look at the implementation. I welcome any comments :).

And thanks for being part of the Ory Community ❤️

David-Wobrock:
Not really @aptgetgit, sorry.

But happy to see this feature to be merged into Kratos 👍 Thank you everyone involved ❤️

PR #2375, opened 2022-04-07
docs: schema fixes

### LINKED ISSUE #2357: Docs: messages are not required
### Preflight checklist

- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

~The `messages` field of the `uiNode` is marked as `required` in the HTTP API documentation. However, the field is not returned when getting a registration flow via `/self-service/registration/flows`.~

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

v0.9.0-alpha.3

_No response_

_No response_

### Additional Context

_No response_

woylie:
This PR updates the API schema.

resolves #2357 

- [x] turn `uiNodeType` into enum
- [x] turn `uiNodeGroup` into enum
- [x] ensure select field value for `attributes.node_type` matches actual node_type value (see screenshot)
- [x] add task for previewing API documentation
- [x] In `api.openapi.json`, the `LoginFlow` and the `VerificationFlow` list `type` as a required property. The other flows don't, and none of the flows in `openapi.json` list it. I assume the type should always be set to either `"api"` or `"browser"` and never be omitted or `null`. Is that so? If so, I'll add `type` as a required property to all the flows.

imagen

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,14 @@ $(call make-lint-dependency)
 docs/cli:
 		go run ./cmd/clidoc/. .
 
+.PHONY: docs/api
+docs/api:
+		npx @redocly/openapi-cli preview-docs spec/api.json
+
+.PHONY: docs/swagger
+docs/swagger:
+		npx @redocly/openapi-cli preview-docs spec/swagger.json
+
 .bin/ory: Makefile
 		bash <(curl https://raw.githubusercontent.com/ory/meta/master/install.sh) -d -b .bin ory v0.1.14
 		touch -a -m .bin/ory
--- a/ui/node/node.go
+++ b/ui/node/node.go
@@ -17,8 +17,42 @@ import (
 	"github.com/ory/x/stringslice"
 )
 
-// swagger:model uiNodeType
+// swagger:enum uiNodeType
 type Type string
+
+const (
+	Text   Type = "text"
+	Input  Type = "input"
+	Image  Type = "img"
+	Anchor Type = "a"
+	Script Type = "script"
+)
+
+var (
+	ErrUnknownNodeType = errors.New("unknown node type")
+)
+
+func (t Type) String() string {
+	return string(t)
+}
+
+func (t *Type) UnmarshalJSON(v []byte) error {
+	switch string(v) {
+	case `"text"`:
+		*t = Text
+	case `"input"`:
+		*t = Input
+	case `"img"`:
+		*t = Image
+	case `"a"`:
+		*t = Anchor
+	case `"script"`:
+		*t = Script
+	default:
+		return ErrUnknownNodeType
+	}
+	return nil
+}
 
 // swagger:model uiNodeGroup
 type Group string
@@ -37,12 +71,6 @@ const (
 	TOTPGroup             Group = "totp"
 	LookupGroup           Group = "lookup_secret"
 	WebAuthnGroup         Group = "webauthn"
-
-	Text   Type = "text"
-	Input  Type = "input"
-	Image  Type = "img"
-	Anchor Type = "a"
-	Script Type = "script"
 )
 
 // swagger:model uiNodes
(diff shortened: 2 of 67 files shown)


zepatrik (changes requested):
Thanks for the PR. Unfortunately, this will not fix the problem as we generate the openapi schema from comments in the code. Instead, you will have to find and fix those.
Specifically you want to adjust https://github.com/ory/kratos/blob/37f2f220ce699e031018777c9976cafa22faa984/ui/node/attributes.go#L84-L114

Enums can be created like this: https://github.com/ory/keto/blob/d64ae29b5aa7359792c3515a771eefc093065f9b/internal/expand/tree.go#L15-L23

So you have to adjust https://github.com/ory/kratos/blob/37f2f220ce699e031018777c9976cafa22faa984/ui/node/node.go#L20-L21 to follow the structure of the other sample. It is specifically important to put the consts just below the type definition in one const group.

You can test if your changes work by running `make sdk`.

woylie:
Thanks, I'll revise this tomorrow.

woylie:
I'm sorry, I'm not very familiar with Go and I don't know what I'm doing. The output still looks wrong.

zepatrik:
> Thanks, I'll revise this tomorrow.

That was a short night :wink: 

Swagger, the tool we use for generating the spec, is unfortunately very buggy. I think it only works if the enum name and the type name is the same. I'll quickly try that.

zepatrik:
> One more question, what is the reason for all the HTML tags like <code type="make/command">make test-docs</code> in the readme, instead of just using backticks and code fences?

That's related to a tool we use for testing documentation https://github.com/kevgo/text-runner

Makefile
```diff
@@ -39,6 +39,14 @@ $(call make-lint-dependency)
 docs/cli:
 		go run ./cmd/clidoc/. .
 
+.PHONY: docs/api
+docs/api:
+		npx @redocly/openapi-cli preview-docs spec/api.json
+
+.PHONY: docs/swagger
+docs/swagger:
+		npx @redocly/openapi-cli preview-docs spec/swagger.json
```
zepatrik, line 48: Oh that's nice :+1:

zepatrik (approved):
Nice, looks good to me. The CI failure does not seem related, but maybe rebase onto current master and let it run again?

ui/node/node.go
```diff
@@ -17,34 +17,39 @@ import (
 	"github.com/ory/x/stringslice"
 )
 
-// swagger:model uiNodeType
+// swagger:enum Type
```
aeneasr, line 20: Could we name this

```suggestion
// swagger:enum UiNodeType
```

please?

ui/node/node.go
```diff
@@ -17,34 +17,39 @@ import (
...
 	Image  Type = "img"
 	Anchor Type = "a"
 	Script Type = "script"
 )
 
+func (t Type) String() string {
+	return string(t)
+}
+
+// swagger:enum Group
```
aeneasr, line 35: Could we name this

```suggestion
// swagger:enum UiNodeGroup
```

please?

aeneasr (changes requested):
This is grand, thank you! Just two comments :)

ui/node/node.go
woylie: I would have to rename both the Swagger enum name and the actual Go type (`type UiNodeType string`), otherwise Swagger will not generate the spec correctly. Is that what you want?

aeneasr: I see, yes that would be ok. The problem is that this will end up with `EnumType` in most generators, which is not unique enough and will cause issues when we have multiple enums called `type` - something that is quite likely to happen :)

woylie: Done!

aeneasr (approved):

vinckr:
Hello @woylie 

Issue #2354, opened 2022-04-02
Substantial memory allocations in DefaultPasswordValidator (HIBP)


harnash:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [x] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

It seems that method `fetch()` in `selfservice/strategy/password/validator.go` is allocating memory without limits. Looking briefly at the code I can see that it stores cached counter for each validated password against HIBP API. This could be a problematic in the long run since all the validated password hashes are stored in the memory indefinitely causing steady memory allocation eventually exhausting all available memory.

I believe this implementation stores this values to save calls to the API which is good. However we would need to limit number of data stored in this "cache" to not "leak" memory and add unnecessary memory pressure for the host. In previous app which had similar concept of cache we used https://pkg.go.dev/github.com/hashicorp/golang-lru?utm_source=godoc and it works well. We could also think of implementing something simpler but I'm not sure if that's going to be reliable on larger set of records.

Another thing is that while this local cache is fine for local testing where there is only one instance of Kratos in production when there are more instances this cache might not be as effective since requests will be spread across multiple instances and all of those instances have to populate this cache individually.

Let me know what you think. This is not a critical bug but it should be fixed rather quickly to avoid unnecessary memory leak hunting :-)

### Reproducing the bug

1. Run Kratos instance for a substantial amount of time (24hrs+)
2. Generate substantial traffic to Kratos endpoints (login/registration)
3. Observe memory usage and notice it is steadily increasing over time

### Relevant log output

```shell
Here is the 24hrs graph https://pasteboard.co/9Iw7qL5NHZb5.png (24+1 instances and the biggest graph shows cumulative memory usage for all instances).
```

### Version

0.8.0

### On which operating system are you observing this issue?

Linux

### In which environment are you deploying?

Kubernetes

[...]

PR #2389, opened 2022-04-11
fix: avoid excessive memory allocations

harnash:
This PR changes internal storage for HIBP API call results to use ristretto cache instead of plain map to avoid excessive memory allocations when handling substantial load (each checked hash is stored indefinitely in memory).

## Related issue(s)
* https://github.com/ory/kratos/issues/2354

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments
This patch was running for a few days on our cluster and it looks like memory usage is not increasing and stabilised around 70MB which is great :-)

### DIFF
--- a/selfservice/strategy/password/validator.go
+++ b/selfservice/strategy/password/validator.go
@@ -3,6 +3,7 @@ package password
 import (
 	"bufio"
 	"context"
+
 	/* #nosec G505 sha1 is used for k-anonymity */
 	"crypto/sha1"
 	"fmt"
--- a/selfservice/strategy/password/validator.go
+++ b/selfservice/strategy/password/validator.go
@@ -3,7 +3,6 @@ package password
 import (
 	"bufio"
 	"context"
-
 	/* #nosec G505 sha1 is used for k-anonymity */
 	"crypto/sha1"
 	"fmt"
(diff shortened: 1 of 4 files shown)


selfservice/strategy/password/validator.go
```diff
@@ -62,12 +62,21 @@ type validatorDependencies interface {
 	config.Provider
 }
 
-func NewDefaultPasswordValidatorStrategy(reg validatorDependencies) *DefaultPasswordValidator {
+func NewDefaultPasswordValidatorStrategy(reg validatorDependencies) (*DefaultPasswordValidator, error) {
+	cache, err := ristretto.NewCache(&ristretto.Config{
+		NumCounters: 10 * 10000,
+		MaxCost:     60 * 10000, // BCrypt hash size is 60 bytes
+		BufferItems: 64,
```
aeneasr, line 69: The cost in ristretto is set is kinda weird. For example, it is probably required to ignore the internal cost and use a custom cost function:

https://github.com/dgraph-io/ristretto/blob/8e850b710d6df0383c375ec6a7beae4ce48fc8d5/cache.go#L137

aeneasr (changes requested):

selfservice/strategy/password/validator.go
harnash: Alright. I've added `ignoreInternalCost=true` but I'm not sure if custom cost function is needed at this point.

selfservice/strategy/password/validator.go
```diff
@@ -109,9 +120,8 @@ func (s *DefaultPasswordValidator) fetch(hpw []byte, apiDNSName string) error {
 		return errors.Wrapf(ErrUnexpectedStatusCode, "%d", res.StatusCode)
 	}
 
-	s.Lock()
-	s.hashes[b20(hpw)] = 0
-	s.Unlock()
+	s.hashes.SetWithTTL(b20(hpw), 0, 1, hashCacheItemTTL)
+	s.hashes.Wait()
```
aeneasr, line 124: Is the wait intentional? I think we could do without!

harnash: Fix in 3c5caa669

Issue #952, opened 2021-01-02
Delete self-service flows after completion


aeneasr:
**Is your feature request related to a problem? Please describe.**

Currently, old flows are not purged from the database. This accumulates stale records in the database and bears potential privacy and security risks.

**Describe the solution you'd like**

Flows should be removed when stale, preferably as a background task.

anonrig:
hey @aeneasr, I've been using Kratos on production for a while now, and this issue is getting out of hand (both in terms of storage and query execution speed). is this issue a priority? I would like to know more since it became a game-changer for us.

harnash:
There are other tables that needs this (i.e. `sessions`). So probably one task that will purge all expired records (in batches!) would be needed.

aeneasr:
I was also thinking of cleaning up successful flows on completion. Expiry still needs batch deletion though

abador:
@aeneasr  I started working on it. I'll ping you to take a look at our branch before porting those changes to your repo

aeneasr:
Nice!

abador:
@aeneasr I've been looking into removing finished flows, but it doesn't look as easy to do as I hoped (or I misread something). There seem to be a few edge cases when a flow can be reused or is needed at some point later in the code. I think I'll try to tackle this in a seperate pr/issue

aeneasr:
It can also be a background worker that checks for expired flows and cleans them up, we probably need that anyways for flows that expire and are never completed

abador:
@aeneasr what do you think about https://github.com/Wikia/kratos/pull/56 ? Is this what you were thinking of? I might be able to port those changes to your repo this week.

aeneasr:
Yes, sweet! I only have a few suggestions of things we can probably omit (e.g. i dont think we need the flag because anyone that runs into this issue will have at least 2 or 3 pods of kratos runing) :)

frederikhors:
Is there a way to avoid a new executable to launch and have something like `--watch courier` within the main executable?

PR #2406, opened 2022-04-19
Remove expired data from the database, ISSUE-952

abador:
Cleaning up the database from old data and flows

## Related issue(s)
#952 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


abador:
Hope It's enough :)

persistence/sql/persister_continuity.go
```diff
@@ -41,3 +42,18 @@ func (p *Persister) DeleteContinuitySession(ctx context.Context, id uuid.UUID) e
 	}
 	return nil
 }
+
+func (p *Persister) DeleteExpiredContinuitySessions(ctx context.Context, expiresAt time.Time, limit int) error {
+	// #nosec G201
+	err := p.GetConnection(ctx).RawQuery(fmt.Sprintf(
+		"DELETE FROM %s WHERE expires_at <= ? LIMIT ?",
```
zepatrik, line 49: Please see https://github.com/gobuffalo/pop/pull/658 and specifically https://github.com/gobuffalo/pop/pull/658/files#diff-5d204b98ec636a0fe75d32bf558fc3c06ff540eb1010ede6144e788c0c3aa57aR131-R140

cmd/cleanup/sql.go
```diff
@@ -0,0 +1,50 @@
...
+### WARNING ###
+Before running this command on an existing database, create a back up!
+`,
+		Run: func(cmd *cobra.Command, args []string) {
+			cliclient.NewCleanupHandler().CleanupSQL(cmd, args)
+		},
+	}
+
+	configx.RegisterFlags(c.PersistentFlags())
+	c.Flags().BoolP("read-from-env", "e", false, "If set, reads the database connection string from the environment variable DSN or config file key dsn.")
```
zepatrik, line 45: I would expect that to be the default, there are probably only few cases where you want to pass the DSN as an arg, if any.

cmd/cleanup/sql.go
```diff
@@ -0,0 +1,50 @@
...
+	c.Flags().IntP(config.ViperKeyDatabaseCleanupBatchSize, "b", 100, "Set the number of records to be cleaned per run")
+	c.Flags().Duration(config.ViperKeyDatabaseCleanupSleepBackground, 30*time.Minute, "How long to wait between each cleanup run")
+	c.Flags().Duration(config.ViperKeyDatabaseCleanupSleepTables, time.Minute, "How long to wait between each table cleanup")
```
zepatrik, line 48: Wouldn't it be easier and also less resource-intense to use something like cron jobs for cleanup? A long-running cleanup process blocks at least one DB connection, and is probably also more time-consuming from a operation perspective.

zepatrik (changes requested):
Thanks for this PR, I have some ideas on how to further improve it :wink:

cmd/cleanup/sql.go
abador: Sorry I think that I ported those changes to fast and I pushed the wrong commit. This port shouldn't have the background worker, but the command only.

persistence/sql/persister.go
```diff
@@ -135,6 +136,57 @@ type node interface {
 	GetNID() uuid.UUID
 }
 
+func (p *Persister) CleanupDatabase(ctx context.Context, wait time.Duration) error {
+	currentTime := time.Now()
+	deleteLimit := p.r.Config(ctx).DatabaseCleanupBatchSize()
+	p.r.Logger().Printf("Cleaning up first %d records older than %s\n", deleteLimit, currentTime)
```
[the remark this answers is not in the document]
abador: I went with the `--keep-last` flag. I'm not sure if the tables flag is needed. We would run multiple crons and drop records without being sure that they don't fire at the same time. Doing tables one after another with a proper sleep value mitigates replication lag issues

persistence/sql/persister_continuity.go
abador: this will be fun. Removing limits in our case will be tricky (or we'll just cleanup with our version where we use limits https://github.com/Wikia/kratos/pull/56 and then just run this versionlater on ). As long as we run it more frequently this shouldn't be an issue

zepatrik: > This will be fun.

Is limit-less deletion for an established system a big performance impact? If so, we should probably add some kind of limit, e.g. using a sub-query, that works on all DBs.

persistence/sql/persister.go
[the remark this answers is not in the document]
zepatrik: The `--tables` flag was just a random idea that came to mind. I think it is fine as is, in case someone needs per-table cleanup times we can still add that later.

cmd/cliclient/cleanup.go
```diff
@@ -0,0 +1,61 @@
...
+		d = driver.NewWithoutInit(
+			cmd.Context(),
+			cmd.ErrOrStderr(),
+			configx.WithFlags(cmd.Flags()),
+			configx.SkipValidation())
+		if len(d.Config(cmd.Context()).DSN()) == 0 {
+			fmt.Println(cmd.UsageString())
+			fmt.Println("")
+			fmt.Println("When using flag -e, environment variable DSN must be set")
+			os.Exit(1)
```
zepatrik, line 36: As tests will always fail if you call `os.Exit` we instead use `cobra.Command.RunE` and propagate errors up. As in this case you already printed the error message, you can just return `cmdx.FailSilently(cmd)`. It will silence the standard cobra output and results in an error exit code being set.

cmd/cliclient/cleanup.go
```diff
@@ -0,0 +1,61 @@
...
+	if flagx.MustGetBool(cmd, "read-from-env") {
+		d = driver.NewWithoutInit(
+			cmd.Context(),
+			cmd.ErrOrStderr(),
+			configx.WithFlags(cmd.Flags()),
+			configx.SkipValidation())
+		if len(d.Config(cmd.Context()).DSN()) == 0 {
+			fmt.Println(cmd.UsageString())
+			fmt.Println("")
+			fmt.Println("When using flag -e, environment variable DSN must be set")
```
zepatrik, line 35:
```suggestion
			fmt.Fprintln(cmd.OutOrStdout(), cmd.UsageString())
			fmt.Fprintln(cmd.OutOrStdout(), "")
			fmt.Fprintln(cmd.OutOrStdout(), "When using flag -e, environment variable DSN must be set")
```

Also here we want to use the `out` from the command such that we can in tests write to a buffer and do assertions on the output.

cmd/cliclient/cleanup.go
```diff
@@ -0,0 +1,61 @@
...
+		d = driver.NewWithoutInit(
+			cmd.Context(),
+			cmd.ErrOrStderr(),
+			configx.WithFlags(cmd.Flags()),
+			configx.SkipValidation(),
+			configx.WithValue(config.ViperKeyDSN, args[0]))
+	}
+
+	err := d.Init(cmd.Context(), driver.SkipNetworkInit)
+	cmdx.Must(err, "An error occurred initializing migrations: %s", err)
```
zepatrik, line 54: The errors down here can just be wrapped and returned.

persistence/sql/persister.go
```diff
@@ -135,6 +136,56 @@ type node interface {
 	GetNID() uuid.UUID
 }
 
+func (p *Persister) CleanupDatabase(ctx context.Context, wait time.Duration, older time.Duration) error {
+	currentTime := time.Now().Add(-older)
```
zepatrik, line 140:
```suggestion
	removeOlderThan := time.Now().Sub(older)
```

persistence/sql/persister_continuity.go
```diff
@@ -41,3 +42,17 @@ func (p *Persister) DeleteContinuitySession(ctx context.Context, id uuid.UUID) e
...
 }
+
+func (p *Persister) DeleteExpiredContinuitySessions(ctx context.Context, expiresAt time.Time) error {
+	// #nosec G201
+	err := p.GetConnection(ctx).RawQuery(fmt.Sprintf(
+		"DELETE FROM %s WHERE expires_at <= ?",
+		new(continuity.Container).TableName(ctx),
+	),
+		expiresAt,
+	).Exec()
```
zepatrik, line 53:
```suggestion
	err := p.GetConnection(ctx)
		.Where("expires_at <= ?", expiresAt)
		.Delete(new(continuity.Container)),
```

zepatrik (changes requested):
Nice, this looks pretty good already :+1: 
I have some small suggestions how to improve it further :wink:

persistence/sql/persister.go
abador: This will cause a type mismatch. Since we're working on small durations I would live it this way

cmd/cliclient/cleanup.go
abador: I think I did what you had in mind

persistence/sql/persister_continuity.go
abador: Yeah this is tricky since we didn't have this ability from the start and we can't remove everything all at once, even with using the dates like in the pr. When switching from one system to another we created a lot of sessions/flows etc at the same moment time.  Dropping those records can cause a big replication lag when the delete query hits that moment in time. Using limit helps us mitigate the problem a little and drops them in batches. This way we don't have to run this jub as often and there is no need to monitor this as much

abador:
@zepatrik I left a few comments. If it's ok to leave it this way please resolve the conversations

persistence/sql/persister_continuity.go
aeneasr: The problem with this query is that it will be extremely slow as it is unbound. Check out this PR for batch-based processing: https://github.com/grantzvolsky/hydra/pull/2/files#diff-6034803b09ef5017e3aa7d3082827dadb4503e6ee3a56853d49edd419a75f864

aeneasr:
@abador looks like there are a couple of compile errors

aeneasr (changes requested):
@abador looks like there are a couple of compile errors

abador:
> @abador looks like there are a couple of compile errors

I've seen them on the master branch. I'll merge master and check if they're gone

abador:
Ok it dodn't help:/ I see similar errors in this pr: https://github.com/ory/kratos/runs/6133235832?check_suite_focus=true

internal/driver.go
```diff
@@ -45,6 +46,8 @@ func NewConfigurationWithDefaults(t *testing.T) *config.Config {
 			config.ViperKeyCourierSMTPURL:                    "smtp://foo:bar@baz.com/",
 			config.ViperKeySelfServiceBrowserDefaultReturnTo: "https://www.ory.sh/redirect-not-set",
 			config.ViperKeySecretsCipher:                     []string{"secret-thirty-two-character-long"},
+			config.ViperKeyDatabaseCleanupSleepTables:        1 * time.Minute,
+			config.ViperKeyDatabaseCleanupBatchSize:          100,
```
zepatrik, line 50: Those defaults will be loaded from the config schema, so there should be no need to add them here.

driver/config/config.go
```diff
@@ -1074,6 +1076,14 @@ func (p *Config) SelfServiceLinkMethodBaseURL() *url.URL {
 	return p.p.RequestURIF(ViperKeyLinkBaseURL, p.SelfPublicURL())
 }
 
+func (p *Config) DatabaseCleanupSleepTables() time.Duration {
+	return p.p.DurationF(ViperKeyDatabaseCleanupSleepTables, 5*time.Second)
```
zepatrik, line 1080: Same here, the default should be provided by the schema.
```suggestion
	return p.p.Duration(ViperKeyDatabaseCleanupSleepTables)
```

driver/config/config.go
```diff
@@ -1074,6 +1076,14 @@ func (p *Config) SelfServiceLinkMethodBaseURL() *url.URL {
...
+}
+
+func (p *Config) DatabaseCleanupBatchSize() int {
+	return p.p.IntF(ViperKeyDatabaseCleanupBatchSize, 100)
```
zepatrik, line 1084:
```suggestion
	return p.p.Int(ViperKeyDatabaseCleanupBatchSize)
```

cmd/cliclient/cleanup.go
```diff
@@ -0,0 +1,69 @@
...
+			fmt.Println(cmd.UsageString())
+			return cmdx.FailSilently(cmd)
+		}
+		d = driver.NewWithoutInit(
+			cmd.Context(),
+			cmd.ErrOrStderr(),
+			configx.WithFlags(cmd.Flags()),
+			configx.SkipValidation(),
+			configx.WithValue(config.ViperKeyDSN, args[0]))
+	}
```
zepatrik, line 50:
```suggestion
	opts := []configx.Options{
		configx.WithFlags(cmd.Flags()),
		configx.SkipValidation(),
	}

	if !flagx.MustGetBool(cmd, "read-from-env") {
		if len(args) != 1 {
			return errors.New(`expected to get the DSN as an argument, or the "read-from-env" flag`)
		}
		opts = append(opts, configx.WithValue(config.ViperKeyDSN, args[0]))
	}
	
	d := driver.NewWithoutInit(
		cmd.Context(),
		cmd.ErrOrStderr(),
		...opts,
	)
	if len(d.Config(cmd.Context()).DSN()) == 0 {
		return errors.New(`required config value "dsn" was not set`)
	}
```

persistence/sql/persister.go
```diff
@@ -138,6 +139,56 @@ type node interface {
...
+	}
+	time.Sleep(wait)
+
+	p.r.Logger().Println("Cleaning up expired verification flows")
+	if err := p.DeleteExpiredVerificationFlows(ctx, currentTime, batchSize); err != nil {
+		return err
+	}
+	time.Sleep(wait)
+
+	p.r.Logger().Println("Successfully cleaned up the SQL database!")
```
zepatrik, line 188: At this point we did not fully clean the database but rather purged `batchSize` from each of the tables. If there are more expired rows, they would not have been deleted. The output should reflect that as we will have to re-run the cleanup multiple times until all data are truly cleaned.

persistence/sql/persister_continuity.go
```diff
@@ -50,3 +51,20 @@ func (p *Persister) DeleteContinuitySession(ctx context.Context, id uuid.UUID) e
 	}
 	return nil
 }
+
+func (p *Persister) DeleteExpiredContinuitySessions(ctx context.Context, expiresAt time.Time, limit int) error {
+	// #nosec G201
+	err := p.GetConnection(ctx).RawQuery(fmt.Sprintf(
+		"DELETE FROM %s WHERE id in (SELECT id FROM (SELECT id FROM %s c WHERE expires_at <= ? and nid = ? ORDER BY expires_at ASC LIMIT %d ) AS s )",
```
zepatrik, line 58: Why is there two sub-queries?

zepatrik (commented):
Nice, this looks very good now :+1: I just have some ideas how to improve it further, and a few questions.

persistence/sql/persister_continuity.go
abador: I needed to do something similar that is metioned here: https://github.com/ory/kratos/pull/2406/#discussion_r860916009 . Going with a single subselect didn't work for SQLite (I tried it few other ways), but this way works.

abador:
@aeneasr  and @zepatrik please take a look when you'll have a chance

persistence/sql/persister_login.go
```diff
@@ -63,3 +65,20 @@ func (p *Persister) ForceLoginFlow(ctx context.Context, id uuid.UUID) error {
 		return tx.Save(lr, "nid")
 	})
 }
+
+func (p *Persister) DeleteExpiredLoginFlows(ctx context.Context, expiresAt time.Time, limit int) error {
+	// #nosec G201
+	err := p.GetConnection(ctx).RawQuery(fmt.Sprintf(
+		"DELETE FROM %s WHERE id in (SELECT id FROM (SELECT id FROM %s c WHERE expires_at <= ? and nid = ? ORDER BY expires_at ASC LIMIT %d ) AS s )",
```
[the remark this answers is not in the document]
aeneasr: Sorry, I missed that this question was asked already previously, but I'm still not 100% sure if I understand. Does SQLite not support this?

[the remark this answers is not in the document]
aeneasr: I see, thank you for the elaboration!

It looks like there are a couple of issues related to this in the SQLite Go driver:

1. https://github.com/mattn/go-sqlite3/pull/802

We can probably still reduce the amount of subqueries to one?

```
DELETE FROM %s WHERE id IN (SELECT id FROM %s c WHERE expires_at <= ? and nid = ? ORDER BY expires_at ASC LIMIT %d )
```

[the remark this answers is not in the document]
abador: I think I tried it before and there was an error with the in statement. I'll try one query for now in https://github.com/ory/kratos/pull/2406/commits/d910a1ee80af5e9120b2b3839e7a886c5c02d742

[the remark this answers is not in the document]
aeneasr: Ok, I see, thank you for the investagion! Then there truly is no way to improve this easily across all DB engines!

persistence/sql/persister_continuity.go
```diff
@@ -55,7 +55,7 @@ func (p *Persister) DeleteContinuitySession(ctx context.Context, id uuid.UUID) e
 func (p *Persister) DeleteExpiredContinuitySessions(ctx context.Context, expiresAt time.Time, limit int) error {
 	// #nosec G201
 	err := p.GetConnection(ctx).RawQuery(fmt.Sprintf(
-		"DELETE FROM %s WHERE id in (SELECT id FROM (SELECT id FROM %s c WHERE expires_at <= ? and nid = ? ORDER BY expires_at ASC LIMIT %d ) AS s )",
+		"DELETE FROM %s WHERE id IN (SELECT id FROM %s WHERE expires_at <= ? and nid = ? ORDER BY expires_at ASC LIMIT %d )",
```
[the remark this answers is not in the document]
abador: done

aeneasr:
Thank you so much for this great contribution!

frederikhors:
Is there any docs about this tool? @abador, @aeneasr?

Issue #2307, opened 2022-03-14
Internal Server Error on Empty `PUT /identities/id` body


theory:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

G'day. I spun up a Kratos instance using the quickstart (using Postgres). Works fine, so was trying some other stuff. I execed into the Postgres image, installed curl, and called

``` sh
curl -s http://kratos:4434/identities/c8d98079-c6c0-49e1-a034-92fa2269724b
```

It nicely returned my identity (created by following the [quickstart](https://www.ory.sh/docs/kratos/quickstart)). Then I edited the JSON and tried updating it. Only I forgot to copy the file first, so `-d @file.json` just inserted an empty file. This caused Kratos to return an internal server error.

### Reproducing the bug

How to replicate:

``` sh
curl -v -X PUT http://kratos:4434/identities/c8d98079-c6c0-49e1-a034-92fa2269724b -H "Accept: application/json" -d @~/Downloads/david.json
```

Output:

```
Warning: Couldn't read data from file "~/Downloads/david.json", this makes an 
Warning: empty POST.
*   Trying 172.19.0.6...
* TCP_NODELAY set
* Connected to kratos (172.19.0.6) port 4434 (#0)
> PUT /identities/c8d98079-c6c0-49e1-a034-92fa2269724b HTTP/1.1
> Host: kratos:4434
> User-Agent: curl/7.52.1
> Accept: application/json
> Content-Length: 0
> Content-Type: application/x-www-form-urlencoded
> 
< HTTP/1.1 500 Internal Server Error
< Content-Type: application/json
< Date: Mon, 14 Mar 2022 22:10:02 GMT
< Content-Length: 72
* HTTP error before end of send, stop sending
< 
[...]

PR #2417, opened 2022-04-22
fix: Internal Server Error on Empty PUT /identities/id body

VeenaInd:
Signed-off-by: Veena Saini <veena.code.assignment@gmail.com>

## Description of the changes
This PR fixes the issue https://github.com/ory/kratos/issues/2307. 

Currently if a user sends HTTP PUT request with empty json data or incorrect json path (means the file does not exist), then that results into "internal server error". This error message is a generic and creates confusion. 

Empty json (or we can say http request without body) triggers io.EOF error.  This PR provides the user a better error message when io.EOF error is encountered.

**Affected Module: Identity Update** 

There can be other scenarios of malformed user requests that may trigger "internal server error" during identity update, but those are not handled in this PR.

## Checklist

      introduces a new feature.
- [x ] I am following the
      the changes.
      works.

### DIFF
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -3,6 +3,7 @@ package identity
 import (
 	"context"
 	"encoding/json"
+	"io"
 	"net/http"
 	"time"
 
@@ -435,10 +436,13 @@ type AdminUpdateIdentityBody struct {
 //       500: jsonError
 func (h *Handler) update(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	var ur AdminUpdateIdentityBody
-	if err := errors.WithStack(jsonx.NewStrictDecoder(r.Body).Decode(&ur)); err != nil {
-		h.r.Writer().WriteError(w, r, err)
-		return
-	}
+        if err := errors.WithStack(jsonx.NewStrictDecoder(r.Body).Decode(&ur)); err == io.EOF {
+               h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf(`Empty input json or file does not exits`).WithWrap(err)))
+               return
+        } else if err != nil {
+                h.r.Writer().WriteError(w, r, err)
+                return
+        }
 
 	id := x.ParseUUID(ps.ByName("id"))
 	identity, err := h.r.PrivilegedIdentityPool().GetIdentityConfidential(r.Context(), id)
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -3,7 +3,6 @@ package identity
 import (
 	"context"
 	"encoding/json"
-	"io"
 	"net/http"
 	"time"
 
@@ -18,6 +17,7 @@ import (
 	"github.com/julienschmidt/httprouter"
 	"github.com/pkg/errors"
 
+	"github.com/ory/x/decoderx"
 	"github.com/ory/x/jsonx"
 	"github.com/ory/x/sqlxx"
 	"github.com/ory/x/urlx"
@@ -43,7 +43,8 @@ type (
 		IdentityHandler() *Handler
 	}
 	Handler struct {
-		r handlerDependencies
+		r  handlerDependencies
+		dx *decoderx.HTTP
 	}
 )
 
@@ -52,7 +53,10 @@ func (h *Handler) Config(ctx context.Context) *config.Config {
 }
 
 func NewHandler(r handlerDependencies) *Handler {
-	return &Handler{r: r}
+	return &Handler{
+		r:  r,
+		dx: decoderx.NewHTTP(),
+	}
 }
 
 func (h *Handler) RegisterPublicRoutes(public *x.RouterPublic) {
@@ -436,13 +440,11 @@ type AdminUpdateIdentityBody struct {
 //       500: jsonError
 func (h *Handler) update(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	var ur AdminUpdateIdentityBody
-        if err := errors.WithStack(jsonx.NewStrictDecoder(r.Body).Decode(&ur)); errors.Is(err, io.EOF) {
-                h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf(`Empty input json or file does not exist`).WithWrap(err)))
-                return
-        } else if err != nil {
-                h.r.Writer().WriteError(w, r, err)
-                return
-        }
+	if err := h.dx.Decode(r, &ur,
+		decoderx.HTTPJSONDecoder()); err != nil {
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
 
 	id := x.ParseUUID(ps.ByName("id"))
 	identity, err := h.r.PrivilegedIdentityPool().GetIdentityConfidential(r.Context(), id)
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -436,9 +436,9 @@ type AdminUpdateIdentityBody struct {
 //       500: jsonError
 func (h *Handler) update(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	var ur AdminUpdateIdentityBody
-        if err := errors.WithStack(jsonx.NewStrictDecoder(r.Body).Decode(&ur)); err == io.EOF {
-               h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf(`Empty input json or file does not exits`).WithWrap(err)))
-               return
+        if err := errors.WithStack(jsonx.NewStrictDecoder(r.Body).Decode(&ur)); errors.Is(err, io.EOF) {
+                h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf(`Empty input json or file does not exist`).WithWrap(err)))
+                return
         } else if err != nil {
                 h.r.Writer().WriteError(w, r, err)
                 return
--- a/identity/handler_test.go
+++ b/identity/handler_test.go
@@ -74,8 +74,8 @@ func TestHandler(t *testing.T) {
 	var send = func(t *testing.T, base *httptest.Server, method, href string, expectCode int, send interface{}) gjson.Result {
 		var b bytes.Buffer
 		if send != nil {
-                        require.NoError(t, json.NewEncoder(&b).Encode(send))
-                }
+			require.NoError(t, json.NewEncoder(&b).Encode(send))
+		}
 		req, err := http.NewRequest(method, base.URL+href, &b)
 		require.NoError(t, err)
 		req.Header.Set("Content-Type", "application/json")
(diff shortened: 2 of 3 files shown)


VeenaInd:
Hi,

Just a quick check here. I was trying to reproduce the issue and got "ERR_DNS_FAIL" when attempting to connect to http://kratos:4434/identities . Also, if instead I do curl to http://127.0.0.1:4434/identities/ then that attempt results into "Temporary Redirect".  

So, before proceeding wanted to confirm if there are any proxy settings in the source code that needs to be taken care.

Thanks,
Veena

VeenaInd:
Also, I started with the quickstart and I am able to follow along with the demo. Only issue is when trying to reach to http://kratos:4434/identities.

zepatrik (commented):
> I was trying to reproduce the issue and got "ERR_DNS_FAIL" when attempting to connect to http://kratos:4434/identities . Also, if instead I do curl to http://127.0.0.1:4434/identities/ then that attempt results into "Temporary Redirect".

This depends a bit on your test setup. If you run everything using docker-compose on localhost, you want to use `http://127.0.0.1:4434`. The OP used this setup:

> I execed into the Postgres image, installed curl, ad called `curl -s http://kratos:4434/identities/c8d98079-c6c0-49e1-a034-92fa2269724b`

With docker networking that will be resolved properly, but you can also just use the loopback interface from outside of docker.

Now the redirect is happening because we recently added the path prefix `/admin` for admin endpoints. Therefore, you can test your fix using `curl -v -X PUT http://127.0.0.1:4434/admin/identities/c8d98079-c6c0-49e1-a034-92fa2269724b -H "Accept: application/json"`.

In the end, it would be great to also have a go unit test for this case. There are already other tests where you can just add one more subtest, without the need to setup everything.

Benehiko:
Hi @VeenaInd 

Thank you for taking the time to work on this :)

You don't need to compile and run Kratos with the whole configuration setup manually to test this. I would suggest adding a test case to the `identity/handler_test.go` file and then running that. 

You can find an example of such a test here
https://github.com/ory/kratos/blob/562e340fe980e7c65ab3fc41f82a2a8899a33bfa/identity/handler_test.go#L428-L459

Also please sign the CLA agreement so that the CI can run the tests.
https://github.com/ory/kratos/pull/2417#issuecomment-1106199018

VeenaInd:
Thanks @zepatrik and @Benehiko for this guidance. 

curl -v -X PUT http://127.0.0.1:4434/admin/identities/c8d98079-c6c0-49e1-a034-92fa2269724b -H "Accept: application/json". is good one to replicate the scenario. I am able to get the exact error with this.

Benehiko (changes requested):
This is looking good, we would just need some handler tests to verify the fix

Benehiko:
Also please rebase with `origin/master` as it seems this branch is out of date.

VeenaInd:
Thanks @zepatrik and @Benehiko for your feedback. I have pushed working code and corresponding unit test. If I can think of better code, I will pushed by tomorrow morning.

identity/handler_test.go
```diff
@@ -73,7 +73,9 @@ func TestHandler(t *testing.T) {
 
 	var send = func(t *testing.T, base *httptest.Server, method, href string, expectCode int, send interface{}) gjson.Result {
 		var b bytes.Buffer
-		require.NoError(t, json.NewEncoder(&b).Encode(send))
+		if send != nil {
+                        require.NoError(t, json.NewEncoder(&b).Encode(send))
+                }
 		req, err := http.NewRequest(method, base.URL+href, &b)
```
VeenaInd, line 79: Here, **&b** value will remain nil by default. That's why when the **send** value is nil, we are sending http request body value as nil. This will create io.EOF error in update() function.

identity/handler_test.go
```diff
@@ -673,6 +675,21 @@ func TestHandler(t *testing.T) {
...
+	t.Run("case=should fail to update identity if input json is empty or json file does not exist", func(t *testing.T) {
+                for name, ts := range map[string]*httptest.Server{"public": publicTS, "admin": adminTS} {
+                        t.Run("endpoint="+name, func(t *testing.T) {
+                                        var cr identity.AdminCreateIdentityBody
+                                        cr.SchemaID = "employee"
+                                        cr.Traits = []byte(`{"email":"` + x.NewUUID().String() + `@ory.sh", "department": "ory"}`)
+                                        res := send(t, ts, "POST", "/identities", http.StatusCreated, &cr)
+
+                                        id := res.Get("id").String()
+                                        res = send(t, ts, "PUT", "/identities/"+id, http.StatusBadRequest, nil)
```
VeenaInd, line 687: Here, I tried to send empty interface (also null interface) as 3rd function parameters. But those values after json encoding does not set **&b** value at line 77 as null. That's why I sent **untyped nil** value.

identity/handler_test.go
```diff
@@ -673,6 +675,21 @@ func TestHandler(t *testing.T) {
 		}
 	})
 
+	t.Run("case=should fail to update identity if input json is empty or json file does not exist", func(t *testing.T) {
```
VeenaInd, line 678: This test case does not explicitly handle the case of non existing json file input. Because even if we have nil file pointer, the **send()** method is expecting **interface type** as 3rd parameter. And converting a file pointer to interface type would be challenging based on my current understanding.

identity/handler_test.go
```diff
@@ -673,6 +675,21 @@ func TestHandler(t *testing.T) {
...
+                for name, ts := range map[string]*httptest.Server{"public": publicTS, "admin": adminTS} {
+                        t.Run("endpoint="+name, func(t *testing.T) {
+                                        var cr identity.AdminCreateIdentityBody
+                                        cr.SchemaID = "employee"
+                                        cr.Traits = []byte(`{"email":"` + x.NewUUID().String() + `@ory.sh", "department": "ory"}`)
+                                        res := send(t, ts, "POST", "/identities", http.StatusCreated, &cr)
+
+                                        id := res.Get("id").String()
+                                        res = send(t, ts, "PUT", "/identities/"+id, http.StatusBadRequest, nil)
+                                        assert.Contains(t, res.Get("error.reason").String(), `Empty input json or file does not exist`, "%s", res.Raw)
```
Benehiko, line 688: Here you can probably use an `assert.ErrorIs` instead, here is an example
https://github.com/ory/kratos/blob/4348b8640a282cd61fe30961faba5753e2af8bb0/session/session_test.go#L34

identity/handler.go
```diff
@@ -435,10 +436,13 @@ type AdminUpdateIdentityBody struct {
...
-		h.r.Writer().WriteError(w, r, err)
-		return
-	}
+        if err := errors.WithStack(jsonx.NewStrictDecoder(r.Body).Decode(&ur)); errors.Is(err, io.EOF) {
+                h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf(`Empty input json or file does not exist`).WithWrap(err)))
+                return
+        } else if err != nil {
+                h.r.Writer().WriteError(w, r, err)
+                return
+        }
```
Benehiko, line 445: We could maybe update this and use `decoderx`.

Here is an example
https://github.com/ory/kratos/blob/686c9ba08ff1db8a310eaed5c4b3aec69e0f84da/selfservice/flow/logout/handler.go#L178-L183

Benehiko: the `decoderx.Decode` function returns a herodot error for us and should also give an error for an empty body.

Benehiko (changes requested):
Hi @VeenaInd 

Thank you for working on this, it's looking great so far :)

There are just a few minor changes that need to be made. I also see the CI doesn't run all the tests due to linting problems, you can resolve this by doing `make format` in the root of the Kratos project. It uses [`goimports`](https://pkg.go.dev/golang.org/x/tools/cmd/goimports)  and npm so you would need to install that for this to work.

https://github.com/ory/kratos/blob/a42a0f772af3625c457032d6dcc34289a62acc61/Makefile#L134-L138

identity/handler.go
VeenaInd: Hi @Benehiko ,

Thanks for reviewing the code. I will try to incorporate your suggestions and push the changes soon.

identity/handler_test.go
```diff
@@ -676,19 +676,20 @@ func TestHandler(t *testing.T) {
...
+		for name, ts := range map[string]*httptest.Server{"public": publicTS, "admin": adminTS} {
+			t.Run("endpoint="+name, func(t *testing.T) {
+				var cr identity.AdminCreateIdentityBody
+				cr.SchemaID = "employee"
+				cr.Traits = []byte(`{"email":"` + x.NewUUID().String() + `@ory.sh", "department": "ory"}`)
+				res := send(t, ts, "POST", "/identities", http.StatusCreated, &cr)
+
+				id := res.Get("id").String()
+				res = send(t, ts, "PUT", "/identities/"+id, http.StatusBadRequest, nil)
+				assert.Contains(t, res.Get("error.reason").String(), `Unable to decode HTTP Request Body because its HTTP `+
```
VeenaInd, line 688: Hi @Benehiko , this version of the test case still uses assert.Contains() method only. 
#### Reason: 
As you can see assert.ErrorIs(t, **err,** session.ErrIdentityDisabled)  method looking for **error type** variable as **second parameter**.   Here res is of type gJson.Result, and I could not find any method for this struct that returns error output. Link that I referred : https://sourcegraph.com/github.com/tidwall/gjson/-/docs#Result. 

Please correct me if I am missing something. 

Thanks.

zepatrik (approved):
Thank you, this looks very good now from my side :+1:

identity/handler_test.go
zepatrik: Yes you are right, this is perfectly fine :+1:

Benehiko: ah i see, then it's fine like this - thank you :)

Benehiko (approved):
Good job! LGTM :)

VeenaInd:
Thanks @zepatrik and @Benehiko for your feedback. Though I did "make format", before pushing the code, but still some test cases are failing here at CI pipeline. Anything I can do to make those test cases pass ?

VeenaInd:
> > Thanks @zepatrik and @Benehiko for your feedback. Though I did "make format", before pushing the code, but still some test cases are failing here at CI pipeline. Anything I can do to make those test cases pass ?
> 
> I think github packages are down at the moment which the docker image scanner relies on afaik.
> 
> The other failures might be flakes https://github.com/ory/kratos/runs/6154892204?check_suite_focus=true#step:16:1398
> 
> We can also just wait a bit until Github is fully operational and re-run everything later.

ok, thank you.

vinckr:
Hello @VeenaInd 

VeenaInd:
> Hello @VeenaInd Congrats on merging your first PR in Ory 🎉 ! Your contribution will soon be helping secure millions of identities around the globe 🌏. As a small token of appreciation we send all our first time contributors a gift package to welcome them to the community. [Please drop me an email and I will forward you the form to claim your Ory swag!](mailto:vincent@ory.sh)

Thank you @vinckr. I am glad, I got this opportunity to work on ory project :) .

PR #2442, opened 2022-05-02
feat: add seach engine

david972:
This feature is for adding the capability to use query parameter to filter identities result on `/identities` handler.

For now couple things to do :

- Review the paging
- Add identifier in the search field
- update the documentation 

## Related issue(s)

#1278 
replace (rebase with master) #1671

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/persister_identity.go
+++ b/persistence/sql/persister_identity.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
@@ -298,12 +299,34 @@ func (p *Persister) CreateIdentity(ctx context.Context, i *identity.Identity) er
 		return p.createIdentityCredentials(ctx, i)
 	})
 }
+func (p *Persister) setIdentitySchemas(ctx context.Context, is []identity.Identity) error {
+	var schemaCache = make(map[string]string)
+
+	for k := range is {
+		i := &is[k]
+		if err := i.ValidateNID(); err != nil {
+			return sqlcon.HandleError(err)
+		}
+
+		if u, ok := schemaCache[i.SchemaID]; ok {
+			i.SchemaURL = u
+		} else {
+			if err := p.injectTraitsSchemaURL(ctx, i); err != nil {
+				return err
+			}
+			schemaCache[i.SchemaID] = i.SchemaURL
+		}
+
+		is[k] = *i
+	}
+
+	return nil
+}
 
 func (p *Persister) ListIdentities(ctx context.Context, page, perPage int) ([]identity.Identity, error) {
 	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.ListIdentities")
 	defer span.End()
-
-	is := make([]identity.Identity, 0)
+	is := []identity.Identity{}
 
 	/* #nosec G201 TableName is static */
 	if err := sqlcon.HandleError(p.GetConnection(ctx).Where("nid = ?", corp.ContextualizeNID(ctx, p.nid)).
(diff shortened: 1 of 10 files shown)


david972:
need a help here
found duplicate entries into response

persistence/sql/persister_identity.go
```diff
@@ -565,3 +598,103 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+var validValues = regexp.MustCompile(`[%]+`)
+
+func (p *Persister) validateValues(values []string) bool {
+	for _, value := range values {
+		if validValues.MatchString(value) {
+			return false
+		}
+	}
+	return true
+}
```
aeneasr, line 700: These are complex functions but they do not have any tests. Can you please add unit tests to ensure that these behave the way you want?

persistence/sql/persister_identity.go
```diff
@@ -565,3 +598,103 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+			return q
+		case "postgres", "cockroach":
+			for _, value := range values {
+				param := fmt.Sprintf(`%s`, value)
+				if key != "" {
+					param = fmt.Sprintf(`{"%s":"%s"}`, key, value)
+				}
+				q = q.Where(fmt.Sprintf(`%s @> ?`, prefix), param)
+			}
+			return q
```
aeneasr, line 629: @zepatrik what do you think about this? It's basically SQL injection, but with some sanization in front (see `!validFields.MatchString(field)`). I'm not 100% convinced though yet that we should merge it as is. Can you PTAL?

persistence/sql/persister_identity.go
```diff
@@ -565,3 +598,103 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+				}
+			}
+			return q
+		case "postgres", "cockroach":
+			for _, value := range values {
+				param := fmt.Sprintf(`%s`, value)
+				if key != "" {
+					param = fmt.Sprintf(`{"%s":"%s"}`, key, value)
+				}
+				q = q.Where(fmt.Sprintf(`%s @> ?`, prefix), param)
```
aeneasr, line 627: This in particular is potentially dangerous as the prefix is not escaped.

persistence/sql/persister_identity.go
```diff
@@ -565,3 +598,103 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+
+		prefix, key := extractFieldAndInnerFields(field)
+
+		switch p.Connection(ctx).Dialect.Name() {
+		case "sqlite3", "mysql", "mariadb":
+			for _, value := range values {
+				if key == "" {
+					q = q.Where(fmt.Sprintf(`json_extract(%s, '$') = ?`, prefix), value)
+				} else {
+					q = q.Where(fmt.Sprintf(`json_extract(%s, '$.%s') = ?`, prefix, key), value)
```
aeneasr, line 617: Same here, otherwise it allows to use commands like `GRANT` or `NOW()` here

persistence/sql/persister_identity.go
```diff
@@ -565,3 +598,103 @@ func (p *Persister) injectTraitsSchemaURL(ctx context.Context, i *identity.Ident
...
+			return q
+		}
+
+		prefix, key := extractFieldAndInnerFields(field)
+
+		switch p.Connection(ctx).Dialect.Name() {
+		case "sqlite3", "mysql", "mariadb":
+			for _, value := range values {
+				if key == "" {
+					q = q.Where(fmt.Sprintf(`json_extract(%s, '$') = ?`, prefix), value)
```
aeneasr, line 615: The `%s` should be escaped, something like:

```suggestion
					q = q.Where(fmt.Sprintf(`json_extract(`%s`, '$') = ?`, prefix), value)
```

aeneasr (changes requested):
Thank you!!

I think there are some issues with potential SQL injections. We also need to add many more tests, preferably also fuzzing tests ( https://go.dev/doc/fuzz/ ) for the input parameters.

Additionally, we will need a dedicated test suite for `persister_identity.go` and we need to update the documentation :)

A good place to add documentation for this feature would be in: https://github.com/ory/docs/tree/master/docs/guides/manage-identities

aeneasr:
Another question - have you been able to test how performant this is when you have, let's say, 1 million identities loaded? I could imagine it being very slow / timing out for a certain store size :/ Can we use some type of index to speed up the lookup?

david972:
> Another question - have you been able to test how performant this is when you have, let's say, 1 million identities loaded? I could imagine it being very slow / timing out for a certain store size :/ Can we use some type of index to speed up the lookup?
maybe 
on `traits` it could be hard but on the other yes
I'll have a look

Sytten:
Anything we can do to help merge that? This is really a big piece missing right now from the API

david972:
> Anything we can do to help merge that? This is really a big piece missing right now from the API

Hi
Just need some help because this implementation returns duplicate results 
And more tests needs too

aeneasr:
The biggest challenge this PR faces is computing the query SQL correctly without the risking SQL injection. Another problem is one of performance as the indices required will quickly grow out of hand at scale. It might work for limited use cases but in general we need to:

1. Implement robust SQL validation
2. Add more tests
3. Improve queries
4. Verify scalability

ofairfoul:
Would it be possible to simply expose FindByCredentialsIdentifier on the Admin Identity API?
I think this would meet the needs of the majority of people who just want to find usernames(emails) which are already registered.
e.g. https://developer.okta.com/docs/reference/api/users/#get-user-with-login

kmherrmann:
IMHO we should scope this clearly to 

1. retrieval by credentials identifier as suggested in https://github.com/ory/kratos/pull/2442#issuecomment-1190867502
2. simple prefix/infix/regex(?) matching on identifiers
3. powerful full text search

If we're after (3), I believe we should explore using a dedicated backend like Elastic Search for the purpose, which can easily handle large data sets and can do scoring and ranking properly.

jonnyijapan:
This is interesting - would it be possible to add exactly what new features this PR would introduce in the top post? Like would it be possible to use the API as a lookup, like "get the identity having this credential identifier: 'test@test.xxx' ?

kmherrmann:
> This is interesting - would it be possible to add exactly what new features this PR would introduce in the top post? Like would it be possible to use the API as a lookup, like "get the identity having this credential identifier: '[test@test.xxx](mailto:test@test.xxx)' ?

This PR, IIUC, introduces (2), using additional SQL conditions. That's useful, but won't scale to (3) and (4) and comes with the mentioned security and performance caveats. 

A getByCredentialIdentifier equivalent (option 1) that's optimized for performance (maybe indexed columns and exact matching?) sounds like a good addition to me when used in serving (vs. administrative / support tools).

aeneasr:
After much consideration, I think that we can't move forward with the PR as is. The problem is that this functionality will work at very low scale but quickly erode in performance as we hit 1M+ entries. In the end, the database will need to load all rows into RAM, parse JSON, and match the fields. This just won't work at volume.

I'm open to have dedicated search for specific fields (such as verified address) but identity traits are out of the question. We'll need to do another set up with elastic or something similar instead.

aeneasr:
Sorry for all the work you have put in @david972 . I think it's per-se a great PR, but the task at hand is much bigger than what we thought initially. There are so many edge cases to cover (fuzzing, sql injection, performance), that this will need a much larger architectural effort involving proven technologies (elastic, lucene, ...) to offer something that can scale.

Sending out much Ory love :)

david972:
Hi I understand
maybe we can just open a discussion on how to do it properly ?

aeneasr:
Absolutely!

zFlux:
@aeneasr @kmherrmann - it sounds like @kmherrmann 's option 1) implementing @ofairfoul 's suggestion i.e. expose FindByCredentialsIdentifier on the Admin Identity API (so long as we're ensuring exact match only and index use) is still ok to attempt a PR on?

david972:
transferred here #2783

PR #2468, opened 2022-05-13
fix: unreliable HIBP caching strategy

zepatrik:


selfservice/strategy/password/validator.go
```diff
@@ -107,20 +111,20 @@ func lcsLength(a, b string) int {
 	return greatestLength
 }
 
-func (s *DefaultPasswordValidator) fetch(hpw []byte, apiDNSName string) error {
+func (s *DefaultPasswordValidator) fetch(hpw []byte, apiDNSName string) (int64, error) {
```
zepatrik, line 114: Returning the count here does not rely on the cache having the value in the recursive run.

selfservice/strategy/password/validator_test.go
```diff
@@ -117,73 +123,111 @@ func TestDefaultPasswordValidationStrategy(t *testing.T) {
...
+			t.Run(fmt.Sprintf("case=%s/expected err=%s", tc.name, tc.expectErr), func(t *testing.T) {
+				pw := randomPassword(t)
+				hash := hashPw(t, pw)
+				hibpResp <- tc.res(t, hash)
+
+				err := s.Validate(context.Background(), "", pw)
+				assert.ErrorIs(t, err, tc.expectErr)
 			})
+
+			// verify the fetch was done, i.e. channel is empty
```
zepatrik, line 223: We had some fishy cases, therefore we ensure here the fetch was done.
Especially the first two cases had the same password, therefore cache hit.

selfservice/strategy/password/validator.go
```diff
@@ -135,18 +139,22 @@ func (s *DefaultPasswordValidator) fetch(hpw []byte, apiDNSName string) error {
 		if len(result) == 2 {
 			count, err = strconv.ParseInt(result[1], 10, 64)
 			if err != nil {
-				return errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Expected password hash to contain a count formatted as int but got: %s", result[1]))
+				return 0, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Expected password hash to contain a count formatted as int but got: %s", result[1]))
 			}
 		}
 
 		s.hashes.SetWithTTL(prefix+result[0], count, 1, hashCacheItemTTL)
+		if prefix+result[0] == b20(hpw) {
```
aeneasr, line 147: What is this check good for?

zepatrik: `thisCount` is the count of the actual password. We fetch a whole range of password hashes and add it to the cache. It could be that there is not a hit at all. In that case we want to return zero. But in case of a hit, we want to update `thisCount` before returning, which this if does.

PR #2471, opened 2022-05-15
feat: add PATCH to adminUpdateIdentity (#2380)

### LINKED ISSUE #2380: Support `PATCH` for identities
### Preflight checklist

### Describe your problem

I think it sholud be possible to use `adminUpdateIdentity` without passing a traits object in it (or a `nil`).

I'm using it only to change state of the identity (active/inactive).

Is there another way?

### Version

0.9.0-alpha.3

jonas-jonas:
Closes #2380

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


jonas-jonas:
Had to force-push, because I messed up the OpenAPI model names, sorry.

It would also be possible to update the PUT version of this endpoint instead to support partial payloads as well, just like my current version. WDYT?

jonas-jonas:
The test `TestDefaultPasswordValidationStrategy/max_breaches/case=should_fail_if_response_contains_more_than_maxBreachesThreshold` (file `selfservice/strategy/password/validator_test.go`) might be flaky? It passes on my machine and seems to be unrelated to the changes. Is there a way to re-run the CI task?

zepatrik:
> It would also be possible to update the PUT version of this endpoint instead to support partial payloads as well, just like my current version.

Is that a breaking change?

> The test `TestDefaultPasswordValidationStrategy/max_breaches/case=should_fail_if_response_contains_more_than_maxBreachesThreshold (file selfservice/strategy/password/validator_test.go)` might be flaky?

See #2468

jonas-jonas:
> Is that a breaking change?

Kind of, it would "break" the validator if the payload does not contain the field which would be the desired behavior, though. E.g. the validator would not fail if a field is not supplied. That would not be huge issue IMO, as the missing field would just not be updated, but that behavior would not comply with the spec of a `PUT` request, [as it should always update the whole entity.](https://restful-api-design.readthedocs.io/en/latest/methods.html#patch-vs-put) So a PATCH method would be more appropriate (read: spec compliant), I guess.

> See #2468
I see, so the failure should be fixed with that PR?

identity/handler.go
```diff
@@ -413,6 +416,40 @@ type AdminUpdateIdentityBody struct {
 	State State `json:"state"`
 }
 
+// swagger:parameters adminPatchIdentity
```
zepatrik, line 419: It would be good to change the order here so that the models are just before their handler.

identity/handler.go
```diff
@@ -524,3 +561,84 @@ func (h *Handler) delete(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+
+		stateChangedAt := sqlxx.NullTime(time.Now())
+
+		identity.State = ur.State
+		identity.StateChangedAt = &stateChangedAt
+	}
+
+	if len(ur.Traits) > 0 {
+		identity.Traits = []byte(ur.Traits)
+	}
```
zepatrik, line 625: The nested traits are not patched but replaced here. We should mention that in the description.

identity/handler_test.go
```diff
@@ -460,6 +460,75 @@ func TestHandler(t *testing.T) {
...
+				t.Run("endpoint="+name, func(t *testing.T) {
+					ur := identity.AdminPatchIdentityBody{
+						Traits: []byte(`{"subject":"updated_via_patch", "foo":"bar"}`),
+					}
+
+					res := send(t, ts, "PATCH", "/identities/"+i.ID.String(), http.StatusOK, &ur)
+					assert.EqualValues(t, "updated_via_patch", res.Get("traits.subject").String(), "%s", res.Raw)
+					assert.EqualValues(t, "bar", res.Get("traits.foo").String(), "%s", res.Raw)
+					assert.EqualValues(t, "", res.Get("metadata_admin.admin").String(), "%s", res.Raw)
+					assert.EqualValues(t, "", res.Get("metadata_public.public").String(), "%s", res.Raw)
```
zepatrik, line 482:
```suggestion
					assert.False(t, res.Get("metadata_admin.admin").Exists(), "%s", res.Raw)
					assert.False(t, res.Get("metadata_public.public").Exists(), "%s", res.Raw)
```

identity/handler_test.go
```diff
@@ -460,6 +460,75 @@ func TestHandler(t *testing.T) {
...
+					assert.EqualValues(t, identity.StateActive, res.Get("state").String(), "%s", res.Raw)
+					assert.EqualValues(t, creationMilli, res.Get("state_changed_at").Time().UnixMilli(), "%s", res.Raw)
+
+					res = get(t, ts, "/identities/"+i.ID.String(), http.StatusOK)
+					assert.EqualValues(t, i.ID.String(), res.Get("id").String(), "%s", res.Raw)
+					assert.EqualValues(t, "updated_via_patch", res.Get("traits.subject").String(), "%s", res.Raw)
+					assert.EqualValues(t, "bar", res.Get("traits.foo").String(), "%s", res.Raw)
+					assert.EqualValues(t, "", res.Get("metadata_admin.admin").String(), "%s", res.Raw)
+					assert.EqualValues(t, "", res.Get("metadata_public.public").String(), "%s", res.Raw)
```
zepatrik, line 491:
```suggestion
					assert.False(t, res.Get("metadata_admin.admin").Exists(), "%s", res.Raw)
					assert.False(t, res.Get("metadata_public.public").Exists(), "%s", res.Raw)
```

identity/handler_test.go
```diff
@@ -460,6 +460,75 @@ func TestHandler(t *testing.T) {
...
+			for name, ts := range map[string]*httptest.Server{"public": publicTS, "admin": adminTS} {
+				t.Run("endpoint="+name, func(t *testing.T) {
+					ur := identity.AdminPatchIdentityBody{
+						State: identity.StateInactive,
+					}
+
+					res := send(t, ts, "PATCH", "/identities/"+i.ID.String(), http.StatusOK, &ur)
+					assert.EqualValues(t, uuid, res.Get("traits.subject").String(), "%s", res.Raw)
+					assert.EqualValues(t, "", res.Get("metadata_admin.admin").String(), "%s", res.Raw)
+					assert.EqualValues(t, "", res.Get("metadata_public.public").String(), "%s", res.Raw)
```
zepatrik, line 517:
```suggestion
					assert.False(t, res.Get("metadata_admin.admin").Exists(), "%s", res.Raw)
					assert.False(t, res.Get("metadata_public.public").Exists(), "%s", res.Raw)
```

identity/handler_test.go
```diff
@@ -460,6 +460,75 @@ func TestHandler(t *testing.T) {
...
+					assert.EqualValues(t, identity.StateInactive, res.Get("state").String(), "%s", res.Raw)
+					assert.Less(t, creationMilli, res.Get("state_changed_at").Time().UnixMilli(), "%s", res.Raw)
+
+					res = get(t, ts, "/identities/"+i.ID.String(), http.StatusOK)
+					assert.EqualValues(t, i.ID.String(), res.Get("id").String(), "%s", res.Raw)
+					assert.EqualValues(t, uuid, res.Get("traits.subject").String(), "%s", res.Raw)
+					assert.EqualValues(t, "", res.Get("metadata_admin.admin").String(), "%s", res.Raw)
+					assert.EqualValues(t, "", res.Get("metadata_public.public").String(), "%s", res.Raw)
```
zepatrik, line 525:
```suggestion
					assert.False(t, res.Get("metadata_admin.admin").Exists(), "%s", res.Raw)
					assert.False(t, res.Get("metadata_public.public").Exists(), "%s", res.Raw)
```

zepatrik (changes requested):
Nice, this looks very good :+1:
Just some minor remarks.

identity/handler.go
jonas-jonas: Right, done.

identity/handler.go
jonas-jonas: True, I added a comment to the documentation comment.

It might make sense to support patching here, though I think that may cause a lot of confusion. Maybe something for later on...

zepatrik: Hm but that will be a potentially breaking change. I'd rather decide now to either support it or not here. We can also add a separate endpoint for that at a later point.

jonas-jonas: I see, if I understand that part correctly, the `traits` field could be any arbitrary data type, e.g. a string? That means someone could put an XML string in there, which would break the patching ability here. So in order to completely support patching on that field, there would need to be a lot of logic involved. Also, I am not sure if there is support for deep merging of JSON fields out of the box here. Which means, that needs to be implemented.

I think an extra endpoint for updating traits could be a good solution later on.

jonas-jonas:
Alright, I added enough tests to cover every path in the new handler function. The coverage should now be sufficient (even though the project coverage is failing?). 

Sorry for all the noisy commits on this PR. Probably best to squash+rebase on merge, if possible.

aeneasr (changes requested):
Thank you! I think code-wise, this looks mergable! Unfortunately, we chose JSON Patch as th option to implement patching in Ory Hydra and Ory Cloud already, which allows you to truly target any element in the JSON and update it and not be constrainted by a missing `nil` check.

Thus, I suggest to either create a new PR using JSON Patch following this discussion https://github.com/ory/kratos/issues/2380#issuecomment-1150309489

I'm sorry that you put in all the effort already :( Please feel free to try and get feedback from maintainers on your ideas before going down the road of implementation :)

@zepatrik let's enforce JSON Patch for all PATCH endpoints, unless there is a good reason not do to it (e.g. because PATCH is used in a different context).

jonas-jonas:
@aeneasr I think JSON Patch could be implemented in the current change as well, right? I have no experience with JSON Patch, but it seems to just be another `if` check in the handler of the `patch` handler? WDYT?

> I'm sorry that you put in all the effort already :( Please feel free to try and get feedback from maintainers on your ideas before going down the road of implementation :)

I regard it as a learning experience, so not an issue :). Initially the issue seemed quite clear-and-cut so I didn't want to bother any one there.

Also unfortunately I am down with a little known sickness going around currently, but getting better already ;). So sorry for the delays on my open PRs/issues.

aeneasr:
> Also unfortunately I am down with a little known sickness going around currently, but getting better already ;). So sorry for the delays on my open PRs/issues.

No problem, I wish you a speedy recovery!

> @aeneasr I think JSON Patch could be implemented in the current change as well, right? I have no experience with JSON Patch, but it seems to just be another if check in the handler of the patch handler? WDYT?

Check the Ory Hydra implementation here, I would follow that one :)

https://github.com/ory/hydra/blob/996258d50ec620c89a8f55a98436320ab99db62b/client/handler.go#L343-L378

jonas-jonas:
I have now added JSON Patch to the endpoint and this PR is now ready for review again. 

I spent quite a bit of time, trying to make the `credentials` field updatable as well. Unfortunately the update of credentials needs to import them, which requires us knowing what actually changed about the `Credentials` object, which seems hard to do nicely. If you have an idea here, let me know and I'll investigate it. For now, I have blocked the field from updating. 

Sidenote: This also has a few benefits, as it would be possible to [`copy`](https://jsonpatch.com/#copy) the hashed password into a readable field and thus expose the password hash on (semi-) public endpoints (`GET /identity/:id`, etc.).

An alternative solution is to provide another endpoint that allows updating credentials directly, without relying on JSON Patch. 

Something like `PUT /identities/:id/credentials/:type` with `{ "password": "newpassword123" }` (or similar for OIDC). IMO this would be cleaner to implement, but potentially tedious to use for a user. What do you think here? 

Also, if the data-structure changes in the future we could just remove the `/credentials` path from the `denyPaths` in JSON Patch and implement the credential import logic there, as it would just add functionality without removing anything existing. This would make disallowing `credentials` updates for now relative risk-free IMO.

x/json.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Applies a JSON Patch to an object while checking for denied paths in the patch
+//
+// If the patch contains an update to a field matching a denied path,
+// the whole patch is "rejected" and an error is returned
+func ApplyJSONPatch(p json.RawMessage, object interface{}, denyPaths ...string) error {
+	patch, err := jsonpatch.DecodePatch(p)
+	if err != nil {
+		return err
+	}
```
Benehiko, line 21: Why do we copy this file from the `ory/x` repository?

https://github.com/ory/x/blob/master/jsonx/patch.go

Is it not the same?

x/json_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+func TestApplyJSONPatch(t *testing.T) {
+	object := TestType{
+		Field1: "foo",
+		Field2: []string{
+			"foo",
+			"bar",
+			"baz",
+			"kaz",
+		},
```
Benehiko, line 27: The same for this file, do we change anything specific here from `ory/x`?
https://github.com/ory/x/blob/master/jsonx/patch_test.go

identity/handler_test.go
```diff
@@ -503,6 +519,112 @@ func TestHandler(t *testing.T) {
 			}
 		})
 
+		t.Run("case=PATCH update of state should update state changed at timestamp", func(t *testing.T) {
```
Benehiko, line 522: We require a test for PATCH validating that an identities `credentials` cannot be updated through this method.

Benehiko (changes requested):
This PR looks quite good so far :)

I think we have some duplicate code here and missing tests, but it's almost there

x/json.go
jonas-jonas: You are absolutely right, I just looked at hydra and copied it from there. I did have a few modifications in mind, but it ended up working without it. I removed the files.

x/json_test.go
jonas-jonas: See https://github.com/ory/kratos/pull/2471#discussion_r917421577

identity/handler_test.go
jonas-jonas: Right, I added the test.

Benehiko (approved):
I think this looks quite good to me :)

Thank you for taking the time to work on this!

jonas-jonas:
@Benehiko Thanks for your review. :) 

I squashed the commits & rebased them so your approval got dismissed. I think @aeneasr needs to approve as well though.

Benehiko (approved):

identity/handler.go
```diff
@@ -540,3 +545,86 @@ func (h *Handler) delete(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+// nolint:deadcode,unused
+type adminPatchIdentity struct {
+	// ID must be set to the ID of identity you want to update
+	//
+	// required: true
+	// in: path
+	ID string `json:"id"`
+
+	// in: body
+	Body jsonpatch.Patch
```
aeneasr, line 559: Unfortunately, this does no generated a correct OpenAPI definition (check the `spec/api.json` file). The problem is that we're using an external library here. This does not know go-swagger (the tool we use to extract the OpenAPI Spec from Go Code) and does not have the correct struct annotations.

To fix this, include this struct instead: https://github.com/ory/x/blob/463cd057b17c6d312dd26dd22987369d7c721c6b/openapix/jsonpatch.go#L6

Please also note that JSON Patch APIs usually allow us to send an array of patches, not just one patch :) That's why we want to use JSONPatchDocument instead!

There's one more thing though. The problem is that go-swagger is incorrectly tagging struct fields that are `interface{}`. To correctly handle the type in OpenAPI, you will need to add a JSON Patch for the OpenAPI spec, like we do in Ory Hydra:

https://github.com/ory/hydra/blob/e3c319eb1b1d62b44139dd583392f4ac9bf8ecc7/.schema/openapi/patches/common.yaml#L1-L2

Lastly, you'll need to add the `github.com/ory/x/openapix` package to the paths to resolve in the Makefile, like we do in Ory Hydra: 

https://github.com/ory/hydra/blob/e3c319eb1b1d62b44139dd583392f4ac9bf8ecc7/Makefile#L112

The API tooling is currently very fiddely. I hope that we can improve this some day!

aeneasr (changes requested):
Thank you! Great tests and the code now also looks very good. One last thing we need to address is the OpenAPI specification. When making changes to the APIs, you'll always need to also check if the new content in `spec/openapi.json` look correct :)

identity/handler.go
jonas-jonas: Thanks for the write-up. I added the `jsonpatch.Patch` struct as a placeholder & then forgot to check what the output is. The Patch body should now generate correctly.

Unfortunately, I now get some seemingly unrelated changes in the `Pagination` & `PaginationHeaders` files when generating with the same setup as you have in hydra. I am not sure what the issue is and could not figure out how they relate to my changes. Could you have a look at that and let me know if there is an issue?

aeneasr: damn :/ that is probably because of https://github.com/ory/x/blob/463cd057b17c6d312dd26dd22987369d7c721c6b/openapix/pagination.go#L3-L42 which I recently changed. I'll fix that and release a new ory/x version

aeneasr: https://github.com/ory/x/compare/token-pagination-naming?expand=1

aeneasr: Updating to ory x [v0.0.439](https://github.com/ory/x/releases/tag/v0.0.439) should fix this problem

jonas-jonas: I used the newest version of `ory/x`. I am still generating unrelated models, but I guess that is intended?
Might want to that in a separate commit to master instead though. I can open a separate PR to update ory/x and generate the necessary models, if needed.

jonas-jonas: The dependency updates seem to have a broken atleast the test `TestNewRecoveryEmailAddress` in  `identity/identity_recovery_test.go` as a [UUID consisting of only zeroes is now considered empty](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L579), which makes the assertion fail. I fixed the test here, but I still think this should be done in a separate PR. Let me know how you do those kind of things here :)

identity/identity_recovery_test.go
```diff
@@ -12,7 +12,7 @@ func TestNewRecoveryEmailAddress(t *testing.T) {
 	iid := x.NewUUID()
 	a := NewRecoveryEmailAddress("foo@ory.sh", iid)
 
-	assert.Equal(t, a.Value, "foo@ory.sh")
-	assert.Equal(t, a.Via, RecoveryAddressTypeEmail)
-	assert.NotEmpty(t, a.ID)
+	assert.Equal(t, "foo@ory.sh", a.Value)
+	assert.Equal(t, RecoveryAddressTypeEmail, a.Via)
+	assert.Equal(t, iid, a.IdentityID)
```
aeneasr, line 17: Should we not keep the test for `a.ID`? You can do e.g. `assert.NotEqual(t, uuid.Nil, a.ID)`

aeneasr (changes requested):
Almost! :) :) :)

identity/identity_recovery_test.go
jonas-jonas: That's the issue though, it _is_ Equal. `uuid.Nil` is in essence an array of length 16 of 0. And that's what is set into the struct here, as the ID isn't set by `NewRecoveryEmailAddress(...)`.

jonas-jonas: I've now also fixed some places where the check was now failing. 

The issue boils down to the following:
By default, nothing is set into the ID fields of the struct when calling their `New...` functions. This means, that they're populated with the default value for `uuid.UUID` which is, in essence, a byte array with 16 elements, all of which are `0x00`. Now the `assert.NotEmpty` check was changed from looking only at the length of array types to determine whether their empty to comparing them to the default value of the declared type. If this comparison yields `true` the test case fails. 

This is the commit in `stretchr/testify` that introduced this behaviour: https://github.com/stretchr/testify/commit/840cb801497147a6d30a568e0874dfefb10867a9

aeneasr: I see, in that case it should be assert.Equal(t, uuid.Nil, id) :)

jonas-jonas: I‘d say that would be testing language features, but the check shouldn’t hurt, so I‘ll add it tomorrow. :) 
Thanks for your patience :)

jonas-jonas: I added a `assert.Equal(t, uuid.Nil, id)` check. Please have another look, thanks :)

aeneasr: It doesn't looks like the assertin ohas been added above, can you please check?

identity/identity_verification_test.go
```diff
@@ -20,5 +20,5 @@ func TestNewVerifiableEmailAddress(t *testing.T) {
 	assert.Equal(t, a.Status, VerifiableAddressStatusPending)
 	assert.Equal(t, a.Verified, false)
 	assert.EqualValues(t, nullTime, a.VerifiedAt)
-	assert.NotEmpty(t, a.ID)
+	assert.Equal(t, iid, a.IdentityID)
```
aeneasr, line 23: Same here, please re-add the empty ID check :)

identity/identity_recovery_test.go
aeneasr: I also just realised that the original assertion was `NotEmpty` while it actually should have been `assert.Empty`, which probably didn't help with confusion. Sorry about that, that git blame is on me ;)

aeneasr (approved):

jonas-jonas:
Sorry, to bump this again, but I have lost track of the open things on this.

I think the things you requested to be changed are done now. Is this correct? I saw, you also started to contribute the changes to the test cases in a separate PR. Should we wait until that is merged, before moving on with this?

PR #2516, opened 2022-06-10
improvement: return enum values in node meta

### LINKED ISSUE #2459: use appropriate input type for enum traits
### Preflight checklist

- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

When a trait is defined as an enum, Kratos should return the node either as a select or a group of radio buttons. At the moment, it returns a text input.

This has been requested before in #175, but the related PR only handled the `pattern` attribute.

see above

none

### Version

v0.9.0-alpha.3

### Additional Context

I'm storing the user's locale in Kratos. With the information returned by Kratos, the form only shows the allowed values in the validation message that is returned after submission.

```json
{
  "properties": {
    "traits": {
      "type": "object",
      "properties": {
        "locale": {
          "type": "string",
          "enum": ["en", "ja"],
          "default": "ja"
        }
      }
    }
  }
}
```

woylie:
Adds `enum` values from identity schema to node meta.

I chose the meta instead of the attributes because there is no official enum attribute for HTML inputs. I didn't change the input type, because I didn't want to make assumptions about how to render a field with enum values - you may want to use radio buttons, a select, a multi select, maybe checkboxes if it's an array, or maybe a data list. Just returning the values gives enough of a hint to modify the UI, I think.

While an enum in a JSON schema may contain any values, I return them as strings. HTML forms only produce strings anyway, so I don't think it matters if non-string values are encoded as strings or not.

I'm not too familiar with Go, so apologies if I did something stupid.

resolves #2459 

## Checklist


aeneasr:
> I chose the meta instead of the attributes because there is no official enum attribute for HTML inputs. I didn't change the input type, because I didn't want to make assumptions about how to render a field with enum values - you may want to use radio buttons, a select, a multi select, maybe checkboxes if it's an array, or maybe a data list. Just returning the values gives enough of a hint to modify the UI, I think.

I understand! I think from an implementation standpoint it's probably better for us to take a stand and just say: we think this is `<select><option></select>`. And if someone thinks differently, they can implement this differently (e.g. with checkboxes).

That way we can easily add a new field to the renderers and it feels natural. Adding a new field "type" that needs interpretation is just one more thing that needs to be understood, and has not been used as such before.

What do you think about making this select/option types? It would also allow us to e.g. add labels (for translation) to the options if that ever becomes a thing.

woylie:
> It would also allow us to e.g. add labels (for translation) to the options if that ever becomes a thing.

If you want to add labels, this couldn't be represented with an `enum` in a JSON schema alone. What do you have in mind?

woylie:
Something like this?

```json
{
  "type": "select",
  "group": "default",
  "attributes": {
    "name": "some_trait",
    "node_type": "select",
    "options": [
        {"label": "Some value", "value": "some_value"},
        {"label": "Some other value", "value": "some_other_value"}
    ],
    "value": "some_value",
    "required": true,
    "disabled": false
  },
  "messages": null,
  "meta": {}
}
```

For enums the option label would be the same as the value, but we keep the option open to support separate labels for other use cases in the future.

aeneasr:
🤔 good question, I'm wondering whether it is warranted to introduce a new children attribute or keep it under attributes as you suggested. It would probably be a long term "good" decision if we ever need to add children somewhere else, but a bit blown out of proportion for what this PR in particular tries to achieve.

woylie:
With children we would have something like this:

```json
{
  "type": "select",
  "group": "default",
  "attributes": {
    "name": "some_trait",
    "node_type": "select",
    "value": "some_value",
    "required": true,
    "disabled": false
  },
  "messages": null,
  "meta": {},
  "children": [
    "type": "option",
    "attributes": {
      "name": "some_trait",
      "node_type": "option",
      "value": "some_value",
      "text": {
          "id": 1234567, (?)
          "text": "Some value"
       }
    }
  ]
}
```

Not sure about the `text` attribute, I just tried to reuse the data structures that are already there. Also, instead of adding the `value` attribute to the select node, there could be a `selected` attribute on the `option`, to mirror HTML.

For me, the most important thing right now is just to get the enum values from the API. text input + meta.enum would already solve that, and adding an additional `select` node type would additive. But if we add that `select` type now, we should decide on the proper format now, I think, to avoid API changes later.

woylie:
After thinking a bit more about this, I'd prefer the solution with the inlined options attribute instead of a generic children property. There will only ever be one allowed child element for selects, which is `<option>`. Building this abstraction around the `select` element makes it easier to implement the UI.  The generic `options` attribute can be easily used to switch out the `select` for radio buttons, or checkboxes, or whatever. Probably easier than parsing a list of generic child elements.

And even if you think beyond select elements, the UI container is not meant to render highly complex user interfaces anyway. Should there ever be more nested node types, they will probably be for specific HTML elements in specific situations, instead of arbitrarily nestable. So even for potential other new node types, it probably makes sense to build some kind of abstraction around them. That makes it easier to customize the UI, compared to figuring out which child elements might be returned at a certain point (and if the children are just UI nodes, you potentially also have to deal with recursion).

aeneasr:
That's a fair point! Enums are also, by definition, a "select one of multiple", so select/option is a really good option to choose from here IMO.

woylie:
I updated the PR to add a `select` node. Some points:

- Since a select is technically not an input, I added a separate module for the select attributes, with a separate `NewSelectFieldFromSchema` function (and similar ones). The switch between an input and a select node is made in `NodesFromJSONSchema` depending on the presence of the `enum` property in the schema.
- I put the current value on the attributes (as in the input attributes) instead of the option. That makes more sense to me, since we're dealing with an abstraction anyway, and you likely use some component for rendering selects which takes the value as an argument.
- Since `NewFromHTTPRequest` does not take the JSON schema as an argument, there is no way of knowing whether to use a select or an input. Not sure whether that causes any issues and we need to make the schema available to that function?
- If you're happy with the PR, I'll open a PR to update the [UI customization docs](https://github.com/ory/docs/blob/master/docs/kratos/concepts/ui-user-interface.mdx).

ui/node/attributes.go
```diff
@@ -217,12 +218,63 @@ type ScriptAttributes struct {
...
+//
+// swagger:model uiNodeSelectAttributes
+type SelectAttributes struct {
+	// The input's element name.
+	//
+	// required: true
+	Name string `json:"name"`
+
+	// The input's value.
+	FieldValue interface{} `json:"value,omitempty" faker:"string"`
```
aeneasr, line 231: Do we need this field?

aeneasr (changes requested):
Thank you, this looks really good to me already! Can you please:

1. [ ] Add a new e2e tests for enums, preferrably in `/test/e2e/cypress/integration/profiles/email/schemas.ts`. You can use [`useConfigProfile` and `useProxy`](https://github.com/ory/kratos/blob/6284a9a5152924018d85f306e5758e9d8d759283/test/e2e/cypress/integration/profiles/mfa/totp.spec.ts#L7-L27) to toggle between the two apps (node and react). Use [`setIdentitySchema`](https://github.com/ory/kratos/blob/6284a9a5152924018d85f306e5758e9d8d759283/test/e2e/cypress/support/commands.ts#L84) to set a custom schema. It's a good idea to run these two selfservice [apps locally while developing](https://github.com/ory/kratos/blob/6284a9a5152924018d85f306e5758e9d8d759283/test/e2e/run.sh#L66-L80).
2. [ ] Update the self-service node & nextjs apps to support select
3. [ ] Update the docs to include the new field type

Thank you :)

ui/node/attributes.go
woylie: Yes, that's the currently selected value. See remarks above.

aeneasr:
@woylie let me know when this is good for another review

woylie:
@aeneasr I added an e2e test, is that what you had in mind?

I tried running the e2e tests locally with `./test/e2e/run.sh --dev sqlite` according to the readme, but I'm getting this error:

```
making HTTP(S) get request to  url:http://localhost:4455/health/ready ...
making HTTP(S) get request to  url:http://localhost:4456/health/alive ...
making HTTP(S) get request to  url:http://localhost:4458/ ...
  HTTP(S) error for http://localhost:4456/health/alive Error: connect ECONNREFUSED 127.0.0.1:4456
  HTTP(S) error for http://localhost:4458/ Error: connect ECONNREFUSED 127.0.0.1:4458
  HTTP(S) error for http://localhost:4455/health/ready Error: Request failed with status code 500
```

UncleSamSwiss:
This PR looks great and is exactly what we are also missing in kratos (for the same reason: language selection).

@woylie Is there anything blocking you? Can we help in any way?
@aeneasr Do we have a chance of getting this feature in 1.0?

woylie:
Hey @UncleSamSwiss, the to do list that @aeneasr posted is still open. I had some trouble with the dev environment and then got busy with other things. If you want to take over and finish this, feel free to take over!

woylie:
stale

Issue #2396, opened 2022-04-14
Improve input attributes for password managers / accessibility


woylie:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

I'd like to add some attributes to help password managers understand Kratos forms better, and improve accessibility in general.

### Describe your ideal solution

- [x] add `autocomplete` attribute to password inputs (`new-password` or `current-password` depending on the form) #2523
- [x] add additional autocomplete attributes to email and name fields etc. ~~(might need option in identity schema to set those)~~ #2523
- [ ] add `minlength` and `maxlength` attributes to password inputs (when setting a new password)
- [ ] add `passwordrules` attribute to password inputs (when setting a new password)
- [ ] add `id` attributes to all input nodes, so that we can easily set the `for` attributes on `<label>` without generating IDs in the UI
- [ ] ensure the selfservice UI applications set the `aria-describedby` attribute on inputs when there are input messages

## References

- https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete
- https://developer.apple.com/password-rules/
- https://boldist.co/development/autofill-form-data/
- https://hiddedevries.nl/en/blog/2018-01-13-making-password-managers-play-ball-with-your-login-form
- https://support.1password.com/compatible-website-design/
- https://support.lastpass.com/help/how-do-i-configure-my-website-to-work-with-lastpass

### Workarounds or alternatives

none

### Version

v0.9.0-alpha.3

vinckr:
This would be a feature request for https://github.com/ory/kratos-selfservice-ui-node, right?

aeneasr:
[...]

PR #2523, opened 2022-06-13
improvement: add autocomplete attributes

woylie:
- adds `autocomplete` attribute to node attributes
- sets the `autocomplete` attribute on password inputs to `new-password` (registration, settings) or `current-password` (login)
- sets the `autocomplete` attribute for email, tel and url inputs depending on the `format` of the trait
- sets input type to `tel` if trait format is `tel`

This relates to two of the points mentioned in #2396.

## Checklist

### DIFF
--- a/spec/swagger.json
+++ b/spec/swagger.json
@@ -4205,6 +4205,18 @@
         "node_type"
       ],
       "properties": {
+        "autocomplete": {
+          "description": "The autocomplete attribute for the input.",
+          "type": "string",
+          "enum": [
+            "email",
+            "tel",
+            "url",
+            "current-password",
+            "new-password",
+            "one-time-code"
+          ]
+        },
         "disabled": {
           "description": "Sets the input's disabled field to true or false.",
           "type": "boolean"
--- a/ui/node/attributes.go
+++ b/ui/node/attributes.go
@@ -9,6 +9,7 @@ const (
 	InputAttributeTypeCheckbox      InputAttributeType = "checkbox"
 	InputAttributeTypeHidden        InputAttributeType = "hidden"
 	InputAttributeTypeEmail         InputAttributeType = "email"
+	InputAttributeTypeTel           InputAttributeType = "tel"
 	InputAttributeTypeSubmit        InputAttributeType = "submit"
 	InputAttributeTypeButton        InputAttributeType = "button"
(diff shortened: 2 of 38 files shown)


ui/node/attributes.go
```diff
@@ -9,16 +9,29 @@ const (
...
 	InputAttributeTypeEmail         InputAttributeType = "email"
+	InputAttributeTypeTel           InputAttributeType = "tel"
 	InputAttributeTypeSubmit        InputAttributeType = "submit"
 	InputAttributeTypeButton        InputAttributeType = "button"
 	InputAttributeTypeDateTimeLocal InputAttributeType = "datetime-local"
 	InputAttributeTypeDate          InputAttributeType = "date"
 	InputAttributeTypeURI           InputAttributeType = "url"
 )
 
+const (
```
aeneasr, line 20: You will need to define these in the swagger spec manually, similar to

https://github.com/ory/kratos/blob/037c0957cca2d6695b83b564efeaf17105d99b19/.schema/openapi/patches/identity.yaml#L6

spec/swagger.json
```diff
@@ -4192,6 +4192,9 @@
         }
       }
     },
+    "uiNodeInputAttributeAutocomplete": {
+      "type": "string"
```
aeneasr, line 4196: So basically we want an ENUM here, see my other comment

woylie:
@aeneasr Thanks for the review. I pushed a change that turns the attribute into an enum in the docs, but without the patch. This is analogous to the `UiNodeGroup` and `UiNodeType` enums. We could do the same for the `IdentityCredentialType` enum as well, actually.

woylie:
I ran `UPDATE_SNAPSHOTS=true go test -p 4 -tags sqlite -short ./...` from the `selfservice/strategy` folder. Running the tests regularly on my machine doesn't return any errors. I don't know why the CI is complaining about snapshots not being up to date.

Also, is there a way to run the webhook integration tests locally?

woylie:
> You need Docker installed and then do:
> 
> ```
> UPDATE_SNAPSHOTS=true go test -p 4 -tags sqlite ./...
> ```
> 
> > We could do the same for the IdentityCredentialType enum as well, actually.
> 
> Let's keep it as is for now, there was a reason for this!

Thanks, everything's green now.

aeneasr (approved):

PR #2529, opened 2022-06-17
Selfservice: change flow sort, oidc below password

afreakk:
Changed the order, to have password group above oidc, to fix so the top submit  button is not the oidc.
So enter = password, not oidc

There might be a reason for the previous order I dont know about.  
But using nextjs  or nodejs selfservice reference  implementation with oidc enabled, it will render oidc on top which causes enter press inside a input to trigger the oidc action.  
Which is very annoying :)  
image
image

## Checklist

## Further Comments
I have not tested webauthn and the others, so, dont know if this is suboptimal in relation to them.

### DIFF
--- a/selfservice/flow/login/sort.go
+++ b/selfservice/flow/login/sort.go
@@ -9,10 +9,10 @@ import (
 func sortNodes(ctx context.Context, n node.Nodes) error {
 	return n.SortBySchema(ctx,
 		node.SortByGroups([]node.UiNodeGroup{
-			node.OpenIDConnectGroup,
 			node.DefaultGroup,
 			node.WebAuthnGroup,
 			node.PasswordGroup,
+			node.OpenIDConnectGroup,
 			node.TOTPGroup,
 			node.LookupGroup,
 		}),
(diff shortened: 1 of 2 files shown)


aeneasr:
Thank you, that is indeed annoying! But instead of re-ordering, could we use [tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) instead?

afreakk:
yeah, that would require supporting code in the frontends though, and would still render the `google login` button above the rest.  
I dont care much about that, but I think normally `login with google` is below the usr/pw input fields.

aeneasr:
> yeah, that would require supporting code in the frontends though

That is relatively straight forward :) Basically just needs to be added:

https://github.com/ory/kratos-selfservice-ui-node/blob/7d5ac8831dc50708d4a0df4495ba6dd450595c21/views/partials/ui_node_input_default.hbs#L8-L14

and here:

https://github.com/ory/kratos-selfservice-ui-react-nextjs/blob/f3c98d927a2481688068c4e262298c05d70411e6/pkg/ui/NodeInputDefault.tsx#L27-L37

afreakk:
Instead of tabindex, could we use `type=button` instead of `type=submit` on oidc login?
That works actually.

aeneasr:
As there hasn't been any progress on this PR, I'm closing it as we do not want to change the order globally. But you're of course always welcomed to contribute changes to tabindex as discussed :)

PR #2548, opened 2022-06-28
fix: return an empty array not null when there are no sessions

kszafran:
The `GET /admin/identities/{id}/sessions` endpoint would return `null` instead of `[]` if the identity had no sessions (or the identity didn't exist). It's very easy to miss this case when deserializing responses into slices in tests.

## Checklist


kszafran:
@aeneasr Somewhat related observation - when I call `DELETE /admin/identities/{id}/sessions` and the identity has no sessions, I get 404. Is that intended behavior? It seems to me that it would be more helpful to always return 204, even if there were no sessions to delete.

aeneasr (approved):

Issue #2345, opened 2022-03-28
Disable mail dispatch on attempted account access


nolifeKangaroo:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

When users recover access to their account, an email is sent for two cases:

1. User email is known to the system: We send out an email containing the recovery link
2. User email is not known to the system: We send out an email that says that the user is unknown.

The second case may be abused by a malicious actor to send out mass emails to accounts. This can lead to bad email reputation, for example.

You can try out the flow yourself:

1. Go to https://playground.projects.oryapis.com/ui/recovery and request a recovery email for your email (`foo@example.org`). Wait for the email to arrive at the inbox and read it.
2. Sign up at https://playground.projects.oryapis.com/ui/registation with said email
3. Try the flow again and see what email arrives now

### Describe your ideal solution

A configuration option which allows the administrator to disable that. In the case that the option is enabled, no email would be sent if the user email is unknown. So the result should be:

1. Go to https://playground.projects.oryapis.com/ui/recovery and request an recovery email for an email that has not signed up yet
2. No email should arrive

### Version

v0.9.0-alpha.2

Issue #1835, opened 2021-10-13
Strategies to prevent mass email sending for several flows


aeneasr:
**Is your feature request related to a problem? Please describe.**

> There is a way to abuse verification e-mails sent by kratos-courier. A client can request frequently re-sending verification e-mail and change e-mail address frequently. In these cases, kratos send an e-mail for every client request. This behaviour leads us to become a spam source for the SMTP provider, also overcharge is possible.
>
> I couldn't find a way to prevent that abuse case by configuring Kratos, so I tried to solve this problem on my App side. But to do this, I need the last e-mail sent date for a verifiable address and I thought the updateAt field can be used for this since the status changed after the e-mail is sent. When I look the courier code I noticed the updateAt filed doesn't update by the Pop library if the status value doesn't change(for example re-sending a verification mail). So in this patch, we force update this field.

**Describe the solution you'd like**

> This is indeed an issue, in fact, the issue touches several parts:
>
>1. Creating mass new accounts if the verify hook is enabled
>2. Requesting recovery / verification for random email addresses
>3. Creating an account and continuously updating the verification email so that the verify hook triggers
>
>Adding an indicator of last delivery is IMO not a good approach though:
>
>1. Some users might re-request sending the email a couple of times if it lands in the e.g. spam folder
>2. Emails will be delivered for on-existing accounts also (so you have no previous delivery time)
>
> I'd love to hear ideas though on how to solve it. One possibility would be adding captchas to these submission forms.

aeneasr:
For more context: https://github.com/ory/kratos/pull/1780

the-tunnster:
What if you set up a small go routine to check when the last request call was made, and the frequency of the calls. Set a rate limit, and when it matches it, just stop making calls for a set timeline. Run a defer to close individual go routines, and clean up, allowing a fresh set of requests again after a set amount of time.

Not sure how effective this would be, considering server load, but in my opinion it would account for users who genuinely need it, and block the server from sending out unlimited emails.

mitar:
I think one option to address this is to provide a way for the recipient to report spam. So e-mail should be something like:

```
Hi, please verify your account by clicking the following link: <link>

If you do not recognize or have not expected this e-mail, it is safe to ignore it, or report it <here>
```

mitar:
Another option is also to include an unsubscribe link as the e-mail header. This can then be used by the mail client to automatically unsubscribe the recipient from more of those e-mails.

Gibheer:
[...]

PR #2585, opened 2022-07-08
feat: disable mail dispatch for invalid recovery email

mszekiel:
Added `send` property for `courier.templates.recovery.invalid` configuration field, suppressing email sending when recovery email does not exist.

## Related issue(s)

#2345

## Checklist

## Further Comments

Included additional configuration field for invalid recovery. I've also considered extending directly email field by `send` property but it would affect all other flows using the same struct.  Extended schema definition. Here I wasn't sure if `$ref` template can be extended, so just overwrote it. For `getEmail` cypress command now it's possible to omit check if email is existing.

### DIFF
--- a/driver/config/config.go
+++ b/driver/config/config.go
@@ -66,6 +66,7 @@ const (
 	ViperKeyCourierSMTPClientKeyPath                         = "courier.smtp.client_key_path"
 	ViperKeyCourierTemplatesPath                             = "courier.template_override_path"
 	ViperKeyCourierTemplatesRecoveryInvalidEmail             = "courier.templates.recovery.invalid.email"
+	ViperKeyCourierTemplatesRecoveryInvalidSend              = "courier.templates.recovery.invalid.send"
 	ViperKeyCourierTemplatesRecoveryValidEmail               = "courier.templates.recovery.valid.email"
 	ViperKeyCourierTemplatesRecoveryCodeInvalidEmail         = "courier.templates.recovery_code.invalid.email"
 	ViperKeyCourierTemplatesRecoveryCodeValidEmail           = "courier.templates.recovery_code.valid.email"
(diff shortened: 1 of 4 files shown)


Sytten:
Also related to https://github.com/ory/kratos/issues/1835, this is absolutely needed for anything serious in production.

mrtndwrd:
We use Kratos in a privacy-focused organization and we would also really like to disable the "invalid" emails. What is needed to move this PR forward?

driver/config/config.go
```diff
@@ -66,6 +66,7 @@ const (
 	ViperKeyCourierSMTPClientKeyPath                         = "courier.smtp.client_key_path"
 	ViperKeyCourierTemplatesPath                             = "courier.template_override_path"
 	ViperKeyCourierTemplatesRecoveryInvalidEmail             = "courier.templates.recovery.invalid.email"
+	ViperKeyCourierTemplatesRecoveryInvalidSend              = "courier.templates.recovery.invalid.send"
```
aeneasr, line 69: I think this key should go in another object because here we are configuring the templates. Given that this affects recovery and the `link` and `code` strategy, it should probably be configured there?

For completeness it would also be good to be able to disable this for both recovery and verification.

selfservice/strategy/link/sender.go
```diff
@@ -67,6 +67,12 @@ func (s *Sender) SendRecoveryLink(ctx context.Context, r *http.Request, f *recov
 
 	address, err := s.r.IdentityPool().FindRecoveryAddressByValue(ctx, identity.RecoveryAddressTypeEmail, to)
 	if err != nil {
+		if !s.r.Config().CourierTemplatesRecoveryInvalidSend(ctx) {
```
aeneasr, line 70: Since this PR was opened, we also added a `code` strategy which is now the default and also needs this feature.

aeneasr (commented):
Thank you! To get this shipped we will need a couple of changes:

1. Make this work in the new `code` strategy.
2. Configure this within the `code` and `link` strategy - or alternatively in `recovery` and `verification`.
3. If this enabled, the user is mislead thinking that an email was sent but they had a typo and never receive one. Shouldn't we show an error that the email is not known?

ariep:
Hi, interested party here as we use kratos.

>     3. If this enabled, the user is mislead thinking that an email was sent but they had a typo and never receive one. Shouldn't we show an error that the email is not known?

It should be possible to configure kratos in such a way that a user cannot see the difference between successful and unsuccessful sending. Having a difference, for example in the form of an error message, reveals information about the set of existing users.

We'll add a message to the reset-link form to warn the user that if they do not receive an email, they should check if they made an error when typing their address, or if they're maybe registered using a different email address.

Sytten:
@mszekiel Do you intend in continuing the PR, still baffles me that this behaviour is the default.

mszekiel:
Moved to https://github.com/ory/kratos/pull/3075

PR #2632, opened 2022-08-01
fix: return empty array instead of nil

aeneasr:
Closes https://github.com/ory/cloud/issues/134

## Checklist

### DIFF
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -47,7 +47,7 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
 	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.ListSessionsByIdentity")
 	defer span.End()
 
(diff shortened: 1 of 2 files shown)


persistence/sql/persister_session.go
```diff
@@ -47,7 +47,7 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
 	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.ListSessionsByIdentity")
 	defer span.End()
 
-	s := make([]*session.Session, 0)
+	var s []*session.Session
```
zepatrik, line 50: This should not change the behavior, the function should previously already return an empty slice.
https://go.dev/play/p/KbJYKpXGsLo

zepatrik:
The issue also says that the sessions disappear from the response with no change to whether the sessions actually exist.

aeneasr:
Ok, can you please take a look? Looks like you have more context / understanding than I :)

Issue #1598, opened 2021-07-27
Courier: Failing message stalls queue 


spg:
**Describe the bug**

When an outgoing message fails to send, the message queue is stalled.

For example, if AWS SES replies with error `554 Message rejected: Email address is not verified`, the oldest message in the queue will be retried forever. Newer messages in the queue are not processed.

A user could also stall the queue by submitting an invalid email address (e.g. `me@notld`).

**Reproducing the bug**

Using the Selfservice UI Node project, setup the courier config (I tried using AWS SES), and try recovering a password using an unverified email address `email1@domain.com`. This should cause the courier to fail:

```
time=2021-07-27T15:06:40Z level=error msg=Unable to send email using SMTP connection. func=github.com/ory/kratos/courier.(*Courier).DispatchMessage file=/home/ory/courier/courier.go:187 audience=application error=map[message:gomail: could not send email 1: 554 Message rejected: Email address is not verified. The following identities failed the check in region US-EAST-1: email1@domain.com trace:stack trace could not be recovered from error type *mail.SendError] message_from=noreply@myapp.com service_name=Ory Kratos service_version=v0.7.0-alpha.1 smtp_server=email-smtp.us-east-1.amazonaws.com:587 smtp_ssl_enabled=false
```

Try recovering another email address (`email2@domain.com`). You will see in the logs that the previous email address (`email1@domain.com`) keeps being repeated in the logs, without mention of the new address (`email2@domain.com`).
This indicates that the courier queue keeps retrying sending to `email1@domain.com` without regard for `email2@domain.com`.

**Expected behavior**

The courier queue should not stall when a particular message fails to send. Other messages in the queue should be processed, even if a given message fails.

**Environment**

- Version: v0.7.1-alpha.1
- Environment: Docker

**Additional context**

N/A

spg:
Yes your idea makes sense, this would definitely solve the retrying problem I'm experiencing.

aeneasr:
Nice, would you be up to make a PR for this? :)

spg:
Sure, I don't have experience in Go, but I can try my hand at it!

zepatrik:
Actually duplicate of #402
So please also consider the discussion there

xeruf:
Please tackle this! Our infrastructure was severely crippled for a few days as a nonexistent receiver email was stuck in the queue.

aeneasr:
[...]

PR #2638, opened 2022-08-02
feat: replace message_ttl with static max retry count

jonas-jonas:
This PR replaces the `courier.message_ttl` configuration option with a `courier.message_retries` option to limit how often the sending of a message is retried before it is marked as `abandoned`. 

This is a breaking change, as it removes the `message_ttl` config key.

## Related issue(s)

https://github.com/ory/kratos/issues/402 https://github.com/ory/kratos/issues/1598

## Checklist

### DIFF
--- a/courier/courier_dispatcher.go
+++ b/courier/courier_dispatcher.go
@@ -50,7 +50,7 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
 	maxRetries := c.deps.CourierConfig(ctx).CourierMessageRetries()
 
 	for k, msg := range messages {
-		if msg.SendCount >= maxRetries {
+		if msg.SendCount > maxRetries {
 			if err := c.deps.CourierPersister().SetMessageStatus(ctx, msg.ID, MessageStatusAbandoned); err != nil {
 				if c.failOnError {
 					return err
@@ -58,7 +58,7 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
 				c.deps.Logger().
 					WithError(err).
 					WithField("message_id", msg.ID).
-					Error(`Unable to reset the timed out message's status to "abandoned".`)
+					Error(`Unable to reset the retried message's status to "abandoned".`)
 			}
 		} else {
 			if err := c.deps.CourierPersister().IncrementMessageSendCount(ctx, msg.ID); err != nil {
--- a/courier/courier_dispatcher.go
+++ b/courier/courier_dispatcher.go
@@ -2,7 +2,6 @@ package courier
 
 import (
 	"context"
-	"time"
 
 	"github.com/pkg/errors"
 )
@@ -48,10 +47,10 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
 		return err
 	}
 
-	ttl := c.deps.CourierConfig(ctx).CourierMessageTTL()
+	maxRetries := c.deps.CourierConfig(ctx).CourierMessageRetries()
 
 	for k, msg := range messages {
-		if time.Now().After(msg.CreatedAt.Add(ttl)) {
+		if msg.SendCount >= maxRetries {
 			if err := c.deps.CourierPersister().SetMessageStatus(ctx, msg.ID, MessageStatusAbandoned); err != nil {
 				if c.failOnError {
 					return err
@@ -62,6 +61,16 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
 					Error(`Unable to reset the timed out message's status to "abandoned".`)
 			}
 		} else {
+			if err := c.deps.CourierPersister().IncrementMessageSendCount(ctx, msg.ID); err != nil {
+				if c.failOnError {
+					return err
+				}
+				c.deps.Logger().
+					WithError(err).
+					WithField("message_id", msg.ID).
+					Error(`Unable to increment the message's "send_count" field`)
+			}
+
 			if err := c.DispatchMessage(ctx, msg); err != nil {
 				for _, replace := range messages[k:] {
 					if err := c.deps.CourierPersister().SetMessageStatus(ctx, replace.ID, MessageStatusQueued); err != nil {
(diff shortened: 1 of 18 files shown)


courier/courier_dispatcher_test.go
```diff
@@ -37,11 +32,13 @@ func TestMessageTTL(t *testing.T) {
 		Body:    "test-body-1",
 	}))
 	require.NoError(t, err)
-	require.NotEqual(t, uuid.Nil, id)
+	require.NotZero(t, id)
```
aeneasr, line 35: Did this assertion fail before? I think

```
require.NotEqual(t, uuid.Nil, id)
```

is clearer than `NotZero` because it is not very obvious to me how `NotZero` works under the hood.

courier/courier_dispatcher.go
```diff
@@ -48,10 +47,10 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
 		return err
 	}
 
-	ttl := c.deps.CourierConfig(ctx).CourierMessageTTL()
+	maxRetries := c.deps.CourierConfig(ctx).CourierMessageRetries()
 
 	for k, msg := range messages {
-		if time.Now().After(msg.CreatedAt.Add(ttl)) {
+		if msg.SendCount >= maxRetries {
```
aeneasr, line 53: I think we should check whether we exceed the max retries, not whether we reach them. That So if we say: please retry 5 times, it should retry 5 times and fail at the sixth time.

```suggestion
		if msg.SendCount > maxRetries {
```

aeneasr, line 53: How can we test that emails won't get send that reach the maximum retry count?

aeneasr (changes requested):
Thank you, looking pretty good! Can you please also add documentation to ory/docs for this new feature? And probably remove the `message_ttl` key if it was documented?

courier/courier_dispatcher_test.go
jonas-jonas: No, wasn't failing, you're right. I think we talked about that in the PATCH PR. I'll change it back, though, as in this case it is more obvious to compare it with `uuid.Nil`.

courier/courier_dispatcher.go
jonas-jonas: Right, done. A bit of an awkward spot to wrap the head around, so I added some comments in the test case explaining the behavior.

jonas-jonas: That is the existing test. Since messages that have the `Abandoned` state are no longer queued, they won't be sent out. And they receive that state if the condition `msg.SendCount > maxRetries` is true.

I would like to have a mock SMTP server here, so we can test this case, because currently we are relying on the fact that no SMTP server is running at baz.com which is not ideal. WDYT about that?

jonas-jonas:
> Can you please also add documentation to ory/docs for this new feature? And probably remove the message_ttl key if it was documented?

There was no documentation about this, I'll add some for the new config.

jonas-jonas:
I created a PR on the docs repo: https://github.com/ory/docs/pull/941

jonas-jonas:
I moved the message abandonment logic to the appropriate method for this to work in cloud as well, as discussed, and restructured the tests a bit. 

Are there tests in cloud that ensure that that's the case?

courier/smtp.go
```diff
@@ -199,6 +200,18 @@ func (c *courier) dispatchEmail(ctx context.Context, msg Message) error {
 			// WithField("email_to", msg.Recipient).
 			WithField("message_from", from).
 			Error("Unable to send email using SMTP connection.")
+
+		if protoErr, success := err.(*textproto.Error); success && protoErr.Code >= 500 {
+			// See https://en.wikipedia.org/wiki/List_of_SMTP_server_return_codes
+			// If the SMTP server responds with 5xx, sending the message should not be retried (without changing something about the request)
+			if err := c.deps.CourierPersister().SetMessageStatus(ctx, msg.ID, MessageStatusAbandoned); err != nil {
```
jonas-jonas, line 207: This part is hard to test right now, as we don't have the ability to control SMTP Server responses. Do you have a suggestion here? @aeneasr

courier/smtp.go
```diff
@@ -199,6 +200,18 @@ func (c *courier) dispatchEmail(ctx context.Context, msg Message) error {
 			// WithField("email_to", msg.Recipient).
 			WithField("message_from", from).
 			Error("Unable to send email using SMTP connection.")
+
+		if protoErr, success := err.(*textproto.Error); success && protoErr.Code >= 500 {
```
aeneasr, line 204: It's probably better to use `errors.As` instead as that works with wrapper errors as well :)

courier/smtp.go
aeneasr: Is there not a condition which would lead posthog to send 500? alternatively this could be mocked using go-mock

courier/smtp.go
jonas-jonas: Right, done.

courier/smtp.go
jonas-jonas: Unfortunately, I didn't find a way to do that. [There is `jim` which is the chaos-monkey type module of mailhog](https://github.com/blueimp/mailhog/blob/master/docs/JIM.md), that we do use, but you set the chance of the server declining every message not that we want to explicitly decline the next message, which we would need here. I guess we could send 100 messages to the server and expect at least one of them to return with 500, but that could be flaky statistically speaking.

To make this really clean we need to mock, which as far as I can tell needs a bit of refactoring, [as the SMTP client is initialized in the courier constructor](https://github.com/ory/kratos/blob/master/courier/courier.go#L58).

aeneasr: Hmmmm, and is there no email format the server rejects, likes using utf8 chars or something?

jonas-jonas: Not AFAICT, no: https://github.com/mailhog/MailHog/tree/master/docs

I think a good thing would be to be able to mock the SMTP server easily, then we can write more sophisticated tests for the whole SMTP layer. But again, that needs some refactoring work.

aeneasr: ok, then let's skip this. it should be fine :)

aeneasr (approved):

PR #2645, opened 2022-08-05
feat: replace magic links with one time codes in recovery flow

### LINKED ISSUE #1451: Replace magic links with TOTP OOB codes

jonas-jonas:
This PR introduces a new `code` strategy to recover an account. 
Currently, if a user needs to initiate a recovery flow to recover a lost password/MFA/etc., they’ll receive an email containing a “magic link”. This link contains a flow_id and a recovery_token. This is problematic because some Anti Virus software opens links in emails to check for malicious content, etc.

Instead of the magic link, we send an 8-digit code that is clearly displayed in the email or SMS. A user can now copy/paste or type it manually into the text-field that is shown after the user clicks “submit” on the initiate flow page.

BREAKING CHANGES: This patch changes the behavior of the recovery flow. It introduces a new strategy for account recovery that sends out short "one-time passwords" (`code`) that a user can use to prove ownership of their account and recovery access to it. This PR also updates the default recovery strategy to `code`.

## Related issue(s)

Closes #1451 

## Checklist

## Further Comments

The new flow would look like this: 
image

The flow looks like this if an admin initiates the recovery:
image


jonas-jonas (commented):
This PR is now ready for review! 🎉 

Still Todo: 
~ [ ] Merge https://github.com/ory/kratos-selfservice-ui-react-native/pull/58 (needs another update of the SDK from this branch @aeneasr, sorry - added the enums for the page names)~ not relevant
- [ ] Add documentation

driver/config/config.go
```diff
@@ -691,7 +696,7 @@ func (p *Config) SelfServiceStrategy(ctx context.Context, strategy string) *Self
 			fallthrough
 		case "profile":
 			fallthrough
-		case "link":
+		case "code":
```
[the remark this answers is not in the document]
aeneasr: No, let's move to "code" explicitly and add a breaking changes notice that you will need to explicitly set it to "link" if you want backwards compatibility.

aeneasr:
@jonas-jonas can you please rebase onto master?

driver/config/config.go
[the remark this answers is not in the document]
jonas-jonas: So this is case change is fine with you?

jonas-jonas:
@aeneasr rebase done - waiting for tests to confirm.

aeneasr:
@jonas-jonas can you please work on documentation for this new feature?

We currently have this document ( https://www.ory.sh/docs/kratos/self-service/flows/account-recovery ) but I don't think it's that good (too technical IMO). Maybe we just create a new document under ["identity management"](https://www.ory.sh/docs/kratos/ory-kratos-intro) called "Account recovery and password reset" which explains how the feature works (what it does), how to configure it, and how to use the SDK to implement it?

driver/config/config.go
```diff
@@ -67,6 +67,8 @@ const (
 	ViperKeyCourierTemplatesPath                             = "courier.template_override_path"
 	ViperKeyCourierTemplatesRecoveryInvalidEmail             = "courier.templates.recovery.invalid.email"
 	ViperKeyCourierTemplatesRecoveryValidEmail               = "courier.templates.recovery.valid.email"
+	ViperKeyCourierTemplatesRecoveryCodeInvalidEmail         = "courier.templates.recovery_code.invalid.email"
+	ViperKeyCourierTemplatesRecoveryCodeValidEmail           = "courier.templates.recovery_code.valid.email"
 	ViperKeyCourierTemplatesVerificationInvalidEmail         = "courier.templates.verification.invalid.email"
 	ViperKeyCourierTemplatesVerificationValidEmail           = "courier.templates.verification.valid.email"
```
aeneasr, line 73: Hm, probably good to not break BC here. At the same time it would be better to have:

```patch
courier:
  templates:
    recovery:
-      valid:
-      invalid:
+      link:
+        valid:
+        invalid:
+      code:
+        valid:
+        invalid:
```

driver/config/config.go
```diff
@@ -711,7 +716,7 @@ func (p *Config) SelfServiceStrategy(ctx context.Context, strategy string) *Self
 			fallthrough
 		case "profile":
 			fallthrough
-		case "link":
+		case "code":
```
aeneasr, line 719: Yes

driver/config/config.go
```diff
@@ -162,6 +164,7 @@ const (
 	ViperKeyDatabaseCleanupBatchSize                         = "database.cleanup.batch_size"
 	ViperKeyLinkLifespan                                     = "selfservice.methods.link.config.lifespan"
 	ViperKeyLinkBaseURL                                      = "selfservice.methods.link.config.base_url"
+	ViperKeyCodeLifespan                                     = "selfservice.methods.code.config.lifespan"
```
aeneasr, line 167: In our discussions I recall that we agreed on the following config layout:


```
selfservice:
  flows:
    verification:
      use: code
    recovery:
      use: code
```

instead of validating:


```
selfservice:
  methods:
    code:
      enabled: true|false
    link:
      enabled: true|false
```

because it makes sense to have only one type enabled at a single time.

driver/config/config_test.go
```diff
@@ -553,7 +553,7 @@ func TestViperProvider_Defaults(t *testing.T) {
 			assert.False(t, p.SelfServiceFlowVerificationEnabled(ctx))
 			assert.True(t, p.SelfServiceStrategy(ctx, "password").Enabled)
 			assert.True(t, p.SelfServiceStrategy(ctx, "profile").Enabled)
-			assert.True(t, p.SelfServiceStrategy(ctx, "link").Enabled)
+			assert.True(t, p.SelfServiceStrategy(ctx, "code").Enabled)
```
aeneasr, line 556: When changing tests, it's ok to add assertions, but very rarely ok to remove or modify existing assertions. Please add the link assertion back in here!

internal/testhelpers/selfservice_verification.go
```diff
@@ -77,7 +84,6 @@ func InitializeRecoveryFlowViaBrowser(t *testing.T, client *http.Client, isSPA b
 	require.NoError(t, res.Body.Close())
 	rs, _, err := publicClient.V0alpha2Api.GetSelfServiceRecoveryFlow(context.Background()).Id(res.Request.URL.Query().Get("flow")).Execute()
 	require.NoError(t, err)
```
aeneasr, line 86: Why remove this assertion?

internal/testhelpers/selfservice_verification.go
```diff
@@ -87,7 +93,6 @@ func InitializeRecoveryFlowViaAPI(t *testing.T, client *http.Client, ts *httptes
 
 	rs, _, err := publicClient.V0alpha2Api.InitializeSelfServiceRecoveryFlowWithoutBrowser(context.Background()).Execute()
 	require.NoError(t, err)
-	assert.Empty(t, rs.Active)
```
aeneasr, line 96: Why remove this assertion?

internal/testhelpers/selfservice_verification.go
```diff
@@ -43,9 +45,14 @@ func GetRecoveryFlow(t *testing.T, client *http.Client, ts *httptest.Server) *kr
...
-	require.NoError(t, err, "%s", res.Request.URL.String())
-	assert.Empty(t, rs.Active)
+	flowID := res.Request.URL.Query().Get("flow")
+	assert.NotEmpty(t, flowID, "expected to receive a flow id, got none")
+
+	rs, _, err := publicClient.V0alpha2Api.
+		GetSelfServiceRecoveryFlow(context.Background()).
+		Id(flowID).
+		Execute()
+	require.NoError(t, err, "expected no error when fetching recovery flow: %s", err)
```
aeneasr, line 55: Please respect existing assertions:

```
	assert.Empty(t, rs.Active)
```

persistence/sql/migratest/fixtures/recovery_code/8f75f5d9-9cb4-4848-9a73-9344f686f8a6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "id": "8f75f5d9-9cb4-4848-9a73-9344f686f8a6",
+  "recovery_address": null,
```
aeneasr, line 3: The `code` is missing, I assume that is on purpose?

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+
+func (p *Persister) UseRecoveryCode(ctx context.Context, fID uuid.UUID, codeVal string) (*code.RecoveryCode, error) {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.UseRecoveryCode")
+	defer span.End()
+
+	var recoveryCode code.RecoveryCode
+
+	nid := p.NetworkID(ctx)
+	if err := sqlcon.HandleError(p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) (err error) {
+		for _, secret := range p.r.Config().SecretsSession(ctx) {
```
aeneasr, line 158: I suggest to use the generic system here instead. The session secret is used to sign cookies :)

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+			if err = tx.Where("code = ? AND nid = ? AND NOT used AND selfservice_recovery_flow_id = ?", p.hmacValueWithSecret(ctx, codeVal, secret), nid, fID).First(&recoveryCode); err != nil {
+				if !errors.Is(sqlcon.HandleError(err), sqlcon.ErrNoRows) {
```
aeneasr, line 160: This makes sure the returned error is also properly wrapped:

```
if err := sqlcon.HandleError(err); !errors.Is(err, sqlcon.ErrNoRows) {
```

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+			} else {
+				break
+			}
+		}
+		if err != nil {
+			return err
+		}
+
+		var ra identity.RecoveryAddress
+		if err := tx.Where("id = ? AND nid = ?", recoveryCode.RecoveryAddressID, nid).First(&ra); err != nil {
```
aeneasr, line 172: Could this instead be done with a JOIN above? You could also use a `RETURNING` clause to combine all three queries into one.

selfservice/flow/recovery/flow.go
```diff
@@ -123,13 +127,22 @@ func NewFlow(conf *config.Config, exp time.Duration, csrf string, r *http.Reques
 		Type:      ft,
 	}
 
-	for _, strategy := range strategies {
-		if err := strategy.PopulateRecoveryMethod(r, req); err != nil {
+	// Currently, there are two recovery methods: "code" and (legacy) "link"
```
aeneasr, line 130: See my comment above in the configuration file. This would avoid such a workaround :)

selfservice/flow/recovery/flow.go
```diff
@@ -167,6 +180,10 @@ func (f Flow) GetNID() uuid.UUID {
 	return f.NID
 }
 
+func (f Flow) ShouldEnforceCSRF() bool {
+	return f.Type.IsBrowser() && f.CSRFToken != ""
```
aeneasr, line 184: The original assertion

```
	if f.Type == flow.TypeBrowser && !nosurf.VerifyToken(h.d.GenerateCSRFToken(r), f.CSRFToken) {
```

was weakened quite a bit, because now an empty `f.CSRFToken` disables CSRF validation. But is that really what this state expresses? Disabling security features should always be explicit - default values (empty string) should not allow insecure behaviour! I suggest to revert this and instead make sure that callers decide when it is ok to not have an CSRF Token

selfservice/flow/recovery/flow_test.go
```diff
@@ -61,6 +63,25 @@ func TestFlow(t *testing.T) {
 		_, err = recovery.NewFlow(conf, 0, "csrf", &http.Request{URL: &url.URL{Path: "/", RawQuery: "return_to=" + urlx.AppendPaths(conf.SelfPublicURL(ctx), "/self-service/login/browser").String()}, Host: "ory.sh"}, nil, flow.TypeBrowser)
 		require.NoError(t, err)
 	})
+
+	t.Run("type=enabled code overrides link", func(t *testing.T) {
```
aeneasr, line 67: Probably obsolete test if we implement the config changes

selfservice/flow/recovery/handler.go
```diff
@@ -269,7 +269,7 @@ func (h *Handler) fetch(w http.ResponseWriter, r *http.Request, _ httprouter.Par
 	// Browser flows must include the CSRF token
 	//
 	// Resolves: https://github.com/ory/kratos/issues/1282
-	if f.Type == flow.TypeBrowser && !nosurf.VerifyToken(h.d.GenerateCSRFToken(r), f.CSRFToken) {
+	if f.ShouldEnforceCSRF() && !nosurf.VerifyToken(h.d.GenerateCSRFToken(r), f.CSRFToken) {
```
aeneasr, line 272: This assertion has been weakened by `f.CSRFToken != ""`, please revert

text/message_recovery.go
```diff
@@ -46,6 +55,24 @@ func NewErrorValidationRecoveryTokenInvalidOrAlreadyUsed() *Message {
...
+		Text:    "The recovery code is invalid or has already been used. Please try again.",
+		Type:    Error,
+		Context: context(nil),
+	}
+}
+
+func NewErrorValidationRecoveryFlowSubmittedTooOften() *Message {
+	return &Message{
+		ID:      ErrorValidationRecoveryCodeUsedTooOften,
+		Text:    "The recovery was submitted too often. Please restart the flow.",
```
aeneasr, line 70: I think we are showing a new flow with a zero submit count and a new code, and can remove the text around "flow" which might be confusing to everyday internet users:

```
		Text:    "The recovery was submitted too often. Please try again.",
```

test/e2e/cypress/integration/profiles/mobile/recovery/errors.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    })
+
+    describe('code', () => {
+      beforeEach(() => {
+        cy.deleteMail()
+        cy.longCodeLifespan()
+      })
+
+      // Re-enable once mobile supports recovery
+      it.skip('fails with validation errors', () => {
```
aeneasr, line 30: Please add an issue to track that we need to re-enable these tests!

test/e2e/cypress/integration/profiles/recovery/code/errors.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      beforeEach(() => {
+        cy.deleteMail()
+        cy.longRecoveryLifespan()
+        cy.longCodeLifespan()
+        cy.disableVerification()
+        cy.enableRecovery('code')
+      })
+
+      it('shows code expired message if expired code is submitted', () => {
```
aeneasr, line 33: Please cover the "tried too often" use case!

text/message_recovery.go
jonas-jonas: Sure, sounds good.

driver/config/config.go
jonas-jonas: Yes, that's true and this layout looks a lot easier to use. 

Initially, I didn't want to break the current mechanisms we have in place for the relation of methods/strategies -> flow (and not the other way around). But your proposal looks way cleaner. Will do.

driver/config/config_test.go
jonas-jonas: You're right, I gotta internalize that sentiment. 

In this case, we are expecting a change in behavior, so changing the assertion to expect `False` should be correct, right?

internal/testhelpers/selfservice_verification.go
jonas-jonas: We are now setting the active strategy when the flow is initialized. So the assertion needs to be inverted. 

Btw.: not sure if that's the best way, but we need a way to distinguish between link & code. Suggestions welcome.

persistence/sql/persister_recovery.go
jonas-jonas: Not sure what you mean by that. This is copied from the `UseRecoveryToken` method. Could you point me to a better solution?

persistence/sql/persister_recovery.go
jonas-jonas: I guess it might be possible, but not sure if it would actually be an improvement especially with the current system of iterating through the secrets?

selfservice/flow/recovery/flow.go
jonas-jonas: The reasoning here is, that we control the creation of the `flow` and thus if a CSRF token is required. And since the new flow now always requires CSRF protection at all steps, _except_ when the flow was created by an admin, it would be fair to assume, that an empty `CSRFToken` means that a flow was created by an admin. So: yes, this is what that state expresses.

I'd say a more explicit solution could be, to add a `CreatedBy` = `ADMIN` | `SELFSERVICE` field to `Flow` and then enforce CSRF on the `SELFSERVICE` flows. Thoughts?

persistence/sql/migratest/fixtures/recovery_code/8f75f5d9-9cb4-4848-9a73-9344f686f8a6.json
jonas-jonas: Yes, same for the `recovery_token` files. I guess that's because it uses `json.MarshalIndent` and `Code` is ignored in that. See https://github.com/jonas-jonas/kratos/blob/feat/replaceMagicLinksWithCode/selfservice/strategy/code/code_recovery.go#L35

selfservice/flow/recovery/flow.go
jonas-jonas: Done.

driver/config/config.go
jonas-jonas: Maybe we can use

```patch
courier:
  templates:
    recovery:
      valid:
      invalid:
+     code:
+       valid:
+       invalid:
```

It's a bit awkward as long as the `link` strategy is still there. But once that's gone should be good, right?

test/e2e/cypress/integration/profiles/mobile/recovery/errors.spec.ts
jonas-jonas: Done https://github.com/ory/kratos/issues/2731

test/e2e/cypress/integration/profiles/recovery/code/errors.spec.ts
jonas-jonas: Done.

driver/config/config.go
jonas-jonas: Done.

Although, I did not rewrite the logic to enable/disable strategies, as that is sufficiently generic. Now you can use `selfservice.flows.recovery.use: code` to use the new strategy. It is enabled by default, which means you'd only need to explicitly enable the link strategy if you want to use it.

selfservice/flow/recovery/flow.go
jonas-jonas: Also, another thought: if there is no `CSRFToken` in the database, how would we validate it anyway? Would it be fine to assume that the flow is FUBAR at that point?

driver/config/config.go
aeneasr: That's ok. It will prevent us from having the same problem again next time we introduce a new recovery template.

driver/config/config.go
aeneasr: great!

persistence/sql/migratest/fixtures/recovery_code/8f75f5d9-9cb4-4848-9a73-9344f686f8a6.json
aeneasr: 👍

driver/config/config_test.go
aeneasr: Yup!

selfservice/flow/recovery/handler.go
aeneasr: Have you updated the assertion @jonas-jonas?

driver/config/config.go
jonas-jonas: So you're fine with my proposal?

selfservice/flow/recovery/handler.go
jonas-jonas: See https://github.com/ory/kratos/pull/2645#discussion_r969242435

We need to figure out a way to handle recovery initiated by admins here, so no. :)

driver/config/config_test.go
jonas-jonas: Updated the assertion [here](https://github.com/ory/kratos/pull/2645/files#diff-deb13d25cad45e9e2975fd0ad5710541fd5b7ae76411f9ce50c4af6a053f7593R558)

.github/workflows/ci.yaml
```diff
@@ -187,6 +187,7 @@ jobs:
         env:
           RN_UI_PATH: react-native-ui
           NODE_UI_PATH: node-ui
+          CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
```
jonas-jonas, line 190: We need this for the recording for branches originating from the original repo `ory/kratos`, so I am leaving it in here.

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -0,0 +1,607 @@
...
+		}
+	}
+
+	if time.Now().Add(expiresIn).Before(time.Now()) {
+		s.deps.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf(`Value from "expires_in" must result to a future time: %s`, p.ExpiresIn)))
+		return
+	}
+
+	// Initially, the CSRF token is empty, as the generated flow + link is not yet tied to the users browser
+	flow, err := recovery.NewFlow(config, expiresIn, "", r, s, flow.TypeBrowser)
```
aeneasr, line 165: Please replace this with an explicit value that says: I am an admin flow and do not need a CSRF token check with a field called something like `DangerousSkipCSRFCheck`.

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -0,0 +1,607 @@
...
+func (s *Strategy) retryRecoveryFlowWithMessage(w http.ResponseWriter, r *http.Request, ft flow.Type, message *text.Message) error {
+	s.deps.Logger().
+		WithRequest(r).
+		WithField("message", message).
+		Debug("A recovery flow is being retried because a validation error occurred.")
+
+	ctx := r.Context()
+	config := s.deps.Config()
+
+	f, err := recovery.NewFlow(config, config.SelfServiceFlowRecoveryRequestLifespan(ctx), "", r, s, ft)
```
aeneasr, line 439: CSRF Token :)

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -0,0 +1,607 @@
...
+		Debug("A recovery flow is being retried because a validation error occurred.")
+
+	ctx := r.Context()
+	config := s.deps.Config()
+
+	if expired := new(flow.ExpiredError); errors.As(recErr, &expired) {
+		return s.retryRecoveryFlowWithMessage(w, r, ft, text.NewErrorValidationRecoveryFlowExpired(expired.Ago))
+	}
+
+	f, err := recovery.NewFlow(config, config.SelfServiceFlowRecoveryRequestLifespan(ctx), "", r, s, ft)
```
aeneasr, line 472: CSRF Token

selfservice/strategy/code/strategy_recovery_test.go
```diff
@@ -0,0 +1,1003 @@
...
+
+		res, err := publicTS.Client().Get(code.RecoveryLink)
+		body := ioutilx.MustReadAll(res.Body)
+
+		action := gjson.GetBytes(body, "ui.action").String()
+		require.NotEmpty(t, action)
+		csrfToken := gjson.GetBytes(body, "ui.nodes.#(attributes.name==csrf_token).attributes.value").String()
+		require.NotEmpty(t, csrfToken)
+
+		res, err = publicTS.Client().PostForm(action, url.Values{
```
aeneasr, line 180: Make this a helper and DRY it up in the tests

aeneasr: Maybe explore testhelpers.Submit* helpers

selfservice/strategy/code/strategy_recovery_test.go
```diff
@@ -0,0 +1,1003 @@
...
+			Id(res.Request.URL.Query().Get("flow")).Execute()
+
+		require.NoError(t, err)
+
+		require.Len(t, rs.Ui.Messages, 1)
+		assert.Equal(t, "The recovery was submitted too often. Please try again.", rs.Ui.Messages[0].Text)
+		json, err := rs.Ui.MarshalJSON()
+		require.NoError(t, err)
+		t.Logf("body: %s", string(json))
+		assert.True(t, gjson.GetBytes(json, "nodes.#(attributes.name==email)").Exists())
```
aeneasr, line 783: Add an assertion that we get a fresh flow :)

selfservice/strategy/code/strategy_recovery_test.go
```diff
@@ -0,0 +1,1003 @@
...
+					assert.Len(t, gjson.GetBytes(json, "ui.messages").Array(), 1)
+					assert.Contains(t, gjson.GetBytes(json, "ui.messages.0.text").String(), "You successfully recovered your account.")
+				} else if testCase.FlowType == RecoveryFlowTypeSPA {
+					assert.Equal(t, http.StatusUnprocessableEntity, res.StatusCode)
+
+					json := ioutilx.MustReadAll(res.Body)
+
+					assert.Equal(t, gjson.GetBytes(json, "error.id").String(), "browser_location_change_required")
+					assert.Contains(t, gjson.GetBytes(json, "redirect_browser_to").String(), "settings-ts?")
+				} else if testCase.FlowType == RecoveryFlowTypeSPA {
```
aeneasr, line 848: Duplicate, remove?

test/e2e/cypress/integration/profiles/recovery/code/errors.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+          cy.noSession()
+        }
+
+        cy.get("input[name='code']").type('123123') // Invalid code
+        cy.get("button[value='code']").click()
+        cy.get('[data-testid="ui/message/4060007"]').should(
+          'have.text',
+          'The recovery was submitted too often. Please try again.'
+        )
+        cy.noSession()
```
aeneasr, line 60: Please add one more assertion which enters the email again, to check if the form is properly reset and retriable :)

test/e2e/cypress/integration/profiles/mobile/recovery/errors.spec.ts
```diff
@@ -0,0 +1,114 @@
+import {
```
aeneasr, line 1: Remove these two files

aeneasr (commented):
Please also add a "retry" button which, for example, creates a new flow where the user can input their email again. Maybe you can pre-fill the email with the email used in the current flow and add a short message.

selfservice/flow/recovery/flow.go
jonas-jonas: Done https://github.com/ory/kratos/pull/2645/commits/d865fa68a978ae7dc6a8127f7e01f6145b22c390

session/session.go
```diff
@@ -121,6 +121,7 @@ func (s *Session) SetAuthenticatorAssuranceLevel() {
 			// be part of the AMR.
 			switch amr.Method {
 			case identity.CredentialsTypeRecoveryLink:
+			case identity.CredentialsTypeRecoveryCode:
```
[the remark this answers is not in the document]
jonas-jonas: Done.

selfservice/strategy/code/strategy_recovery.go
jonas-jonas: Done https://github.com/ory/kratos/pull/2645/commits/d865fa68a978ae7dc6a8127f7e01f6145b22c390

selfservice/strategy/code/strategy_recovery.go
jonas-jonas: Done https://github.com/ory/kratos/pull/2645/commits/d865fa68a978ae7dc6a8127f7e01f6145b22c390

selfservice/strategy/code/strategy_recovery.go
jonas-jonas: Done https://github.com/ory/kratos/pull/2645/commits/d865fa68a978ae7dc6a8127f7e01f6145b22c390

selfservice/strategy/code/strategy_recovery_test.go
jonas-jonas: Done.

test/e2e/cypress/integration/profiles/recovery/code/errors.spec.ts
jonas-jonas: Done.

test/e2e/cypress/integration/profiles/mobile/recovery/errors.spec.ts
jonas-jonas: Done.

selfservice/strategy/code/strategy_recovery_test.go
jonas-jonas: Done.

selfservice/strategy/code/strategy_recovery_test.go
jonas-jonas: I DRYed up the tests as best, as I could without refactoring everything :D please have another look.

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -0,0 +1,605 @@
...
+//	Consumes:
+//	- application/json
+//
+//	Produces:
+//	- application/json
+//
+//	Schemes: http, https
+//
+//	Responses:
+//		200: selfServiceRecoveryCode
```
zepatrik, line 132: As this creates the code, why not return 201?

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -0,0 +1,605 @@
...
+			WithMetaLabel(text.NewInfoNodeLabelSubmit()))
+
+	if err := s.deps.RecoveryFlowPersister().CreateRecoveryFlow(ctx, flow); err != nil {
+		s.deps.Writer().WriteError(w, r, err)
+		return
+	}
+
+	id, err := s.deps.IdentityPool().GetIdentity(ctx, p.IdentityID)
+	if errors.Is(err, sqlcon.ErrNoRows) {
+		s.deps.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf("The requested identity id does not exist.").WithWrap(err)))
```
zepatrik, line 186: I guess returning 404 would be more meaningful, as the request is not malformed?
```suggestion
		s.deps.Writer().WriteError(w, r, err)
```
You can then add more details further down in the persister.

zepatrik (changes requested):
Nice this looks pretty solid :+1: 
I just found some details I want to clarify. Will take another look tomorrow, as I don't have more time to finish it today.

courier/template/courier/builtin/templates/recovery_code/valid/email.body.gotmpl
```diff
@@ -0,0 +1,5 @@
+Hi,
+
+please recover access to your account by entering the following code:
+
+{{ .RecoveryCode }}
```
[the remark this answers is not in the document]
jonas-jonas: We could, but the purpose of the new strategy is that the user never leaves the window/tab in which they initiated the flow. 

Also, if a user opens the flow in a different browser, they will get a CSRF violation error and that is terrible UX. I'd leave it out to avoid these confusions.

internal/testhelpers/courier.go
```diff
@@ -33,3 +33,13 @@ func CourierExpectLinkInMessage(t *testing.T, message *courier.Message, offset i
 
 	return match[offset]
 }
+
+func CourierExpectCodeInMessage(t *testing.T, message *courier.Message, offset int) string {
+	if offset == 0 {
+		offset++
```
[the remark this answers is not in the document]
jonas-jonas: Same thing, right? :D 
Done. :)

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
 	}
 	return nil
 }
+
+func (p *Persister) CreateRecoveryCode(ctx context.Context, recoveryCode *code.RecoveryCode) error {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.CreateRecoveryCode")
+	defer span.End()
+
+	code := recoveryCode.Code
+	recoveryCode.Code = p.hmacValue(ctx, code)
```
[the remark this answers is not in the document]
jonas-jonas: _This was copied from the link strategy_.

I'd even go as far as to say we should have a separate DTO struct that is used to create the code and an actual entity struct that represents the saved code (the current `code.RecoveryCode` struct), as we have no way of turning an HMACed value back into the code. WDYT?

So this method would then have this signature:

```golang
func (p *Persister) CreateRecoveryCode(ctx context.Context, recoveryCodeDTO code.RecoveryCodeDTO) (*code.RecoveryCode, error)
```

and the DTO struct:

```golang
type RecoveryCodeDTO struct {
	// Code represents the recovery code
	Code string `json:"-" db:"code"`

	// RecoveryAddress links this code to a recovery address.
	// required: true
	RecoveryAddress *identity.RecoveryAddress `json:"recovery_address" belongs_to:"identity_recovery_addresses" fk_id:"RecoveryAddressID"`

	// CodeType is the type of the code - either "admin" or "selfservice"
	CodeType RecoveryCodeType `json:"-" faker:"-" db:"code_type"`

	// ExpiresAt is the time (UTC) when the code expires.
	// required: true
	ExpiresAt time.Time `json:"expires_at" faker:"time_type" db:"expires_at"`

	// RecoveryAddressID is a helper struct field for gobuffalo.pop.
	RecoveryAddressID uuid.NullUUID `json:"-" faker:"-" db:"identity_recovery_address_id"`
	// FlowID is a helper struct field for gobuffalo.pop.
	FlowID     uuid.UUID `json:"-" faker:"-" db:"selfservice_recovery_flow_id"`

	IdentityID uuid.UUID `json:"identity_id" faker:"-" db:"identity_id"`
}
```

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+	recoveryCode.NID = p.NetworkID(ctx)
+
+	// This should not create the request eagerly because otherwise we might accidentally create an address that isn't
+	// supposed to be in the database.
+	if err := p.GetConnection(ctx).Create(recoveryCode); err != nil {
+		return err
+	}
+
+	recoveryCode.Code = code
```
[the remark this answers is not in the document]
jonas-jonas: Would be gone with my DTO suggestion.

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+func (p *Persister) UseRecoveryCode(ctx context.Context, fID uuid.UUID, codeVal string) (*code.RecoveryCode, error) {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.UseRecoveryCode")
+	defer span.End()
+
+	var recoveryCode code.RecoveryCode
+
+	nid := p.NetworkID(ctx)
+	if err := sqlcon.HandleError(p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) (err error) {
+		for _, secret := range p.r.Config().SecretsSession(ctx) {
+			if err = tx.Where("code = ? AND nid = ? AND NOT used AND selfservice_recovery_flow_id = ?", p.hmacValueWithSecret(ctx, codeVal, secret), nid, fID).First(&recoveryCode); err != nil {
```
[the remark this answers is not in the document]
jonas-jonas: > Is this using a different way than during creation to compute the hmac?

Yes. `p.hmacValue(ctx, code)` uses the **current** secret and here we need to use all known secrets.

> I propose to instead do a single query

Good idea! Small correction: `nid = ? AND selfservice_recovery_flow_id = ?` can return multiple rows, because you can create multiple codes from one recovery flow. So we'd need to iterate over the rows and do the checks on each one of them. Most of the time, it will just be one code, though. With this change, we can be more specific with the errors as well.

persistence/sql/migrations/sql/20220901123209000000_recovery_code.up.sql
```diff
@@ -0,0 +1,33 @@
...
+    used_at timestamp,
+    identity_recovery_address_id UUID,
+    code_type INT NOT NULL,
+    expires_at timestamp NOT NULL DEFAULT '2000-01-01 00:00:00',
+    issued_at timestamp NOT NULL DEFAULT '2000-01-01 00:00:00',
+    selfservice_recovery_flow_id UUID NOT NULL,
+    created_at timestamp NOT NULL,
+    updated_at timestamp NOT NULL,
+    nid UUID NOT NULL,
+    identity_id UUID NOT NULL,
```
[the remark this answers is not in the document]
jonas-jonas: Used here https://github.com/ory/kratos/pull/2645/files#diff-d52bb39b1612240c9cccd9e25074a30c77c6d81fd806c52943e92a174b56b197R414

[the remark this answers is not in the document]
jonas-jonas: The reason being, that there are recovery flows initiated by the admin that do not require a recovery address.

persistence/sql/migrations/sql/20220901123209000000_recovery_code.up.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE identity_recovery_codes
+(
+    id UUID NOT NULL PRIMARY KEY,
+    code VARCHAR (64) NOT NULL, -- HMACed value of the actual code
+    used bool NOT NULL DEFAULT 'false',
+    used_at timestamp,
+    identity_recovery_address_id UUID,
```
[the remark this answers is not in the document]
jonas-jonas: > The reason being, that there are recovery flows initiated by the admin that do not require a recovery address.

Same reason here.

And this is needed because a recovery address that is not verified will be marked as `verified` if it was used to recover access to the account.

selfservice/flow/recovery/error.go
```diff
@@ -69,8 +69,14 @@ func (s *ErrorHandler) WriteFlowError(
...
 	if e := new(flow.ExpiredError); errors.As(err, &e) {
+		strategy, err := s.d.GetActiveRecoveryStrategy(r.Context())
+		// Can't retry the recovery if no strategy has been set
+		if err != nil {
+			s.d.SelfServiceErrorManager().Forward(r.Context(), w, r, err)
+			return
+		}
 		// create new flow because the old one is not valid
-		a, err := FromOldFlow(s.d.Config(), s.d.Config().SelfServiceFlowRecoveryRequestLifespan(r.Context()), s.d.GenerateCSRFToken(r), r, s.d.RecoveryStrategies(r.Context()), *f)
+		a, err := FromOldFlow(s.d.Config(), s.d.Config().SelfServiceFlowRecoveryRequestLifespan(r.Context()), s.d.GenerateCSRFToken(r), r, strategy, *f)
```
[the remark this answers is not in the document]
jonas-jonas: Only one strategy is allowed, but we do support the other strategy for older flows to still remain functional.

I added some code to extract the strategy of the old flow and fall back to the active strategy.

selfservice/strategy/code/.schema/recovery.schema.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$id": "https://schemas.ory.sh/kratos/selfservice/strategy/profile/settings.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "method": {
+      "type": "string"
```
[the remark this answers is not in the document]
jonas-jonas: Yup, done.

selfservice/strategy/code/.schema/recovery.schema.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "code": {
+      "type": "string"
```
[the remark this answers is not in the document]
jonas-jonas: Yup, done.

selfservice/strategy/code/sender.go
```diff
@@ -0,0 +1,118 @@
...
+		WithField("via", via).
+		WithSensitiveField("address", to).
+		Debug("Preparing recovery code.")
+
+	address, err := s.deps.IdentityPool().FindRecoveryAddressByValue(ctx, identity.RecoveryAddressTypeEmail, to)
+	if err != nil {
+		if err := s.send(ctx, string(via), email.NewRecoveryInvalid(s.deps, &email.RecoveryInvalidModel{To: to})); err != nil {
+			return err
+		}
+		return errors.Cause(ErrUnknownAddress)
```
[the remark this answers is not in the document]
jonas-jonas: Not sure, was copied from `link`.

I changed it, as it doesn't do anything here.

selfservice/strategy/code/sender.go
```diff
@@ -0,0 +1,118 @@
...
+	RecoveryCodeSenderProvider interface {
+		RecoveryCodeSender() *RecoveryCodeSender
+	}
+
+	RecoveryCodeSender struct {
+		deps senderDependencies
+	}
+)
+
+var ErrUnknownAddress = errors.New("recovery requested for unknown address")
```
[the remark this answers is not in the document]
jonas-jonas: Done.

selfservice/strategy/code/sender_test.go
```diff
@@ -0,0 +1,64 @@
...
+	hr := httptest.NewRequest("GET", "https://www.ory.sh", nil)
+
+	t.Run("method=SendRecoveryCode", func(t *testing.T) {
+		f, err := recovery.NewFlow(conf, time.Hour, "", u, code.NewStrategy(reg), flow.TypeBrowser)
+		require.NoError(t, err)
+
+		require.NoError(t, reg.RecoveryFlowPersister().CreateRecoveryFlow(context.Background(), f))
+
+		require.NoError(t, reg.RecoveryCodeSender().SendRecoveryCode(context.Background(), hr, f, "email", "tracked@ory.sh"))
+		require.EqualError(t, reg.RecoveryCodeSender().SendRecoveryCode(context.Background(), hr, f, "email", "not-tracked@ory.sh"), code.ErrUnknownAddress.Error())
```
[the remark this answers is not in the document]
jonas-jonas: Done.

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -0,0 +1,605 @@
...
+	// required: true
+	IdentityID uuid.UUID `json:"identity_id"`
+
+	// Code Expires In
+	//
+	// The recovery code will expire at that point in time. Defaults to the configuration value of
+	// `selfservice.flows.recovery.request_lifespan`.
+	//
+	//
+	// pattern: ^[0-9]+(ns|us|ms|s|m|h)$
```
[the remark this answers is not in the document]
jonas-jonas: This should do the trick: `^([0-9]+(ns|us|ms|s|m|h))*$`
Done.

selfservice/strategy/code/strategy_recovery.go
jonas-jonas: True, done.

selfservice/strategy/code/strategy_recovery.go
jonas-jonas: > You can then add more details further down in the persister.

What do you mean by that? 
I'd just change `ErrBadRequest` to `ErrNotFound`.

persistence/sql/persister_recovery.go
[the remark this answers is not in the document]
zepatrik: SGTM

persistence/sql/persister_recovery.go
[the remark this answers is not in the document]
zepatrik: OK nice, then let's go with that single query!

selfservice/strategy/code/strategy_recovery.go
zepatrik: `sqlcon.ErrNoRows` already is `herodot.ErrNotFound`. You can add more details by specifying that the identity was not found, instead of just the "generic" 404. This handler is about the code, but it needs the identity as a prerequisite. Adding something like `.WithReadson("requested identity does not exist")` makes it clear that the error is not related to the code itself.

zepatrik: I guess here, as `GetIdentity` is used in many places. But whatever makes more sense. You can probably use `errors.As` to unwrap the herodot error.

persistence/sql/persister_recovery.go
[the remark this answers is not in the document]
jonas-jonas: Done.

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+				if err := sqlcon.HandleError(err); !errors.Is(err, sqlcon.ErrNoRows) {
+					return err
+				}
+			} else {
+				break
+			}
+		}
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
jonas-jonas: As this method now changed a lot now, could you have another look if the error handling is fine now?

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+		}
+
+		var ra identity.RecoveryAddress
+		if err := tx.Where("id = ? AND nid = ?", recoveryCode.RecoveryAddressID, nid).First(&ra); err != nil {
+			if !errors.Is(sqlcon.HandleError(err), sqlcon.ErrNoRows) {
+				return err
```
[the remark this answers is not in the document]
jonas-jonas: Done, I think. :)

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+
+	return &recoveryCode, nil
+}
+
+func (p *Persister) DeleteRecoveryCode(ctx context.Context, codeStr string) error {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.DeleteRecoveryCode")
+	defer span.End()
+
+	/* #nosec G201 TableName is static */
+	return p.GetConnection(ctx).RawQuery(fmt.Sprintf("DELETE FROM %s WHERE token=? AND nid = ?", new(code.RecoveryCode).TableName(ctx)), codeStr, p.NetworkID(ctx)).Exec()
```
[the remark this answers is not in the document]
jonas-jonas: We aren't using the method, so I just deleted it for now. 

(It also wouldn't work, as we are not HMACing the code 😅)

selfservice/flow/recovery/flow.go
```diff
@@ -91,9 +94,15 @@ type Flow struct {
 	// RecoveredIdentityID is a helper struct field for gobuffalo.pop.
 	RecoveredIdentityID uuid.NullUUID `json:"-" faker:"-" db:"recovered_identity_id"`
 	NID                 uuid.UUID     `json:"-"  faker:"-" db:"nid"`
+
+	// DangerousSkipCSRFCheck indicates whether anti CSRF measures should be enforced in this flow
+	//
+	// This is needed, because we can not enforce these measures, if the flow has been initialized by some else than
```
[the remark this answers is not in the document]
jonas-jonas: Done.

selfservice/strategy/code/strategy_recovery.go
jonas-jonas: Done.

persistence/sql/persister_recovery.go
```diff
@@ -127,3 +128,67 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+		var ra identity.RecoveryAddress
+		if err := tx.Where("id = ? AND nid = ?", recoveryCode.RecoveryAddressID, nid).First(&ra); err != nil {
+			if !errors.Is(sqlcon.HandleError(err), sqlcon.ErrNoRows) {
+				return err
+			}
+		}
+		recoveryCode.RecoveryAddress = &ra
+
+		/* #nosec G201 TableName is static */
+		return tx.RawQuery(fmt.Sprintf("UPDATE %s SET used=true, used_at=? WHERE id=? AND nid = ?", recoveryCode.TableName(ctx)), time.Now().UTC(), recoveryCode.ID, nid).Exec()
```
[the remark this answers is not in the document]
jonas-jonas: done.

jonas-jonas:
>Instead of iterating over secrets you could consider using just one IN query. Depends though on the overhead of computing potentially unnecessary HMACs.
@zepatrik 

Not sure why I can't answer directly under this comment, but I thought I'd address this: I changed the implementation to fetch all possible codes from the DB as discussed and iterate through them to check if their codes match. If we find a matching row, we break the loop early. So in the best case this has `O(1)` and in the worst `O(n^2)` time complexity.

internal/testhelpers/selfservice_verification.go
jonas-jonas: Done, by flipping the assertion, because now the field is always set.

persistence/sql/persister_recovery.go
jonas-jonas: I have refactored this quite a bit, after @zepatrik pointed out some improvements here. The initial question still stands though, as I don't know which secrets we can use instead.

persistence/sql/migrations/sql/20220901123209000000_recovery_code.mysql.up.sql
```diff
@@ -0,0 +1,33 @@
...
+    code VARCHAR (64) NOT NULL, -- HMACed value of the actual code
+    used_at timestamp NULL DEFAULT NULL,
+    identity_recovery_address_id CHAR(36),
+    code_type int NOT NULL,
+    expires_at timestamp NOT NULL DEFAULT '2000-01-01 00:00:00',
+    issued_at timestamp NOT NULL DEFAULT '2000-01-01 00:00:00',
+    selfservice_recovery_flow_id CHAR(36),
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    nid CHAR(36) NOT NULL,
```
zepatrik, line 13: Shouldn't this be an FK as well?

selfservice/flow/recovery/error_test.go
```diff
@@ -226,5 +229,47 @@ func TestHandleError(t *testing.T) {
...
+			recoveryFlow = &recovery.Flow{Type: flow.TypeBrowser, Active: "not-valid"}
+			flowError = flow.NewFlowExpiredError(anHourAgo)
+
+			lf, _ := expectRecoveryUI(t)
+			require.Len(t, lf.UI.Messages, 1, "%s", jsonx.TestMarshalJSONString(t, lf))
+			assert.Equal(t, int(text.ErrorValidationRecoveryFlowExpired), int(lf.UI.Messages[0].ID))
+			assert.Equal(t, "code", lf.Active.String())
+		})
+
+		t.Run("case=XX", func(t *testing.T) {
```
zepatrik, line 259: Missing description?

selfservice/strategy/code/code_recovery.go
```diff
@@ -0,0 +1,125 @@
...
+type RecoveryCodeType int
+
+const (
+	RecoveryCodeTypeAdmin RecoveryCodeType = iota + 1
+	RecoveryCodeTypeSelfService
+)
+
+var (
+	ErrCodeNotFound    = herodot.ErrNotFound.WithReasonf("unknown recovery code")
+	ErrCodeAlreadyUsed = herodot.ErrNotFound.WithReasonf("recovery code was already used")
```
zepatrik, line 25: Should this really be `404`? Not rather `Bad Request` or `Precondition Failed`?

selfservice/strategy/code/code_recovery.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (r RecoveryCode) WasUsed() bool {
+	return r.UsedAt.Valid
+}
+
+func GenerateRecoveryCode() string {
+	return randx.MustString(8, randx.Numeric)
+}
+
+type RecoveryCodeDTO struct {
```
zepatrik, line 85: So this struct is only used internally by the persister right? Why does it not have `db` tags? Why does the other struct have them?

selfservice/strategy/code/code_recovery_test.go
```diff
@@ -0,0 +1,91 @@
...
+			ID:        x.NewUUID(),
+			FlowID:    f.ID,
+			ExpiresAt: time.Now().Add(expiresIn),
+		}
+	}
+
+	req := &http.Request{URL: urlx.ParseOrPanic("https://www.ory.sh/")}
+	t.Run("func=GenerateRecoveryCode", func(t *testing.T) {
+		t.Run("case=creates unique tokens", func(t *testing.T) {
+			codes := make([]string, 100)
```
zepatrik, line 37: The chance that two codes are the same is `100/10^8 = 1/10^6` which should be low enough to not cause flakes here. If my statistics are correct, and I am bad in statistics :joy:

selfservice/strategy/code/code_recovery_test.go
```diff
@@ -0,0 +1,91 @@
...
+			token := newCode(-time.Hour, f)
+			require.True(t, token.IsExpired())
+		})
+		t.Run("case=returns false if flow is not expired", func(t *testing.T) {
+			f, err := recovery.NewFlow(conf, time.Hour, "", req, nil, flow.TypeBrowser)
+			require.NoError(t, err)
+
+			token := newCode(time.Hour, f)
+			require.False(t, token.IsExpired())
+		})
```
zepatrik, line 60: What about a combination of flow is not expired, but code is? Or the other way around? Could that actually happen?

zepatrik (commented):
This looks very good now :+1:
Still didn't finish all files, especially e2e :sweat_smile: soo many code changes :zzz: but I'll continue soon.

persistence/sql/migrations/sql/20220901123209000000_recovery_code.mysql.up.sql
jonas-jonas: Yes, added the constraints below.

selfservice/flow/recovery/error_test.go
jonas-jonas: Duno, I XX very descriptive. ;)

Fixed.

selfservice/strategy/code/code_recovery.go
jonas-jonas: Right, I like bad request. Done.

selfservice/strategy/code/code_recovery.go
jonas-jonas: This struct is only used as a Data Transfer Object, meaning we are populating its values with the stuff we know when creating a new code. Then it is passed into the persister method, which then "transforms" it into the _actual_ entity by taking all the stuff from this struct into the new `RecoveryCode` object. This struct does not represent the entity, but is rather just "parameters" to a method.

>Why does the other struct have them?

Because `RecoveryCode` is what actually represents the entity.

selfservice/strategy/code/code_recovery_test.go
jonas-jonas: This test only tests the method on the `RecoveryCode` struct, which does not take the flow into account.

The flow can expire before the code, in which case the flow will be invalidated first and the code is never checked.

In case the code expires first, this method is called. We have tests for both cases. 

- flow expired: https://github.com/ory/kratos/pull/2645/files#diff-d3976c6ed314fd1f2372a82a50ef612c974597c85bb49031eecd3f50952c3c40R829
- code expired: https://github.com/ory/kratos/pull/2645/files#diff-d3976c6ed314fd1f2372a82a50ef612c974597c85bb49031eecd3f50952c3c40R161

Also E2e: https://github.com/ory/kratos/pull/2645/files#diff-4156044df819e73198f46caad2012dc54e4d204a92f2417ff94d29f4a4676d61R75

selfservice/strategy/code/code_recovery_test.go
jonas-jonas: Should be fine ™️

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -0,0 +1,609 @@
...
+	sID := s.RecoveryStrategyID()
+
+	// If the email is present in the submission body, the user needs a new code via resend
+	if f.State != recovery.StateChooseMethod && len(body.Email) == 0 {
+		if err := flow.MethodEnabledAndAllowed(ctx, sID, sID, s.deps); err != nil {
+			return s.HandleRecoveryError(w, r, nil, body, err)
+		}
+
+		f.SubmitCount++
+		if err := s.deps.RecoveryFlowPersister().UpdateRecoveryFlow(ctx, f); err != nil {
```
hperl, line 283: There were a lot of changes, and a lot of comments raised and addressed, so I'll just focus on this: Are you really sure that the `SubmitCount` is updated atomically? What we want to prevent is that multiple requests come in concurrently with different brute-forced codes, and then concurrently `SubmitCount` is set to `1` and the database is updated.

I would have expected some kind of database transaction beginning from when the `Recover` function starts to until we know whether the code matched.

You have a test "description=should not be able to use an invalid code more than 5 times", that tests that the count works, but it performs the checks concurrently.

In order to protect us from brute-forcing 8 digits, I'd really want to have the concurrent case covered. Am I missing something about the flow works? Still new to the Kratos code base.

jonas-jonas: Yes, you're right. I moved the `submit_count > 5` check inside the transaction.

However, I am not sure how to test the concurrent submission issue, or rather, how we should handle concurrent submissions in general. I wrote a quick test case that executes the same recovery request concurrently, but that leads to `500 database table is locked: selfservice_recovery_flows`. I think this is only an issue on SQLite, as the others should have better (no) locking mechanisms. But I am no expert on this... We are running all tests against SQLite. Since you know the code base better, WDYT @aeneasr?.

persistence/sql/persister_recovery.go
jonas-jonas: `IN` queries seem to be very difficult with `pop`. I saw `sqlx.In` but that is targeted at full queries, IDK how it works with `pop`. Why is iterating over the secrets an issue? 

>You could also use a RETURNING clause to combine all three queries into one.

`RETURNING` is not supported on `mysql`...

driver/config/config.go
jonas-jonas: I think @zepatrik resolved this by accident, this is not done.

aeneasr:
@jonas-jonas please resolve the merge conflicts :)

persistence/sql/persister_recovery.go
```diff
@@ -129,66 +131,127 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+
+			return err
+		}
+
+		if submitCount > 5 {
+			return code.ErrCodeSubmittedTooOften
+		}
+
+		var recoveryCodes []code.RecoveryCode
+		if err = sqlcon.HandleError(tx.Where("nid = ? AND selfservice_recovery_flow_id = ?", nid, fID).All(&recoveryCodes)); err != nil {
```
aeneasr, line 202: Do we expect to have more then one recovery code per flow?

persistence/sql/persister_recovery.go
```diff
@@ -129,66 +131,127 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+			if errors.Is(err, sqlcon.ErrNoRows) {
+				// Return no error, as that would roll back the transaction
+				return nil
+			}
+
+			return err
+		}
+
+		if submitCount > 5 {
+			return code.ErrCodeSubmittedTooOften
```
aeneasr, line 198:
```suggestion
			return errors.WithStack(code.ErrCodeSubmittedTooOften)
```

persistence/sql/persister_recovery.go
```diff
@@ -129,66 +131,127 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+		if err := sqlcon.HandleError(tx.RawQuery(fmt.Sprintf("SELECT submit_count FROM %s WHERE id = ? AND nid = ?", flowTableName), fID, nid).First(&submitCount)); err != nil {
+			if errors.Is(err, sqlcon.ErrNoRows) {
+				// Return no error, as that would roll back the transaction
+				return nil
+			}
+
+			return err
+		}
+
+		if submitCount > 5 {
```
aeneasr, line 197: Handling this here IMO breaks the isolation of the persister, because now we are performing business logic within something responsible of fecthing / writing to the datastore. For the sake of shipping this, I would be OK to keep it as is if we add a big warning to the function godoc.

If you were to extract this logic, I think we would need some method in the persister such as: `IncreaseSubmitCount() (count int, err error)` and use that to perform our business logic in some other service / place.

A good example for this is the session persister, and session manager:

* session persister: writes and reads from the data store
* session manager: does things like setting cookies, changing AAL of a session, ...

persistence/sql/persister_recovery.go
```diff
@@ -129,66 +131,127 @@ func (p *Persister) DeleteExpiredRecoveryFlows(ctx context.Context, expiresAt ti
...
+
+	secrets:
 		for _, secret := range p.r.Config().SecretsSession(ctx) {
-			if err = tx.Where("code = ? AND nid = ? AND NOT used AND selfservice_recovery_flow_id = ?", p.hmacValueWithSecret(ctx, codeVal, secret), nid, fID).First(&recoveryCode); err != nil {
-				if !errors.Is(sqlcon.HandleError(err), sqlcon.ErrNoRows) {
-					return err
+			suppliedCode := p.hmacValueWithSecret(ctx, codeVal, secret)
+			for i := range recoveryCodes {
+				code := recoveryCodes[i]
+				if code.Code != suppliedCode {
```
aeneasr, line 216: Please use `subtle.ConstantTimeCompare` to avoid timing attacks

persistence/sql/persister_recovery.go
aeneasr: The solution is OK!

selfservice/flow/recovery/error.go
```diff
@@ -69,8 +69,17 @@ func (s *ErrorHandler) WriteFlowError(
...
 	if e := new(flow.ExpiredError); errors.As(err, &e) {
+		strategy, err := s.d.RecoveryStrategies(r.Context()).Strategy(f.Active.String())
+		if err != nil {
+			strategy, err = s.d.GetActiveRecoveryStrategy(r.Context())
+			// Can't retry the recovery if no strategy has been set
+			if err != nil {
+				s.d.SelfServiceErrorManager().Forward(r.Context(), w, r, err)
+				return
+			}
+		}
```
aeneasr, line 80: Is there no easier way to get this?

selfservice/flow/recovery/strategy.go
```diff
@@ -45,7 +45,7 @@ func (s Strategies) Strategy(id string) (Strategy, error) {
 		}
 	}
 
-	return nil, errors.Errorf(`unable to find strategy for %s have %v`, id, ids)
+	return nil, herodot.ErrInternalServerError.WithReasonf("unable to find strategy for %s have %v", id, ids)
```
aeneasr, line 48: Wrap in `errors.WithStack` please

selfservice/strategy/code/code_recovery.go
aeneasr: Looking at the code, it does not seem to be neccessary to have this object. The fields are all the same data types as far as I can tell. That means that we can do the transformations on the `RecoveryCode` struct itself and don't need this indirection.

If you were to have other or less fields, the struct would be called something `CreateRecoveryCodeParams` :)

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -182,23 +182,25 @@ func (s *Strategy) createRecoveryCode(w http.ResponseWriter, r *http.Request, _
 	}
 
 	id, err := s.deps.IdentityPool().GetIdentity(ctx, p.IdentityID)
-	if errors.Is(err, sqlcon.ErrNoRows) {
-		s.deps.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf("The requested identity id does not exist.").WithWrap(err)))
+	if notFoundErr := sqlcon.ErrNoRows; errors.As(err, &notFoundErr) {
```
aeneasr, line 185:
```suggestion
	if errors.Is(err, notFoundErr) {
```

test/e2e/cypress/integration/profiles/recovery/code/errors.spec.ts
```diff
@@ -75,75 +117,67 @@ context('Account Recovery Errors', () => {
...
-        cy.get('[data-testid="ui/message/4060005"]').should(
-          'contain.text',
-          'The recovery flow expired'
-        )
-
-        cy.noSession()
-      })
-
-      it('is unable to recover the account if the code is incorrect', () => {
+      it("invalid remote recovery email template", () => {
```
aeneasr, line 166: Test `is unable to recover the email address if the code is expired` is now missing - for a reason?

aeneasr (commented):
Great job, many improvements added! I've got a few comments but nothing too wild :) We're very close to getting this shipped!

persistence/sql/persister_recovery.go
jonas-jonas: Yes, I agree. 

I'd like to revisit this, when (if) we are refactoring the recovery flow or when I add verification to the OTP strategy, I have a few ideas to improve this whole data layout a bit more.

persistence/sql/persister_recovery.go
jonas-jonas: Yes, "resend" is one case where this can happen.

persistence/sql/persister_recovery.go
aeneasr: Ok, could you please add a codedoc for the function, a todo entry, and an issue for it with you assigned? :)

test/e2e/cypress/integration/profiles/recovery/code/errors.spec.ts
jonas-jonas: Just renamed, to better reflect what we are testing, IMO: https://github.com/ory/kratos/pull/2645/files/b48710507909abefc579dfa05c5789b4c3fc91ff#diff-4156044df819e73198f46caad2012dc54e4d204a92f2417ff94d29f4a4676d61R75

persistence/sql/persister_recovery.go
jonas-jonas: Done.

persistence/sql/persister_recovery.go
jonas-jonas: Done.

https://github.com/ory/kratos/issues/2785

persistence/sql/persister_recovery.go
jonas-jonas: Done.

selfservice/flow/recovery/strategy.go
jonas-jonas: Done.

jonas-jonas: Done.

selfservice/flow/recovery/error.go
jonas-jonas: Not really sure how, tbh.

We first need to check if the strategy of the "old" flow is still available. If it isn't, we fall back to the active strategy, and if that's not available we need to fail the flow.

persistence/sql/persister_recovery.go
aeneasr: Would that invalidate the older code?

selfservice/strategy/code/code_recovery.go
aeneasr: Ok, then maybe call it `RawCode` or `CodePlaintext` to make it clear in `CreateRecoveryCodeParams`

jonas-jonas: Alright, I named it `CodeHMAC` on the `RecoveryCode` and `RawCode` on `CreateRecoveryCodeParams`.

persistence/sql/persister_recovery.go
jonas-jonas: No, it does not, but I also thought of that. Maybe we should do that?

UX: 
- Gmail sometimes groups these emails together, which makes it hard to distinguish between the first and second email, so it would be nice if both codes work

Security:
- Having multiple codes does open up the possibility to brute force these codes. Especially, since we don't have a limit on codes. That might also be an option: `if "count(codes) > 5" then "retry the flow"` e.g. if a user clicked "resend" more than 5 times, we create a new flow. 

Overall, I think security is more important here, and the easy fix is to just delete all codes from the database if a new code is requested. WDYT?

aeneasr: > Overall, I think security is more important here, and the easy fix is to just delete all codes from the database if a new code is requested. WDYT?

Let's do that!

selfservice/strategy/code/code_recovery.go
aeneasr: 👍

selfservice/strategy/code/strategy_recovery.go
jonas-jonas: I am now looking at writing a test to reproduce this. There are several issues with "just spawn a bunch of goroutines that execute the same request". I have talked to @hperl, and we discussed a few of them, but these are still there, and I am not sure how to solve these:
1. The `NewErrorTestServer` reads errors from the DB, which fails with "database is locked". I assume this is due to the many requests, but can't verify this.
2. Some requests fail with 
```
Error:      	Received unexpected error:
Post "http://127.0.0.1:54193/recovery-ts?flow=12a1c322-dd4b-43ce-8da6-0870c2a9f4fd": EOF
```
I assume, this is due to the test servers being overloaded. 

At this point, I am not even sure if `strategy_recovery.go` is the correct place to do this. Do you know if we have done this type of testing before, or do you have experience with it? I could use some pointers here.

persistence/sql/persister_recovery.go
jonas-jonas: Done.

selfservice/strategy/code/strategy_recovery.go
aeneasr: Since we have logic in SQL in a transaction to solve this, I would be fine with not testing it and instead focus on getting this shipped. Maybe add some code comments to make sure that people understand why this is needed!

aeneasr:
I'll check out the code locally and play around, but code-wise this looks stellar :) Once tests CI is green and I've finished my testing this can land on production next!

aeneasr (approved):
Epic, great job @jonas-jonas !

aeneasr (approved):

Issue #2639, opened 2022-08-02
Add API Endpoint to fetch emails


jonas-jonas:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Currently, there is no way to fetch emails that were generated by kratos via the API. This also applies to SMS messages and messages that failed to deliver.

### Describe your ideal solution

The admin API should provide an endpoint to fetch all messages generated by kratos at `/admin/courier/messages`

The endpoint would return an array of items with the following structure. Example:
```json
{
  "id": "1ff05f9e-0684-4d45-adcf-ef896667ba84",
  "status": "ABANDONED",
  "type": "EMAIL",
  "recipient": "test-recipient-1@example.org",
  "subject": "stub email subject test-subject-1",
  "template_type": "stub",
  "send_count": 1,
  "created_at": "2022-08-02T16:29:26.978027+02:00",
  "updated_at": "2022-08-02T16:29:26.978027+02:00"
}
```

For the fields `status` and `type` we need some kind of string enums to make them better understandable. Currently, these fields are of type integer in the database, which results in `status=4` (abandoned) and `type=1` (Email).

The endpoint should also provide a way to filter for the message's `status` field to only fetch, e.g. `MessageStatusAbandoned` messages.

It should also have pagination support.

### Workarounds or alternatives

None, currently.

### Version

v0.10.1

aeneasr:
> The admin API should provide an endpoint to fetch all messages generated by kratos at /courier/messages

```
- The admin API should provide an endpoint to fetch all messages generated by kratos at /courier/messages`
+ The admin API should provide an endpoint to fetch all messages generated by kratos at /admin/courier/message
```

or alternatively:

```
/admin/courier/deliveries
```

> The endpoint would return an array of items with the following example:

[...]

PR #2651, opened 2022-08-08
feat: add api endpoint to fetch messages

erolkskn:
## Related issue(s)
#2639 

## Checklist


aeneasr:
@erolkskn let me know when this is good for a review :)

erolkskn:
@aeneasr Hi aeneas, i think it's ready for review now :) Btw, about the last commit, autogen bot somehow formatted swagger doc blocks wrong (see: 4159b93ae3f8175cf7ccf77d34e4a7a2d0181d4f). That commit caused swagger to be not able to generate spec. So i needed to make that change in order to make sdk generation work again (see: https://github.com/ory/kratos/pull/2651/commits/b51a916fd67385ae8f4c44f0ce654c33e720711c).

aeneasr:
Great, @erolkskn ! Yeah the problem is that Go 1.19 changed the formatter, which now breaks the go-swagger generation. I fixed the issues on master, so if you merge / rebase they should be resolved.

@jonas-jonas can you please take a look at the PR? :)

Also, we probably need a bit of documentation for this feature, maybe in a new doc in https://github.com/ory/docs/tree/master/docs/kratos/emails-sms ?

persistence/sql/persister_courier.go
```diff
@@ -25,6 +25,33 @@ func (p *Persister) AddMessage(ctx context.Context, m *courier.Message) error {
 	return sqlcon.HandleError(p.GetConnection(ctx).Create(m)) // do not create eager to avoid identity injection.
 }
 
+func (p *Persister) ListMessages(ctx context.Context, filter courier.MessagesFilter) ([]courier.Message, int64, error) {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.ListMessages")
+	defer span.End()
+
+	q := p.GetConnection(ctx).Where("nid=?", p.NetworkID(ctx))
```
jonas-jonas, line 32: I think, this should be `corp.ContextualizeNID(ctx, p.nid)` to keep it consistent with the other methods in the persister.

persistence/sql/persister_courier.go
```diff
@@ -25,6 +25,33 @@ func (p *Persister) AddMessage(ctx context.Context, m *courier.Message) error {
...
+	if filter.Status != nil {
+		q = q.Where("status=?", *filter.Status)
+	}
+
+	if filter.Recipient != "" {
+		q = q.Where("recipient=?", filter.Recipient)
+	}
+
+	messages := make([]courier.Message, 0)
+	if err := q.Paginate(filter.Page, filter.PerPage).Order("created_at DESC").All(&messages); err != nil {
```
jonas-jonas, line 43: I know this wasn't part of the issue, but seeing this line made me think of this: It might be worth adding a sort-direction to the `courier.MessagesFilter` struct to set the direction (`DESC` or `ASC`) here. This could be used to get the first sent emails matching the other filters. cc @aeneasr

jonas-jonas (changes requested):
Very nice work! Thank you for your contribution. This looks pretty good so far, I just have two ideas on how to improve it.

persistence/sql/persister_courier.go
erolkskn: Yeah it was actually like that before, but after i rebased this branch onto master yesterday, i seen that `corp.ContextualizeNID(ctx, pid)` is refactored to `p.NetworkID(ctx)` in commit b0d8f3853886228a64e82437643a82b3970d6ff7. That's why i did go with `p.NetworkID(ctx)`

jonas-jonas: Huh, interesting, I missed that—sorry about that. This looks good to me then.

persistence/sql/persister_courier.go
erolkskn: @jonas-jonas I think that's a good idea. Should i proceed to implement this now ? Or are we waiting for @aeneasr 's approval ?

aeneasr: I agree that it would be useful, but since we do not have ordering yet in our REST APIs and we have no design document describing how such ordering would look like, i'd like to defer this for another time :)

jonas-jonas: Then LGTM.

jonas-jonas (dismissed):
LGTM!

persistence/sql/persister_test.go
```diff
@@ -240,8 +239,7 @@ func TestPersister(t *testing.T) {
 			})
 			t.Run("contract=courier.TestPersister", func(t *testing.T) {
 				pop.SetLogger(pl(t))
-				upsert, insert := sqltesthelpers.DefaultNetworkWrapper(p)
```
aeneasr, line 243: We use this helper elswhere so it's needed to have this little function. Can you please revert those changes? Thanks! :)

aeneasr (changes requested):
Thank you! This looks great, just one topic we need to change :)

persistence/sql/persister_test.go
erolkskn: @aeneasr Alright, i reverted the change. How does it look now ? :)

aeneasr (commented):
It looks fantastic! I will do another pass next week but I think this is good to be merged :)

courier/handler.go
```diff
@@ -0,0 +1,128 @@
...
+		return
+	}
+
+	l, tc, err := h.r.CourierPersister().ListMessages(r.Context(), filter)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
+
+	if h.r.Config().IsInsecureDevMode(r.Context()) {
```
zepatrik, line 96: Shouldn't this be 
```suggestion
	if !h.r.Config().IsInsecureDevMode(r.Context()) {
```

courier/handler_test.go
```diff
@@ -0,0 +1,193 @@
...
+	t.Run("case=list messages", func(t *testing.T) {
+		// Arrange test data
+		const msgCount = 10    // total message count
+		const procCount = 5    // how many messages' status should be equal to `processing`
+		const rcptOryCount = 2 // how many messages' recipient should be equal to `noreply@ory.sh`
+		messages := make([]courier.Message, msgCount)
+
+		for i := range messages {
+			require.NoError(t, faker.FakeData(&messages[i]))
+			messages[i].Type = courier.MessageTypeEmail
```
zepatrik, line 80:
```suggestion
			messages[i].Type = courier.MessageTypeEmail
			messages[i].Body = "body content"
```

courier/handler_test.go
```diff
@@ -0,0 +1,193 @@
...
+		})
+		t.Run("case=body should be redacted if kratos is in dev mode", func(t *testing.T) {
+			conf.MustSet(ctx, "dev", true)
+			for _, name := range tss {
+				t.Run("endpoint="+name, func(t *testing.T) {
+					parsed := getList(t, name, "")
+					require.Len(t, parsed.Array(), msgCount, "%s", parsed.Raw)
+
+					for _, item := range parsed.Array() {
+						assert.Equal(t, "<redacted-unless-dev-mode>", item.Get("body").String())
```
zepatrik, line 166:
```suggestion
						assert.Equal(t, "body content", item.Get("body").String())
```

courier/handler_test.go
```diff
@@ -0,0 +1,193 @@
...
+		})
+		t.Run("case=body should not be redacted if kratos is not in dev mode", func(t *testing.T) {
+			conf.MustSet(ctx, "dev", false)
+			for _, name := range tss {
+				t.Run("endpoint="+name, func(t *testing.T) {
+					parsed := getList(t, name, "")
+					require.Len(t, parsed.Array(), msgCount, "%s", parsed.Raw)
+
+					for _, item := range parsed.Array() {
+						assert.NotEqual(t, "<redacted-unless-dev-mode>", item.Get("body").String())
```
zepatrik, line 179:
```suggestion
						assert.Equal(t, "<redacted-unless-dev-mode>", item.Get("body").String())
```

courier/message.go
```diff
@@ -16,14 +21,146 @@ const (
...
+		return MessageStatusQueued, nil
+	case messageStatusSentText:
+		return MessageStatusSent, nil
+	case messageStatusProcessingText:
+		return MessageStatusProcessing, nil
+	case messageStatusAbandonedText:
+		return MessageStatusAbandoned, nil
+	default:
+		return MessageStatus(0), errors.New("Message status is not valid")
+	}
```
zepatrik, line 43:
```suggestion
	switch s := stringsx.SwitchExact(str); {
	case s.AddCase(MessageStatusQueued.String()):
		return MessageStatusQueued, nil
	case s.AddCase(MessageStatusSent.String()):
		return MessageStatusSent, nil
	case s.AddCase(MessageStatusProcessing.String()):
		return MessageStatusProcessing, nil
	case s.AddCase(MessageStatusAbandoned.String()):
		return MessageStatusAbandoned, nil
	default:
		return 0, errors.WithStack(herodot.ErrBadRequest.WithError(s.ToUnknownCaseErr()).WithReason("Message status is not valid"))
	}
```

zepatrik: Using `herodot` errors automatically sets the status code correctly.

courier/message.go
```diff
@@ -16,14 +21,146 @@ const (
...
+
+func ToMessageType(str string) (MessageType, error) {
+	switch str {
+	case messageTypeEmailText:
+		return MessageTypeEmail, nil
+	case messageTypePhoneText:
+		return MessageTypePhone, nil
+	default:
+		return MessageType(0), errors.New("Message type is not valid")
+	}
```
zepatrik, line 118: Similar to above suggestion :wink:

courier/message_test.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"testing"
+
+	"github.com/ory/kratos/courier"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMessageStatusValidity(t *testing.T) {
+	invalid := courier.MessageStatus(0)
+	require.Error(t, invalid.IsValid(), "IsValid() should return an error when message status is invalid")
```
zepatrik, line 12:
```suggestion
	require.ErrorIs(t, invalid.IsValid(), herodot.ErrBadRequest, "IsValid() should return an error when message status is invalid")
```

courier/message_test.go
```diff
@@ -0,0 +1,56 @@
...
+		} {
+			result, err := courier.ToMessageStatus(str)
+			require.NoError(t, err)
+			require.Equal(t, exp, result)
+		}
+	})
+	t.Run("case=should return error for invalid message status str", func(t *testing.T) {
+		result, err := courier.ToMessageStatus("invalid")
+		require.Error(t, err)
+		require.Error(t, result.IsValid())
```
zepatrik, line 31:
```suggestion
		require.ErrorIs(t, result.IsValid(), herodot.ErrBadRequest)
```

courier/message_test.go
```diff
@@ -0,0 +1,56 @@
...
+	})
+}
+
+func TestMessageTypeValidity(t *testing.T) {
+	invalid := courier.MessageType(0)
+	require.Error(t, invalid.IsValid(), "IsValid() should return an error when message type is invalid")
```
zepatrik, line 37:
```suggestion
	require.ErrorIs(t, invalid.IsValid(), herodot.ErrBadRequest, "IsValid() should return an error when message type is invalid")
```

courier/message_test.go
```diff
@@ -0,0 +1,56 @@
...
+		} {
+			result, err := courier.ToMessageType(str)
+			require.NoError(t, err)
+			require.Equal(t, exp, result)
+		}
+	})
+	t.Run("case=should return error for invalid message type str", func(t *testing.T) {
+		result, err := courier.ToMessageType("invalid")
+		require.Error(t, err)
+		require.Error(t, result.IsValid())
```
zepatrik, line 54:
```suggestion
		require.ErrorIs(t, result.IsValid(), herodot.ErrBadRequest)
```

courier/test/persistence.go
```diff
@@ -107,6 +107,32 @@ func TestPersister(ctx context.Context, newNetworkUnlessExisting NetworkWrapper,
...
+					Page:    1,
+					PerPage: 100,
+				},
+			}
+			ms, tc, err := p.ListMessages(ctx, filter)
+
+			require.NoError(t, err)
+			assert.Len(t, ms, len(messages))
+			assert.Equal(t, int64(len(messages)), tc)
+			assert.Equal(t, messages[len(messages)-1].ID, ms[0].ID)
```
zepatrik, line 124:
```suggestion
			assert.Equal(t, messages[len(messages)-1].ID, ms[0].ID)
			assert.ElementsMatch(t, messages, ms)
```
`assert.ElementsMatch` ignores the order.`

courier/message.go
```diff
@@ -16,14 +21,146 @@ const (
 	MessageStatusAbandoned
 )
 
+const (
+	messageStatusQueuedText     = "queued"
+	messageStatusSentText       = "sent"
+	messageStatusProcessingText = "processing"
+	messageStatusAbandonedText  = "abandoned"
+)
```
zepatrik, line 29:
```suggestion
// A Message's Status
//
// swagger:enum messageStatus
type messageStatus string

const (
	messageStatusQueuedText messageStatus     = "queued"
	messageStatusSentText messageStatus       = "sent"
	messageStatusProcessingText messageStatus = "processing"
	messageStatusAbandonedText messageStatus  = "abandoned"
)
```

You might be able to avoid the patches if you do it this way. What is important is that the type name is the same as the `// swagger:enum name`, and that all values are in one block immediately after the type for this to work. But you can also just keep the patch if this does not work properly.

.schema/openapi/patches/courier.yaml
```diff
@@ -0,0 +1,29 @@
...
+- op: remove
+  path: /components/schemas/messageType/format
+- op: replace
+  path: /components/schemas/messageType/type
+  value: string
+- op: add
+  path: /components/schemas/messageType/enum
+  value:
+    - email
+    - phone
```
zepatrik, line 24: Maybe make this just one op?
```suggestion
- op: replace
  path: /components/schemas/messageStatus
  value:
    type: string
    enum:
      - queued
      - sent
      - processing
      - abandoned
# Makes messageType a string enum
- op: replace
  path: /components/schemas/messageType
  value:
    type: string
    enum:
      - email
      - phone
```
That way it is exactly what you want.

zepatrik (changes requested):
Nice, just found some minor bugs :wink: 
But overall nearly there :+1:

courier/handler.go
erolkskn: Most definitely!

courier/test/persistence.go
erolkskn: Didn't know that, thank you :-)

.schema/openapi/patches/courier.yaml
erolkskn: Yeah that's way better approach, idk what i was thinking writing that piece of code :-)

courier/message.go
erolkskn: @zepatrik I guess `ErrBadRequest.WithError` should be `ErrBadRequest.WithWrap` right ?

courier/test/persistence.go
erolkskn: @zepatrik On a second thought, this can't work on this scenario. Test cases above the `case=list messages` either updates the messages with `SetMessageStatus` or `IncrementMessageSendCount` which none of them takes a reference to the `courier.Message` itself. So `ListMessages` method always returns messages with different `MessageStatus` or `SendCount` than what we have in `messages` variable. So `assert.ElementsMatch` statement bound to be failed due to the reasons i listed above.

courier/message.go
erolkskn: @zepatrik That would definitely make us avoid the patches, but is it really necessary to introduce another type just for this ? I think that might complicate code-base a little bit more (all the need for casting `messageStatus` to `string` and vice versa, new `IsValid()` method just for `messageStatus` etc.) ? We already have 150 lines of code just for two enumeration, i think expanding it might be a little bit confusing to future developers of this project. What do you think ?

aeneasr (dismissed):

aeneasr (dismissed):
Great job!

PR #2664, opened 2022-08-12
fix: replace io/util usage by io and os package

jonas-jonas:
## Checklist

### DIFF
--- a/internal/httpclient/api_metadata.go
+++ b/internal/httpclient/api_metadata.go
@@ -14,7 +14,7 @@ package client
 import (
 	"bytes"
 	"context"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"net/url"
 )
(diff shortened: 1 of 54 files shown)


internal/httpclient/api_metadata.go
```diff
@@ -14,7 +14,7 @@ package client
 import (
 	"bytes"
 	"context"
-	"io/ioutil"
+	"io"
```
zepatrik, line 17: Hm, the auto-generated files will always use `io/ioutil`. I propose to add some automation to `make sdk` that fixes this after the SDK is regenerated.

zepatrik (commented):
Nice, looking pretty good :+1:

internal/httpclient/api_metadata.go
aeneasr: It would make more sense to update the OpenAPI Generator template

jonas-jonas: Yup, you are right, I forgot about the generated code.. 

I'll have a look at them and try to fix it there. Are they defined by us or upstream in the generator?

I would like to avoid some automatic tooling for this and rather fix it in the generation logic.

jonas-jonas: Done.

aeneasr (approved):

PR #2679, opened 2022-08-19
fix: include flow id in use recovery token query

jonas-jonas:
This PR adds the `selfservice_recovery_flow_id` to the query used when "using" a token in the recovery flow.

This PR also adds a new enum field for `identity_recovery_tokens` to distinguish the two flows: admin versus self-service recovery.

BREAKING CHANGES: This patch invalidates recovery flows initiated using the Admin API. Please re-generate any admin-generated recovery flows and tokens.


selfservice/strategy/link/test/persistence.go
```diff
@@ -34,12 +34,8 @@ func TestPersister(ctx context.Context, conf *config.Config, p interface {
 		conf.MustSet(config.ViperKeySecretsDefault, []string{"secret-a", "secret-b"})
 
 		t.Run("token=recovery", func(t *testing.T) {
-			t.Run("case=should error when the recovery token does not exist", func(t *testing.T) {
```
jonas-jonas, line 37: FYI - this test was duplicated below (L63-L66)

persistence/sql/migrations/sql/20220819132309000000_add_flow_type_to_identity_recovery_tokens.cockroach.up.sql
```diff
@@ -0,0 +1,17 @@
+ALTER TABLE identity_recovery_tokens
```
aeneasr, line 1: Please remove the COMMIT; BEGIN transaction statements and move the DDL and DSL statements into separate files

aeneasr (changes requested):

persistence/sql/migrations/sql/20220819132309000000_add_flow_type_to_identity_recovery_tokens.cockroach.up.sql
jonas-jonas: Done.

persistence/sql/persister_recovery.go
```diff
@@ -74,7 +74,7 @@ func (p *Persister) UseRecoveryToken(ctx context.Context, token string) (*link.R
 	nid := p.NetworkID(ctx)
 	if err := sqlcon.HandleError(p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) (err error) {
 		for _, secret := range p.r.Config().SecretsSession(ctx) {
-			if err = tx.Where("token = ? AND nid = ? AND NOT used", p.hmacValueWithSecret(ctx, token, secret), nid).First(&rt); err != nil {
+			if err = tx.Where("token = ? AND nid = ? AND NOT used AND selfservice_recovery_flow_id = ?", p.hmacValueWithSecret(ctx, token, secret), nid, fID).First(&rt); err != nil {
```
aeneasr, line 77: Is `selfservice_recovery_flow_id` nullable? If so, this query here will never be `selfservice_recovery_flow_id = NULL`.

aeneasr (commented):
Looking good! Just one question then we're ✅

persistence/sql/persister_recovery.go
jonas-jonas: It is nullable and `selfservice_recovery_flow_id = NULL` can be true for older flows, as we can't infer the flow ID in a migration. Though with this implementation the parameter `fID` will _never_ be null, as there _is_ always a flow. In case of an admin initiated flow, it was just not set in the `selfservice_recovery_flow_id` column.

Which means these flows would no longer work with this change. I guess, I could add a check to see if the flow is an `Admin` flow, and if so, not pass a `fID` in, but that would kinda defeat the purpose of this change, IMO.

I am not sure how to handle these "legacy" flows.

aeneasr: This would mean that all admin recovery flows initiated before an upgrade and completed after the upgrade would fail. I think that is not pretty, but acceptable, given that admin recovery flows are not that frequent and the admin knows about the upgrade. Can you please add in the PR message body a `BREAKING CHANGE` note to inform users about this change and impact?

jonas-jonas: Done.

aeneasr (approved):

PR #2695, opened 2022-08-25
feat: add flow id check to use verification token

jonas-jonas:
## Related issue(s)
#2679 

## Checklist


persistence/sql/migrations/sql/20220825134336000000_delete_verification_token_without_flow_id.up.sql
```diff
@@ -0,0 +1,2 @@
+DELETE FROM identity_verification_tokens
+WHERE selfservice_verification_flow_id IS NULL;
```
zepatrik, line 2: Why not add a migration that sets this column to `NOT NULL`?

jonas-jonas: True, will add another migration.

zepatrik: :+1: the rest looks very good

jonas-jonas: Urgh, I added the migrations. Turned out to be a lot of work for what it is ^^😓

persistence/sql/migratest/testdata/20200830154602_testdata.sql
```diff
@@ -1,5 +1,3 @@
 INSERT INTO identity_verification_tokens (id, token, used, used_at, identity_verifiable_address_id, selfservice_verification_flow_id, created_at, updated_at, expires_at, issued_at)
 VALUES ('ee56574d-2f0c-43f6-8d26-0062938ae330', '1001ba7ddd644cb68478e8947e4jfhc', false, null, '45e867e9-2745-4f16-8dd4-84334a252b61', '5385c962-0295-4575-9b1b-d7eef13c0eda', '2013-10-07 08:23:19', '2013-10-07 08:23:19', '2013-10-07 08:23:19', '2013-10-07 08:23:19');
 
-INSERT INTO identity_verification_tokens (id, token, used, used_at, identity_verifiable_address_id, selfservice_verification_flow_id, created_at, updated_at, expires_at, issued_at)
```
aeneasr, line 4: Please re-add this or give a reason why it was removed

persistence/sql/migrations/sql/20220825134336000000_not_null_constraint_verification_token_flow_id.sqlite3.down.sql
```diff
@@ -0,0 +1,37 @@
+ALTER TABLE identity_verification_tokens
```
aeneasr, line 1: How confident are you that this migration is correct and not, for example, missing some keys?

aeneasr (changes requested):
Thank you! This looks very good! The last remaining piece is to add a new file to migratest/testdata to insert another row that uses your new schema :)

persistence/sql/migratest/testdata/20200830154602_testdata.sql
jonas-jonas: Because the only difference between the two rows was that the `selfservice_flow_id` was null in the second, which doesn't make sense anymore, as there will always be a `selfservice_flow_id` following this change.

persistence/sql/migrations/sql/20220825134336000000_not_null_constraint_verification_token_flow_id.sqlite3.down.sql
jonas-jonas: I copied the schema + indices from the SQLite database that I ran the migrations against. So, pretty confident. 

Also, we should see this breaking something in the E2E tests, right?

persistence/sql/migratest/testdata/20200830154602_testdata.sql
jonas-jonas: Or rather, since the schema now does not allow `NULL`s in the column, this testdata statement would just fail^^

jonas-jonas: >The last remaining piece is to add a new file to migratest/testdata to insert another row that uses your new schema :)

Also, the schema is not really "new" there is just one more not null constraint. So we already have testdata desciribing this.

jonas-jonas: Sorry, I forgot how the testdata works. I updated the files.

persistence/sql/migrations/sql/20220825134336000001_not_null_constraint_verification_token_flow_id.sqlite3.down.sql
```diff
@@ -0,0 +1,37 @@
+ALTER TABLE identity_verification_tokens
+RENAME TO identity_verification_tokens_;
+
+CREATE TABLE "identity_verification_tokens" (
```
aeneasr, line 4: You could probably make this a bit easier with:

```
ALTER TABLE ... ADD COLUMN old NOT NULL
UPDATE ... SET new = old
ALTER TABLE DROP COLUMN old
ALTER TABLE RENAME COLUMN new, old
CREATE INDEX ...
```

aeneasr: just make sure to add the indices correctly :)

aeneasr: Oh, or is it not possible to add foreign keys like this?

aeneasr: nope, should be possible:

https://www.sqlite.org/lang_altertable.html


```
ALTER TABLE ... ADD COLUMN old NOT NULL REFERENCES ...
```

jonas-jonas: You can't add foreign keys to an existing table in SQLite, apparently. I'd prefer your solution too, but that isn't possible AFAICT, since `selfservice_verification_flow_id` is an FK.

persistence/sql/migrations/sql/20220825134336000000_not_null_constraint_verification_token_flow_id.sqlite3.down.sql
aeneasr: If the schema is incompatible yes, if it's like a small bug (missing index) then probalby not! So yes, some cases will be caught but some won't :/

persistence/sql/migratest/testdata/20220902141902_testdata.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO identity_verification_tokens (id, token, used, used_at, identity_verifiable_address_id, selfservice_verification_flow_id, created_at, updated_at, expires_at, issued_at, nid)
```
aeneasr, line 1: Please add another row where the column that you changed (selfservice_verification_flow_id) has a NULL value, so that we see if the migratino does what it should (i.e. not create the snapshot because the row was deleted) :) Or is that already the case?

persistence/sql/migrations/sql/20220825134336000001_not_null_constraint_verification_token_flow_id.sqlite3.down.sql
jonas-jonas: Huh, interesting. Then I had a different issue, I guess. The error messages of SQLite aren't that helpful ^^ Thanks, I'll try that.

persistence/sql/migratest/testdata/20220902141902_testdata.sql
jonas-jonas: That fails with `Error 1048: Column 'selfservice_verification_flow_id' cannot be null
` which then fails the whole test suite.

persistence/sql/migrations/sql/20220825134336000001_not_null_constraint_verification_token_flow_id.sqlite3.down.sql
jonas-jonas: Frustrating stuff: Adding a foreign key works, removing the old one doesn't. That's why we apparently need to copy the whole table... Dropping the (nullable) `selfservice_verification_flow_id` column leaves behind its foreign key, and SQLite complains about that: `error in table identity_verification_tokens after drop column: unknown column "selfservice_verification_flow_id_" in foreign key definition`

aeneasr: Oh, I see ... damn :(

persistence/sql/migratest/testdata/20220902141902_testdata.sql
jonas-jonas: You mean, so that we can see, that the migration actually works on existing data? We have initial data from when this feature was initially added [here](https://github.com/ory/kratos/blob/db8c345f52670c8a2ec2181e62c638c063e73a7a/persistence/sql/migratest/testdata/20200830154602_testdata.sql). Should I add another one just in case?

aeneasr: Nope, thank you for the reference - that is enough!

persistence/sql/migrations/sql/20220825134336000001_not_null_constraint_verification_token_flow_id.sqlite3.up.sql
```diff
@@ -0,0 +1,37 @@
...
+DROP INDEX identity_verification_tokens_id_nid_idx;
+DROP INDEX identity_verification_tokens_nid_id_idx;
+DROP INDEX identity_verification_tokens_token_nid_used_idx;
+DROP INDEX identity_verification_tokens_token_uq_idx;
+DROP INDEX identity_verification_tokens_verifiable_address_id_idx;
+DROP INDEX identity_verification_tokens_verification_flow_id_idx;
+
+CREATE INDEX identity_verification_tokens_id_nid_idx ON identity_verification_tokens (id, nid);
+CREATE INDEX identity_verification_tokens_nid_id_idx ON identity_verification_tokens (nid, id);
+CREATE INDEX identity_verification_tokens_token_nid_used_flow_id_idx ON identity_verification_tokens (nid, token, used, selfservice_verification_flow_id);
```
aeneasr, line 29: This index has changed for SQLite, but the other migrations did not update this index. Is that intentional and if so why? I see that one query has changed to always use the key, so we probably need to add the missing index for the other databases too?

selfservice/strategy/link/test/persistence.go
```diff
@@ -117,44 +112,44 @@ func TestPersister(ctx context.Context, conf *config.Config, p interface {
...
 				require.Error(t, err)
 			})
 
 			t.Run("case=should create a new verification token", func(t *testing.T) {
-				token := newVerificationToken(t, "foo-user@ory.sh")
+				_, token := newVerificationToken(t, "foo-user@ory.sh")
 				require.NoError(t, p.CreateVerificationToken(ctx, token))
 			})
 
 			t.Run("case=should create a verification token and use it", func(t *testing.T) {
```
aeneasr, line 134: Please add a test which wants to use a verification token that exists, but with the incorrect flow ID. That should return an error!

aeneasr (changes requested):
Awesome, almost there!

persistence/sql/migrations/sql/20220825134336000001_not_null_constraint_verification_token_flow_id.sqlite3.up.sql
jonas-jonas: True, missed that - sorry.

selfservice/strategy/link/test/persistence.go
jonas-jonas: Done.

aeneasr (approved):
Thank you, it's perfect now :)

PR #2699, opened 2022-08-26
fix: do not invalidate recovery addr on update

### LINKED ISSUE #2433: Admin API identity update invalidates recovery link
### Preflight checklist

- [ ] #2646

### Describe the bug

Hey,

I encountered a problem: when an identity gets updated (via admin endpoint) during account recovery,
the recovery flow get invalidated. Is this working as expected?

### Reproducing the bug

1. Create an identity via admin endpoint
2. Initiate self service recovery flow so an email with recovery link gets sent out
3. Update same identity via admin endpoint with no changes (same traits, state == active, same schema)
4. Follow the recovery link received in email
5. Get a flow response that the recovery token is invalid

### Relevant log output

_No response_

### Relevant configuration

```yml
identity schema:

{
  "$id": "identity0",
  "title": "Test",
  "type": "object",
  "properties": {
    "traits": {
      "type": "object",
      "properties": {
        "email": {
          "type": "string",
          "format": "email",
          "title": "E-Mail",
          "ory.sh/kratos": {
            "credentials": {
              "password": {
                "identifier": true
              }
            },
            "recovery": {
              "via": "email"
            },
            "verification": {
              "via": "email"
            }
          }
        }
      },
      "required": [
        "email"
      ],
      "additionalProperties": false
    }
  }
}
```

### Version

v0.9.0-alpha.3

Linux

Docker

### Additional Context

If there is no identity update via admin endpoint between the recovery link getting sent out and clicked, the whole recovery and password reset flow works correctly.

If this is a problem because the admin endpoint updates the whole identity, maybe the recovery link could stay valid if there are no changes in the update.

hperl:
This PR modifies persister.UpdateIdentity to only update the recovery addresses that changed, leaving the unchanged once (and their associated recovery tokens) valid.

Shout-out to @thcyron for the analysis.

Fixes #2433

---

## TODO

- [x] Is everything Go 1.18 ready? (CI, Docker, ...)

### DIFF
--- a/persistence/sql/persister_identity.go
+++ b/persistence/sql/persister_identity.go
@@ -208,7 +208,7 @@ func updateAssociation[T interface {
 		Order("id ASC").
 		All(&inDB); err != nil {
 
-		return err
+		return sqlcon.HandleError(err)
 	}
 
 	newAssocs := make(map[string]*T)
@@ -227,7 +227,7 @@ func updateAssociation[T interface {
 			newAssocs[h] = nil // Ignore associations that are already in the db.
 		} else {
 			if err := p.GetConnection(ctx).Destroy(a); err != nil {
-				return err
+				return sqlcon.HandleError(err)
 			}
 		}
 	}
@@ -235,7 +235,7 @@ func updateAssociation[T interface {
 	for _, a := range newAssocs {
 		if a != nil {
 			if err := p.GetConnection(ctx).Create(a); err != nil {
-				return err
+				return sqlcon.HandleError(err)
 			}
 		}
 	}
@@ -426,7 +426,7 @@ func (p *Persister) UpdateIdentity(ctx context.Context, i *identity.Identity) er
 				new(identity.Credentials).TableName(ctx)),
 			i.ID, p.NetworkID(ctx)).Exec(); err != nil {
 
-			return err
+			return sqlcon.HandleError(err)
 		}
 
 		if err := p.update(WithTransaction(ctx, tx), i); err != nil {
--- a/persistence/sql/persister_identity.go
+++ b/persistence/sql/persister_identity.go
@@ -202,6 +202,43 @@ func (p *Persister) createVerifiableAddresses(ctx context.Context, i *identity.I
 	return nil
 }
 
+func (p *Persister) updateRecoveryAddresses(ctx context.Context, i *identity.Identity) error {
+	var addressesInDb []identity.RecoveryAddress
+	if err := p.GetConnection(ctx).Where("identity_id = ? AND nid = ?", i.ID, p.NetworkID(ctx)).Order("id ASC").All(&addressesInDb); err != nil {
+		return err
+	}
+
+	newAddresses := make(map[string]*identity.RecoveryAddress)
+	oldAddresses := make(map[string]*identity.RecoveryAddress)
+	for j, a := range i.RecoveryAddresses {
+		i.RecoveryAddresses[j].IdentityID = i.ID
+		i.RecoveryAddresses[j].NID = p.NetworkID(ctx)
+		i.RecoveryAddresses[j].Value = stringToLowerTrim(i.RecoveryAddresses[j].Value)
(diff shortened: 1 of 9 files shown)


persistence/sql/persister_identity.go
```diff
@@ -202,6 +202,43 @@ func (p *Persister) createVerifiableAddresses(ctx context.Context, i *identity.I
 	return nil
 }
 
+func (p *Persister) updateRecoveryAddresses(ctx context.Context, i *identity.Identity) error {
```
aeneasr, line 205: This probably needs some tests to ensure the diff is applied correctly :)

aeneasr (changes requested):
Thank you for the quick turnaround! We have the problem both in recovery as well as verification :) It would also be awesome if we could write a small test to ensure that the diff is working correctly. Speaking about this, it's not uncommon that we need to check whether a `1:n` relationship has changed in Ory Kratos (same goes for credentials for example) so maybe we can generalize this problem using a (potentially generic) diff function to return the IDs that need to be removed and inserted?

persistence/sql/persister_identity.go
hperl: Thanks, makes sense. I only tested for the actual bug for now, but I can extend this to all relationships.

hperl: I just looked at the existing tests, and we already test that `UpdateIdentity` propagates updates to the recovery address here:

https://github.com/ory/kratos/blob/a059ef079ceffdec6364e174e782b171cbbdc25b/identity/test/pool.go#L769-L803

aeneasr: 👍 perfect

aeneasr (dismissed):

persistence/sql/persister_identity.go
```diff
@@ -192,24 +192,83 @@ func (p *Persister) createVerifiableAddresses(ctx context.Context, i *identity.I
...
+	}
+
+	// Subtle: we delete the old associations from the DB first, because else
+	// they could cause UNIQUE constraints to fail on insert.
+	for h, a := range oldAssocs {
+		if _, found := newAssocs[h]; found {
+			newAssocs[h] = nil // Ignore associations that are already in the db.
+		} else {
+			if err := p.GetConnection(ctx).Destroy(a); err != nil {
+				return err
```
aeneasr, line 230: Please use `sqlcon.HandleError()`

persistence/sql/persister_identity.go
```diff
@@ -192,24 +192,83 @@ func (p *Persister) createVerifiableAddresses(ctx context.Context, i *identity.I
...
+			}
+		}
+	}
+
+	for _, a := range newAssocs {
+		if a != nil {
+			if err := p.GetConnection(ctx).Create(a); err != nil {
+				return err
```
aeneasr, line 238: Please use `sqlcon.HandleError()`

persistence/sql/persister_identity.go
```diff
@@ -350,27 +411,25 @@ func (p *Persister) UpdateIdentity(ctx context.Context, i *identity.Identity) er
...
 
-		if err := p.createVerifiableAddresses(ctx, i); err != nil {
+		/* #nosec G201 TableName is static */
+		if err := tx.RawQuery(
+			fmt.Sprintf(
+				`DELETE FROM %s WHERE identity_id = ? AND nid = ?`,
+				new(identity.Credentials).TableName(ctx)),
+			i.ID, p.NetworkID(ctx)).Exec(); err != nil {
+
 			return err
```
aeneasr, line 429: `sqlcon.HandleError()`

hperl:
@aeneasr all review comments addressed. As far as the test for the verified address go, the suite is very thorough:

https://github.com/ory/kratos/blob/bece07609f1d46f8e24da4f51468edbb4a8847b3/identity/test/pool.go#L555-L719

aeneasr (approved):
Great job :)

Issue #2091, opened 2021-12-30
Store additional metadata with sessions


joekrill:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

There isn't a great way to distinguish between multiple sessions beyond the date it was created (and perhaps the authentication method). Once #655 (Admin/Selfservice session management #655 ) is implemented, we should be able to show user's a list of their sessions and allow them to revoke individual sessions. Most implementations will show additional information about a session, like the IP address and user agent/browser that was used. But Kratos currently does not store this information alongside each session. 

For example this is what Gmail shows me when I view a list of active sessions:
<img width="867" alt="Screen Shot 2021-12-30 at 12 45 12 PM" src="https://user-images.githubusercontent.com/393220/147776936-6dcef6df-3ff2-43ce-ab2d-fd3c6677fe68.png">

### Describe your ideal solution

Ideally when a session is created (or extended?) at least the IP address and User Agent would be captured and stored. The IP address could add some complications because of proxies and request forwarding. 

Perhaps there is just a metadata field that can be populated using jsonnet? Maybe a hook could allow customizing what sort of metadata is stored?

It would also be ideal if we could provide a physical location. Something like the [IP2Location](https://lite.ip2location.com/ip2location-lite) databases could be utilized. I'm not sure if this is something that would be out of scope for Kratos, though. It could certainly be handled outside of that.

### Workarounds or alternatives

It might be possible to use a web hook that then stores this information externally, since a web hook can include the request headers. But it doesn't seem that you can ascertain a particular session ID from the payload of a webhook -- only the current identity is available? So I'm not sure this is a real great workaround.

### Version

0.8.2-alpha.1

aeneasr:
[...]

PR #2715, opened 2022-09-06
feat: adding device information to the session

kelkarajay:
This PR introduces saving device information where the session is in use - IP Address, UA, Location (if available)

The priority sequence for Client IP derivation is -

- Cloudflare injected header - True-Client-IP
- Real IP header injected by reverse proxies like nginx X-Real-IP
- X-Forwarded-For header that's a comma separated list with each node's host IP in the request chain
- If none of the above work, fallback to Go's request RemoteAddr

## Related issue(s)
- https://github.com/ory-corp/cloud/issues/3011
- https://github.com/ory/kratos/issues/2091

## Checklist


internal/testhelpers/handler_mock.go
```diff
@@ -39,7 +39,7 @@ func MockSetSession(t *testing.T, reg mockDeps, conf *config.Config) httprouter.
 
 func MockSetSessionWithIdentity(t *testing.T, reg mockDeps, conf *config.Config, i *identity.Identity) httprouter.Handle {
 	return func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
-		activeSession, _ := session.NewActiveSession(r.Context(), i, conf, time.Now().UTC(), identity.CredentialsTypePassword, identity.AuthenticatorAssuranceLevel1)
+		activeSession, _ := session.NewActiveSession(r.Context(), r.Header, i, conf, time.Now().UTC(), identity.CredentialsTypePassword, identity.AuthenticatorAssuranceLevel1)
```
aeneasr, line 42: Maybe just pass the request as a whole? That would simplify the function (we use both context and header).

session/session.go
```diff
@@ -179,6 +190,22 @@ func (s *Session) Activate(ctx context.Context, i *identity.Identity, c lifespan
 	s.Identity = i
 	s.IdentityID = i.ID
 
+	agent := requestHeaders["User-Agent"]
```
aeneasr, line 193: It would be better to use `Get` here, no?

session/session.go
```diff
@@ -179,6 +190,22 @@ func (s *Session) Activate(ctx context.Context, i *identity.Identity, c lifespan
...
+		s.UserAgent = strings.Join(agent, " ")
+	}
+
+	clientIP := requestHeaders.Get("True-Client-IP")
+	if clientIP != "" {
+		s.ClientIPAddress = clientIP
+	} else {
+		// TODO: Use x lib implementation to parse client IP address from the header string
+		s.ClientIPAddress = requestHeaders.Get("X-Forwarded-For")
+	}
```
aeneasr, line 204: What about the "ClientIP" field in the request?

session/session.go
kelkarajay: yes, although `User-Agent` header has multiple values since it appears to be broken into an array of strings delimited by the whitespace char. As a result, `Get` would get only the first bit of the UA

internal/testhelpers/handler_mock.go
kelkarajay: Sounds good, I'll change that

session/session.go
kelkarajay: Corrected the sequence by which the IP is derived

persistence/sql/migrations/sql/20220907132836000000_sessions_add_client_fields.cockroach.down.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE sessions
```
aeneasr, line 1: If many migrations share the same code, you can also create "global" migrations and then override only for databases that have a different DSL:

```patch
- persistence/sql/migrations/sql/20220907132836000000_sessions_add_client_fields.cockroach.down.sql
+ persistence/sql/migrations/sql/20220907132836000000_sessions_add_client_fields.down.sql
```

session/session.go
```diff
@@ -167,18 +181,45 @@ func NewInactiveSession() *Session {
...
 
 	s.Active = true
-	s.ExpiresAt = authenticatedAt.Add(c.SessionLifespan(ctx))
+	s.ExpiresAt = authenticatedAt.Add(c.SessionLifespan(r.Context()))
 	s.AuthenticatedAt = authenticatedAt
 	s.IssuedAt = authenticatedAt
 	s.Identity = i
 	s.IdentityID = i.ID
 
+	agent := r.Header["User-Agent"]
```
aeneasr, line 196: So this code sets the session information *at the time of session activation*, so when the session is usually issued.

The naming however does not indicate that, as it just says `client_ip_address`. In the original issue we discuss an idea for how to track a device over it's lifetime:

https://github.com/ory/kratos/issues/2091#issuecomment-1003414305

At the least I think these fields should be arrays! We don't have to track these values over their lifetime _just yet_ and could just add the IP during login, but the API should be including this use case which we might add in the future!

aeneasr (changes requested):

session/session.go
kelkarajay: Sounds good, I've tweaked the design a bit. We should be able to track list of the combo of UserAgent (client) - IP - Location. The model looks like below which is then embedded as a list in the Session.  

```go
type Metadata struct {
	ID uuid.UUID `json:"id" faker:"-" db:"id"`
	SessionID uuid.UUID `json:"-" faker:"-" db:"session_id"`
	IPAddress *string `json:"ip_address" db:"ip_address"`
	UserAgent *string `json:"user_agent" db:"user_agent"`
	Location *string `json:"location" db:"location"`
	CreatedAt time.Time `json:"seen_at" faker:"-" db:"created_at"`
        LastSeen time.Time `json:"last_seen" faker:"-" db:"last_seen"`
	NID uuid.UUID `json:"-"  faker:"-" db:"nid"`
}
```

```go
	// Metadata has history of all clients where the session was used
	Metadata []Metadata `json:"metadata" faker:"-" db:"-"`
```

The metadata records are inserted in a new table. The PR still needs work.

TODO - 

- [ ] `Upsert` should save metadata records
- [ ] `SaveSessionMetadata` should update the last seen only on metadata record that already exists for the same IP.

persistence/sql/migratest/fixtures/session/7458af86-c1d8-401c-978a-8da89133f98b.json
```diff
@@ -0,0 +1,53 @@
...
+    "updated_at": "2013-10-07T08:23:19Z"
+  },
+  "metadata": [
+    {
+      "id": "884f556e-eb3a-4b9f-bee3-11763642c6c0",
+      "ip_address": "54.155.246.232",
+      "user_agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36",
+      "location": "Munich, Germany",
+      "seen_at": "2022-08-07T08:23:19Z",
+      "last_seen": "2022-08-09T08:35:19Z"
```
aeneasr, line 50: Last seen would implicate that we need to update the session on every whoami call. That makes the calls very write heavy and can clog up the DB at scale. I think we should avoid this for now!

persistence/sql/migratest/fixtures/session/7458af86-c1d8-401c-978a-8da89133f98b.json
```diff
@@ -0,0 +1,53 @@
...
+        "status": "pending",
+        "created_at": "2013-10-07T08:23:19Z",
+        "updated_at": "2013-10-07T08:23:19Z"
+      }
+    ],
+    "metadata_public": null,
+    "created_at": "2013-10-07T08:23:19Z",
+    "updated_at": "2013-10-07T08:23:19Z"
+  },
+  "metadata": [
```
aeneasr, line 43: What about `devices` or `clients` like in the original proposal? Metadata is very generic and I expect that we will have user-specified metadata in the future, so I don't think that this key name is a good fit for the functionality.

persistence/sql/migratest/fixtures/session/7458af86-c1d8-401c-978a-8da89133f98b.json
```diff
@@ -0,0 +1,53 @@
...
+    {
+      "id": "884f556e-eb3a-4b9f-bee3-11763642c6c0",
+      "ip_address": "54.155.246.232",
+      "user_agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36",
```
aeneasr, line 47: Typically, one user agent would have multiple IPs as you sign in to a browser, and then use that browser across networks. I'm wondering if that layout is a good fit for that use case.

aeneasr (changes requested):
Thank you, it's an improvement to the previous iteration! Unfortunately I still have doubts that the data model will work at scale / long-term. Can you please take a look at the comments?

Before doing any additional coding work, please post a comment with the proposed data layout (feel free to ping me in Slack) that you want to implement so that we can get it right before spending a lot of coding time. Thanks! :)

In the future, we should also stick to our design document process ( https://github.com/ory/kratos/issues/new?assignees=&labels=rfc&template=DESIGN-DOC.yml ) when foundational API and data design is being done.

session/session.go
aeneasr: Can you please address the following problems in the design:

1. Last seen is very hard to implement at scale, I would remove it for now
2. The session will get a metadata field which can be used to store arbitrary data (eg campaign source=facebook) Please rename the field to e.g device or agent or client
3. One user agent can have multiple IPs and locations, but one session usually has only one user agent (we should still allow multiple though)

kelkarajay:
@aeneasr First the minor stuff, 

> What about devices or clients like in the original proposal? Metadata is very generic and I expect that we will have user-specified metadata in the future, so I don't think that this key name is a good fit for the functionality.

_I'll rename the metadata struct to Client. The initial idea was to call it a Log and the DB table session_logs (https://github.com/ory/kratos/pull/2715/commits/5bb40d89cf442cd70878b79c84aaee6e811915af). This however could cause confusion in terms of how people interpreted it and I renamed it after discussion with @zepatrik._ 

> The model looks like below which is then embedded as a list in the Session.
> 
> ```go
> type Metadata struct {
>	ID uuid.UUID `json:"id" faker:"-" db:"id"`
>	SessionID uuid.UUID `json:"-" faker:"-" db:"session_id"`
>	IPAddress *string `json:"ip_address" db:"ip_address"`
>	UserAgent *string `json:"user_agent" db:"user_agent"`
>	Location *string `json:"location" db:"location"`
>	CreatedAt time.Time `json:"seen_at" faker:"-" db:"created_at"`
>   LastSeen time.Time `json:"last_seen" faker:"-" db:"last_seen"`
>	NID uuid.UUID `json:"-"  faker:"-" db:"nid"`
>}
>```
>
>```go
>	// Metadata has history of all clients where the session was used
>	Metadata []Metadata `json:"metadata" faker:"-" db:"-"`
>```	
>
> The metadata records are inserted in a new table.

_The core idea is that it's a 1-X relationship. One session can have multiple Client/logs based on IP address assuming its a machine that is probably switching between a VPN or direct connection or perhaps different locations like coffee shops/home. The unique index for records would be SessionID + IP. (The User Agent is an unreliable piece of information to uniquely identify a device since it changes when the browser is updated for example.)_

> Last seen would implicate that we need to update the session on every whoami call. That makes the calls very write heavy and can clog up the DB at scale. I think we should avoid this for now!

_The last seen field field could be updated only if it exceeds a previous duration threshold. This could be done to save on the number of writes to the database. For example, Github doesn't give you the exact timestamp, rather that the token was used within the last week or two weeks._

Bildschirmfoto 2022-09-12 um 12 09 52


_The current model helps represent all the IP's the session has been active at. An example below is what could be represented using the current model -_

1. _Session being used from 2 IP's. Could be VPN that disconnects on a machine._
2. _Another session probably used months ago_

| Session | IP Address | User Agent | Last Seen |
| --- | --- | --- | --- |
| ABCSession1 | 54.12.145.12 (Munich, Germany) |  Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36 | 5 July 2022 |
| ABCSession1 | 25.12.145.12 (Dresden, Germany) |  Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36 | 1 September 2022 |
| WERSession3 | 86.123.211.123 (Leipzig, Germany) | Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36 | 5 April 2021 |
| RTESession2 | 86.123.211.123 (FFM, Germany) | Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36 | 5 December 2020 |


----

> Data cleanup

_A cascade delete would help keep the records clean with history being removed along with an expired session record._

Thoughts?

persistence/sql/persister_session.go
```diff
@@ -36,10 +36,29 @@
 		return nil, err
 	}
 
+	devices, err := p.GetSessionDevices(ctx, sid)
```
aeneasr, line 39: It makes probably more sense to use `.Eager()` for this. For the identity we don't use Eager() because it has some custom validation built in.

You can pass arguments to eager which say what relation you want to be hydrated. Saves you some code :)

persistence/sql/persister_session.go
```diff
@@ -66,7 +85,13 @@
 				return err
 			}
 
+			devices, err := p.GetSessionDevices(ctx, s.ID)
```
aeneasr, line 88: Same here with eager

aeneasr (commented):
Perfect, just two things that could be improved in the SQL logic!

persistence/sql/migrations/sql/20220913091948000000_sessions_devices_add_index.postgres.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX "session_devices_session_id_nid_idx" ON "session_devices" (session_id, nid);
```
aeneasr, line 1: I think the index down migration (all db versions) should remove the index if possible, or the command should be `CREATE INDEX IF NOT EXIST`. Otherwise it will not be possible to roll back this migration and apply it again

persistence/sql/persister_session.go
```diff
@@ -82,15 +83,34 @@ func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) error
...
 	s.NID = p.NetworkID(ctx)
 
-	if err := p.Connection(ctx).Find(new(session.Session), s.ID); errors.Is(err, sql.ErrNoRows) {
-		// This must not be eager or identities will be created / updated
-		return errors.WithStack(p.GetConnection(ctx).Create(s))
-	} else if err != nil {
-		return errors.WithStack(err)
-	}
+	return errors.WithStack(p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) error {
+		if err := tx.Find(new(session.Session), s.ID); errors.Is(err, sql.ErrNoRows) {
```
aeneasr, line 87: This needs to respect `nid`, it currently doesn't.

Can you also please use `sqlcon.HandleError` and `sqlcon.ErrNotFound()`

persistence/sql/persister_session.go
```diff
@@ -82,15 +83,34 @@ func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) error
...
+			// This must not be eager or identities will be created / updated
+			err = errors.WithStack(tx.Create(s))
+			if err != nil {
+				return err
+			}
```
aeneasr, line 92:
```suggestion
			if err := sqlcon.HandleError(tx.Create(s)); err != nil {
				return err
			}
```

persistence/sql/persister_session.go
```diff
@@ -82,15 +83,34 @@ func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) error
...
+			for i := range s.Devices {
+				device := &(s.Devices[i])
+				device.SessionID = s.ID
+				device.NID = s.NID
 
-	// This must not be eager or identities will be created / updated
-	return p.GetConnection(ctx).Update(s)
+				if err := tx.Create(device); err != nil {
+					return err
+				}
```
aeneasr, line 101: Wrap with `sqlcon.HandleError()`

persistence/sql/persister_session.go
```diff
@@ -82,15 +83,34 @@ func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) error
...
+				s.Devices[i] = *device
+			}
+
+			return nil
+		} else if err != nil {
+			return errors.WithStack(err)
+		}
+
+		// This must not be eager or identities will be created / updated
+		return tx.Update(s)
```
aeneasr, line 112: So this won't update devices if I read this correctly. If that's the case, plase document it in the code doc comment(s) of the method / interface and here!

aeneasr (changes requested):
minor changes to match Ory's code architecture, then it's good to go

persistence/sql/migrations/sql/20220913091948000000_sessions_devices_add_index.postgres.up.sql
kelkarajay: Sure. I'll fix it. My thought was that the migration preceding this one is to drop the table considering that it's shipped together

kelkarajay: Added a migration to drop the index (using IF EXISTS where possible)

persistence/sql/persister_session.go
kelkarajay: Added notes to the func

persistence/sql/migrations/sql/20220913091948000000_sessions_devices_add_index.postgres.up.sql
kelkarajay: @aeneasr Scratch that. Just merged the index creation statement with the create table statement. The down migration simply drops the table, removing the need for "drop index" statements.

kelkarajay:
@aeneasr PR is ready for review now

zepatrik (changes requested):
Very nice, just found some minor stuff.
As is right now, we only add a device entry when we "activate" a session, right? This happens (if configured)

1. after login (incl. raising AAL)
2. after registration
3. after recovery

This means we don't keep track of location changes or when the session was last used at all. That would require (sporadic) writes as a side-effect of the whoami call. As that is a non-goal right now, it is important that the API makes it clear that it is not always updated. A `last_seen_at` timestamp, or list of devices in general, suggests we do keep that constantly updated.
This turned out to be a bit more complex than expected :sweat_smile: but let's find a good first iteration.

persistence/sql/migrations/sql/20220907132836000000_sessions_add_client_fields.cockroach.up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE "session_devices"
+(
+  "id"         UUID      NOT NULL,
+  PRIMARY KEY ("id"),
+  "ip_address" STRING DEFAULT '',
+  "user_agent" STRING DEFAULT '',
+  "location"   STRING DEFAULT '',
```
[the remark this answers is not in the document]
kelkarajay: Sounds good. I'll fix this and bump limits for user agent and location fields to 512 chars and add trim in the code.

session/session.go
```diff
@@ -167,26 +209,57 @@ func NewInactiveSession() *Session {
...
+	} else if realClientIP := r.Header.Get("X-Real-IP"); realClientIP != "" {
+		device.IPAddress = &realClientIP
+	} else if forwardedIP := r.Header["X-Forwarded-For"]; len(forwardedIP) != 0 {
+		ip, _ := httpx.GetClientIPAddress(forwardedIP, httpx.InternalIPSet)
+		device.IPAddress = &ip
+	} else {
+		device.IPAddress = &r.RemoteAddr
+	}
+
+	clientGeoLocation := []string{r.Header.Get("Cf-Ipcity"), r.Header.Get("Cf-Ipcountry")}
```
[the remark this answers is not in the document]
kelkarajay: We begin with supporting Cloudflare initially (Transform documented here - https://developers.cloudflare.com/rules/transform/managed-transforms/reference/).

A snippet from the link incase it 404's later - 

> Adds HTTP request headers with location information for the visitor's IP address. 
> The added headers are:
> - `cf-ipcity` The visitor's city (value from the ip.src.city field).
> - `cf-ipcountry` The visitor's country (value from the ip.src.country field).
> - `cf-ipcontinent` The visitor's continent (value from the ip.geoip.continent field).
> - `cf-iplongitude` The visitor's longitude (value from the ip.src.lon field).
> - `cf-iplatitude` The visitor's latitude (value from the ip.src.lat field).

Support for other location headers could be added incrementally (like the parser for IP address header in ory/x - https://github.com/ory/x/pull/593)

persistence/sql/migrations/sql/20220907132836000000_sessions_add_client_fields.mysql.up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE `session_devices`
+(
+  `id`         char(36) NOT NULL,
+  PRIMARY KEY (`id`),
+  `ip_address` VARCHAR(50)  DEFAULT '',
+  `user_agent` VARCHAR(255) DEFAULT '',
+  `location`   VARCHAR(255) DEFAULT '',
```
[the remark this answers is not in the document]
kelkarajay: Limits are bumped to 512. Pending trim

persistence/sql/migrations/sql/20220907132836000000_sessions_add_client_fields.cockroach.up.sql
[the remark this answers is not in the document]
kelkarajay: https://github.com/ory/kratos/commit/6129d3b99d003f144bb3e88e89be9ff4c4097acb - Trim added in the persister func prior to save.

kelkarajay:
Summary of what we just discussed (@zepatrik) - 

- The `last_seen` timestamp will not be exposed in the JSON representation in its current form
- `Eager` query for devices will be based on an option passed to the persister

Follow up issue for completing the updates for device last used timestamp and capturing changes in IP address, location from `whoami` requests

jonas-jonas (changes requested):
Just a few minor remarks. Apart from that, looks fantastic!

session/session.go
```diff
@@ -27,6 +32,40 @@ type refreshWindowProvider interface {
...
+	// UserAgent of the client
+	UserAgent *string `json:"user_agent" faker:"-" db:"user_agent"`
+
+	// Geo Location corresponding to the IP Address
+	Location *string `json:"location" faker:"ptr_geo_location" db:"location"`
+
+	// Time of capture
+	CreatedAt time.Time `json:"-" faker:"-" db:"created_at"`
+
+	// Last seen
```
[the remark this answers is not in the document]
kelkarajay: Will be sporadic updates imo but for now hidden until the implementation of the `whoami` updates https://github.com/ory/kratos/pull/2715#discussion_r974375516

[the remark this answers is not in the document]
jonas-jonas: Yes, but then this comment is not correct, right?

persistence/sql/persister_session.go
```diff
@@ -76,21 +90,50 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
...
+					device.UserAgent = stringsx.GetPointer(stringsx.TruncateByteLen(*device.UserAgent, SessionDeviceUserAgentMaxLength))
+				}
 
-	// This must not be eager or identities will be created / updated
-	return p.GetConnection(ctx).Update(s)
+				if err := sqlcon.HandleError(tx.Create(device)); err != nil {
+					return err
+				}
+
+				s.Devices[i] = *device
```
zepatrik, line 125: I find the pointer handling of `device` a bit confusing here :sweat_smile: 
So you first take the pointer of `s.Devices[i]` and then modify that struct a bit. But that would modify the original struct, so you don't really have to write it back here? Only if you where to do `for i, device := range s.Devices` you would create a copy of the struct and therefore need to write it back here?

persistence/sql/persister_session.go
```diff
@@ -76,21 +90,50 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
...
+			}
+
+			return nil
+		} else if err != nil {
```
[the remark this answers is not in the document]
zepatrik: Doing this here will not modify the original error, so you return it in the next line without handling again :wink: 
You would have to do
```suggestion
		} else if err = sqlcon.HandleError(err);  err != nil {
```
which does not help with readability at all :sweat_smile: 
Also, wouldn't `exists` be false (i.e. zero value of bool) in case there is an error, and therefore you incorrectly create the session first? Moving error handling so far from where the error is returned is always tricky and error prone.

session/handler.go
```diff
@@ -304,7 +304,7 @@ func (h *Handler) adminListIdentitySessions(w http.ResponseWriter, r *http.Reque
 	}
 
 	page, perPage := x.ParsePagination(r)
-	sess, err := h.r.SessionPersister().ListSessionsByIdentity(r.Context(), iID, active, page, perPage, uuid.Nil)
+	sess, err := h.r.SessionPersister().ListSessionsByIdentity(r.Context(), iID, active, page, perPage, uuid.Nil, ExpandEverything)
```
zepatrik, line 307: Can we add or adjust a test case to verify the devices are correctly returned here?

zepatrik (changes requested):
This is coming along very nicely :+1:
Just have some minor remarks and comments. With a few more tests (in case they are not already there and I missed them) it should be done soon :tada:

persistence/sql/persister_session.go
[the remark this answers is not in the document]
jonas-jonas: Yep, you're right. I wasn't sure either, but was really thrown off by the original solution of checking if the error is `sqlcon.ErrNoRows`. I guess it comes down to experience in the codebase, because we do that a lot in general, but I had to really think about what that `if` statement does, which IMO is not a good sign...

persistence/sql/persister_session.go
```diff
@@ -76,21 +90,50 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
...
 	s.NID = p.NetworkID(ctx)
 
-	if err := p.Connection(ctx).Find(new(session.Session), s.ID); errors.Is(err, sql.ErrNoRows) {
-		// This must not be eager or identities will be created / updated
-		return errors.WithStack(p.GetConnection(ctx).Create(s))
-	} else if err != nil {
-		return errors.WithStack(err)
-	}
+	return errors.WithStack(p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) error {
+		if err := sqlcon.HandleError(tx.Where("id = ? AND nid = ?", s.ID, s.NID).First(new(session.Session))); errors.Is(err, sqlcon.ErrNoRows) {
```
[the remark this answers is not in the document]
jonas-jonas: Maybe instead, we could do this. I know that it's hard to review this snippet, but maybe @kelkarajay can work with this? I find this a lot more readable. Sorry, for making such a fuzz about this, but as a relative newcomer to the codebase, I find the persister code hard to read as is, because of all the pop magic, and making it clearer is a win, right?

```go

	return errors.WithStack(p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) error {

		exists, err := tx.Where("id = ? AND nid = ?", s.ID, s.NID).Exists(new(session.Session))
		if err != nil {
		}

		if exists {
			// This must not be eager or identities will be created / updated
			// Only update session and not corresponding session device records
			return sqlcon.HandleError(tx.Update(s)) // should be correct?
		}

		// This must not be eager or identities will be created / updated
		if err := sqlcon.HandleError(tx.Create(s)); err != nil {
			return err
		}

		for i := range s.Devices {
			device := &(s.Devices[i])
			device.SessionID = s.ID
			device.NID = s.NID

			if device.Location != nil {
				device.Location = stringsx.GetPointer(stringsx.TruncateByteLen(*device.Location, SessionDeviceLocationMaxLength))
			}
			if device.UserAgent != nil {
				device.UserAgent = stringsx.GetPointer(stringsx.TruncateByteLen(*device.UserAgent, SessionDeviceUserAgentMaxLength))
			}

			if err := sqlcon.HandleError(tx.Create(device)); err != nil {
				return err
			}

			s.Devices[i] = *device
		}
		return nil

	}))
```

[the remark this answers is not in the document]
kelkarajay: It would have been nice to have the insert function like https://gorm.io/docs/create.html#Upsert-x2F-On-Conflict. Worth considering upstream contribution to pop, yeah. In the interim, the snippet above looks good tbh

session/session.go
```diff
@@ -27,6 +32,40 @@ type refreshWindowProvider interface {
...
+	// Device record ID
+	//
+	// required: true
+	ID uuid.UUID `json:"id" faker:"-" db:"id"`
+
+	// SessionID is a helper struct field for gobuffalo.pop.
+	SessionID uuid.UUID `json:"-" faker:"-" db:"session_id"`
+
+	// IPAddress of the client
+	IPAddress *string `json:"ip_address" faker:"ptr_ipv4" db:"ip_address"`
```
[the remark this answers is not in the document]
kelkarajay: Yep, should be nullable. I did consider that initially but saw not much value over the primitive type with a pointer

persistence/sql/persister_session.go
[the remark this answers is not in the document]
kelkarajay: Went with this ^

persistence/sql/migrations/sql/20220907132836000000_sessions_add_client_fields.postgres.up.sql
```diff
@@ -0,0 +1,16 @@
...
+  "id"         UUID      NOT NULL,
+  PRIMARY KEY ("id"),
+  "ip_address" VARCHAR(50)  DEFAULT '',
+  "user_agent" VARCHAR(512) DEFAULT '',
+  "location"   VARCHAR(512) DEFAULT '',
+  "nid"        UUID      NOT NULL,
+  "session_id" UUID      NOT NULL,
+  "created_at" timestamp NOT NULL,
+  "updated_at" timestamp NOT NULL,
+  FOREIGN KEY ("session_id") REFERENCES "sessions" ("id") ON DELETE cascade,
```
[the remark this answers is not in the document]
kelkarajay: Good point, I'll add naming while also condensing migrations for multiple db flavours where possible as Patrik suggested

internal/testhelpers/identity.go
```diff
@@ -14,11 +14,11 @@ import (
 )
 
 func CreateSession(t *testing.T, reg driver.Registry) *session.Session {
-	ctx := context.Background()
+	req := httptest.NewRequest("GET", "/sessions/whoami", nil)
```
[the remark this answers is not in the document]
kelkarajay: Fixed

session/session.go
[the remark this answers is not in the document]
kelkarajay: Corrected to - `Last Updated at`.

persistence/sql/migrations/sql/20220907132836000000_sessions_add_client_fields.cockroach.up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE "session_devices"
```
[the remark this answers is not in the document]
kelkarajay: Done

session/persistence.go
```diff
@@ -160,9 +160,9 @@ func TestPersister(ctx context.Context, conf *config.Config, p interface {
 			require.NoError(t, p.UpsertSession(ctx, &expected2))
 
 			require.NoError(t, p.DeleteSessionsByIdentity(ctx, expected2.IdentityID))
-			_, err := p.GetSession(ctx, expected1.ID)
+			_, err := p.GetSession(ctx, expected1.ID, ExpandNothing)
 			require.Error(t, err)
-			_, err = p.GetSession(ctx, expected2.ID)
+			_, err = p.GetSession(ctx, expected2.ID, ExpandNothing)
```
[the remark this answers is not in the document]
kelkarajay: Test exists here - https://github.com/ory/kratos/blob/7713be7e2c34fb45acf1748819af6a1fe636b7b8/session/test/persistence.go#L90

session/handler.go
kelkarajay: Added devices and assertion to the existing test - https://github.com/ory/kratos/blob/7713be7e2c34fb45acf1748819af6a1fe636b7b8/session/test/persistence.go#L188

persistence/sql/persister_session.go
kelkarajay: you're right. I removed the overwrite

aeneasr:
Could you PTAL at the merge conflict? :)

persistence/sql/migrations/sql/20220907132836000000_add_session_devices_table.mysql.up.sql
```diff
@@ -0,0 +1,16 @@
...
+  `location`   VARCHAR(512) DEFAULT '',
+  `session_id` char(36) NOT NULL,
+  `nid`        char(36) NOT NULL,
+  `created_at` DATETIME NOT NULL,
+  `updated_at` DATETIME NOT NULL,
+  FOREIGN KEY (`session_id`) REFERENCES `sessions` (`id`) ON DELETE cascade,
+  FOREIGN KEY (`nid`) REFERENCES `networks` (`id`) ON DELETE cascade,
+  CONSTRAINT unique_session_device UNIQUE (nid, session_id, ip_address, user_agent)
+) ENGINE = InnoDB;
+CREATE INDEX `session_devices_session_id_nid_idx` ON `session_devices` (`session_id`, `nid`);
```
aeneasr, line 16: I suggest to add an index over `id, nid` as well

persistence/sql/migrations/sql/20220907132836000000_add_session_devices_table.up.sql
```diff
@@ -0,0 +1,15 @@
...
+  "location"   VARCHAR(512) DEFAULT '',
+  "nid"        UUID             NOT NULL,
+  "session_id" UUID             NOT NULL,
+  "created_at" timestamp        NOT NULL,
+  "updated_at" timestamp        NOT NULL,
+  CONSTRAINT "session_metadata_sessions_id_fk" FOREIGN KEY ("session_id") REFERENCES "sessions" ("id") ON DELETE cascade,
+  CONSTRAINT "session_metadata_nid_fk" FOREIGN KEY ("nid") REFERENCES "networks" ("id") ON DELETE cascade,
+  CONSTRAINT unique_session_device UNIQUE (nid, session_id, ip_address, user_agent)
+);
+CREATE INDEX "session_devices_session_id_nid_idx" ON "session_devices" (session_id, nid);
```
aeneasr, line 15: I suggest to add an index over `id, nid` as well

persistence/sql/persister_session.go
```diff
@@ -76,21 +90,51 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
...
 	s.NID = p.NetworkID(ctx)
 
-	if err := p.Connection(ctx).Find(new(session.Session), s.ID); errors.Is(err, sql.ErrNoRows) {
+	return errors.WithStack(p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) error {
+		exists, err := tx.Where("id = ? AND nid = ?", s.ID, s.NID).Exists(new(session.Session))
+		if err != nil {
+			return sqlcon.HandleError(err)
+		}
+
+		if exists {
```
aeneasr, line 108: So upsert will ignore any device updates, is that intentional?

persistence/sql/persister_session.go
```diff
@@ -76,21 +90,51 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
...
+			if device.Location != nil {
+				device.Location = stringsx.GetPointer(stringsx.TruncateByteLen(*device.Location, SessionDeviceLocationMaxLength))
+			}
+			if device.UserAgent != nil {
+				device.UserAgent = stringsx.GetPointer(stringsx.TruncateByteLen(*device.UserAgent, SessionDeviceUserAgentMaxLength))
+			}
+
+			if err := sqlcon.HandleError(tx.Create(device)); err != nil {
+				return err
+			}
```
aeneasr, line 133: It might make sense to put this into a dedicated function such as `AddDeviceToSession()`

session/expand.go
```diff
@@ -0,0 +1,43 @@
...
+const (
+	ExpandSessionDevices Expandable = "Devices"
+)
+
+// ExpandNothing expands nothing
+var ExpandNothing []Expandable
+
+// ExpandEverything expands all the fields of a project.
+var ExpandEverything = Expandables{
+	ExpandSessionDevices,
```
aeneasr, line 42: I would expect that this also expands the identity and it's sub-fields.

session/expand.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	return false
+}
+
+const (
+	ExpandSessionDevices Expandable = "Devices"
+)
+
+// ExpandNothing expands nothing
+var ExpandNothing []Expandable
```
aeneasr, line 38: I would expect this to not expand the identity

session/session.go
```diff
@@ -167,26 +209,57 @@ func NewInactiveSession() *Session {
...
+
+	clientGeoLocation := []string{r.Header.Get("Cf-Ipcity"), r.Header.Get("Cf-Ipcountry")}
+
+	var sb strings.Builder
+	for _, i := range clientGeoLocation {
+		if sb.Len() != 0 && i != "" {
+			sb.WriteString(", ")
+		}
+		sb.WriteString(i)
+	}
```
aeneasr, line 259: Why not `strings.Join()`?

aeneasr (changes requested):
Great job! Just a few minor comments left, then we're good to ship!

persistence/sql/migrations/sql/20220907132836000000_add_session_devices_table.mysql.up.sql
kelkarajay: Added

persistence/sql/persister_session.go
kelkarajay: Yes, updating `last_seen` for a device will be on a single row and will not update the session. Documented - 

```go
// UpsertSession creates a session if not found else updates.
// This operation also inserts Session device records when a session is being created.
// The update operation skips updating Session device records since only one record would need to be updated in this case.
func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) error {
```

session/expand.go
kelkarajay: @aeneasr I didn't add `Identity` as an Expandable due to this comment in the code -

```go
	// This is needed because of how identities are fetched from the store (if we use eager not all fields are
	// available!).
	i, err := p.GetIdentity(ctx, s.IdentityID)
```

session/expand.go
kelkarajay: I'll fix and not query `Identity` when expandables slice is empty

aeneasr: 👍

session/expand.go
kelkarajay: Added identity to the `everything` set and excluded from the `ToEager` output. Wrote tests to cover this

session/session.go
kelkarajay: Fixed

aeneasr (approved):
Awesome! If we move https://github.com/ory/kratos/pull/2715/files#diff-b496b7e4f89db676e41b419d69c98c5b1336cf9b6029e0d819e07be0a79cea28R124 to a separate function (can be private for now!) we can merge this :)

PR #2725, opened 2022-09-11
feat: support importing MD5 hashes

erolkskn:
## Related issue(s)
#2716 

## Checklist

### DIFF
--- a/hash/hash_comparator.go
+++ b/hash/hash_comparator.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/aes"
 	"crypto/cipher"
+	"crypto/md5" // #nosec G501
 	"crypto/subtle"
 	"encoding/base64"
 	"fmt"
@@ -38,6 +39,8 @@ func Compare(ctx context.Context, password []byte, hash []byte) error {
 		return CompareScrypt(ctx, password, hash)
 	case IsFirebaseScryptHash(hash):
 		return CompareFirebaseScrypt(ctx, password, hash)
+	case IsMD5Hash(hash):
+		return CompareMD5(ctx, password, hash)
 	default:
 		return errors.WithStack(ErrUnknownHashAlgorithm)
 	}
@@ -173,6 +176,30 @@ func CompareFirebaseScrypt(_ context.Context, password []byte, hash []byte) erro
 	return errors.WithStack(ErrMismatchedHashAndPassword)
 }
 
+func CompareMD5(_ context.Context, password []byte, hash []byte) error {
+	// Extract the hash from the encoded password
+	pf, salt, hash, err := decodeMD5Hash(string(hash))
+	if err != nil {
+		return err
+	}
+
+	arg := password
+	if salt != nil {
+		r := strings.NewReplacer("{SALT}", string(salt), "{PASSWORD}", string(password))
+		arg = []byte(r.Replace(string(pf)))
+	}
+	// #nosec G401
+	otherHash := md5.Sum(arg)
+
+	// Check that the contents of the hashed passwords are identical. Note
+	// that we are using the subtle.ConstantTimeCompare() function for this
+	// to help prevent timing attacks.
+	if subtle.ConstantTimeCompare(hash, otherHash[:]) == 1 {
+		return nil
+	}
+	return errors.WithStack(ErrMismatchedHashAndPassword)
+}
+
 var (
 	isBcryptHash         = regexp.MustCompile(`^\$2[abzy]?\$`)
 	isArgon2idHash       = regexp.MustCompile(`^\$argon2id\$`)
@@ -180,6 +207,7 @@ var (
 	isPbkdf2Hash         = regexp.MustCompile(`^\$pbkdf2-sha[0-9]{1,3}\$`)
 	isScryptHash         = regexp.MustCompile(`^\$scrypt\$`)
 	isFirebaseScryptHash = regexp.MustCompile(`^\$firescrypt\$`)
+	isMD5Hash            = regexp.MustCompile(`^\$md5\$`)
 )
 
 func IsBcryptHash(hash []byte) bool {
@@ -204,6 +232,10 @@ func IsScryptHash(hash []byte) bool {
 
 func IsFirebaseScryptHash(hash []byte) bool {
 	return isFirebaseScryptHash.Match(hash)
+}
+
+func IsMD5Hash(hash []byte) bool {
+	return isMD5Hash.Match(hash)
 }
 
 func decodeArgon2idHash(encodedHash string) (p *config.Argon2, salt, hash []byte, err error) {
@@ -349,4 +381,41 @@ func decodeFirebaseScryptHash(encodedHash string) (p *Scrypt, salt, saltSeparato
 	}
 
 	return p, salt, saltSeparator, hash, signerKey, nil
+}
+
+// decodeMD5Hash decodes MD5 encoded password hash.
+// format without salt: $md5$<hash>
+// format with salt $md5$pf=<salting-format>$<salt>$<hash>
+func decodeMD5Hash(encodedHash string) (pf, salt, hash []byte, err error) {
+	parts := strings.Split(encodedHash, "$")
+
+	switch len(parts) {
+	case 3:
+		hash, err := base64.StdEncoding.Strict().DecodeString(parts[2])
+		return nil, nil, hash, err
+	case 5:
+		_, err = fmt.Sscanf(parts[2], "pf=%s", &pf)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		pf, err := base64.StdEncoding.Strict().DecodeString(string(pf))
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		salt, err = base64.StdEncoding.Strict().DecodeString(parts[3])
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		hash, err = base64.StdEncoding.Strict().DecodeString(parts[4])
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		return pf, salt, hash, nil
+	default:
+		return nil, nil, nil, ErrInvalidHash
+	}
 }
--- a/hash/hasher_test.go
+++ b/hash/hasher_test.go
@@ -6,6 +6,7 @@ package hash_test
 import (
 	"context"
 	"crypto/rand"
+	"encoding/base64"
 	"fmt"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


hash/hash_comparator.go
```diff
@@ -260,3 +290,14 @@ func decodeScryptHash(encodedHash string) (p *Scrypt, salt, hash []byte, err err
 
 	return p, salt, hash, nil
 }
+
+// decodeMD5Hash decodes MD5 encoded password hash.
+// format: $md5$<hash>
+func decodeMD5Hash(encodedHash string) (hash []byte, err error) {
+	parts := strings.Split(encodedHash, "$")
```
aeneasr, line 297: Most md5 hashers probably use salting, but there's probably not one way to add the salt (as MD5 itself has no salting). How could we solve this?

erolkskn: Yeah that would be great addition. In most of the scenarios i have seen, salt is either appended or prepended to the password. And again probably not everyone uses salting. I think format like below, can accommodate up to that needs:
`$md5$<hash>$<salt base64>,<mode>`
- **mode** parameter can be either **a** *(abbreviation for append)* or **p** *(abbreviation for prepend)*. In the case of **a** salt will be appended to the password before it's being hashed. In the case of **p** salt will be prepended to the password before it's being hashed.
- The section after the third dollar sign (inclusive) is optional. That means if someone wants to store a password which has no salt, it can be formatted like this: `$md5$<hash>`. By putting **hash** before **salt** we can easily make *salting* optional.

erolkskn: In that case,
Password without a salt can look like this: `$md5$v=1$m=n$$202cb962ac59075b964b07152d234b70`
Password with salt can look like this: `$md5$v=1$m=a$MTIzNDU2Nw==$8a3379ae294935afa4195fa27600fd43`
Just wanted to make sure we are on the same page, am i getting it right ?

aeneasr: 🤔 I was also thinking that we could use sprintf to set the salt:

```
$md5$v=1$pf=%[0]s%[1]s$salt$hash
$md5$v=1$hash
```

where `pf` is `printf` and the first argument being the salt, the second the hash. If `pf` is not set, we do not have a salt. This would allow us to define the salt format any way we like:

```
%[0]s%[1]s
%[1]s%[0]s
%[0]sSOME_DELIMITER%[1]s
....
```

WDYT?

erolkskn: That's a good idea, this would make salting process more flexible. But i think formatting by index might come as *too implicit* for some users. How about formatting with named parameters like this: 
```
%[salt]s%[password]s
%[salt]sSOME_DELIMITER%[password]s
```
We might replace `%[salt]s` to `%[0]s` and `%[password]s` to `%[1]s` before passing format string to `sprintf`. For example format string `%[salt]s%[password]s` can turn into `%[0]s%[1]s` before passing it to `sprintf`. 
This way we both can use indexed formatting with sprintf and make the definition of format more explicit to the users.

What do you think ?

aeneasr: That's true, but I think it makes sense to just rely of `Sprintf`. In v2 we can think about introducing parsing here :)

erolkskn: Or we can make `$pf` variable enclosed by double quotes, and only thing the users would have to escape could be the double quotes itself and `%` char. 
For example:
```
<salt>[$<password> => $md5$v=1$pf="%[0]s[$%[1]s"$salt$hash
<salt>"%<password> => $md5$v=1$pf="%[0]s\"%%%[1]s"$salt$hash
```

erolkskn: As a third option we may not need any escaping at all. We can assume that part starts after `$pf=` and ends before `$salt` is the format string. For example:
```
<salt>[$<password> => $md5$v=1$pf=%[0]s[$$salt$hash
````
Even though that's an option, users would still need to escape `%` character, because of how Sprintf works.

Comment (erolkskn) [reply hash/hash_comparator.go:297]: > As a third option we may not need any escaping at all. We can assume that part starts after `$pf=` and ends before `$salt` is the format string. For example:
> 
> ```
> <salt>[$<password> => $md5$v=1$pf=%[0]s[$%[1]s$salt$hash
> ```
> 
> Even though that's an option, users would still have to escape `%` character, because of how Sprintf works.

We can capture the input with this regex: `\$md5\$v\=1\$pf\=(.*)\$(.*)\$(.*)`

Comment (aeneasr) [reply hash/hash_comparator.go:297]: I'm not sure if we have to escape it! Splitting by `$` and then processing each parameter in it's own context, it should work IMO :) The one problem I see is if the template itself uses `$`: `pf=%[0]s$%[1]s` 🤔 

In which case escaping it in double quotes might make sense: `pf="%[0]s$%[1]s"`

Comment (erolkskn) [reply hash/hash_comparator.go:297]: I think we are on different page now, so i'm just going to update the code according to thoughts i have in my mind, may be this can illuminate our conversation about the topic.

File: hash/hash_comparator.go [hunk 2]
```diff
@@ -260,3 +293,37 @@
...
+// format without salt: $md5$<hash>
+// format with salt $md5$v=1$pf=<salting-format>$<salt-base64>$<hash>
+func decodeMD5Hash(encodedHash string) (pf, salt, hash []byte, err error) {
+	if parts := md5WithSalt.FindStringSubmatch(encodedHash); len(parts) == 4 {
+		salt, err := base64.StdEncoding.DecodeString(parts[2])
+		if err != nil {
+			return nil, nil, nil, errors.WithStack(err)
+		}
+
+		hash, err := hex.DecodeString(parts[3])
```
hperl, line 312: I think we are using base64 encoding everywhere else in password format strings. Why use hex encoding here?

hperl (commented):
Hi, and thanks a lot for your PR. It looks really good already, I have just left some comments.

You can improve the coverage by adding more test cases to parse for formats that don't match.

hash/hasher_test.go
```diff
@@ -242,4 +242,29 @@ func TestCompare(t *testing.T) {
 	assert.Error(t, hash.Compare(context.Background(), []byte("test"), []byte("$scrypt$ln=16385,r=8,p=1$2npRo7P03Mt8keSoMbyD/tKFWyUzjiQf2svUaNDSrhA=$MiCzNcIplSMqSBrm4HckjYqYhaVPPjTARTzwB1cVNYE=")))
 	assert.Error(t, hash.Compare(context.Background(), []byte("tesu"), []byte("$scrypt$ln=16384,r=8,p=1$2npRo7P03Mt8keSoMbyD/tKFWyUzjiQf2svUaNDSrhA=$MiCzNcIplSMqSBrm4HckjYqYhaVPPjTARTzwB1cVNYE=")))
 	assert.Error(t, hash.Compare(context.Background(), []byte("tesu"), []byte("$scrypt$ln=abc,r=8,p=1$2npRo7P03Mt8keSoMbyD/tKFWyUzjiQf2svUaNDSrhA=$MiCzNcIplSMqSBrm4HckjYqYhaVPPjTARTzwB1cVNYE=")))
+
+	assert.Nil(t, hash.Compare(context.Background(), []byte("test"), []byte("$md5$v=1$098f6bcd4621d373cade4e832627b4f6")))
+	assert.Nil(t, hash.CompareMD5(context.Background(), []byte("test"), []byte("$md5$v=1$098f6bcd4621d373cade4e832627b4f6")))
+	assert.Nil(t, hash.CompareMD5(context.Background(), []byte("test"), []byte("$md5$v=1$098F6BCD4621D373CADE4E832627B4F6")))
+	assert.Error(t, hash.Compare(context.Background(), []byte("test"), []byte("$md5$v=1$5a105e8b9d40e1329780d62ea2265d8a")))
```
hperl, line 249:
```suggestion
	assert.Error(t, hash.Compare(context.Background(), []byte("test"), []byte("$md5$v=1$5a105e8b9d40e1329780d62ea2265d8a")))
	assert.Error(t, hash.Compare(context.Background(), []byte("test"), []byte("$md5$v=1$ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ")))

```

hperl: To improve test coverage ;)

hash/hash_comparator.go
```diff
@@ -132,12 +137,36 @@
...
+	return errors.WithStack(ErrMismatchedHashAndPassword)
+}
+
 var (
 	isBcryptHash   = regexp.MustCompile(`^\$2[abzy]?\$`)
 	isArgon2idHash = regexp.MustCompile(`^\$argon2id\$`)
 	isArgon2iHash  = regexp.MustCompile(`^\$argon2i\$`)
 	isPbkdf2Hash   = regexp.MustCompile(`^\$pbkdf2-sha[0-9]{1,3}\$`)
 	isScryptHash   = regexp.MustCompile(`^\$scrypt\$`)
+	isMD5Hash      = regexp.MustCompile(`^\$md5\$v=1\$`)
```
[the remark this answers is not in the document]
erolkskn: It's based on the comment of Aeneas, see: https://github.com/ory/kratos/pull/2725#discussion_r968211034 It's just parameter for specifying which format version to use. It's not used for distinguishing salting from plain hash.

hash/hash_comparator.go
erolkskn: Honestly, I'm in between on this subject. In the one hand almost all of the products I've seen to this day keeps MD5 hashes in hexadecimal string format. On the other hand using hex encoding here contradicts with out formats as you say. My main point of concern in here is using base64 encoding might be a little confusing for some users. If we ask them to give it in base64 encoding instead of hex string, following questions might be arose:

- Should the user give base64 output of hex string itself ? 
- Should the user give base64 output of bytes itself ? 

Currently on other formats, we are using the strategy of encoding bytes with base64, so i'm assuming in that scenario we will go with the second option. In that case almost all of the users should convert hexadecimal formatted string to bytes first, and then they should encode those bytes with base64

hash/hash_comparator.go
```diff
@@ -132,12 +137,36 @@
...
+func CompareMD5(_ context.Context, password []byte, hash []byte) error {
+	// Extract the hash from the encoded password
+	pf, salt, hash, err := decodeMD5Hash(string(hash))
+	if err != nil {
+		return err
+	}
+
+	arg := password
+	if salt != nil {
+		arg = []byte(fmt.Sprintf(string(pf), salt, password))
```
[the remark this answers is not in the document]
erolkskn: When i took a look around, i hadn't seen any security issues regarding this. But i think this subject should be investigated in more depth.

hash/hasher_test.go
erolkskn: Oh thanks for that :) I'm going to merge them.

hash/hash_comparator.go
hperl: I read the implementation yesterday and I still find the usage of the untrusted format string a little scary. At best it is way more powerful than we need it to be.

How about we just specify the format string as some string that contains `PASS` and `SALT` template strings, and then we `strings.Replace` the actual hash into that? The examples cover all use cases that I can think of (password or salt first, delimiters, etc.), e.g.:

* `PASS:SALT`
* `SALTPASS`
* `PASS$SALT`

To support literal `$`s in the template we can either base64-encode the whole string, or wrap the template with double quotes. I tend towards base64, because this is consistent with other fields in the format (salt and hash), and because the parsing format is then just a `strings.Split(s, "$")`, which is way easier to read than a regular expression.

WDYT?

hash/hash_comparator.go
hperl: Yes, I'd prefer that. The output of the MD5 hash itself should be base64 encoded.

hash/hash_comparator.go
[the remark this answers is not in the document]
hperl: So it is to distinguish the format string from some other thing we might want to introduce in the future?

If so, can't we worry about that then and introduce a new field?

So today we write
* `$md5$pf=some_v1_format_string$salt$hash`

If we want to change the format of `pf` in the future (if ever), we can introduce a new field, e.g.:

* `$md5$pfV2=some_v2_format_string$salt$hash`

The parser would ignore `pf` if there is also a `pfV2`.

Am I missing something? As I read it I found the version number somewhat confusing and didn't know what it related to, as MD5 itself does not have multiple versions.

hash/hash_comparator.go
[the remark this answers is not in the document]
hperl: I suggested a simpler format here: https://github.com/ory/kratos/pull/2725#discussion_r977279806

hash/hash_comparator.go
erolkskn: @hperl Definitely +1 on this. This feels way more secure than the `Printf` approach. About the base64 encoded format +1 on this too. This can ease the parsing process. Instead of using regex we can use `Split` approach as you suggested.

hash/hash_comparator.go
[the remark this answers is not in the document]
erolkskn: I think there is a little confusion here. The format is like the below in the current version:
`$md5$v=1$pf=...`
And if we introduce another version in the future, it's going to look like this:
`$md5$v=2$pf=...`

hash/hash_comparator.go
erolkskn: @hperl Alright then, just to be clear on this, users should be encoding 16 byte md5 hash with base64, right ? Not the 32 byte md5 hex string representation.

hperl: yes :)

hash/hash_comparator.go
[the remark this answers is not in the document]
hperl: My comment was that we don't need the version, and if in the future something changes we can version just that one field.

hash/hash_comparator.go
hperl: nice!

hash/hash_comparator.go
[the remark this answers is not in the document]
erolkskn: Oh, i think i got confused. I totally misread that, sorry about this. Yeah it's probably more convenient to do :-)

hperl (dismissed):
Thanks @erolkskn, LGTM! :tada: 
@zepatrik can we merge this now or should we wait for final approval by @aeneasr?

/cc @jonas-jonas

zepatrik (approved):
Looks good IMO. We will merge it beginning of next week :+1:

aeneasr (dismissed):

aeneasr:
Awesome! This is mergable!

The last missing piece is no adding a bit of docs for this feature :) https://www.ory.sh/docs/kratos/manage-identities/import-user-accounts-identities#bcrypt-pkbdf2-argon2-scrypt-family-hashed-password

We probably need a dedicated section for MD5 as our implementation is not standardized, meaning that we'll need to explain how this works :)

hugotiburtino (approved):
Beautiful!

erolkskn:
Oh sorry for this took too long. I couldn't find appropriate time to write down docs. It should be ready in the next few days. Going to rebase with master tonight.

aeneasr:
Awesome! And there have since been a few merge conflicts - could you please also take a look into those? :)

erolkskn:
> Awesome! And there have since been a few merge conflicts - could you please also take a look into those? :)

Of course :-) Going to look at them as soon as I can today

aeneasr:
Thank you! We still need to explain this feature in the documentation. Do you need help for that? :)

erolkskn:
> Thank you! We still need to explain this feature in the documentation. Do you need help for that? :)

I'm on it now :-) I'm going to create a PR regarding this on "ory/docs" this evening.

aeneasr:
awesome!!!

hperl (approved):
LGTM! Thanks for the good work, @erolkskn!

graysontvbs:
hi , I am using the "import" command to import the identity, and the password is encrypted using the md5 method. However, I cannot log in correctly. After checking, I found that if the previous password was encrypted using a different programming language, it cannot be properly compared and verified.

This is because the md5 encryption method in Go differs from other programming languages (Go uses bytes). However, in older systems, different programming languages such as PHP or C# may have been used. This prevents me from using those md5 passwords for login.

In summary, it seems that there are two different md5 encryption methods, and it appears that they are not supported at the moment. Can we open a separate issue to address this matter?

jonas-jonas:
>I found that if the previous password was encrypted using a different programming language, it cannot be properly compared and verified.

With which programming language were your passwords hashed before? In theory the implementation in Kratos should follow the spec, but if the passwords were hashed with some custom implementation, that deviates from the spec, Kratos won't be able to verify the passwords.

Please also make sure, [to include the salt](https://www.ory.sh/docs/kratos/manage-identities/import-user-accounts-identities#hashed-passwords) if a salt was used to hash the passwords.

>Can we open a separate issue to address this matter? 

Please do!

graysontvbs:
I think I have reached some conclusions. The unit test cases for this PR are incorrect. Taking the first test case as an example, after encrypting with md5 and base64, the expected result should be `MDk4ZjZiY2Q0NjIxZDM3M2NhZGU0ZTgzMjYyN2I0ZjY=`, but the test case uses `CY9rzUYh03PK3k6DJie09g==`.

The reason for this is that the encryption process involves passing the byte[:] result of md5 into the base64 encode function, rather than passing the string result of md5. If you use the following sample code, you can clearly see that the base64 output will be different depending on the input data:


```go
md5Hash := md5.Sum([]byte("test")) // 098f6bcd4621d373cade4e832627b4f6

//use byte[:] to base64 encode (X)
base64Encoded := base64.StdEncoding.EncodeToString(md5Hash[:])
fmt.Println(base64Encoded) //CY9rzUYh03PK3k6DJie09g==

//use string to base64 encode (O)
base64Encoded1 := base64.StdEncoding.EncodeToString([]byte(hex.EncodeToString(md5Hash[:])))
fmt.Println(base64Encoded1) //MDk4ZjZiY2Q0NjIxZDM3M2NhZGU0ZTgzMjYyN2I0ZjY=
```

PR #2733, opened 2022-09-14
feat: complete session devices upsert

kelkarajay:
This PR fixes the eager query for Session devices and adds the creation of the devices in the Upsert. The update operation for devices doesn't belong with the upsert as at most one device would need an update (Due to change of IP or simply updating the `last_seen` timestamp).

Fixes - https://github.com/ory/kratos/pull/2715#discussion_r970537554

### DIFF
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -91,7 +91,7 @@ func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) error
 				return err
 			}
 
-			for i := 0; i < len(s.Devices); i++ {
+			for i := range s.Devices {
 				device := &(s.Devices[i])
 				device.SessionID = s.ID
 				device.NID = s.NID
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -96,8 +96,7 @@ func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) error
 				device.SessionID = s.ID
 				device.NID = s.NID
 
-				err = tx.Create(device)
-				if err != nil {
+				if err := tx.Create(device); err != nil {
 					return err
 				}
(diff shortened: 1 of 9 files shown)


persistence/sql/persister_session.go
```diff
@@ -107,15 +83,35 @@ func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) error
...
+		if err := tx.Find(new(session.Session), s.ID); errors.Is(err, sql.ErrNoRows) {
+			// This must not be eager or identities will be created / updated
+			err = errors.WithStack(tx.Create(s))
+			if err != nil {
+				return err
+			}
 
-	// This must not be eager or identities will be created / updated
-	return p.GetConnection(ctx).Update(s)
+			for i := 0; i < len(s.Devices); i++ {
```
aeneasr, line 94: Easier:

```
for i := range s.Devices {
  device := &s.Devices[i]
}
```

persistence/sql/persister_session.go
```diff
@@ -107,15 +83,35 @@ func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) error
...
+				device := &(s.Devices[i])
+				device.SessionID = s.ID
+				device.NID = s.NID
+
+				err = tx.Create(device)
+				if err != nil {
```
aeneasr, line 100: We tend to usse this style for single error checking, to prevent scoping issues:

```suggestion
				if err := tx.Create(device); err != nil {
```

aeneasr (changes requested):
Two minor things! :)

persistence/sql/persister_session.go
kelkarajay: Agree. changed this

aeneasr (approved):

Issue #2422, opened 2022-04-24
Review and add more commonly used password hashing algorithms for importing users


ravikiran438:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

As per earlier discussion here [605](https://github.com/ory/kratos/issues/605) and the subsequent implementation [2256](https://github.com/ory/kratos/pull/2256) its possible now to import user credentials with passwords hashed using the following 3 algorithms.

1. Argon2
2. PKBDF2
3. BCrypt

But most systems still use legacy hashing algorithms and need a way to support importing of user credentials as sending a recovery link to hundreds of thousands of users could be bothersome to some of the users.

### Describe your ideal solution

**Solution 1:** We may want to review and add few more commonly used algorithms to facilitate easy migration from other systems. As an example [Auth0](https://auth0.com/docs/manage-users/user-migration/bulk-user-import-database-schema-and-examples#supported-hash-algorithms) provides the following hashing algorithms.

- [Argon2](https://github.com/p-h-c/phc-winner-argon2)
- [bcrypt](https://auth0.com/blog/hashing-in-action-understanding-bcrypt/)
- [LDAP](https://tools.ietf.org/html/rfc2307#section-5.3) (RFC-2307 "userPassword")
- [HMAC](https://tools.ietf.org/html/rfc2104)
- [MD4](https://tools.ietf.org/html/rfc1320)
- [MD5](https://tools.ietf.org/html/rfc1321)
- [SHA1](https://tools.ietf.org/html/rfc3174)
- [SHA256 and SHA512](https://tools.ietf.org/html/rfc4634)
- [PBKDF2](https://tools.ietf.org/html/rfc2898#section-5.2)
- [scrypt](https://datatracker.ietf.org/doc/rfc7914/)

[...]

PR #2741, opened 2022-09-16
feat: allow importing salted sha hashing algorithms

zambien:
Allows importing of salted SHA  (SSHA, SSHA256, and SSHA512) passwords into Ory Kratos

## Related issue(s)

https://github.com/ory/kratos/issues/2422

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


zambien:
I'll get started on updating the docs soon!

hash/hash_comparator.go
```diff
@@ -132,12 +141,42 @@ func CompareScrypt(_ context.Context, password []byte, hash []byte) error {
...
+	return nil
+}
+
 var (
 	isBcryptHash   = regexp.MustCompile(`^\$2[abzy]?\$`)
 	isArgon2idHash = regexp.MustCompile(`^\$argon2id\$`)
 	isArgon2iHash  = regexp.MustCompile(`^\$argon2i\$`)
 	isPbkdf2Hash   = regexp.MustCompile(`^\$pbkdf2-sha[0-9]{1,3}\$`)
 	isScryptHash   = regexp.MustCompile(`^\$scrypt\$`)
+	isSSHAHash     = regexp.MustCompile(`^{SSHA}.*`)
```
hperl, line 177: Why are we not follwing the $-separator format here?

hash/hash_comparator.go
```diff
@@ -132,12 +141,42 @@ func CompareScrypt(_ context.Context, password []byte, hash []byte) error {
 	return errors.WithStack(ErrMismatchedHashAndPassword)
 }
 
+func CompareSSHA(_ context.Context, password []byte, hash []byte) error {
+
+	if _, err := ssha.Validate(string(password), string(hash)); err != nil {
```
hperl, line 146: The `ssha` library does not do a constant time comparison of the password with the hash. Is this not an issue?

hperl (changes requested):

hperl (changes requested):
Thank you very much for your PR. It looks very good already. I have just left some comments for you.

hash/hash_comparator.go
zambien: SHA and SSHA are often stored with, for example, {SSHA512} prepending the value.  An example function being something like this:

`$hash = "{SSHA256}" . base64_encode(hash('sha256', $password . $salt) . $salt);`

This is pretty standard for these methods.  We could change this but that would burden users with transforming their inputs when importing users.

I'm happy to make the change to use $SSHA$ where appropriate if everyone prefers that approach.

hash/hash_comparator.go
zambien: Interesting, I hadn't considered this.  Thanks for pointing it out.  I will look into this and mark this in a draft state for now.

zambien: I've opened https://github.com/jsimonetti/pwscheme/pull/4 to resolve this.  This PR will remain in draft mode for now as I await review there.  Thanks for the direction!

zambien: Constant time comparison has been merged into the included library.  Setting this back to ready for review!

jsimonetti: I believe you still need to update the go.mod file to use these latest changes.

hperl: Sleeping over it I think we should just call directly to `crypto/sha` and friends instead of using a library for the following reasons:

* I really want to stick to the PHC format specified here: https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md, which means using `$` as delimitiers
* We will need some flexibility in the handling of the salt for hash functions that do not use a salt as part of their spec (such as bcrypt). Applications may have computed the hash by appending or prepending the salt, or by inserting some separator between hash and salt. See the discussion and my comment on a related PR here: https://github.com/ory/kratos/pull/2725#discussion_r977279806. I think we should use some common solution for all non-password hashes (MD5, SHA-family).
* Lastly, the library still seems quite new, and I'd rather just depend on standard Go libraries for something security critical from a software supply chain standpoint. This is not to say that I hugely respect the effort to create a nice library that automatically handles salting in a nice way, even though these hashes should not be used if you get a choice, because they are just too fast to brute-force, if not broken (like MD5).

So in short, I think we should inline the functionality and make the salting more flexible. WDYT?

hperl (changes requested):
Thanks for the update! It looks very good now :tada:, but I think we still need to iterate over the salting. :salt:

hash/hash_comparator.go
zambien: I definitely understand the sentiments around not including newer/lesser known external security libraries and have thought about that some over the past few days.  I'm good with that approach and the other recommendations you made and will get started on understanding and implementing those changes.

For the various SHA and SSHA implementations, is the desire to be more generic and DRY when handling these or is it a valid approach to still have functions for validating each implementation separately?

zambien: After reading the linked comments and code more carefully I think I understand now.  The functions to handle SHA, SSHA, SHA256, SSHA256, etc. should handle the presence or absence of a salt dynamically and to your point, while there are norms around how these are persisted, the spec is flexible so the implementation should also be in Kratos.  Thanks!

zambien: Thinking about this a bit more, a real world example is the string `testing123`.   The SSHA512 sum of this string is:

`xPUl/px+1cG55rUH4rzcwxdOIPSB2TingLpiJJumN2xyDWN4Ix1WQG3ihnvHaWUE8MYNkvMi5rf0C9NYixHsE6Yh59M=`

The salt is included in that SSHA512 sum of course.  The actual salt is:

`piHn0w==`

This could be represented using phc format as:

$SSHA512$hash$salt

for example:

`$SSHA512$xPUl/px+1cG55rUH4rzcwxdOIPSB2TingLpiJJumN2xyDWN4Ix1WQG3ihnvHaWUE8MYNkvMi5rf0C9NYixHsE6Yh59M=$piHn0w==`

I'm not sure if this is the right approach... it feels counter-intuitive to append the salt when it is already part of the hash.  I have some code working with the format.  Since we can't be sure of the formatting I think what is captured above will work but wanted to get some other eyes on it before finishing the rest of the code.

zambien: Thinking on this more, if the main point of this PR is to be able to import passwords from existing data stores, it becomes problematic to assume that the salt is added in any particular way.  If we are constructing a new hash there is no issue but for preexisting hashes the problem is hairy.  This is discussed some in https://github.com/ory/kratos/pull/2725#discussion_r977279806 but there doesn't seem to be a solution that is clear.

Could we discuss what this would look like in PHC format with real examples?

zambien: @hperl  and @aeneasr any thoughts on the above?

elbotho: If someone whats to read up on standards for storing password hashes:
- [SO: what is the standard for password hash string encoding?](https://stackoverflow.com/a/73598323/3294973)
- [PHC string format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md)
- [PassLib SHA-1 Crypt Formating](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.sha1_crypt.html#passlib.hash.sha1_crypt)

elbotho: PHC formating: `$<id>[$v=<version>][$<param>=<value>(,<param>=<value>)*][$<salt>[$<hash>]]`

So an example using the formating parameter @hperl proposed over in the [MD5 PR](https://github.com/ory/kratos/pull/2725) would look something like this, right?

`$sha-1$f=SALTPASS$<salt>$<hash>`

Where the format string ("SALTPASS"), salt and hash are base64 encoded.

hash/hash_comparator.go
elbotho: I guess the question is if kratos wants to support only SSHA or also custom implementations (for example reversed salt and password order).

hugotiburtino:
> So in short, I think we should inline the functionality and make the salting more flexible. WDYT?

Definitely having the possibility of prepending the salt would save many lifes (i.e. many password resets)

hash/hash_comparator.go
aeneasr: Could you please provide a source for this statement?

I still think that we'll should use the PHC format here, because SHA for passwords is not standardized and many implementations with different salting mechanisms exist. So I must insist that this importer follows suit here. We can always write a CLI helper (`grep` :D) which converts whatever SHA format into PHC.

You can take a look at the MD5 PR ( #2725 ) to see how we solved it there!

elbotho: @aeneasr I think the confusion here is, that SSHA (and SSHA256, and SSHA512) is what @zambien wanted to implement with this PR in the first place and that is standardized (e.g. https://community.canvaslms.com/t5/SIS-User-Articles/SSHA-Password-Generation/ta-p/243730) and you and @serlo for example would prefer a more flexible implementation of all sorts of SHA approaches.

elbotho: The other thread (https://github.com/ory/kratos/pull/2741#discussion_r980357651) is more up to date on this.

hash/hash_comparator.go
aeneasr: yup, that sounds straight forward @elbotho

hash/hash_comparator.go
zambien: > @aeneasr I think the confusion here is, that SSHA (and SSHA256, and SSHA512) is what @zambien wanted to implement with this PR in the first place and that is standardized (e.g. https://community.canvaslms.com/t5/SIS-User-Articles/SSHA-Password-Generation/ta-p/243730) and you and others (e.g we at @serlo) would prefer a more flexible implementation for all sorts of SHA approaches.

I've started on some code which will implement both SHA and SSHA and also will dynamically handle the presence or absence of a salt based on the comments in the various threads.  Been a busy couple weeks but I'll commit it once I feel it is close to ready to be reviewed.

aeneasr: Ok, for SSHA I think it makes sense to use the format as standardized `{SSHA}...`! For SSH as discussed with the PHC string format :) Sorry for confusing this here a bit.

hash/hash_comparator.go
jsimonetti: > * Lastly, the library still seems quite new, and I'd rather just depend on standard Go libraries for something security critical from a software supply chain standpoint.

While I completely understand your supply chain argument, I would like to dispute your claim the library is new. In fact, it is even a lot older then kratos is.

zambien:
> > So in short, I think we should inline the functionality and make the salting more flexible. WDYT?
> 
> Definitely having the possibility of prepending the salt would save many lifes (i.e. many password resets)

@hugotiburtino you have a use case where you are using SSHA, SSHA256, or SSHA512 where the salt is prepended instead of being appended?  There is another thread where the latest comment @aeneasr thought to use the initial solution for SSHA (not using PHC.  I want to be sure to consider everyone's needs.  Even though using the `{SSHA}...` format is easier I agree that isn't really a standard, but is a norm/more.  Although every implementation I've seen for SSHA uses the appended implementation.

elbotho:
> @hugotiburtino you have a use case where you are using SSHA, SSHA256, or SSHA512 where the salt is prepended instead of being appended? There is another thread where the latest comment @aeneasr thought to use the initial solution for SSHA (not using PHC). I want to be sure to consider everyone's needs. Even though using the `{SSHA}...` format is easier I agree that isn't really a standard, but is a norm/more. Although every implementation I've seen for SSHA uses the appended implementation.
> 
> Edit: punctuation

Our use case (same org. as @hugotiburtino) is actually not SSHA but a custom flavour of SHA so we would be well served with the PHC syntax. Thanks for your consideration.

If we can help you with this PR in any way, just get in touch.

hugotiburtino:
hey @hperl 
Maybe you could review it again, since the code now adresses your [concerns](https://github.com/ory/kratos/pull/2741#discussion_r978717809).

hperl (dismissed):
Thank you for the changes, LGTM!

There was a bug in the error handling for short SSHA passwords that I fixed in https://github.com/ory/kratos/pull/2741/commits/fea0bbdabc66f36cc1267e1e28c1bd23ad0f7f58. PTAL.

aeneasr:
We probably also need to update the documentation to now include regular SHA?

elbotho:
> We probably also need to update the documentation to now include regular SHA?

@aeneasr that's right. I'm awaiting the review of https://github.com/ory/docs/pull/1090 so we can explain the formatting in the same way and can add the changes then.

aeneasr:
That's merged now! There's a few conflicts left that need resolving.

@hperl are the changes good to go now?

hperl (approved):
This looks very good now :tada:!

I just merged master, if CI passes we should be ready to merge :muscle:

hperl (approved):

aeneasr:
epic!

PR #2744, opened 2022-09-22
feat: hot-reload TLS certificates

alnr:
ory/hydra#2850 but for Kratos


driver/config/config.go
```diff
@@ -1304,16 +1308,31 @@ func (p *Config) GetTSLCertificatesForAdmin(ctx context.Context) []tls.Certifica
...
+		getCert, err := tlsx.GetCertificate(ctx, certPath, keyPath, errs)
+		if err != nil {
+			p.l.WithError(err).Fatalf("Unable to load HTTPS TLS Certificate")
+			return nil // reachable in unit tests when Fatalf is hooked
+		}
+		go func() {
+			for err := range errs {
+				p.l.WithError(err).Error("Failed to reload TLS certificates, using previous certificates")
+			}
+		}()
```
zepatrik, line 1332: This goroutine should also listen on `ctx.Done()` and return in that case.
```suggestion
		go func() {
			for {
				select {
					case err := <-errs:
						p.l.WithError(err).Error("Failed to reload TLS certificates, using previous certificates")
					case <-ctx.Done():
						return
				}
			}
		}()
```

driver/config/config_test.go
```diff
@@ -815,34 +832,38 @@ func TestLoadingTLSConfig(t *testing.T) {
...
 		logger.Logger.Hooks.Add(hook)
 
 		p := config.MustNew(t, logger, os.Stderr, configx.SkipValidation())
 		p.MustSet(ctx, config.ViperKeyPublicTLSKeyBase64, "empty")
 		p.MustSet(ctx, config.ViperKeyPublicTLSCertBase64, certBase64)
-		assert.Nil(t, p.GetTSLCertificatesForPublic(ctx))
-		assert.Equal(t, "TLS has not been configured for public, skipping", hook.LastEntry().Message)
+		assert.Nil(t, p.GetTLSCertificatesForPublic(ctx))
+		assert.Equal(t, "Unable to load HTTPS TLS Certificate", hook.LastEntry().Message)
+		assert.True(t, exited)
```
zepatrik, line 851: Nice improvement to the test here :+1:

driver/config/config_test.go
```diff
@@ -867,34 +888,38 @@ func TestLoadingTLSConfig(t *testing.T) {
 		p := config.MustNew(t, logger, os.Stderr, configx.SkipValidation())
 		p.MustSet(ctx, config.ViperKeyAdminTLSKeyPath, keyPath)
 		p.MustSet(ctx, config.ViperKeyAdminTLSCertPath, certPath)
-		assert.NotNil(t, p.GetTSLCertificatesForAdmin(ctx))
-		assert.Equal(t, "Setting up HTTPS for admin", hook.LastEntry().Message)
+		assert.NotNil(t, p.GetTLSCertificatesForAdmin(ctx))
+		assert.Equal(t, "Setting up HTTPS for admin (automatic certificate reloading active)", hook.LastEntry().Message)
```
zepatrik, line 892: You could also use
```suggestion
		assert.Contains(t, hook.LastEntry().Message, "Setting up HTTPS for admin", hook.LastEntry().Message)
```
just FYI

driver/config/config_test.go
```diff
@@ -867,34 +888,38 @@ func TestLoadingTLSConfig(t *testing.T) {
...
+		assert.True(t, exited)
 	})
 
 	t.Run("case=admin: failing to load certificate from a file", func(t *testing.T) {
 		logger := logrusx.New("", "")
-		logger.Logger.ExitFunc = func(code int) {}
+		exited := false
+		logger.Logger.ExitFunc = func(code int) { exited = true }
 		hook := new(test.Hook)
 		logger.Logger.Hooks.Add(hook)
```
zepatrik, line 915: You could make this a small local test helper function here, as it is used so often. Something like
```go 
func newTestLogger() (l *logrusx.Logger, h *test.Hook, exited *bool) {
  ...
}
```

internal/testhelpers/e2e_server.go
```diff
@@ -181,35 +181,44 @@ func CheckE2EServerOnHTTPS(t *testing.T, publicPort, adminPort int) (publicUrl,
 	return
 }
 
-func GenerateTLSCertificateFilesForTests(t *testing.T, certPath, keyPath string) {
+// GenerateTLSCertificateFilesForTests writes a new, self-signed TLS
+// certificate+key (in PEM format) to a temporary location on disk and returns
+// the paths to both. The files are automatically cleaned up when the given
+// *testing.T concludes its tests.
+func GenerateTLSCertificateFilesForTests(t *testing.T) (certPath, keyPath string) {
```
zepatrik, line 188: As we also often need the base64 encoding of the PEM content, why not return it here as well and avoid repeatedly encoding it?

zepatrik (changes requested):
Very nice :tada: 
Sorry for the long review time, somehow forgot about this PR :sweat_smile: 
I just have some ideas how to simplify and improve the code a bit further :wink:

driver/config/config.go
aeneasr: Agree!

aeneasr (changes requested):

internal/testhelpers/e2e_server.go
alnr: streamlined the helper

driver/config/config_test.go
alnr: Added helper func

zepatrik (approved):
Thanks, this looks great now :heart_eyes:

driver/config/config.go
aeneasr: Makes sense!

driver/config/config.go
```diff
@@ -1304,16 +1308,31 @@ func (p *Config) GetTSLCertificatesForAdmin(ctx context.Context) []tls.Certifica
...
-	} else if !errors.Is(err, tlsx.ErrNoCertificatesConfigured) {
-		p.l.WithError(err).Fatalf("Unable to load HTTPS TLS Certificate")
+		return func(*tls.ClientHelloInfo) (*tls.Certificate, error) { return &cert, nil }
+	}
+	if certPath != "" && keyPath != "" {
+		errs := make(chan error, 1)
+		getCert, err := tlsx.GetCertificate(ctx, certPath, keyPath, errs)
+		if err != nil {
+			p.l.WithError(err).Fatalf("Unable to load HTTPS TLS Certificate")
+			return nil // reachable in unit tests when Fatalf is hooked
```
aeneasr, line 1326: Shouldn't this return the error in that case?

driver/config/config.go
```diff
@@ -1304,16 +1308,31 @@ func (p *Config) GetTSLCertificatesForAdmin(ctx context.Context) []tls.Certifica
...
 		p.l.Infof("Setting up HTTPS for %s", daemon)
-		return cert
-	} else if !errors.Is(err, tlsx.ErrNoCertificatesConfigured) {
-		p.l.WithError(err).Fatalf("Unable to load HTTPS TLS Certificate")
+		return func(*tls.ClientHelloInfo) (*tls.Certificate, error) { return &cert, nil }
+	}
+	if certPath != "" && keyPath != "" {
+		errs := make(chan error, 1)
+		getCert, err := tlsx.GetCertificate(ctx, certPath, keyPath, errs)
+		if err != nil {
```
aeneasr, line 1324: Is `!errors.Is(err, tlsx.ErrNoCertificatesConfigured)` no longer a valid error state, or why was it ignored?

aeneasr (commented):
Awesome, thanks! Just a two questions :)

driver/config/config.go
alnr: This function does not return an error if something goes wrong but instead exits the whole process through the logger's `Fatal` method. Not sure that's necessarily the best design, but I didn't want to change that behavior in this PR. 

This `return` is reachable only in unit tests when the logger's ExitFunc is overridden.

driver/config/config.go
alnr: `tlsx.ErrNoCertificatesConfigured` is only returned by `tlsx.Certificate()`, which has little bit of a clunky interface. Instead of having the whole logic governing where certificates are read from (disk or base64-encoded in config value) inside a call to a library function, I moved this decision into this function. We call either `tlsx.CertificateFromBase64` or `tlsx.GetCertificate` depending on which config keys are set. If none are set, we return nil (no TLS).

It might make sense to make this logic even more strict and explicitly fail when the user tries to configure both a path and a base64 value (or a mixture of the two).

aeneasr: Thanks!

driver/config/config.go
aeneasr: Thanks! I overlooked that the return value is not an error, but a func

PR #2745, opened 2022-09-22
chore: add node version check to test/e2e/run.sh

### LINKED ISSUE #2738: test(e2e): check `node -v` when starting E2E tests
The E2E tests currently fail on newer Node versions, because node now prefers IPv6 which not all our example projects support. This leads to an endless loop of checking for the availability of these apps. 

We need to add a check to the `test/e2e/run.sh` script that validates that the callers node version is smaller than 18 and panic/exit if the version is incompatible.

jonas-jonas:
resolves #2738

This PR adds a (temporary) version check to the e2e run script to avoid timeout issues starting with node version 17 and up.

## Related issue(s)

#2738

## Checklist


piotrmsc (commented):
Change itself works, could be maybe bit simpler by grabbing and checking only major version but I have a question to GHA running e2e 
https://github.com/ory/kratos/blob/master/.github/workflows/ci.yaml#L109
it is using ubuntu-latest, won't this cause problems? I don't see nodejs version pinning there?

jonas-jonas:
@piotrmsc, yeah, I tried a lot of stuff to just get the major version, but couldn't figure out how to get regex capture groups working in bash. If you have an idea, lmk.

> it is using ubuntu-latest, won't this cause problems? I don't see nodejs version pinning there?

I think it is using the LTS version, which is 16.17, currently. But yes, for the time being, we should also pin the version somehow. I'll look into it.

jonas-jonas:
@piotrmsc I have cleaned up the check in the script and pinned the node version to 16 in the workflow file.

Could you take another look? :)

piotrmsc (approved):
works, but general question: maybe it's better to just adjust the scripts to work with newer versions of node?

jonas-jonas:
>works, but general question: why we don't in general adjust the scripts to work with newer versions of node?

Because that will require updating a bunch of dependencies in a bunch of projects (React, React Native, express examples) and then hope that they fixed the issue downstream. There is just too much going on to tell if this is an easy fix or if it will take longer. This change is just to save everyone some headache if they forget to change the Node version to 16.

PR #2769, opened 2022-09-30
chore: remove dead code

kelkarajay:
Removing tests dead code. These tests are currently not run and a more extensive test suite exists (including the ones in `session/persistence.go`) under `session/test/persistence.go`

### DIFF
--- a/session/persistence.go
+++ b/session/persistence.go
@@ -2,17 +2,9 @@ package session
 
 import (
 	"context"
-	"testing"
 	"time"
 
-	"github.com/bxcodec/faker/v3"
 	"github.com/gofrs/uuid"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-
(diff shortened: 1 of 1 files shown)


session/persistence.go
```diff
@@ -59,111 +51,3 @@ type Persister interface {
 	// RevokeSessionsIdentityExcept marks all except the given session of an identity inactive. It returns the number of sessions that were revoked.
 	RevokeSessionsIdentityExcept(ctx context.Context, iID, sID uuid.UUID) (int, error)
 }
-
-func TestPersister(ctx context.Context, conf *config.Config, p interface {
```
aeneasr, line 63: Please instead make sure that this test is executed!

aeneasr (changes requested):

session/persistence.go
aeneasr: Ah, misread the opening!

PR #2787, opened 2022-10-07
fix: take over return_to param from unauthorized settings to login flow

mmeller-wikia:
## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments
Preserve `return_to` query param when initiating the flow of browser settings.
Follow documented behaviour by actually initiating a new login flow (currently user is redirected to UI login page and redirect_to param is lost)

> // This endpoint initializes a browser-based user settings flow. Once initialized, the browser will be redirected to
// `selfservice.flows.settings.ui_url` with the flow ID set as the query parameter `?flow=`. If no valid
// Ory Kratos Session Cookie is included in the request, a login flow will be initialized.

**Failing testcase:** https://github.com/ory/kratos/pull/2788


selfservice/flow/settings/handler_test.go
```diff
@@ -142,10 +142,24 @@ func TestHandler(t *testing.T) {
 		})
 
 		t.Run("description=init a flow as browser", func(t *testing.T) {
-			t.Run("description=without privileges", func(t *testing.T) {
-				res, body := initSPAFlow(t, new(http.Client))
-				assert.Equal(t, http.StatusUnauthorized, res.StatusCode, "%s", body)
-				assert.Equal(t, text.ErrNoActiveSession, gjson.GetBytes(body, "error.id").String(), "%s", body)
```
mmeller-wikia, line 148: This test was moved to SPA flows

mmeller-wikia:
Hi @aeneasr
Can I ask for review?
Failing test case: https://github.com/ory/kratos/pull/2788

mmeller-wikia:
Hi @aeneasr 
I see you are back, so let me ping you :)

selfservice/flow/settings/handler_test.go
```diff
@@ -142,10 +142,24 @@ func TestHandler(t *testing.T) {
...
+				require.NoError(t, err)
+
+				res, err := c.Do(req)
+				require.NoError(t, err)
+				// here we check that the redirect status is 303
+				require.Equal(t, http.StatusSeeOther, res.StatusCode)
+				location, err := res.Location()
+				require.NoError(t, err)
+				require.Equal(t, publicTS.URL+login.RouteInitBrowserFlow+returnTo, location.String())
+				defer res.Body.Close()
```
jonas-jonas, line 162: If the `require`s prior to this fail, this is never executed, and thus the `Body` is never closed. 

I don't think we even do this a lot in the other tests, as they're quite short-lived processes and these resource leaks shouldn't really pile up. Nevertheless, it doesn't hurt to do this properly, but it should be moved to right after the `require.NoError` check of the response. See also https://pkg.go.dev/net/http#Client.Do ("On error, any Response can be ignored.").

jonas-jonas (changes requested):
Sorry for the delay. A lot going on with the Ory Summit :) 

Nice work! Thanks for the PR. I just found one issue in the tests.

selfservice/flow/settings/handler_test.go
mmeller-wikia: Fixed in https://github.com/ory/kratos/pull/2787/commits/cd252af32c90d6f979286e9ac2bc1d4f1cecd71e

mmeller-wikia:
Hi 
@jonas-jonas @aeneasr @zepatrik 
I'm pinging you because linter checks on master are failing (see https://github.com/ory/kratos/pull/2823 -> https://github.com/ory/kratos/actions/runs/3281486381/jobs/5403601490 )
On my last commit https://github.com/ory/kratos/pull/2787/commits/cd252af32c90d6f979286e9ac2bc1d4f1cecd71e only e2e tests failed (which may happen from time to time).
When I merged master to my branch linter checks starts failing out of nowhere.
It's very inconvenient for external contributors

jonas-jonas:
Hi @mmeller-wikia, I saw the failures, but didn't get to looking into it yet. I'll look into it after the Ory summit, sorry for the delay.

FWIW I guess the failures are due to unintentional updates in the ci tooling. Sorry about that.

selfservice/flow/settings/handler_test.go
```diff
@@ -142,10 +142,24 @@ func TestHandler(t *testing.T) {
...
+				require.NoError(t, err)
+
+				res, err := c.Do(req)
+				require.NoError(t, err)
+				defer res.Body.Close()
+				// here we check that the redirect status is 303
+				require.Equal(t, http.StatusSeeOther, res.StatusCode)
+				location, err := res.Location()
+				require.NoError(t, err)
+				require.Equal(t, publicTS.URL+login.RouteInitBrowserFlow+returnTo, location.String())
```
zepatrik, line 162: Maybe add a second case to verify it works without a `return_to` param as well. Just to prevent regressions.

selfservice/flow/settings/handler_test.go
```diff
@@ -173,13 +187,19 @@ func TestHandler(t *testing.T) {
 
 				res, err := c.Do(req)
 				require.NoError(t, err)
+				defer res.Body.Close()
 				// here we check that the redirect status is 303
 				require.Equal(t, http.StatusSeeOther, res.StatusCode)
-				defer res.Body.Close()
```
zepatrik, line 178: I guess just add a check for the location header here?

zepatrik (changes requested):
Nice, this looks very good :tada:
I'd just like to add one tiny test case (that should have been there already; or I missed it).

zepatrik:
The test failure seems to be just a flake... sorry for that.

selfservice/flow/settings/handler_test.go
mmeller-wikia: Done in https://github.com/ory/kratos/pull/2787/commits/656a838ad81ed455d508f1d32d7f92acfa6ada55

mmeller-wikia:
@zepatrik 
PR is ready for CR :)

zepatrik (approved):
Awesome :tada:

Issue #2633, opened 2022-08-01
Privileged Session state should be included in `whoami` response


zepatrik:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Clients currently have no easy way to determine whether a session is privileged.

From https://ory-community.slack.com/archives/C012RJ2MQ1H/p1659099978177669

> 1. I am calling self-service/settings/browser
> 2.  I use the return value to get the flow id and the csrf token
> 3.  I call /self-service/settings with the flow id, the csrf token and a NEW PASSWORD
> 4.  I receive error session_refresh_required with a redirect_browser_to url
> 5.  I redirect to the specified url /self-service/login/browser which contains query params for refresh=true and a return_to url that is /self-service/settings
> 6.  I log in with the user's credentials with refresh true
> 7.  I call the /self-service/settings
> 8.  The PASSWORD doesn't seem to be updated

Also requested by @tc-amorevino

### Describe your ideal solution

It would be nice for the UX to be able to see whether the `Update Password` field should be displayed, or instead a `login?refresh=true` flow should be initiated. A `privileged_until` timestamp will probably be better for the technical side than a boolean, because the client could preemptively trigger the re-login if the session would be non-privileged within a short time. It also means the response does not change over time (assuming the session is not revoked/prolonged). 

### Workarounds or alternatives

Determine the "privileged" status on the client side by knowing the privileged session lifespan.

### Version

all

brahmlower:
I started working on implementing this feature this weekend. I've got a branch in progress with the implementation working, but it's taking longer than I anticipated to get the tests updated. I think I'm relatively close to getting the tests finished though- hopefully in the next day or two.

zepatrik:
Awesome!

brahmlower:
[...]

Issue #2620, opened 2022-07-27
Inconsistent Timezones


mike-capyh:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

Kratos sometimes uses UTC and sometimes uses system time for times that probably should be in the same timezone.
This can cause immediately-stale flows when system time has a sufficiently negative UTC offset
(eg bare-metal installation of Kratos on OS set to "America/Los_Angeles (PDT, -0700)"
and might cause unintentionally-long-lived flows when the OS has a positive UTC offset.

An example of this is in the selfservice registration flow, note the Now().UTC() vs naked Now() below:

https://github.com/ory/kratos/blob/master/selfservice/flow/registration/flow.go#L88
```
func NewFlow(conf *config.Config, exp time.Duration, csrf string, r *http.Request, ft flow.Type) (*Flow, error) {
	now := time.Now().UTC()
```

https://github.com/ory/kratos/blob/master/selfservice/flow/registration/handler.go#L342
```
	if ar.ExpiresAt.Before(time.Now()) {
```

On Linux, `timedatectl set-timezone UTC` can be used to work around the issue,
and `timedatectl set-timezone America/Los_Angeles` can be used to reproduce the stale registration flow.

### Reproducing the bug

1. Set system time to America/Los_Angeles (or any other sufficiently negative UTC offset timezone).
2. Initiate and attempt to complete registration browser flow.
3. Kratos errors with 410 Gone and a message to the effect that the flow is expired.  For the example LA timezone, the flow will be about 6 hours expired (if PDT) due to the -7 offset and the apparent 1 hour default expiration time.

### Version

v0.10.1 (build commit: ab16580b4326250885b920198b280456eb873a6b)

### On which operating system are you observing this issue?

Linux

### In which environment are you deploying?

Binary

[...]

PR #2793, opened 2022-10-11
feat: #2633 session includes privileged_until timestamp

brahmlower:
Added the `privileged_until` property to sessions. The source of truth for "is this session privileged?" is now the `session.Session.IsPrivileged` method, away from the impromptu comparisons of `time.Now()` with `privileged_session_max_age` setting. This change requires a database migration.

## Related issue(s)

https://github.com/ory/kratos/issues/2633

I forgot to check for a design conversation in the github discussions, but it looked like the feature was well considered based on the notes in the original issue. Happy to do re-work if changes to behavior are desired.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

1. An edge case exists where sessions that would be considered privileged before migrating will be considered unprivileged after migrating, since existing sessions will not be retroactively updated with a privileged until timestamp.
2. The `session.Session.IsPrivileged` helper method was useful for reducing repetition in checking if the session is privileged or not. It currently makes no assertion about the session being expired or active. I'm happy to include those comparisons in the conditional if folks prefer?

I left my commits unsquashed to help with the review process, in case folks prefer to review commit-by-commit. I can squash these all down to a single commit before starting review if that's preferable.

Remaining items
- [ ] Update documentation including api spec file (I completely forgot about this- will have to get to it tomorrow)
- [ ] Double check date ordering on migrations
- [x] Circle back and address clarifying questions from development (left a few TODO/FIXMEs)


brahmlower:
This PR is larger than I expected it'd be when I first started 😅 I will do a self review of this tomorrow evening (Tuesday, Oct 11) to find and address anything errors/cleanup items I might've missed during development.

selfservice/strategy/lookup/settings_test.go
```diff
@@ -126,7 +136,14 @@ func TestCompleteSettings(t *testing.T) {
...
 		id, _ := createIdentity(t, reg)
-		browserClient := testhelpers.NewHTTPClientWithIdentitySessionToken(t, reg, id)
+		// TODO: Review this particular diff- I suspect there was a bug since `browserClient` was
+		// actually a token client, not a cookie client
+		browserClient := testhelpers.NewHTTPClientBuilder(t).
+			SetReqestFromWhoAmI().
+			SetIdentity(id).
+			SetSessionDefault().
+			ClientWithSessionCookie(reg)
+			// ClientWithSessionToken(reg)
```
brahmlower, line 146: I did want to call this change out. Previously, the variable name implied the client should be created with a cookie (for the browser), but was calling the token function.

I just wanted to double check I should be building the client with the cookie and not a token here?

selfservice/strategy/lookup/settings_test.go
```diff
@@ -139,15 +156,26 @@ func TestCompleteSettings(t *testing.T) {
...
 		id := createIdentityWithoutLookup(t, reg)
-		browserClient := testhelpers.NewHTTPClientWithIdentitySessionToken(t, reg, id)
+		// TODO: Review this particular diff- I suspect there was a bug since `browserClient` was
+		// actually a token client, not a cookie client
+		browserClient := testhelpers.NewHTTPClientBuilder(t).
+			SetReqestFromWhoAmI().
+			SetIdentity(id).
+			SetSessionDefault().
+			ClientWithSessionCookie(reg)
+			// ClientWithSessionToken(reg)
```
brahmlower, line 178: Same question as https://github.com/ory/kratos/pull/2793/files#r992046174

persistence/sql/migrations/sql/20221009000000000000_add_session_privileged_until_column.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "sessions" ADD COLUMN "privileged_until" DATETIME DEFAULT NULL;
```
brahmlower, line 1: Hmm, tests are failing on `DATETIME` for sqlite. I thought I had verified this locally since the short tests use sqlite 🤔

selfservice/strategy/lookup/settings_test.go
```diff
@@ -160,7 +188,11 @@ func TestCompleteSettings(t *testing.T) {
 		})
 
 		t.Run("case=api", func(t *testing.T) {
-			apiClient := testhelpers.NewHTTPClientWithIdentitySessionToken(t, reg, id)
+			apiClient := testhelpers.NewHTTPClientBuilder(t).
+				SetReqestFromWhoAmI().
+				SetIdentity(id).
+				SetSessionDefault().
+				ClientWithSessionToken(reg)
```
brahmlower, line 195: This should be using the `NewIdentityClientWithSessionToken` helper

selfservice/flow/settings/handler_test.go
```diff
@@ -367,8 +399,8 @@ func TestHandler(t *testing.T) {
 
 		t.Run("description=fail to submit form as another user", func(t *testing.T) {
 			t.Run("type=api", func(t *testing.T) {
-				user1 := testhelpers.NewHTTPClientWithArbitrarySessionToken(t, reg)
-				user2 := testhelpers.NewHTTPClientWithArbitrarySessionToken(t, reg)
+				user1 := testhelpers.NewDefaultClientWithSessionCookie(t, reg)
+				user2 := testhelpers.NewDefaultClientWithSessionCookie(t, reg)
```
brahmlower, line 403: Should be using `NewDefaultClientWithSessionToken`

selfservice/flow/settings/handler_test.go
```diff
@@ -256,8 +288,8 @@ func TestHandler(t *testing.T) {
 
 		t.Run("description=should fail to fetch request if identity changed", func(t *testing.T) {
 			t.Run("type=api", func(t *testing.T) {
-				user1 := testhelpers.NewHTTPClientWithArbitrarySessionToken(t, reg)
-				user2 := testhelpers.NewHTTPClientWithArbitrarySessionToken(t, reg)
+				user1 := testhelpers.NewDefaultClientWithSessionCookie(t, reg)
+				user2 := testhelpers.NewDefaultClientWithSessionCookie(t, reg)
```
brahmlower, line 292: Should be using NewDefaultClientWithSessionToken

selfservice/flow/settings/handler_test.go
```diff
@@ -228,7 +260,7 @@ func TestHandler(t *testing.T) {
 			returnTo := "https://www.ory.sh"
 			conf.MustSet(ctx, config.ViperKeyURLsAllowedReturnToDomains, []string{returnTo})
 
-			client := testhelpers.NewHTTPClientWithArbitrarySessionToken(t, reg)
+			client := testhelpers.NewDefaultClientWithSessionCookie(t, reg)
```
brahmlower, line 263: Should be using `NewDefaultClientWithSessionToken`

I missed a lot of these 😬

selfservice/flow/settings/handler_test.go
```diff
@@ -181,7 +200,7 @@ func TestHandler(t *testing.T) {
 
 		t.Run("description=init a flow as SPA", func(t *testing.T) {
 			t.Run("description=success", func(t *testing.T) {
-				user1 := testhelpers.NewHTTPClientWithArbitrarySessionToken(t, reg)
+				user1 := testhelpers.NewDefaultClientWithSessionCookie(t, reg)
```
brahmlower, line 203: Should be using `NewDefaultClientWithSessionToken`

selfservice/flow/settings/handler_test.go
```diff
@@ -127,7 +146,7 @@ func TestHandler(t *testing.T) {
 			})
 
 			t.Run("description=success", func(t *testing.T) {
-				user1 := testhelpers.NewHTTPClientWithArbitrarySessionToken(t, reg)
+				user1 := testhelpers.NewDefaultClientWithSessionCookie(t, reg)
```
brahmlower, line 149: Should be using `NewDefaultClientWithSessionToken`

selfservice/flow/settings/handler_test.go
brahmlower: Actually, this might've been incorrect in the original test- the SPA tests generally use the cookie clients, so I think this change is okay.

selfservice/flow/settings/handler_test.go
brahmlower: I think this change from token to cookie client is okay, since the we're testing browser flows.

selfservice/flow/settings/handler_test.go
```diff
@@ -60,12 +60,18 @@ func TestHandler(t *testing.T) {
 	_ = testhelpers.NewErrorTestServer(t, reg)
 	_ = testhelpers.NewSettingsUIFlowEchoServer(t, reg)
 
-	conf.MustSet(ctx, config.ViperKeySelfServiceSettingsPrivilegedAuthenticationAfter, "1ns")
```
brahmlower, line 63: This set all sessions in this test suite to be unprivileged by default. This similar behavior is achieved by creating the `primaryUser` and `otherUser` clients using the `testhelpers.UnprivilegedProvider()`

selfservice/strategy/password/settings_test.go
```diff
@@ -158,20 +157,30 @@ func TestSettings(t *testing.T) {
...
-			t.Cleanup(func() {
-				conf.MustSet(ctx, config.ViperKeySelfServiceSettingsPrivilegedAuthenticationAfter, "5m")
-			})
+
+			var sessionMutator func(mutateFunc func(session *session.Session))
+			browserUser1 := testhelpers.NewHTTPClientBuilder(t).
+				SetReqestFromWhoAmI().
+				SetIdentity(browserIdentity1).
+				SetSessionWithProvider(testhelpers.UnprivilegedProvider()).
+				RecordSessionMutator(reg, &sessionMutator).
```
brahmlower, line 167: Here we're able to get a pointer to a function (saved to `sessionMutator`) for mutating the session at a later point. It's used on like 201 to later set the session to privileged, as if the user had refreshed their session.

selfservice/strategy/profile/strategy_test.go
```diff
@@ -88,14 +72,35 @@ func TestStrategyTraits(t *testing.T) {
 	browserIdentity2 := &identity.Identity{ID: x.NewUUID(), Traits: identity.Traits(`{}`), State: identity.StateActive}
 	apiIdentity2 := &identity.Identity{ID: x.NewUUID(), Traits: identity.Traits(`{}`), State: identity.StateActive}
 
-	browserUser1 := testhelpers.NewHTTPClientWithIdentitySessionCookie(t, reg, browserIdentity1)
-	browserUser2 := testhelpers.NewHTTPClientWithIdentitySessionCookie(t, reg, browserIdentity2)
-	apiUser1 := testhelpers.NewHTTPClientWithIdentitySessionToken(t, reg, apiIdentity1)
-	apiUser2 := testhelpers.NewHTTPClientWithIdentitySessionToken(t, reg, apiIdentity2)
```
brahmlower, line 94: It was easier to just create these clients as needed in each test, rather than using a shared value and having to reset the privileged state each time.

selfservice/strategy/profile/strategy_test.go
```diff
@@ -296,8 +306,6 @@ func TestStrategyTraits(t *testing.T) {
 
 	t.Run("description=should end up at the login endpoint if trying to update protected field without sudo mode", func(t *testing.T) {
 		var run = func(t *testing.T, config *kratos.SelfServiceSettingsFlow, isAPI bool, c *http.Client) *http.Response {
-			time.Sleep(time.Millisecond)
```
brahmlower, line 299: Saved ourselves an extra 1 millisecond on the tests 😉

selfservice/strategy/profile/strategy_test.go
```diff
@@ -309,30 +317,31 @@ func TestStrategyTraits(t *testing.T) {
...
+			browserUser1 := getBrowserUser(false, browserIdentity1)
 			f := testhelpers.InitializeSettingsFlowViaBrowser(t, browserUser1, false, publicTS)
 			res := run(t, f, false, browserUser1)
 			assert.EqualValues(t, http.StatusUnauthorized, res.StatusCode)
 			assert.Contains(t, res.Request.URL.String(), conf.GetProvider(ctx).String(config.ViperKeySelfServiceLoginUI))
 
 			t.Run("should update when signed back in", func(t *testing.T) {
-				setPrivileged(t)
+				browserUser1 := getBrowserUser(true, browserIdentity1)
+				_ = run(t, f, false, browserUser1)
```
brahmlower, line 344: IIRC there's a side effect in the run function that's needed for the `browserUser1.get(f.Ui.Action)` on the next line to work correctly. This "solution" of creating a new client and then essentially re-doing the first part of this test (from line 338) is a little hacky, and difficult to understand in the long run.

I wrote this before I added the `RecordSessionMutator` functionality to the session test helpers. This test should be changed to use that session mutator and just change the session to privileged here rather than re-creating a client

selfservice/strategy/profile/strategy_test.go
```diff
@@ -512,10 +525,13 @@ func TestStrategyTraits(t *testing.T) {
 	})
 
 	// Update the login endpoint to auto-accept any incoming login request!
-	_ = testhelpers.NewSettingsLoginAcceptAPIServer(t, testhelpers.NewSDKCustomClient(publicTS, browserUser1), conf)
+	// TODO: This causes some side effect, possibly on the old browserUser1 client- I'm not sure what the implications
+	// of commenting this out are. I'll follow up on this before opening this for final review
+	// _ = testhelpers.NewSettingsLoginAcceptAPIServer(t, testhelpers.NewSDKCustomClient(publicTS, browserUser1), conf)
```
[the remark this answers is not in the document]
brahmlower: Yup, this looks to be the case. Will remove this comment 👍

session/session.go
```diff
@@ -271,8 +293,23 @@ func (s *Session) IsActive() bool {
...
 }
 
-func (s *Session) Refresh(ctx context.Context, c lifespanProvider) *Session {
+func (s *Session) SetPrivilegedUntil(t time.Time) {
+	until := sqlxx.NullTime(t)
+	s.PrivilegedUntil = &until
+}
+
+func (s *Session) IsPrivileged() bool {
+	fmt.Printf("IsPrivileged is handling a null value: %v\n", s.PrivilegedUntil == nil)
```
brahmlower, line 302: Remove debug output

session/session.go
```diff
@@ -271,8 +293,23 @@ func (s *Session) IsActive() bool {
...
+	// TODO: I think s.PriviledgedUntil can be nil because the sqlxx.NullTime can be a nil pointer,
+	// and we'd be doing a nil pointer dereference
```
brahmlower, line 304: add nil guard

selfservice/flow/settings/error_test.go
```diff
@@ -144,7 +144,15 @@ func TestHandleError(t *testing.T) {
...
+				s, err := session.NewActiveSession(
+					req,
+					&id,
+					testhelpers.NewSessionLifespanAndPrivilegedMaxAgeProvider(time.Hour, time.Minute),
+					time.Now(),
+					identity.CredentialsTypePassword,
+					identity.AuthenticatorAssuranceLevel1,
+				)
 				require.NoError(t, err)
 				c := testhelpers.NewHTTPClientWithSessionToken(t, reg, s)
```
brahmlower, line 157: `c := NewIdentityClientWithSessionToken(t, reg, id)`

internal/testhelpers/session.go
```diff
@@ -33,6 +33,48 @@ func NewSessionLifespanProvider(expiresIn time.Duration) *SessionLifespanProvide
...
+}
+
+func NewSessionPrivilegedMaxAgeProvider(privilegedMaxAge time.Duration) *SessionPrivilegedMaxAgeProvider {
+	return &SessionPrivilegedMaxAgeProvider{e: privilegedMaxAge}
+}
+
+type SessionLifespanAndPrivilegedMaxAgeProvider struct {
+	l *SessionLifespanProvider
+	p *SessionPrivilegedMaxAgeProvider
+}
```
brahmlower, line 51: I joined the lifespan and privileged max age providers to avoid having to essentially pass the same config in twice. But the name of this provider is large and unwieldy 🤔

internal/testhelpers/session.go
```diff
@@ -136,92 +178,261 @@ func NewHTTPClientWithSessionToken(t *testing.T, reg *driver.RegistryDefault, se
...
+func (c *HTTPClientBuilder) ClientNoRedirectWithSessionCookie(reg *driver.RegistryDefault) *http.Client {
+	c.nilSessionCheck()
+	return c.ClientCallback(func(t *testing.T, s *session.Session) *http.Client {
+		return NewNoRedirectHTTPClientWithSessionCookie(t, reg, s)
 	})
 }
+
+// Common configurations
+
+func NewIdentityClientWithSessionToken(t *testing.T, reg *driver.RegistryDefault, id *identity.Identity) *http.Client {
```
brahmlower, line 392: These functions provide essentially the same interface as the the `NewHTTPClientWithIdentitySessionCookie` functions that were removed. The change in name is supposed to help communicate how the client is being built.

internal/testhelpers/session.go
```diff
@@ -136,92 +178,261 @@ func NewHTTPClientWithSessionToken(t *testing.T, reg *driver.RegistryDefault, se
...
-		time.Now(),
-		identity.CredentialsTypePassword,
-		identity.AuthenticatorAssuranceLevel1,
+// A builder struct to help customize an authenticated http client
+type HTTPClientBuilder struct {
+	t        *testing.T
+	request  *http.Request
+	identity *identity.Identity
+	session  *session.Session
+}
```
brahmlower, line 207: I opted to use the builder pattern for these clients because it gives a more flexibility interface while building the clients. With it, we can present the same interface that most consumers of these helpers were already using (using the wrappers at the bottom of this file), but in the rare cases where more customization is needed we can use the builder methods directly however we need.

internal/testhelpers/session.go
```diff
@@ -136,92 +178,261 @@ func NewHTTPClientWithSessionToken(t *testing.T, reg *driver.RegistryDefault, se
...
+//
+//		// Pretend the session has been refreshed, and is now privileged for an additional minute
+//		sessionMutator(func(session *session.Session) {
+//			session.SetPrivilegedUntil(time.Now().Add(time.Minute))
+//		})
+//
+//		// Run some tests using `client` requiring the session to be privileged
+//		...
+//	}
+func (c *HTTPClientBuilder) RecordSessionMutator(reg *driver.RegistryDefault, funcRef *func(func(session *session.Session))) *HTTPClientBuilder {
```
brahmlower, line 348: I figured this function could be confusing to come across in the few instances where we us it, so I wanted to make sure I provided some detailed docs on its usage and purpose for existing.

internal/testhelpers/session.go
```diff
@@ -136,92 +178,261 @@ func NewHTTPClientWithSessionToken(t *testing.T, reg *driver.RegistryDefault, se
...
+		c.t.Errorf("creating a Session requires an identity to be present on the builder, but identity was nil")
+	}
+}
+
+// Raise a test error if the session hasn't been assigned yet.
+func (c *HTTPClientBuilder) nilSessionCheck() {
+	if c.session == nil {
+		c.t.Errorf("consuming the HTTPClientBuilder requires a session to be present on the builder, but session was nil")
+	}
+}
```
brahmlower, line 232: Nil checks all error out immediately because it means the test has been written incorrectly.

internal/testhelpers/session.go
```diff
@@ -136,92 +178,261 @@ func NewHTTPClientWithSessionToken(t *testing.T, reg *driver.RegistryDefault, se
...
+		SetSessionDefault().
+		ClientWithSessionToken(reg)
+}
+
+func NewIdentityClientWithSessionCookie(t *testing.T, reg *driver.RegistryDefault, id *identity.Identity) *http.Client {
+	return NewHTTPClientBuilder(t).
+		SetReqestFromWhoAmI().
+		SetIdentity(id).
+		SetSessionDefault().
+		ClientWithSessionToken(reg)
```
brahmlower, line 405: This need to be calling `ClientWithSessionCookie`

persistence/sql/migrations/sql/20221009000000000000_add_session_privileged_until_column.up.sql
brahmlower: I might've just misread the error last night- it looks like the sqlite tests were working, but we needed migrations specific for cockroach and postgres.

selfservice/strategy/lookup/settings_test.go
brahmlower: I've gone ahead and stuck with using the client w/ cookies on this. Tests seemed to pass 🤷

selfservice/strategy/lookup/settings_test.go
brahmlower: (same change here as the linked comment)

brahmlower:
I still need to update docs, but I'll have to get to that tomorrow- already burnt too much of the midnight oil today 😁

All of my remaining comments are just adding context to my decision making, or pointing out noteworthy changes, and don't necessarily require any followup.

brahmlower:
Hmm those CI failures are unfortunate too :/ I distinctly remember looking at those `conf.MustSet(ctx, config.ViperKeySelfServiceSettingsPrivilegedAuthenticationAfter, time.Millisecond)` lines. I thought I had stripped those all out and replaced them with the session helpers, but maybe I lost changes to the `strategy_settings_test.go` file to poor git usage or something 🤔 I dunno. Regardless, I'll fix up those tests tomorrow

brahmlower:
Rebased off the master branch (thus the force push). I couldn't replicate the test failures from CI locally, so I'm not 100% sure my latest commit will resolve the failures. I also started on the doc updates by regenerated the sdk, but that that cause a lot of tests to break again, so I'm still working on getting that figured out. I doubt I'll have this ready for review until Monday at the earliest

brahmlower:
I've been stretched a bit thin this past week and haven't been able to find the time/energy needed for this PR. Haven't forgotten about it though 🙂

brahmlower:
I've been trying to fix the failing CI tests that I mentioned in the first part of [this message](https://github.com/ory/kratos/pull/2793#issuecomment-1278543456), but it's been tricky. I was able to reproduce the test failure locally (I learned I should be running `make test-coverage`), and pushed the attempted fix in [e199c54](https://github.com/ory/kratos/pull/2793/commits/e199c5495cec454acd47a3e360dee9ebff5970d5), but clearly missed another few instances in that file that needed to be updated. I'm starting to wonder if the [inconsistent timezone issue](https://github.com/ory/kratos/issues/2620) is playing a role here 🤔

brahmlower:
Okay the unit tests are working again, but the e2e tests look like they're timing out. I thought I'd seen this a few times before so I'm not sure if this is just a flakey test, or if I've actually broken something. I tried to replicate the failure locally, but the MDM software we use is preventing me from running the e2e tests, so that's my blocker for now. I'll try to get that sorted out with our IT folks on Monday.

aeneasr (commented):
Sorry for the super late review!

I'm wondering whether we need to store this field in the database. Would it not be enough to just hydrate this extra field when reading the session from the database? This would not work very well for eager relationships but I'm not sure whether we actually need that field in that scenario.

If we store it in the database, it means that the sessions will need to use this field over the global config. If an admin messes up, sets time to 10d, then quickly resets it to 10m, there would be some sessions with 10d privileged time.

shoaibmerchant:
@brahmlower any progress on this feature, this will be helpful in implementing something like a 'sudo' mode on frontend

brahmlower:
Hey @shoaibmerchant and @aeneasr 👋 Super sorry for my lack of updates here- my team changed projects pretty abruptly around Oct/Nov 2022 and I pretty much forgot I had this PR open. I haven't worked on this branch since my last commit, so I don't have any useful updates. I am hoping to be switching back to projects that'll give me time to drive this PR more within the next couple months, but can't make any promises.

To answer @aeneasr's questions:

> I'm wondering whether we need to store this field in the database. Would it not be enough to just hydrate this extra field when reading the session from the database?

At the time, my thinking was that having an extra step of indirection between the settings and the value read from the session object would be valuable. That way the duration is set on creation so that a session doesn't unexpectedly become unprivileged after the setting is changed to a really short value.

> If we store it in the database, it means that the sessions will need to use this field over the global config. If an admin messes up, sets time to 10d, then quickly resets it to 10m, there would be some sessions with 10d privileged time.

This is an excellent point! I think we should definitely take that approach instead of the storing the value in the db. This PR will need a fair bit of rework then.


On Monday (March 13) I'll see if I can plan some time for this in the next couple months, and will let y'll know what my timeline looks like from there.

aeneasr:
@brahmlower great to see you back! I believe that you would have an easier time creating a new PR for this. My thinking is that you probably will just include the `privileged_until` timestamp in the whoami endpoint, and compute it dynamically. When naming/documenting it, make sure that it is clear that the time indicated is an **estimate** as it can change every time the admin changes the config value. So it's more a estimate than a hard value.

brahmlower:
Oh that's a great point- cutting a fresh branch would probably be way less work 😅

I haven't had a chance to formally include this work in my next couple sprints (was dealing with covid last week, and have been catching all this week), but we have a small project winding down next week so I think I can dedicate some time for this issue then.

I'm going to close this PR, and should have a fresh one open next week 👍
 
(cc folks watching https://github.com/ory/kratos/issues/2633)

Issue #2513, opened 2022-06-09
Settings flow should not return UI nodes for metadata fields


woylie:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

I added a field to the public metadata:

```json
{
  "properties": {
    "metadata_public": {
      "type": "object",
      "properties": {
        "picture": {
          "type": "string",
          "format": "uri"
        }
      },
      "additionalProperties": false
    },
    "traits": {}
}
```

Since the metadata shouldn't be editable by the user, I would expect that the UI container for the settings flow does not contain any input nodes for the picture field. However, Kratos is returning an input for the metadata field:

```json
{
  "attributes": {
    "name": "metadata_public.picture",
    "node_type": "input",
    "type": "url"
  },
  "group": "profile",
  "messages": [],
  "meta": {},
  "type": "input"
}
```

### Reproducing the bug

1. add a metadata field to an identity schema
2. initiate a settings flow or registration flow

### Version

0.10.1

### In which environment are you deploying?

Docker Compose

aeneasr:
The field `public_metadata` is not validated against the JSON Schema. Adding it to the schema has no effect :)

woylie:
That's good to know. But wouldn't we want that?

aeneasr:
Not sure, the metadata fields are set arbitrary and have no implication on Ory Kratos' business logic. The JSON Schema for traits is used when validating UI input, generating UI nodes, etc.

The only reason could be enforcement of a schema, but at the cost of added learning complexity when starting out with Ory Kratos.

woylie:
Kratos is the last thing before the database. I think it would be reasonable to enforce the schema before persisting the data at that point. Apart from validation, it would also be nice to have the metadata format as part of the schema, so that anybody who handles identities knows what to expect.

[...]

PR #2820, opened 2022-10-18
fix: validate identity metadata against schema

zepatrik:
Currently `metadata_admin` and `metadata_public` are not validated against the identity schema.


aeneasr:
Please also address:

- [ ] https://github.com/ory/kratos/issues/2513

aeneasr (changes requested):
@zepatrik will you continue working on this?

PR #2855, opened 2022-11-02
feat: add admin get api for session

kelkarajay:
## Changes
- Adding an API to get session with expand options
- Refactor Session Listing to add expand options to query params and default to expand nothing

## Related issue(s)

https://github.com/ory-corp/cloud/issues/2007

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -36,7 +36,8 @@ func (p *Persister) GetSession(ctx context.Context, sid uuid.UUID, expandables s
 	nid := p.NetworkID(ctx)
 
 	q := p.GetConnection(ctx).Q()
-	if len(expandables) > 0 {
+	// if len(expandables) > 0 {
+	if expandables.Has(session.ExpandSessionDevices) {
 		q = q.Eager(expandables.ToEager()...)
 	}
 
@@ -75,7 +76,9 @@ func (p *Persister) ListSessions(ctx context.Context, active *bool, paginatorOpt
 		if active != nil {
 			q = q.Where("active = ?", *active)
 		}
-		if len(expandables) > 0 {
+
+		// if len(expandables) > 0 {
+		if expandables.Has(session.ExpandSessionDevices) {
 			q = q.Eager(expandables.ToEager()...)
 		}
 
--- a/session/handler.go
+++ b/session/handler.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/ory/x/pagination/keysetpagination"
@@ -344,7 +343,7 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
 
 	var expandables Expandables
 	if urlValues.Has("expand") {
-		es := strings.Split(urlValues.Get("expand"), ",")
+		es := urlValues["expand"]
 		for _, e := range es {
 			expand, ok := ParseExpandable(e)
 			if !ok {
--- a/session/handler.go
+++ b/session/handler.go
@@ -277,7 +277,9 @@ type adminListSessionsRequest struct {
 	// in: query
 	Active bool `json:"active"`
 
-	// ExpandOptions is a comma seperated list of all properties that must be expanded in the Session. If no value is provided, the expandable properties are skipped.
+	// ExpandOptions is a query parameter encoded list of all properties that must be expanded in the Session.
+	// Example - ?expand=Identity&expand=Devices
+	// If no value is provided, the expandable properties are skipped.
 	//
 	// enum: Identity,Devices
 	// required: false
@@ -372,7 +374,9 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
 // swagger:parameters adminGetSession
 // nolint:deadcode,unused
 type adminGetSessionRequest struct {
-	// ExpandOptions is a comma seperated list of all properties that must be expanded in the Session. If no value is provided, the expandable properties are skipped.
+	// ExpandOptions is a query parameter encoded list of all properties that must be expanded in the Session.
+	// Example - ?expand=Identity&expand=Devices
+	// If no value is provided, the expandable properties are skipped.
 	//
 	// enum: Identity,Devices
 	// required: false
--- a/session/handler.go
+++ b/session/handler.go
@@ -344,8 +344,7 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
 	}
 
 	var expandables Expandables
-	if urlValues.Has("expand") {
-		es := urlValues["expand"]
+	if es, ok := urlValues["expand"]; ok {
 		for _, e := range es {
 			expand, ok := ParseExpandable(e)
 			if !ok {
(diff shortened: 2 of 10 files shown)


session/handler.go
```diff
@@ -349,6 +366,83 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
...
+//	Responses:
+//	  200: session
+//	  400: jsonError
+//	  401: jsonError
+//	  404: jsonError
+//	  500: jsonError
+func (h *Handler) adminGetSession(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	if ps.ByName("id") == "whoami" {
+		// for /admin/sessions/whoami redirect to the public route
+		x.RedirectToPublicRoute(h.r)(w, r, ps)
```
aeneasr, line 412: Is this a replacement for https://github.com/ory/kratos/pull/2855/files#diff-abfcc5544632668aab2f1162de82ec56e8ec7444f6ce377cb3a30a9eafbed198L75 ?

session/handler.go
```diff
@@ -349,6 +366,83 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
...
+
+	var expandables Expandables
+
+	urlValues := r.URL.Query()
+	if urlValues.Has("expand") {
+		es := strings.Split(urlValues.Get("expand"), ",")
+		for _, e := range es {
+			expand, ok := ParseExpandable(e)
+			if !ok {
+				h.r.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError("could not parse expand option").WithDebug("could not parse expand option"))
```
aeneasr, line 430: Since the error message and debug message are the same, it's probably enough to just have one of the two. And we usually wrap things in `errors.WithStack()` and use `WithReason` to add error context :)

```suggestion
				h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf("Could not parse expand option: %s", e))))
```

session/handler.go
```diff
@@ -349,6 +366,83 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
...
+	if err != nil {
+		h.r.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(err.Error()).WithDebug("could not parse UUID"))
+		return
+	}
+
+	var expandables Expandables
+
+	urlValues := r.URL.Query()
+	if urlValues.Has("expand") {
+		es := strings.Split(urlValues.Get("expand"), ",")
```
aeneasr, line 426: An alternative would be to design the API like this:

```
?expand=foo&expand=bar&expand=baz
```

WDYT?

session/handler.go
```diff
@@ -349,6 +366,83 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
...
+//
+//	Security:
+//	  oryAccessToken:
+//
+//	Responses:
+//	  200: session
+//	  400: jsonError
+//	  401: jsonError
+//	  404: jsonError
+//	  500: jsonError
```
aeneasr, line 408:
```suggestion
//	  default: jsonError
```

aeneasr (commented):
This looks great!

session/handler.go
kelkarajay: Yes, the router panics on adding a `/admin/sessions/whoami` because it can't tell it apart from `/admin/sessions/<some string>` and this is a workaround to deal with that

session/handler.go
kelkarajay: @aeneasr I went through some reading for this design. (https://medium.com/raml-api/arrays-in-query-params-33189628fa68).

> What it looks like:
> 
> `?fields=kind,items(title,characteristics/length)`

The [Google drive approach](https://developers.google.com/drive/api/guides/performance#partial-response) enables us to have fine grained fields in addition to property expands (not currently supported). If this isn't something we need, then what you suggested seems better

aeneasr: Cool, that makes sense! Can you please add this to the API design document once the work around the session listing is done? :)

persistence/sql/persister_session.go
```diff
@@ -36,7 +36,8 @@ func (p *Persister) GetSession(ctx context.Context, sid uuid.UUID, expandables s
 	nid := p.NetworkID(ctx)
 
 	q := p.GetConnection(ctx).Q()
-	if len(expandables) > 0 {
+	// if len(expandables) > 0 {
+	if expandables.Has(session.ExpandSessionDevices) {
```
kelkarajay, line 40: This is a temporary hack to get around the `Devices` eager query despite `expandables.ToEager()...` returning an empty slice (`Identity` expandable does not count as a regular expandable).

kelkarajay: @zepatrik context to the slack conversation ^

session/handler.go
```diff
@@ -349,6 +367,83 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
...
+// The request object for getting a session in an administrative context.
+//
+// swagger:parameters adminGetSession
+// nolint:deadcode,unused
+type adminGetSessionRequest struct {
+	// ExpandOptions is a query parameter encoded list of all properties that must be expanded in the Session.
+	// Example - ?expand=Identity&expand=Devices
+	// If no value is provided, the expandable properties are skipped.
+	//
+	// enum: Identity,Devices
```
aeneasr, line 381: Wow! I did not know this works! Ping @zepatrik !

```suggestion
	// enum: identity,devices
```

session/handler.go
```diff
@@ -338,7 +343,20 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
 		return
 	}
 
-	sess, total, nextPage, err := h.r.SessionPersister().ListSessions(r.Context(), active, opts, ExpandEverything)
+	var expandables Expandables
+	if urlValues.Has("expand") {
+		es := urlValues["expand"]
```
aeneasr, line 348: You *could* reduce nesting here:

```suggestion
	if es, ok := urlValues["expand"]; ok {
```

session/handler.go
```diff
@@ -349,6 +367,83 @@ func (h *Handler) adminListSessions(w http.ResponseWriter, r *http.Request, ps h
...
+	if err != nil {
+		h.r.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(err.Error()).WithDebug("could not parse UUID"))
+		return
+	}
+
+	var expandables Expandables
+
+	urlValues := r.URL.Query()
+	if urlValues.Has("expand") {
+		es := urlValues["expand"]
```
aeneasr, line 427:
```suggestion
	if es, ok := urlValues["expand"]; ok {
		
```

aeneasr (commented):
Very nice, just some minor and quickly addressable comments :)

session/handler.go
kelkarajay: Screenshot 2022-11-04 at 09 53 13

Yes, it does put this out in the spec like in the screenshot. But it still doesn't work when I try to get this to link to the definition. So ```
ExpandOptions []Expandable `json:"expand"` ``` would be the right thing to do imo to restrict the list options to the enum ones.

session/handler.go
kelkarajay: done

session/handler.go
kelkarajay: done

session/handler.go
aeneasr: I see, you can take a look into our internal docs to see how we usually solve enums: https://intranet.ory.sh/docs/dev/api-guidelines#enums

PR #2867, opened 2022-11-07
feat: add handler with openapi def for admin revoke session

kelkarajay:
### Changes

- Adds an admin revoke session API

### Related

https://github.com/ory-corp/cloud/pull/3298


session/handler.go
```diff
@@ -440,6 +440,29 @@ func (h *Handler) adminGetSession(w http.ResponseWriter, r *http.Request, ps htt
...
+//
+//	Schemes: http, https
+//
+//	Responses:
+//	  204: emptyResponse
+//	  400: jsonError
+//	  401: jsonError
+//	  500: jsonError
+func (h *Handler) adminRevokeSession(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	x.RedirectToPublicRoute(h.r)(w, r, ps)
```
jonas-jonas, line 462: I don't think this will work for your use case. You still need the session cookie if you do it like this, which you don't have in the backoffice, right? See https://github.com/ory/kratos/pull/2867/files#diff-abfcc5544632668aab2f1162de82ec56e8ec7444f6ce377cb3a30a9eafbed198L585

IIUC, you will need to write the logic from https://github.com/ory/kratos/pull/2867/files#diff-abfcc5544632668aab2f1162de82ec56e8ec7444f6ce377cb3a30a9eafbed198L602 here and fetch `SessionID` from this `r`s body.

kelkarajay: Yes, I don't think so either 😅. I just threw in that copy-pasta to finish the sdk gen

jonas-jonas: Ah, just wanted to make sure you don't waste the time while integrating this into the backoffice 😅

aeneasr (approved):

PR #2899, opened 2022-11-16
feat: add session cache header feature flag

kelkarajay:
- Added config key `session.whoami.caching` with default being disabled
- Set cache headers based on configuration

## Related issue(s)

https://github.com/ory-corp/cloud/issues/3283

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/driver/config/config.go
+++ b/driver/config/config.go
@@ -118,6 +118,7 @@ const (
 	ViperKeySessionPath                                      = "session.cookie.path"
 	ViperKeySessionPersistentCookie                          = "session.cookie.persistent"
 	ViperKeySessionWhoAmIAAL                                 = "session.whoami.required_aal"
+	ViperKeySessionWhoAmICaching                             = "session.whoami.caching"
 	ViperKeySessionRefreshMinTimeLeft                        = "session.earliest_possible_extend"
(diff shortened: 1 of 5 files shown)


driver/config/config.go
```diff
@@ -118,6 +118,7 @@ const (
 	ViperKeySessionPath                                      = "session.cookie.path"
 	ViperKeySessionPersistentCookie                          = "session.cookie.persistent"
 	ViperKeySessionWhoAmIAAL                                 = "session.whoami.required_aal"
+	ViperKeySessionWhoAmICaching                             = "session.whoami.caching"
```
aeneasr, line 121: This feature is meant to be enabled for all systems starting today. Legacy systems may choose to revert to the old behavior, but will be forced to upgrade at some point. 

Putting this here signals to me that the configuration option will stay forever. That's why I chose to add a new section for this:

https://github.com/ory/kratos/pull/2833/files#diff-4a79b8650ca1bb0046229c6603ab7458e7711fc69812f13bbb8fc73b81b833e6R2430

In the medium term, we want to **reduce** the number of configuration options. Please implement this using the feature-flag naming!

Otherwise, this looks really good to me already :)

aeneasr (approved):

PR #2902, opened 2022-11-18
fix(sdk): rename OAuth2Client models to be kratos specific

jonas-jonas:
Fixes an SDK generation issue where both Hydra's and Kratos' `OAuth2Client` were present in the merged spec. Now, the `OAuth2Client` from Kratos' is prefixed with `kratos`.

## Related issue(s)

https://github.com/ory/sdk/issues/226

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


jonas-jonas:
I verified that this fixes the SDK generation by generating the client SDK spec from the Network Project and using that to generate the python client, that is currently failing. I'd appreciate anyone that could confirm that this does not break any existing integrations/examples.

.schema/openapi/patches/hydra.yaml
```diff
@@ -0,0 +1,35 @@
+# These patches are needed, because we merge the Kratos spec with the Hydra spec (and others) to create the 
+# Ory Client spec.
+# Hydra defines "OAuth2Client" (and the other schemas mentioned here) as well, which is where Kratos' "OAuth2Client"
+# comes from originally. It is generated by our SDK generator. This leads to a clash and some generators can not
+# deal with this. So we rename these fields to have a "kratos" prefix to distignuish them
+# Also see https://github.com/ory/sdk/issues/226
+- op: copy
+  from: /components/schemas/OAuth2Client
+  path: /components/schemas/kratosOAuth2Client
```
aeneasr, line 9: I would really like to avoid using `kratos` as a prefix in here.

aeneasr (commented):
I assume the problem here is that `OAuth2Client` and `oAuth2Client` both end up in the spec, and the generators have a hard time dealing with this because they are case insensitive. Could that be true? If so, I would prefer to fix this instead. So we could, for example, rename `OAuth2Client` to `oAuth2Client` in here. We might need to check how this would work when merging the hydra and kratos definitions in ory cloud, but it could work. WDYT?

jonas-jonas:
That was my first idea as well, but I figured it might hinder debugging if you don't know which is which. 

Another issue is that both specs are not completely equal, as the "doubly generated" one does not contain our custom types (`NullX`, etc.), but only `string`s, etc. I have run the generation in the `cloud` repo with 2 `oAuth2Client`s (same casing). See [1], I don't think you can use `$ref` *and* other fields, so this yields an invalid spec. 

1: 
image

jonas-jonas:
>I assume the problem here is that OAuth2Client and oAuth2Client both end up in the spec, and the generators have a hard time dealing with this because they are case insensitive. Could that be true?

Yes, that is my assumption as well, though I haven't looked into why some generators cannot deal with that.

.schema/openapi/patches/hydra.yaml
```diff
@@ -1,31 +1,15 @@
...
-  path: /components/schemas/OAuth2ConsentRequestOpenIDConnectContext
-- op: replace
-  path: /components/schemas/OAuth2LoginRequest/properties/oidc_context/$ref
-  value: '#/components/schemas/oAuth2ConsentRequestOpenIDConnectContext'
-
-- op: copy
-  from: /components/schemas/OAuth2LoginRequest
+- op: add
   path: /components/schemas/oAuth2LoginRequest
+  value:
```
aeneasr, line 10: @jonas-jonas looks like this is the only thing we need

aeneasr: Can you please check whether that fixes the problem in the sdk generator?

jonas-jonas: No, that causes the same errors, because there are still two `oauth2Client`s present in the spec. We somehow need to get rid of one of them or distinguish them with more than just casing.

jonas-jonas:
Replaced and fixed via cb1c33d46c40926c52994ef6e88108008206e23f

Issue #2824, opened 2022-10-19
Verification via one time `code`s


jonas-jonas:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Currently, the email verification process is only possible via magic links.
These have multiple drawbacks, which is why we want to replace them with one-time-codes.

### Describe your ideal solution

Implement the `code` strategy for verification flows as well.

### Workarounds or alternatives

none

### Version

v0.10.1

### Additional Context

Most of the code is already there in the recovery part of the strategy. Some could be adjusted to shared helpers.

jonas-jonas:
There are multiple paths, a user could take here.

### **the happy path (verification enabled, but not required for login)**

1. The user registers with the required traits
2. A verification flow is created and a verification email with code is sent
3. The user is redirected to the flow and a verification form is shown
4. The user enters the correct code from the email and is redirected to the specified after registration page

### **verification enabled, required for login, but user does not verify**

1. The user registers an account
2. A verification flow is created and a verification email with code is sent
3. The user is redirected to the flow and a verification form is shown
4. The user closes the tab and tries to log in with the credentials they set up (*without verifying the email)*
    1. **Either**: the last verification flow with a resend code button is shown
    2. **Or**: a new verification flow is automatically created, a new verification email is automatically sent

### **verification enabled, but not required for login, but user does not verify**

1. The user registers an account
2. A verification flow is created and a verification email with code is sent
3. The user is redirected to the flow and a verification form is shown
4. The user closes the tab and tries to log in with the credentials they set up (*without verifying the email)*
5. The login flow is completed, and they are redirected to the default “after login” page

*for this case we need some way to re-trigger a verification flow, AFAICT, that is not currently possible, but it might make sense to add such a box to the settings page [1]*

### **verification disabled**

[...]

PR #2904, opened 2022-11-21
feat: phone verification via sms code

nrutherford:
Added support for verifying phone numbers using verification code strategy and courier SMS support.

## Related issue(s)

#2824 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

1. How should attempts to verify an unknown phone number be handled?
2. Should the resend logic be enhanced to prevent abuse of SMS?
3. Should recovery be supported via phone?

Our use-case is that we will allow signup/login using either email or mobile number and will verify either/both during our "signup" flow using code strategy. We service markets where customers don't necessarily have an email address.


aeneasr:
> How should attempts to verify an unknown phone number be handled?

Good question, I don't think we should send an SMS in that case. Maybe we change the labels to explain this:

> An SMS will be sent to the phone number you supplied. If the number is unknown, no SMS will be sent. Please check that the number you provided is indeed correct.

> Should the resend logic be enhanced to prevent abuse of SMS?

Absolutely, but I don't know how :/ Do you have ideas?

> Should recovery be supported via phone?

I don't think we should, given the security profile of SMS.

DrBu7cher:
Is something needed to get this done/merged? :)
I would love to see this feature in action.

DrBu7cher:
Bumping to ask again: Is something necessary here?

aeneasr:
My comment / review is waiting for an answer from the author :)

DrBu7cher:
Except the questions - my suggestion: Let's look at how others handle this and "borrow" the ideas from there. :)
Resending should be with a configurable delay. Maybe even max per day?

Recovery via phone is maybe a feature in the future, I think verifying the phone number is a huge plus already, so let's keep the focus on that for this PR.

And regarding the unknown number, I agree with @aeneasr s idea :)

nrutherford:
Apologies for not replying sooner - was hoping to have better answers before replying, but alas:

> > How should attempts to verify an unknown phone number be handled?
> 
> Good question, I don't think we should send an SMS in that case. Maybe we change the labels to explain this:
> 
> > An SMS will be sent to the phone number you supplied. If the number is unknown, no SMS will be sent. Please check that the number you provided is indeed correct.
> 
I agree.

> > Should the resend logic be enhanced to prevent abuse of SMS?
> 
> Absolutely, but I don't know how :/ Do you have ideas?
> 

Effectively the same ideas as @DrBu7cher, see how others handle it, like:

- enforcing a time delay between resends
- configurable max sends per phone number (per day, forever)

> > Should recovery be supported via phone?
> 
> I don't think we should, given the security profile of SMS.

Definitely agree with this.

martinffx:
> Effectively the same ideas as @DrBu7cher, see how others handle it, like:
>
>   - enforcing a time delay between resends
>   - configurable max sends per phone number (per day, forever)

@aeneasr these seem like reasonable protections against abuse

congngale:
Hi All, 

Any update on this feature? I really like to have phone verification on user registration.

Beside, I could not find any document for configuration SMS provider.

Thanks,
Cong Nga Le

soltanoff:
Hi All,

Any update on this feature? I really like to have phone verification on user registration too.

Thanks!

siddmoitra:
Hi @nrutherford, thank you for the contribution. I was bit late to see this DRAFT PR, so created exactly a similar solution on a private branch to be used. Guess not needed after your PR is merged. Do you have a timeline to this one ?

Agree to the comments made regarding `NOT sending SMS to unknown recipients` and `not using SMS for recovery`. Had a same implementation myself.

cc @aeneasr

PR #2914, opened 2022-11-28
feat: store errors of courier message

jonas-jonas:
BREAKING CHANGE: The `/admin/courier/messages` endpoint now uses `keysetpagination` instead. 

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE courier_message_dispatches (
+  dispatch_id UUID PRIMARY KEY,
+  message_id UUID NOT NULL,
+  status TEXT NOT NULL,
+  error TEXT,
+  nid UUID NOT NULL,
+  created_at timestamp NOT NULL,
+  updated_at timestamp NOT NULL,
```
aeneasr, line 8: Why don't these have defaults like mysql?

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.mysql.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE courier_message_dispatches (
```
aeneasr, line 1: Please add migration tests for these

courier/courier_dispatcher.go
```diff
@@ -77,7 +61,27 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
...
+				return err
+			}
+			// Skip the message
+			c.deps.Logger().
+				WithField("message_id", msg.ID).
+				Warnf(`Message was abandoned because it did not deliver after %d attempts`, msg.SendCount)
+
+		} else if err := c.DispatchMessage(ctx, msg); err != nil {
+
+			if err := c.deps.CourierPersister().RecordDispatch(ctx, msg.ID, CourierMessageDispatchStatusFailed, err.Error()); err != nil {
```
aeneasr, line 79: `err.Error()` will be pretty meaningless, it will show something like "Internal Server Error". I think it makes more sense to serialize this to JSON (as we do when sending error responses in APIs) and record that.

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE courier_message_dispatches (
+  dispatch_id UUID PRIMARY KEY,
+  message_id UUID NOT NULL,
+  status TEXT NOT NULL,
+  error TEXT,
```
aeneasr, line 5: If we use JSON based serialization this should be `json`

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE courier_message_dispatches (
+  dispatch_id UUID PRIMARY KEY,
+  message_id UUID NOT NULL,
+  status TEXT NOT NULL,
```
aeneasr, line 4: Set this to a maximum length

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE courier_message_dispatches (
+  dispatch_id UUID PRIMARY KEY,
```
aeneasr, line 2: Please use `id` for the primary key

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.mysql.up.sql
```diff
@@ -0,0 +1,11 @@
...
+  dispatch_id CHAR(36) PRIMARY KEY,
+  message_id CHAR(36) NOT NULL,
+  status TEXT NOT NULL,
+  error TEXT,
+  nid CHAR(36) NOT NULL,
+  created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  CONSTRAINT courier_message_dispatches_message_id_fk FOREIGN KEY (message_id) REFERENCES courier_messages (id) ON DELETE cascade,
+  CONSTRAINT courier_message_dispatches_nid_fk FOREIGN KEY (nid) REFERENCES networks (id) ON DELETE cascade
```
aeneasr, line 10: See comments from other file

persistence/sql/persister_courier.go
```diff
@@ -170,3 +179,75 @@ func (p *Persister) IncrementMessageSendCount(ctx context.Context, id uuid.UUID)
...
+
+	if err := p.GetConnection(ctx).Create(&dispatch); err != nil {
+		return sqlcon.HandleError(err)
+	}
+
+	return nil
+}
+
+func (p *Persister) CreateTestData(ctx context.Context) error {
+
```
aeneasr, line 218: I don't think this method belongs here. Why not put it into internal/testhelpers?

courier/courier_dispatcher_test.go
```diff
@@ -54,24 +54,9 @@ func TestDispatchMessageWithInvalidSMTP(t *testing.T) {
 		messages, err := reg.CourierPersister().NextMessages(ctx, 10)
 		require.Len(t, messages, 1)
 	})
-
-	t.Run("case=max retries reached", func(t *testing.T) {
```
aeneasr, line 58: Where did this test move to?

courier/handler.go
```diff
@@ -62,10 +70,9 @@ type listCourierMessagesResponse struct {
 
 // Paginated List Courier Message Parameters
 //
-// nolint:deadcode,unused
 // swagger:parameters listCourierMessages
 type ListCourierMessagesParameters struct {
-	migrationpagination.RequestParameters
+	keysetpagination.RequestParameters
```
aeneasr, line 75: Is this a breaking change? Is this endpoint released already?

aeneasr (commented):
Looking pretty good already! I think that we will need to change this medium-term to an event-based stream on top of opentelemetry, but this is a good get-it-done patch :)

courier/handler.go
jonas-jonas: It is a breaking change, but this endpoint is not released in OS and in the network it wasn't reachable.

So I'd argue this is fine.

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.up.sql
jonas-jonas: The `TEXT` type doesn't have a maximum length. Its specification is ["a variable text type of unlimited length"](https://www.postgresql.org/docs/current/datatype-character.html#:~:text=variable%20unlimited%20length). We could change this to a `VARCHAR` and provide a maximum length, but since we don't know specifically what the error string could be I'd prefer `TEXT` here.

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.up.sql
jonas-jonas: It's not, we just dump whatever we get from the SMTP library in here as plaintext. I don't think we need it to be JSON, as SMTP doesn't communicate with JSON, afaik.

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.up.sql
jonas-jonas: Nvm, though you were talking about the "error" field. 

This of course can be a char/varchar, sorry for the noise.

courier/courier_dispatcher.go
jonas-jonas: `err.Error()` returns `dial tcp: lookup my.custom.mailserver.host on 127.0.0.11:53: no such host` for example. I think that's all the errors users need, right?

> I think it makes more sense to serialize this to JSON 

Hm, interesting. 

How do we display them then? Just showing plain JSON is not great...

I updated this PR and will try it out in the network next. Good idea!

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.up.sql
jonas-jonas: (y)

courier/courier_dispatcher.go
aeneasr: You can still parse the JSON in the UI :)

courier/courier_dispatcher_test.go
jonas-jonas: I moved the logic of `abandoning` messages back to `DispatchQueue`, so this can't be tested here anymore. Instead the existing test at https://github.com/ory/kratos/pull/2914/files#diff-0b4d4586068f4dbc5b91bea49de83394f45726e7525e01a1c315cd2046562827R59 tests this now.

persistence/sql/persister_courier.go
jonas-jonas: I needed a way in cloud to generate a bunch of messages to test the pagination. So I added this + an endpoint to do that... Obviously wasn't going to propose it :D

This PR wasn't actually ready for PR yet, but I realized too late that it's not in Draft anymore. Sorry.

persistence/sql/migrations/sql/20221109114303000000_add_courier_send_attempts_table.mysql.up.sql
jonas-jonas: Done.

jonas-jonas:
@aeneasr this is now ready for round two, but we need https://github.com/ory/x/pull/640 to be merged before we merge this, just FYI.

driver/config/config_test.go
```diff
@@ -946,7 +947,8 @@ func TestIdentitySchemaValidation(t *testing.T) {
 		assert.NoError(t,
 			os.MkdirAll(tdir, // DO NOT CHANGE THIS: https://github.com/fsnotify/fsnotify/issues/340
 				os.ModePerm))
-		tmpConfig, err := os.Create(filepath.Join(tdir, "config.yaml"))
+		configFileName := randx.MustString(8, randx.Alpha)
```
jonas-jonas, line 950: This test https://github.com/ory/kratos/actions/runs/3605104374/jobs/6075180290#step:15:1429 has been flaky a few times. I couldn't repro it locally, but suspect that using just `config.yaml` in the same directory might lead to flakyness. This should solve this. Other opinions welcome! :)

aeneasr (commented):
Awesome!

courier/handler.go
```diff
@@ -119,30 +125,82 @@ func (h *Handler) listCourierMessages(w http.ResponseWriter, r *http.Request, _
...
+	}
+
+	message, err := h.r.CourierPersister().FetchMessage(r.Context(), msgID)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
+
+	if !h.r.Config().IsInsecureDevMode(r.Context()) {
+		message.Body = "<redacted-unless-dev-mode>"
```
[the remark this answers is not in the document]
aeneasr: @jonas-jonas can you create a follow up issue for this? I don't know where we discussed a possible solution, but it would be great to add the idea we had there.

courier/handler.go
```diff
@@ -119,30 +125,82 @@ func (h *Handler) listCourierMessages(w http.ResponseWriter, r *http.Request, _
 		}
 	}
 
-	x.PaginationHeader(w, urlx.AppendPaths(h.r.Config().SelfAdminURL(r.Context()), AdminRouteMessages), int64(tc), filter.Page, filter.PerPage)
+	w.Header().Set("x-total-count", fmt.Sprint(tc))
```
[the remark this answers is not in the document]
jonas-jonas: Sure, done: https://github.com/ory/x/issues/645

persistence/sql/migrations/sql/20221205092803000000_add_courier_send_attempts_table.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE courier_message_dispatches (
+  id UUID PRIMARY KEY,
+  message_id UUID NOT NULL,
+  status VARCHAR(7) NOT NULL,
+  error JSON,
+  nid UUID NOT NULL,
+  created_at timestamp NOT NULL,
+  updated_at timestamp NOT NULL,
```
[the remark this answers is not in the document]
jonas-jonas: Done.

aeneasr (approved):

jonas-jonas:
@aeneasr since Kratos was released in the meantime, this is now a breaking change. I added the note to the PR description, FYI.

aeneasr:
TY

PR #2917, opened 2022-11-30
feat: forward parsed request cookies to webhook Jsonnet snippet

alnr:
Request cookies were already available in raw form in the `ctx.request_headers` top-level argument to the Jsonnet snippet. Parsing cookies in Jsonnet is tedious and error-prone, though, so we parse them internally for convenience.

Ref https://github.com/ory-corp/cloud/issues/3434

### DIFF
--- a/selfservice/hook/web_hook.go
+++ b/selfservice/hook/web_hook.go
@@ -57,6 +57,7 @@ type (
 		RequestHeaders http.Header        `json:"request_headers"`
 		RequestMethod  string             `json:"request_method"`
 		RequestURL     string             `json:"request_url"`
+		RequestCookies map[string]string  `json:"request_cookies"`
 		Identity       *identity.Identity `json:"identity,omitempty"`
 	}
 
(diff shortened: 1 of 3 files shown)


selfservice/hook/web_hook.go
```diff
@@ -57,6 +57,7 @@ type (
 		RequestHeaders http.Header        `json:"request_headers"`
 		RequestMethod  string             `json:"request_method"`
 		RequestURL     string             `json:"request_url"`
+		Cookies        map[string]string  `json:"cookies"`
```
aeneasr, line 60:
```suggestion
		RequestCookies        map[string]string  `json:"request_cookies"`
```

aeneasr (changes requested):
Awesome, just one small naming change _)

selfservice/hook/web_hook.go
alnr: fixed

aeneasr (approved):

Issue #2930, opened 2022-12-06
GetSessionsByID not working properly with active=true query parameter


vsumit89:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [x] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

while getting the active sessions of a user by sending request to kratos admin endpoint - /admin/identities/{identity_id}/sessions?active=true, I am correctly getting all the active sessions but along with it I am getting one inactive (active=false) session. 

### Reproducing the bug

After running kratos hit the admin endpoint - /admin/identities/{identity_id}/sessions?active=true

### Relevant log output

```shell
[
  {
    "id": "6fb620fa-4e03-484a-aa40-779ee883ef22",
    "active": false,
    "expires_at": "2022-11-30T12:52:05.179595Z",
    "authenticated_at": "2022-11-29T12:52:05.179595Z",
    "authenticator_assurance_level": "aal1",
    "authentication_methods": [
      {
        "method": "password",
        "aal": "aal1",
        "completed_at": "2022-11-29T12:52:05.17959476Z"
      }
    ],
    "issued_at": "2022-11-29T12:52:05.179595Z",
    "identity": {
      "id": "049e524f-d460-4325-904a-cadd16037a95",
      "schema_id": "default",
      "schema_url": "http://127.0.0.1:4455/.ory/kratos/public/schemas/ZGVmYXVsdA",
      "state": "active",
      "state_changed_at": "2022-11-29T12:27:12.389522Z",
      "traits": {
        "email": "kavach-develop@factly.in"
      },
      "verifiable_addresses": [
        {
          "id": "11ea6fbf-a9ed-4526-a7c4-4351b1a26fb5",
          "value": "kavach-develop@factly.in",
          "verified": true,
          "via": "email",
          "status": "completed",
          "verified_at": "2022-12-05T09:02:07.247184Z",
[...]

PR #2958, opened 2022-12-15
fix: listing sessions query

kelkarajay:
## Changes

- `Where` clause added to also check session expiry timestamp to return correct list
- Fixed to query identity only once
- Remove marshalling mutation 

## Related issue(s)

https://github.com/ory/kratos/issues/2930

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -78,6 +78,12 @@ func (p *Persister) ListSessions(ctx context.Context, active *bool, paginatorOpt
 		q := c.Where("nid = ?", nid)
 		if active != nil {
 			q = q.Where("active = ?", *active)
+
+			if *active {
+				q.Where("expires_at >= ?", time.Now())
+			} else {
+				q.Where("expires_at < ?", time.Now())
+			}
 		}
 
 		// if len(expandables) > 0 {
@@ -134,6 +140,12 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
 		}
 		if active != nil {
 			q = q.Where("active = ?", *active)
+
+			if *active {
+				q.Where("expires_at >= ?", time.Now())
+			} else {
+				q.Where("expires_at < ?", time.Now())
+			}
 		}
 		if len(expandables) > 0 {
 			q = q.Eager(expandables.ToEager()...)
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -78,12 +78,10 @@ func (p *Persister) ListSessions(ctx context.Context, active *bool, paginatorOpt
 	if err := p.Transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
 		q := c.Where("nid = ?", nid)
 		if active != nil {
-			q = q.Where("active = ?", *active)
-
 			if *active {
-				q.Where("expires_at >= ?", time.Now())
+				q.Where("active = ? AND expires_at >= ?", *active, time.Now().UTC())
 			} else {
-				q.Where("expires_at < ?", time.Now())
+				q.Where("active = ? OR expires_at < ?", *active, time.Now().UTC())
 			}
 		}
 
@@ -113,6 +111,7 @@ func (p *Persister) ListSessions(ctx context.Context, active *bool, paginatorOpt
 					return err
 				}
 
+				sess.Active = sess.IsActive()
 				sess.Identity = i
 			}
 		}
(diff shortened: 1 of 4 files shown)


persistence/sql/persister_session.go
```diff
@@ -58,6 +58,7 @@ func (p *Persister) GetSession(ctx context.Context, sid uuid.UUID, expandables s
 		s.Identity = i
 	}
 
+	s.Active = s.IsActive()
```
aeneasr, line 61: Do we need this because IsActive returns false when expired, but Active might not be false?

persistence/sql/persister_session.go
```diff
@@ -78,6 +79,12 @@ func (p *Persister) ListSessions(ctx context.Context, active *bool, paginatorOpt
 		q := c.Where("nid = ?", nid)
 		if active != nil {
 			q = q.Where("active = ?", *active)
+
+			if *active {
+				q.Where("expires_at >= ?", time.Now())
```
aeneasr, line 84:
```suggestion
				q.Where("expires_at >= ?", time.Now().UTC())
```

persistence/sql/persister_session.go
```diff
@@ -78,6 +79,12 @@ func (p *Persister) ListSessions(ctx context.Context, active *bool, paginatorOpt
...
+			} else {
+				q.Where("expires_at < ?", time.Now())
```
aeneasr, line 86:
```suggestion
				q.Where("expires_at < ?", time.Now().UTC())
```

persistence/sql/persister_session.go
```diff
@@ -128,12 +135,23 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
 	nid := p.NetworkID(ctx)
 
 	if err := p.Transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
+		i, err := p.GetIdentity(ctx, iID)
```
aeneasr, line 138: We're now no longer respecting the expand property and instead fetch the identity always. I don't think that's intended? Please move it back to where it was, but of course outside of the loop to prevent unneccessary queries

```

 		if expandables.Has(session.ExpandSessionIdentity) {
 			i, err := p.GetIdentity(ctx, iID)
 			// ...
 			for _, s := range s {

```

persistence/sql/persister_session.go
```diff
@@ -128,12 +135,23 @@ func (p *Persister) ListSessionsByIdentity(ctx context.Context, iID uuid.UUID, a
...
 		if except != uuid.Nil {
 			q = q.Where("id != ?", except)
 		}
 		if active != nil {
 			q = q.Where("active = ?", *active)
+
+			if *active {
+				q.Where("expires_at >= ?", time.Now())
+			} else {
+				q.Where("expires_at < ?", time.Now())
```
aeneasr, line 153: At some point we realized that all DB timestamps need to be UTC

```suggestion
				q.Where("expires_at >= ?", time.Now().UTC())
			} else {
				q.Where("expires_at < ?", time.Now().UTC())
```

session/session.go
```diff
@@ -153,17 +153,6 @@ func (s Session) TableName(ctx context.Context) string {
 	return "sessions"
 }
 
-func (s Session) MarshalJSON() ([]byte, error) {
```
aeneasr, line 156: I think this is a cleaner solution than setting the field upon loading it from the database, because expiry is time based, and we always want the "freshest" value for `Active`. Also a bit confused as to why no tests are failing if this is removed

session/session.go
```diff
@@ -153,17 +153,6 @@ func (s Session) TableName(ctx context.Context) string {
...
-	type sl Session
-	s.Active = s.IsActive()
-
-	result, err := json.Marshal(sl(s))
-	if err != nil {
-		return nil, err
-	}
-	return result, nil
```
aeneasr, line 164: Could just be:

```
return json.Marshal(sl(s))
```

aeneasr (changes requested):

persistence/sql/persister_session.go
aeneasr: I think yes

session/session.go
kelkarajay: This was removed because it alters the value of list items based on Identity state after the listing query has run. 

As in, a request for `?active=true` yields active session records but then this code block alters the value. This may result in result set with `active: false` due to `s.Active && s.ExpiresAt.After(time.Now()) && (s.Identity == nil || s.Identity.IsActive())`, which will confuse the API consumer. To try and solve this problem, I've made corrections to the query in https://github.com/ory/kratos/pull/2958/commits/eddece5b94834a4d559e9ca8bece589408a782d6

kelkarajay:
Re-organized listing tests to do seed data setup once and run against that

aeneasr (approved):

Issue #2680, opened 2022-08-22
Getting 500 server error on sending "totp_code"


rostyk110:
### Preflight checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [x] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [x] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

Auth app - **Google Authenticator** 

When I try to add Auth app, I get 500 Server Error. It happens after: connect Auth App (_successfully_), disconnect (_successfully_), connect again (_getting Error_). 

**Response:** 
```
{
    "code": 500,
    "status": "Internal Server Error",
    "request": "65abe04a-f758-9d46-b662-36482f537b32",
    "reason": "Could not find they TOTP key in the internal context. This is a code bug and should be reported to https://github.com/ory/kratos/.",
    "message": "An internal server error occurred, please contact the system administrator"
}
```

### Reproducing the bug
Sorry, recording is 24.2MB, so I compressed it. 

[Screen Recording 2022-08-22 at 22.09.45.mov.zip](https://github.com/ory/kratos/files/9396994/Screen.Recording.2022-08-22.at.22.09.45.mov.zip)

Or here is youtube link:
https://www.youtube.com/watch?v=_37Fns4AkT0

### Relevant log output
```
{
   "audience":"application",
   "error":{
      "debug":"",
      "message":"An internal server error occurred, please contact the system administrator",
      "reason":"Could not find they TOTP key in the internal context. This is a code bug and should be reported to https://github.com/ory/kratos/.",
      "status":"Internal Server Error",
      "status_code":500
   },
   "http_request":{
      "headers":{
         "accept":"application/json, text/plain, */*",
         "accept-encoding":"gzip, deflate, br",
         "content-length":"142",
         "content-type":"application/json",
[...]

PR #2960, opened 2022-12-15
fix: TOTP invalid code and/or 500 error

kszafran:
Fixes a 500 error if TOTP is removed and then re-added in the same flow.

It also fixes another related bug, that can be reproduced with https://github.com/ory/kratos-selfservice-ui-react-nextjs:
1. go to settings
2. change your name (or email, or whatever)
3. try to set up TOTP
4. "invalid code"

The reason is that because the `InternalContext` is not reset properly in `PostSettingsHook`, the flow will have a new QR code and secret key (`UI` is reset), but the internal context will keep the old TOTP secret.

In the case of the original issue, `InternalContextKeyURL` is deleted from `InternalContext` when TOTP is removed, and since it's not set properly in `PostSettingsHook`, when we try to re-add TOTP, we get 500, because `InternalContextKeyURL` is missing.

## Related issue(s)

https://github.com/ory/kratos/issues/2680

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

WebAuthn is most likely affected by the same bug, but I didn't any tests there. I wasn't able to run WebAuthn tests locally, because of this error:
```
cy.task('resetCRI') failed with the following error:

The 'task' event has not been registered in the plugins file. You must register it before using cy.task()
```

### DIFF
--- a/selfservice/flow/settings/hook.go
+++ b/selfservice/flow/settings/hook.go
@@ -242,6 +242,7 @@ func (e *HookExecutor) PostSettingsHook(w http.ResponseWriter, r *http.Request, 
 	ctxUpdate.Flow.UI = newFlow.UI
 	ctxUpdate.Flow.UI.ResetMessages()
 	ctxUpdate.Flow.UI.AddMessage(node.DefaultGroup, text.NewInfoSelfServiceSettingsUpdateSuccess())
+	ctxUpdate.Flow.InternalContext = newFlow.InternalContext
 	if err := e.d.SettingsFlowPersister().UpdateSettingsFlow(r.Context(), ctxUpdate.Flow); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


aeneasr (approved):
Great find, thank you!

kszafran:
@aeneasr I fixed the failing `lookup/settings_test.go` by changing the assertion from:
```go
assert.Equal(t, "{}", gjson.GetBytes(actualFlow.InternalContext, ...
```
to
```go
assert.Empty(t, gjson.GetBytes(actualFlow.InternalContext, ...
```
I haven't looked too deep into it, but my initial guess is that the test was relying on the buggy behavior, where the internal context was not reset properly. There is already an assertion like this (using `assert.Empty(t, ...`) in another place in the file, so I'm assuming the behavior is correct.

kszafran:
And now I also fixed `webauthn/settings_test.go` by changing an assertion. I hope it makes sense, but since I'm not familiar with the WebAuthn flow, I'm not exactly sure.

kszafran:
And now some snapshot comparisons fail in WebAuthn tests, even though they passed before 😞 .

aeneasr (dismissed):

aeneasr:
Yeah sorry about that, we had some breaks on master :( Should all be fixed now

PR #2961, opened 2022-12-16
fix: flaky test

CaptainStandby:
This fixes a problematic unit test, that monitors log messages and expected a specific message at the end of the list. This has been corrected by searching for the expected message in the list.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/driver/config/config_test.go
+++ b/driver/config/config_test.go
@@ -14,6 +14,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -1042,14 +1043,35 @@ func TestIdentitySchemaValidation(t *testing.T) {
 					marshalAndWrite(t, ctx, tmpConfig, i)
 				}(t, ctx, tmpConfig, i)
 
-				select {
-				case <-ctx.Done():
-					panic("the test could not complete as the context timed out before the file watcher updated")
-				case <-c:
-					lastHook, err := hook.LastEntry().String()
-					assert.NoError(t, err)
+				go func() {
+					for range c {
+						// If we don't drain c the sender will block
+					}
+				}()
 
-					assert.Contains(t, lastHook, "The changed identity schema configuration is invalid and could not be loaded.")
+				// There are a bunch of log messages beeing logged. We are looking for a specific one.
+				timeout := time.After(time.Millisecond * 500)
+			outer:
+				for {
+					for _, v := range hook.AllEntries() {
+						s, err := v.String()
+						if err != nil {
+							t.Errorf("Unexpected Error: %s", err.Error())
+							continue
+						}
+
+						if strings.Contains(s, "The changed identity schema configuration is invalid and could not be loaded.") {
+							break outer
+						}
+					}
+
+					select {
+					case <-ctx.Done():
+						panic("the test could not complete as the context timed out before the file watcher updated")
+					case <-timeout:
+						t.Fatal("Expected log line was not encountered within specified timeout")
+					default: //nothing
+					}
 				}
 
 				wg.Wait()
--- a/driver/config/config_test.go
+++ b/driver/config/config_test.go
@@ -27,8 +27,6 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/spf13/cobra"
 
-	"github.com/ory/x/watcherx"
-
 	"github.com/ory/kratos/internal/testhelpers"
 
 	"github.com/ory/x/configx"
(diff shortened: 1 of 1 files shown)


driver/config/config_test.go
```diff
@@ -941,32 +940,31 @@ func TestIdentitySchemaValidation(t *testing.T) {
...
-		c := make(chan bool, 1)
+	testWatch := func(t *testing.T, ctx context.Context, cmd *cobra.Command, identity *configFile) (*config.Config, *test.Hook, func([]map[string]string)) {
 		tdir := t.TempDir()
 		assert.NoError(t,
 			os.MkdirAll(tdir, // DO NOT CHANGE THIS: https://github.com/fsnotify/fsnotify/issues/340
 				os.ModePerm))
 		configFileName := randx.MustString(8, randx.Alpha)
 		tmpConfig, err := os.Create(filepath.Join(tdir, configFileName+".config.yaml"))
 		assert.NoError(t, err)
+		t.Cleanup(func() { tmpConfig.Close() })
```
alnr, line 951: nit
```suggestion
		t.Cleanup(tmpConfig.Close)
```

alnr (approved):

driver/config/config_test.go
alnr: I'm an idiot

alnr (approved):

driver/config/config_test.go
aeneasr: The suggestion is correct :)

driver/config/config_test.go
```diff
@@ -1022,34 +1020,44 @@ func TestIdentitySchemaValidation(t *testing.T) {
...
-					assert.NoError(t, err)
-
-					assert.Contains(t, lastHook, "The changed identity schema configuration is invalid and could not be loaded.")
+					// Change the identity config to an invalid file
+					writeSchema(invalidIdentity.Identity.Schemas)
+				}()
+
+				// There are a bunch of log messages beeing logged. We are looking for a specific one.
+				timeout := time.After(time.Millisecond * 500)
+			outer:
```
aeneasr, line 1040: If possible try to avoid labels for better readability :)

driver/config/config_test.go
```diff
@@ -1022,34 +1020,44 @@ func TestIdentitySchemaValidation(t *testing.T) {
...
+						}
+
+						if strings.Contains(s, "The changed identity schema configuration is invalid and could not be loaded.") {
+							break outer
+						}
+					}
+
+					select {
+					case <-ctx.Done():
+						panic("the test could not complete as the context timed out before the file watcher updated")
```
aeneasr, line 1056: No panics ever please

driver/config/config_test.go
```diff
@@ -1022,34 +1020,44 @@ func TestIdentitySchemaValidation(t *testing.T) {
...
+				}()
+
+				// There are a bunch of log messages beeing logged. We are looking for a specific one.
+				timeout := time.After(time.Millisecond * 500)
+			outer:
+				for {
+					for _, v := range hook.AllEntries() {
+						s, err := v.String()
+						if err != nil {
+							t.Errorf("Unexpected Error: %s", err.Error())
```
aeneasr, line 1045: Please use `require.NoError` or `assert.NoError`

aeneasr:
@CaptainStandby can you please address the comments in a follow-up PR? Thank you!

CaptainStandby:
Sure thing #2964

zepatrik (commented):
I also investigated this already. The underlying issue is that `marshalAndWrite` causes two notifications (one for `.Truncate` and one for `.Write`).
The file content is then read after the first event. Depending on the speed of the system/disk this can then flake, returning invalid content (i.e. empty file), or the expected content. Ideally we would do an [atomic file write](https://github.com/kubernetes/kubernetes/blob/03bfbdd8aa04739a5d99bdefeb9fcf379ffb970d/pkg/volume/util/atomic_writer.go) in the test instead, but this is also OK.

CaptainStandby:
Yeah, we noticed this too and the first take on fixing this, was doing atomic file write. But we decided to go with this approach, because we wanted to retain the "flakyness" of the write to test that the system could correctly deal with that situation.
Maybe a specific test, that simulates a problematic write could be added, now that I think about it.

alnr:
Yes, the non-atomic write is almost certainly the problem.

Like @CaptainStandby said, we purposefully kept that test behavior because non-atomic writes could easily happen in real-world use, too. Instead of changing the test to use an atomic move, we more broadly inspect what's logged.

A similar flakyness could happen if the logger internally re-ordered log lines.

Issue #2505, opened 2022-06-05
Allow admin to remove MFA credentials (TOTP, lookup secrets, webauthn)


halms:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

If a user loses all their configured MFA credentials (when `reuired_aal` is set to `highest_available` for the profile) there is currently no exposed way to remove the MFA credentials.

I consider it a sensible feature to expose an endpoint on the admin API to remove/reset MFA.

There are currently a few open PRs that lay the groundwork for this:
#2423 will allow updating credentials in general
#2380 will allow _selectively_ updating the only credentials
#2438 will allow _adding_ `totp` and `lookup_secrets`

What I consider still missing, is a way to remove the credentials via an admin API endpoint.

### Describe your ideal solution

Solution 1:
Allow the `PUT /admin/identities/{id}` (and the future `PATCH /admin/identities/{id}`) to remove credentials
Possibly by passing something like (YAML for simpler notation)
```yaml
credentials:
  totp: null
  webauthn: null
  lookup_secrets: null
```

Solution 2:
Create a new `DELETE /admin/identities/{id}/mfa` or similar endpoint

Solution 3:
Create a new `DELETE /admin/identities/{id}/credentials/{type}` (type being `totp`, `webauthn`, `lookup_secrets`) or similar endpoint.

### Workarounds or alternatives

Any of the three solutions would fulfill the need. My personal preference would be to offer both solution 1 and one of 2 or 3, with 3 being probably the cleaner one.

### Version

v0.10.1

maaartyyynaa:
Hi! This feature would be also useful for us and we are starting to work on the Solution 3 implementation

aeneasr:
I think this is a great idea and solution 3 looks very good to me

supercairos:
[...]

PR #2962, opened 2022-12-16
feat: add a new admin API to remove a specific 2nd factor credential

supercairos:
This PR provide a way for an admin to remove a specific credential method.
For now, only 2nd factor methods are allowed (TOTP, Lookup Secrets, WebAuthn)
This endpoint can be used by support services to unlock a custom account in case they've lost they MFA device.

Following the discussion with issue #2505 i've decided to implement in the following way:
`DELETE /admin/identities/{id}/credential/{type}`
id: being the targeted user id
type: being one of the 2FA kratos supports (totp, lookup, webauthn)

## Related issue(s)
https://github.com/ory/kratos/issues/2505

## Checklist
      introduces a new feature.
- [X] I am following the
      the changes.
      works.

## Further Comments
I've moved the credentials config models found previously in the `selfservice/strategy` pkg to the `identity` package to be able to use them from the admin endpoints and to avoid some cyclic pkg inclusion


identity/handler.go
```diff
@@ -675,3 +682,101 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
...
+
+	// Type is the credential's Type.
+	// One of totp, webauthn, lookup
+	//
+	// required: true
+	// in: path
+	Type string `json:"type"`
+}
+
+// swagger:route DELETE /admin/identities/{id}/credential/{type} identity deleteCredential
```
aeneasr, line 705:
```suggestion
// swagger:route DELETE /admin/identities/{id}/credential/{type} identity deleteIdentityCredential
```

identity/handler.go
```diff
@@ -675,3 +682,101 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
 
 	h.r.Writer().Write(w, r, WithCredentialsMetadataAndAdminMetadataInJSON(*identity))
 }
+
+// Delete Credential Parameters
+//
+// swagger:parameters deleteCredential
```
aeneasr, line 688:
```suggestion
// swagger:parameters deleteIdentityCredential
```

identity/handler.go
```diff
@@ -675,3 +682,101 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
...
+// nolint:deadcode,unused
+type deleteCredential struct {
```
aeneasr, line 690:
```suggestion
type deleteIdentityCredential struct {
```

identity/handler.go
```diff
@@ -675,3 +682,101 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
...
+			}
+		}
+
+		count, err := h.r.IdentityManager().CountActiveFirstFactorCredentials(r.Context(), identity)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithWrap(err)))
+		}
+
+		if count < 2 && wasPasswordless {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReason("Unable to remove this security key because it would lock this user out of his account")))
```
aeneasr, line 761:
```suggestion
			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReason("Unable to remove this security key because it would lock this user out of their account.")))
```

identity/handler.go
```diff
@@ -675,3 +682,101 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
...
+		var cc CredentialsWebAuthnConfig
+		if err := json.Unmarshal(cred.Config, &cc); err != nil {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to decode identity credentials.").WithDebug(err.Error())))
+			return
+		}
+
+		var wasPasswordless bool
+		for _, cred := range cc.Credentials {
+			if cred.IsPasswordless {
+				wasPasswordless = true
```
aeneasr, line 751: I think we should iterate over all credentials and remove those with `IsPasswordless == false` and keep the rest

aeneasr (commented):
Very nice! Only a few nit picks :)

supercairos:
Hi @aeneasr or @zepatrik any chance you had time to check this PR ? :)

identity/handler_test.go
```diff
@@ -991,4 +992,108 @@ func TestHandler(t *testing.T) {
...
+			})
+			t.Run("type=remove webauthn passwordless type/"+name, func(t *testing.T) {
+				expected := `{"credentials":[{"id":"THTndqZP5Mjvae1BFvJMaMfEMm7O7HE1ju+7PBaYA7Y=","added_at":"2022-12-16T14:11:55Z","public_key":"pQECAyYgASFYIMJLQhJxQRzhnKPTcPCUODOmxYDYo2obrm9bhp5lvSZ3IlggXjhZvJaPUqF9PXqZqTdWYPR7R+b2n/Wi+IxKKXsS4rU=","display_name":"test","authenticator":{"aaguid":"rc4AAjW8xgpkiwsl8fBVAw==","sign_count":0,"clone_warning":false},"is_passwordless":true,"attestation_type":"none"}],"user_handle":"Ef5JiMpMRwuzauWs/9J0gQ=="}`
+				i := createIdentities(map[identity.CredentialsType]string{identity.CredentialsTypeWebAuthn: expected})(t)
+				remove(t, ts, "/identities/"+i.ID.String()+"/credentials/webauthn", http.StatusNoContent)
+				// Check that webauthn has not been deleted
+				res := get(t, ts, "/identities/"+i.ID.String(), http.StatusOK)
+				assert.EqualValues(t, i.ID.String(), res.Get("id").String(), "%s", res.Raw)
+				assert.True(t, res.Get("credentials").Exists())
+				assert.True(t, res.Get("credentials").Get("webauthn").Exists())
```
aeneasr, line 1036: Could you please refactor this to use `GetIdentityConfidential` instead, and then ensure that the credentials config is correct - it should have the same value as `expected`. That ensures this works properly

identity/handler.go
```diff
@@ -675,3 +682,112 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
...
+		}
+
+		cc.Credentials = updated
+		cred.Config, err = json.Marshal(cc)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to encode identity credentials.").WithDebug(err.Error())))
+			return
+		}
+
+		identity.SetCredentials(CredentialsTypeWebAuthn, *cred)
```
aeneasr, line 772: This is missing one test in the test suite, having webauthn credentials with both passwordless and 2fa credentials, and then testing that we do indeed get the correct result:

**before**
* webauthn key 1 passwordless
* webauthn key 2 passwordless
* webauthn key 3 2fa
* webauthn key 4 2fa

**expected result**
* webauthn key 3 2fa
* webauthn key 4 2fa

My recommendation would be to move this code into  a separate function, write unit tests there. This will make the testing in handler_test a bit easier

aeneasr (commented):
Awesome, this is very neat! Also great tests! There's two things that need to be improved in the tests, but I think those changes should be doable in little time :) Once these are addressed, this is good to go to master!

identity/handler.go
supercairos: I did move the webauthn removal to his own function but I didn't tested that function per-se as it was complexe to build and my test function above had all the bits need to test the webauthn removal;

(If I want to keep the `deletCredentialWebAuthFromIdentity` private I would need to trick the testing to code to make it public as the handler_test has it own pkg name)

But if needed I can write the code.

supercairos: I did move the webauthn removal to his own function but I didn't tested that function per-se as it was complexe to build and my test function above had all the bits need to test the webauthn removal;

(If I want to keep the `deletCredentialWebAuthFromIdentity` private I would need to trick the testing to code to make it public as the handler_test has it own pkg name)

But if needed I can write the code.

identity/handler_test.go
supercairos: Done, I used the snapshot method (rather than comparing to expected) as it seams that's what's done on other tests

supercairos:
@aeneasr thanks for the review, I should have fixed all your raised concerns.

Let me know if it looks good :) 🙏

identity/handler.go
aeneasr: you can create a new file `handler_unit_test`, use the proper package there, and write the test :) But the handler test now is sufficient, so it's OK imo :)

aeneasr:
Great work!

Linyhazel:
Hellooo the type is actually `lookup_secret` instead of `lookup`, maybe you can update the documentation :)

PR #2965, opened 2022-12-17
fix: dont reset to false

CaptainStandby:
Addendum to #2964

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/driver/config/config_test.go
+++ b/driver/config/config_test.go
@@ -1042,7 +1042,7 @@ func TestIdentitySchemaValidation(t *testing.T) {
 					for _, v := range hook.AllEntries() {
 						s, err := v.String()
 						assert.NoErrorf(t, err, "Unexpected Error")
-						success = strings.Contains(s, "The changed identity schema configuration is invalid and could not be loaded.")
+						success = success || strings.Contains(s, "The changed identity schema configuration is invalid and could not be loaded.")
 					}
 
 					select {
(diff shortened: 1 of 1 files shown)


aeneasr (dismissed):

driver/config/config_test.go
```diff
@@ -1042,7 +1042,7 @@ func TestIdentitySchemaValidation(t *testing.T) {
 					for _, v := range hook.AllEntries() {
 						s, err := v.String()
 						assert.NoErrorf(t, err, "Unexpected Error")
```
aeneasr, line 1044: nitpick: Only use formatted errors `Error**f**` when formatting strings a la `"foo %s bar"`. Since the assertion itself already says that we assert no error, we also don't need the additional error message `Unexpected Error` as it doesn't provide additional value:

```suggestion
						require.NoError(t, err)
```

aeneasr (commented):
Nice! By the way, you could probably do something similar now with generics a la

```go

import (
	"github.com/samber/lo"
	"strings"
)


		_, found := lo.Find(hook.AllEntries(), func(i HookType) bool {
			s, err := v.String()
			require.NoErrorf(t, err, "Unexpected Error")
			return strings.Contains(s, "The changed identity schema configuration is invalid and could not be loaded.")
		})

		// ...

		default: //nothing
			if found {
				break
			}
		}
```

although they way you wrote it here is more Go-esque than that (and doesn't need a library). Just an alternative way to express this

aeneasr (approved):

PR #2969, opened 2022-12-19
fix: spurious cancelation of async webhooks, better tracing

alnr:
Previously, async webhooks (`response.ignore=true`) would be canceled early once the incoming Kratos request was served and it's associated context released. We now dissociate the cancellation of async hooks from the normal request processing flow.

I've also improved the OpenTelemetry instrumentation.

### DIFF
--- a/selfservice/hook/web_hook.go
+++ b/selfservice/hook/web_hook.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/tidwall/gjson"
@@ -282,6 +283,8 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
 	}
 
 	errChan := make(chan error, 1)
+	e.deps.Logger().WithRequest(req.Request).Info("Dispatching webhook")
+	t0 := time.Now()
 	go func() {
 		defer close(errChan)
 
@@ -307,9 +310,16 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
 	}()
 
 	if gjson.GetBytes(e.conf, "response.ignore").Bool() {
+		traceID, spanID := span.SpanContext().TraceID(), span.SpanContext().SpanID()
 		go func() {
-			err := <-errChan
-			e.deps.Logger().WithError(err).Warning("A web hook request failed but the error was ignored because the configuration indicated that the upstream response should be ignored.")
+			if err := <-errChan; err != nil {
+				e.deps.Logger().WithField("otel", map[string]string{
+					"trace_id": traceID.String(),
+					"span_id":  spanID.String(),
+				}).WithError(err).Warning("Webhook request failed but the error was ignored because the configuration indicated that the upstream response should be ignored.")
+			} else {
+				e.deps.Logger().WithField("duration", time.Since(t0)).Info("Webhook request succeeded")
+			}
 		}()
 		return nil
 	}
@@ -323,7 +333,7 @@ func parseWebhookResponse(resp *http.Response) (err error) {
 	}
 	var hookResponse rawHookResponse
 	if err := json.NewDecoder(resp.Body).Decode(&hookResponse); err != nil {
-		return errors.Wrap(err, "hook response could not be unmarshalled properly from JSON")
+		return errors.Wrap(err, "webhook response could not be unmarshalled properly from JSON")
 	}
 
 	var validationErrs []*schema.ValidationError
@@ -343,11 +353,11 @@ func parseWebhookResponse(resp *http.Response) (err error) {
 				Context: detail.Context,
 			})
 		}
-		validationErrs = append(validationErrs, schema.NewHookValidationError(msg.InstancePtr, "a web-hook target returned an error", messages))
+		validationErrs = append(validationErrs, schema.NewHookValidationError(msg.InstancePtr, "a webhook target returned an error", messages))
 	}
 
 	if len(validationErrs) == 0 {
-		return errors.New("error while parsing hook response: got no validation errors")
+		return errors.New("error while parsing webhook response: got no validation errors")
 	}
 
 	return schema.NewValidationListError(validationErrs)
--- a/selfservice/hook/web_hook.go
+++ b/selfservice/hook/web_hook.go
@@ -12,7 +12,9 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/tidwall/gjson"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	semconv "go.opentelemetry.io/otel/semconv/v1.11.0"
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/ory/kratos/ui/node"
@@ -30,7 +32,6 @@ import (
 	"github.com/ory/kratos/session"
 	"github.com/ory/kratos/text"
 	"github.com/ory/kratos/x"
-	"github.com/ory/x/otelx"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


selfservice/hook/web_hook.go
```diff
@@ -281,35 +267,66 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
...
+		httpClient                       = e.deps.HTTPClient(ctx)
+		async                            = gjson.GetBytes(e.conf, "response.ignore").Bool()
+		parseResponse                    = gjson.GetBytes(e.conf, "can_interrupt").Bool()
+		tracer                           = trace.SpanFromContext(ctx).TracerProvider().Tracer("kratos-webhooks")
+		cancel        context.CancelFunc = func() {}
+		spanOpts                         = []trace.SpanStartOption{trace.WithAttributes(attrs...)}
+		errChan                          = make(chan error, 1)
+	)
+	if async {
+		ctx, cancel = context.WithTimeout(context.Background(), 5*time.Minute)
```
aeneasr, line 289: I don't think this would fix the issue, the underlying HTTP client may be canceling the request, not the context? It has a timeout of I think 10 seconds.

Using a background context is also a possible security issue. Please make sure that we're not confusing the context and name this, for example, `asyncContext`, and only use it in the correct execution branch

selfservice/hook/web_hook.go
```diff
@@ -281,35 +267,66 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
...
+		spanOpts = append(spanOpts, trace.WithNewRoot())
+	}
+	ctx, span := tracer.Start(ctx, "Webhook", spanOpts...)
```
aeneasr, line 292: Here for example is the first instance where the incorrect context is used if async is true. The trace would be lost and without context!

aeneasr (commented):
Can you please add a test that shows this change to be effective? You can probably add a test which executes a webhook where the target takes ~5 seconds to respond or something similar. Even better, if test without your patch fails

alnr:
> Can you please add a test that shows this change to be effective? You can probably add a test which executes a webhook where the target takes ~5 seconds to respond or something similar. Even better, if test without your patch fails

Will add a test.

alnr:
Added failing test as first commit and my fix as second.

selfservice/hook/web_hook.go
aeneasr: What do you think of doing something like this:

```
req := // ...
var cancel // ...
if (async) {
  makeRequestContext, cancel = context.WithTimeout(context.Background(), time.Minute())
  req = req.WithContext(ctx)
}
```

that way, we ensure that the context is never leaving the async execution path :)

aeneasr: > The code is correct. For async webhooks, we start a new root span which is not associated with the incoming span.

But then the span would not appear as part of the request, and it would not be possible to correlate it to an incoming HTTP request? Is that really intentional?

selfservice/hook/web_hook_integration_test.go
```diff
@@ -839,3 +840,95 @@ func TestDisallowPrivateIPRanges(t *testing.T) {
...
+	select {
+	case <-time.After(200 * time.Millisecond):
+		t.Fatal("timed out waiting for webhook request to reach test handler")
+	case <-handlerEntered:
+		// ok
+	}
+	// at this point, a goroutine is in the middle of the call to our test handler and waiting for a response
+	incomingCancel() // simulate the incoming Kratos request having finished
+	timeout := time.After(200 * time.Millisecond)
+	for done := false; !done; {
```
aeneasr, line 903: `handlerEntered` should indicate that we already entered the HTTP webhook target. I think we don't need to also check the log entry

selfservice/hook/web_hook.go
```diff
@@ -274,29 +274,36 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
 			attribute.String("webhook.identity.nid", data.Identity.NID.String()),
 		)
 	}
+
 	var (
-		httpClient                       = e.deps.HTTPClient(ctx)
-		async                            = gjson.GetBytes(e.conf, "response.ignore").Bool()
-		parseResponse                    = gjson.GetBytes(e.conf, "can_interrupt").Bool()
-		tracer                           = trace.SpanFromContext(ctx).TracerProvider().Tracer("kratos-webhooks")
-		cancel        context.CancelFunc = func() {}
```
aeneasr: The cancel function was not needed IMO. We only used it to cancel the time out, but the underlying HTTP client already has a timeout itself (it's 30 seconds)

selfservice/hook/web_hook.go
```diff
@@ -274,29 +274,36 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
...
-	t0 := time.Now()
+
+	req = req.WithContext(ctx)
+	if ignoreResponse {
+		// This is one of the few places where spawning a context.Background() is ok. We need to do this
+		// because the function runs asynchronously and we don't want to cancel the request if the
+		// incoming request context is cancelled.
+		//
+		// The webhook will still cancel after 30 seconds as that is the configured timeout for the HTTP client.
+		req = req.WithContext(context.Background())
```
aeneasr, line 297: Instead of overshadowing the context, we now use `context.Background()` only where it is needed - in the request context. An alternative would be to not include a context if the request is async:

```req = req.WithContext(ctx)
 	if !ignoreResponse {
 	 	req = req.WithContext(ctx)
 	}
```

selfservice/hook/web_hook.go
```diff
@@ -320,16 +327,17 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
 		errChan <- nil
 	}()
 
-	if async {
+	if ignoreResponse {
 		traceID, spanID := span.SpanContext().TraceID(), span.SpanContext().SpanID()
+		logger := e.deps.Logger().WithField("otel", map[string]string{
+			"trace_id": traceID.String(),
+			"span_id":  spanID.String(),
+		})
```
aeneasr, line 335: Always include trace info

selfservice/hook/web_hook_integration_test.go
```diff
@@ -899,36 +901,23 @@ func TestAsyncWebhook(t *testing.T) {
 	}
 	// at this point, a goroutine is in the middle of the call to our test handler and waiting for a response
 	incomingCancel() // simulate the incoming Kratos request having finished
+	close(blockHandlerOnExit)
 	timeout := time.After(200 * time.Millisecond)
-	for done := false; !done; {
```
aeneasr: I don't think this was needed, as we already wait for `handlerEntered`

selfservice/hook/web_hook_integration_test.go
```diff
@@ -899,36 +901,23 @@ func TestAsyncWebhook(t *testing.T) {
...
-		if last := logHook.LastEntry(); last != nil {
-			msg, err := last.String()
-			require.NoError(t, err)
-			assert.Contains(t, msg, "Dispatching webhook")
+	var found bool
+	for !found {
+		for _, entry := range logHook.AllEntries() {
+			if entry.Message == "Webhook request succeeded" {
```
aeneasr, line 909: Make this independent of ordering and just look whether we find the succeed hook :)

aeneasr (approved):

selfservice/hook/web_hook.go
alnr: Yes, the span would be orphaned in that sense. But still easy enough to find.

If you made the async webhook a child span of the incoming Kratos request span, the child will outlive the parent span. That will trip up many tools and generally violate OpenTelemetry semantics.

OpenTracing used to have a special trace-trace relationship (`FollowsFrom`) which was perfect for cases like this. AFAIK, that did not make it into the OpenTelemetry specification, though, and tooling support was sketchy.

aeneasr: Oh I see, that is a very good explanation - thank you!

So I did not include this in the merge, would you be open to make a follow-up PR with the span fixed?

alnr: The merged patch is fine as-is. The OpenTelemetry people themselves are not sure how they want to models this: https://github.com/open-telemetry/opentelemetry-specification/issues/65

I'd suggest looking at some traces in Jaeger and Tempo, and deciding whether we're OK with how those tools interpret the data in practice. If there's a major hiccup, we can then change the instrumentation.

aeneasr: Perfect, that sounds like a plan!

PR #2989, opened 2022-12-27
fix: get session api doc

kelkarajay:


spec/api.json
```diff
@@ -4130,7 +4130,7 @@
             "oryAccessToken": []
           }
         ],
-        "summary": "This endpoint returns the session object with expandables specified.",
```
kelkarajay, line 4133: main change here

PR #2995, opened 2023-01-02
feat: remove duplicate queries from whoami calls

### LINKED ISSUE #2402: whoami performance
### Preflight checklist

### Describe your problem

I was doing some whoami load testing on oryd/kratos:v0.8.0-alpha.2 with (same datacenter in google cloud but not a localhost network) 2cpu 8gb postgres with 100s of sessions, 10s of identities.

min was 15ms, 50-percentile was 20ms, 99-percentile was 200ms under a regular traffic load -- variable 0-10qps.

------------

[whoami handler](https://github.com/ory/kratos/blob/51cdc341ef34eac96069cd6371d847e70d16b830/session/handler.go#L176)  calls [GetSessionByToken](https://github.com/ory/kratos/blob/51cdc341ef34eac96069cd6371d847e70d16b830/session/manager_http.go#L139)

[GetSessionByToken](https://github.com/ory/kratos/blob/51cdc341ef34eac96069cd6371d847e70d16b830/persistence/sql/persister_session.go#L110) does a `select * from sessions where token = ? and nid = ? limit 1` and then calls [GetIdentity](https://github.com/ory/kratos/blob/51cdc341ef34eac96069cd6371d847e70d16b830/persistence/sql/persister_identity.go#L353) contingent on session being found.

GetIdentity sequentially issues: 
`select * from identities where id = ? and nid = ?`
then
p.findVerifiableAddresses : `select ...  [verifiiable addresses] ...`
then
p.findRecoveryAddresses : `select ... [recovery addresses] ...`
then
p.injectTraitsSchemaURL
which is 
`ms, err := m.Config(ctx).IdentityTraitsSchemas()` [marshal + get bytes + decode](https://github.com/ory/kratos/blob/51cdc341ef34eac96069cd6371d847e70d16b830/driver/config/config.go#L521), quick loop to find matching schema, and then a read out of a schema.

I'm a javascript guy but I assume there is something akin to async / await in go. 

There are 4 sequential db round trips in the code path currently. Read session + read identities + read verifiable addresses + read recovery addresses + followed by what appears to be some schema parsing.

I'd be curious if this could be turned into (read session) + parallel (read identities, read addresses, read recovery addresses, do schema lookup). I think this would roughly half the median whoami latency. Session has identity id + nid which I think are good enough to parallelize all 3 of those other calls.

I could be wrong -- but the schema parsing step sounds expensive; I'm curious if that could be cached in some kind of global scope or registry on start-up. Maybe it is already cached and I can't read go -- hard for me to follow golang with stuff like `p.p.Marshal(kjson.Parser())`

-

### Version

oryd/kratos:v0.8.0-alpha.2 + postgres

### Additional Context

I saw what I think are a similar set of changes in the identities list call: https://github.com/ory/kratos/commit/14762d24a3844be982f3b4208dcf5239239a4475

Temporary schema cache in the local func scope (presumably because it actually is expensive?)
+ `Eager("VerifiableAddresses", "RecoveryAddresses")`

aeneasr:
Closes #2402

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


session/manager_http.go
```diff
@@ -178,13 +193,16 @@ func (s *ManagerHTTP) extractToken(r *http.Request) string {
...
+	ctx, span := s.r.Tracer(ctx).Tracer().Start(ctx, "sessions.ManagerHTTP.FetchFromRequest")
+	defer otelx.End(span, &err)
+
 	token := s.extractToken(r)
 	if token == "" {
 		return nil, errors.WithStack(NewErrNoCredentialsForSession())
 	}
 
-	se, err := s.r.SessionPersister().GetSessionByToken(ctx, token, ExpandEverything)
+	se, err := s.r.SessionPersister().GetSessionByToken(ctx, token, ExpandEverything, identity.ExpandDefault)
```
kelkarajay, line 205: We can optimize this further here. Session expand doesn't have to be `ExpandEverything` every time. The expand for devices is only ever needed in the `whoami` handler

kelkarajay (changes requested):

session/manager_http.go
aeneasr: That's true, ideally it would be `ExpandNothing`. But chaning this is a serious breaking change, which I did not want to address API versioning in this PR.

identity/handler.go
```diff
@@ -136,7 +136,7 @@ type listIdentitiesParameters struct {
 //	  default: errorGeneric
 func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	page, itemsPerPage := x.ParsePagination(r)
-	is, err := h.r.IdentityPool().ListIdentities(r.Context(), page, itemsPerPage)
+	is, err := h.r.IdentityPool().ListIdentities(r.Context(), ExpandDefault, page, itemsPerPage)
```
kelkarajay, line 139: Consider expand as query parameter here for the handler. Could be used to optimize the list op in the console view

session/manager_http.go
kelkarajay: I was thinking of a safer - `ExpandIdentity` rather than nothing since the Identity state is also needed to determine if the session may even be used/active

aeneasr: Ah true, that makes sense. We could however also optimize the SQL query to return the active state from the identity without needing to expand the identity itself using a JOIN

identity/handler.go
aeneasr: Definitely, but will need to track this as a follow up. Right now my focus is on improving performance of the default api request path :)

session/manager_http.go
kelkarajay: Yeah, makes sense. I'll also create an issue to track the `active` attribute relying on `expires_at` like we discussed earlier

aeneasr: 👍 would you be open to create such an issue? :)

identity/handler.go
aeneasr: 👍 would you be open to create such an issue? :)

alnr:
LGTM

session/manager_http.go
kelkarajay: - Use expiry timestamp vs boolean flag GH issue opened - https://github.com/ory-corp/cloud/issues/3895
- Fix Identity expand for session query https://github.com/ory-corp/cloud/issues/3896

@aeneasr

PR #3002, opened 2023-01-04
chore: fix missing index on courier list count

jonas-jonas:
Changes the order in which the count of all `courier_message`s is fetched from the database, to avoid scoping in the query and thus doing full table scans. Also adds missing indexes for the `status` and `recipient` filter.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


persistence/sql/migrations/sql/20230104193739000000_courier_list_index.up.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX courier_messages_status_idx ON courier_messages (status);
+
+CREATE INDEX courier_messages_status_idx ON courier_messages (recipient);
```
aeneasr, line 3: Indices are a way to tell the database which queries it should expect, so that it can optimize the lookup for these. That's why indices should always include all columns for a specific query. Optimally, indices are built in such a way that they can be re-used elsewhere.

- https://www.cockroachlabs.com/docs/v22.2/schema-design-indexes
- https://www.cockroachlabs.com/docs/v22.2/create-index
- https://www.cockroachlabs.com/docs/stable/indexes.html
- https://www.cockroachlabs.com/docs/v22.2/schema-design-indexes#best-practices
- https://www.cockroachlabs.com/docs/v22.2/performance-best-practices-overview#unique-id-best-practices

Queries can benefit from an index even if they only filter a prefix of its columns. For example, if you create an index of columns (A, B, C), queries filtering (A) or (A, B) can use the index, so you don't need to also index (A).

The more variety a column has, the better it is in the index, because it avoids hot spots. So always put the column with the highest "entropy" first:

```suggestion
CREATE INDEX courier_messages_recipient_status_idx ON courier_messages (nid, recipient, status);
```

In the example below, if recipient is not required, you *should be able to* change the query to

```
	if filter.Recipient != "" {
 		q = q.Where("recipient=?", filter.Recipient)
 	} else {
 		q = q.Where("recipient != ?", "")
  }
```

so that you don't need to add another index (please verify this first though).

Furthermore, you're creating the same index name twice, which will result in an error.

jonas-jonas: Thanks for the write-up. I have now read up a bit more on indexes, and in the process discovered the issue with the pagination (https://github.com/ory/x/pull/654). That also led to the wrong index being used, according to `EXPLAIN` (on postgres). Before, the execution plan included a `seq scan` ([the equivalent to a full table scan on postgres](https://use-the-index-luke.com/sql/explain-plan/postgresql/operations)):

```sql
EXPLAIN
SELECT
  *
FROM
  courier_messages
WHERE
  nid = $1
  AND status = 3
  AND recipient = ''
  AND "created_at" < now()
  OR (
    "created_at" = now()
    AND "id" > $1
  )
ORDER BY
  "created_at" DESC,
  "id" ASC
LIMIT
  11
```

| QUERY PLAN                                                                                                                                                                                                                                    |
| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Limit  (cost=34.56..34.56 rows=1 width=546)                                                                                                                                                                                                   |
|   ->  Sort  (cost=34.56..34.56 rows=1 width=546)                                                                                                                                                                                              |
|         Sort Key: created_at DESC, id                                                                                                                                                                                                         |
|         ->  **Seq Scan on courier_messages**  (cost=0.00..34.55 rows=1 width=546)                                                                                                                                                                 |
|               Filter: (((nid = '409449ca-186b-45fd-ad5d-9bf3923e33a1'::uuid) AND (status = 3) AND ((recipient)::text = ''::text) AND (created_at < now())) OR ((created_at = now()) AND (id > '409449ca-186b-45fd-ad5d-9bf3923e33a1'::uuid))) |


With the fix in place, it uses the existing `courier_messages_nid_id_idx` index, even with the `status` and `recipient` filters.

```sql
EXPLAIN
SELECT
  *
FROM
  courier_messages
WHERE
  nid = $1
  AND status = 3
  AND recipient = ''
  AND (
    "created_at" < now()
    OR (
      "created_at" = now()
      AND "id" > $1
    )
  )
ORDER BY
  "created_at" DESC,
  "id" ASC
LIMIT
  11
```

| QUERY PLAN                                                                                                                                                                           |
| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| Limit  (cost=6.07..6.07 rows=1 width=546)                                                                                                                                            |
|   ->  Sort  (cost=6.07..6.07 rows=1 width=546)                                                                                                                                       |
|         Sort Key: created_at DESC, id                                                                                                                                                |
|         ->  **Index Scan using courier_messages_nid_id_idx on courier_messages**  (cost=0.27..6.06 rows=1 width=546)                                                                     |
|               Index Cond: (nid = '409449ca-186b-45fd-ad5d-9bf3923e33a1'::uuid)                                                                                                       |
|               Filter: ((status = 3) AND ((recipient)::text = ''::text) AND ((created_at < now()) OR ((created_at = now()) AND (id > '409449ca-186b-45fd-ad5d-9bf3923e33a1'::uuid)))) |

My understanding is, that the `list` query itself does not actually need another index and https://github.com/ory/x/pull/654 should solve that issue on its own.

The initial issue was, however, that the "Count" query did not use an index. This is fixed by calling `Count` first, before executing the paginated statement. The reason is, that the pagination helper modifies the original `pop.Q` (statement builder) instance, which is thus also used in the `Count` query. 

Please let me know if I misunderstood anything here.

jonas-jonas: I have now done some more testing, and benchmarking on a CRDB instance.

The following results have been gathered using `EXPLAIN (opt, verbose)` on CRDB.

To cover all possible queries in the "ListMessages" method with an index, we would need 4 indexes:
- `CREATE INDEX … (nid, created_at DESC, id)` if neither `status` nor `recipient` is supplied
- `CREATE INDEX … (nid, status, created_at DESC, id)` if only `status` is supplied
- `CREATE INDEX … (nid, recipient, created_at DESC, id)` if only `recipient` is supplied
- `CREATE INDEX … (nid, status, recipient, created_at DESC, id)` if both are supplied

IIUC, this would probably hinder the performance, because it would make writes slower. It is also not scalable, if we want to add a filter for say `template_type` in the future.

I'd propose to only create `CREATE INDEX … (nid, created_at DESC, id)` to cover the case where neither `status` nor `recipient` is supplied, as that is probably the most common case. Additionally, we can add separate indexes on `status` and `recipient`: `CREATE INDEX ... (status)` (and `(recipient)` respectively).

I have looked at the query plans for the cases and these are the results:

| filter | cost
|--- | ---
| _`no filter`_ | `cost=14.82`
| `status = 3` | `cost=18.47`
| `recipient = 'x@ory.sh'` | `cost=18.47`
| `status = 3 AND recipient = 'x@ory.sh'` | `cost=18.48`

_(Without the separate indices on `status` and `recipient` the cost would be `cost=21.37` for the queries filtering for them.)_

Another improvement would be to use the `STORING` clause in the index (`CREATE INDEX … (nid, created_at DESC, id) STORING (body, template_data, etc...)`), to avoid the index join with the table `courier_messages` and return the data from the index directly, which would bring the cost for the query with no filters down to `9.37`. But this requires us to define the columns in the index definition, which is possible, but IMO out of the scope of this PR. Let me know how you want me to proceed here.

persistence/sql/migrations/sql/20230104193739000000_courier_list_index.up.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX courier_messages_nid_created_at_id_idx ON courier_messages (nid, created_at, id);
+
+CREATE INDEX courier_messages_recipient_at_id_idx ON courier_messages (recipient);
```
jonas-jonas, line 3: The `status` variant was already added earlier in `persistence/sql/migrations/sql/20210307130558000000_courier_status_index.cockroach.up.sql`

jonas-jonas:
This PR is now ready for another review. Please have a look at https://github.com/ory/kratos/pull/3002#discussion_r1063547037 where I have written down the analysis of the costs of these indices.

courier/test/persistence.go
```diff
@@ -126,12 +126,60 @@ func TestPersister(ctx context.Context, newNetworkUnlessExisting NetworkWrapper,
...
 			t.Run("on another network", func(t *testing.T) {
-				_, p := newNetwork(t, ctx)
-				ms, tc, _, err := p.ListMessages(ctx, filter, []keysetpagination.Option{})
+				nid1, p1 := newNetwork(t, ctx)
+				ms, tc, _, err := p1.ListMessages(ctx, filter, []keysetpagination.Option{})
 
 				require.NoError(t, err)
 				require.Len(t, ms, 0)
 				require.Equal(t, int64(0), tc)
+
```
jonas-jonas, line 135: This is what a failure of the test case looks like: https://github.com/ory/kratos/actions/runs/3856498251/jobs/6572792385#step:15:3563

persistence/sql/migrations/sql/20230104193739000000_courier_list_index.up.sql
jonas-jonas: And to add, since I didn't it include in the comments yet: Adding `recipient != ''` to the query and creating the index with `CREATE INDEX ... (nid, recipient, status, created_at DESC, id)` does have an effect:

filter | cost
-- | --
`recipient != '' AND status != -1` | `cost=17.32`
`recipient != '' AND status = 3` | `cost=17.32`
`recipient = 'x@ory.sh' AND status != -1` | `cost=18.48`
`status = 3 AND recipient = 'x@ory.sh'` | `cost=14.87`

So in essence, it slows down the query with no filters, and speeds up the others a bit. IMO that's not really what we want, because it's more likely that users want to see _just any_ message, until we have proper filtering, using a search engine. But do let me know what you think!

alnr: What's the expected query pattern here, realistically? The `status` column has low entropy and, if combined with a search for `recipient`, a scan through the result set is not a problem. A sequential scan isn't a problem if the set to scan is small. Combined indices beyond these:

- `CREATE INDEX … (nid, created_at DESC, id)`
- `CREATE INDEX … (nid, status, created_at DESC, id)`
- `CREATE INDEX … (nid, recipient, created_at DESC, id)`

probably don't give us anything useful.

I recommend you check if those indices get used in `ListMessages` (even if more filters are active) and call it a day.

jonas-jonas: Thanks @alnr. From my POV, users would initially go to the email delivery view (where we show this data) with no filters active. Then they might filter for `Abandoned` messages and _might_ search for a specific recipient. So we definitely need an index on the query with no filters, might need an index on the `status` query, and might need one on the `recipient` filter query. While possible, a query with both a `status` and a `recipient` filter is not that useful, so probably not to be expected.

I have settled on creating `CREATE INDEX … (nid, created_at DESC, id)` and indices for `status` (already exists) and `recipient` (new in this PR). 

That way, we get these costs:

| filter | cost
|--- | ---
| _`no filter`_ | `cost=14.82`
| `status = 3` | `cost=18.47`
| `recipient = 'x@ory.sh'` | `cost=18.47`
| `status = 3 AND recipient = 'x@ory.sh'` | `cost=18.48`

If we additionally create: 

- `CREATE INDEX … (nid, status, created_at DESC, id)`
- `CREATE INDEX … (nid, recipient, created_at DESC, id)`

We get `cost=14.82` on the queries that have either of the fields, but not for both, which would be fine, IMO. Question is, what is the overhead for writes, if we have all 3 of these indices active? I'd say we will do many more writes to this table, than this admin query is executed.

alnr: Updating 3 indices shouldn't be a problem. I'd say go for it and revisit if there's a problem.

courier/test/persistence.go
alnr: That's terrifying.

jonas-jonas: Yup, luckily this was _very_ unlikely, as there had to have been three preconditions: 

1. two messages with a timestamp collision (microsecond precision)
2. the first message needed to be a page token
3. the second message needed to have a UUID that had a higher "string" value (e.g. 1: "00...", 2: "11...")

persistence/sql/migrations/sql/20230104193739000000_courier_list_index.up.sql
jonas-jonas: Alright, I added the 2 missing indices to the PR. Should be good to go now.

persistence/sql/migrations/sql/20230104193739000000_courier_list_index.up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE INDEX courier_messages_nid_created_at_id_idx ON courier_messages (nid, created_at DESC, id);
+
+CREATE INDEX courier_messages_nid_status_created_at_id_idx ON courier_messages (nid, status, created_at DESC, id);
+
+CREATE INDEX courier_messages_nid_recipient_created_at_id_idx ON courier_messages (nid, recipient, created_at DESC, id);
```
aeneasr, line 5: Is `id` on this index neccessary? Do we also need an index for `nid, status, recipient`? By the waay, it would be more efficient to change the index order to `nid, recipient, status`. That would require us to change

```
	q := p.GetConnection(ctx).Where("nid=?", p.NetworkID(ctx))
	if filter.Status != nil {
		q = q.Where("status=?", *filter.Status)
	}
	if filter.Recipient != "" {
 		q = q.Where("recipient=?", filter.Recipient)
 	}
```

to


```
	q := p.GetConnection(ctx).Where("nid=?", p.NetworkID(ctx))
	if filter.Recipient != "" {
 		q = q.Where("recipient=?", filter.Recipient)
 	}
	if filter.Status != nil {
		q = q.Where("status=?", *filter.Status)
	}
```

Recipient has a higher cardinality than status, making it a better candidate for the index

courier/test/persistence.go
```diff
@@ -126,12 +126,60 @@ func TestPersister(ctx context.Context, newNetworkUnlessExisting NetworkWrapper,
 			assert.Equal(t, messages[len(messages)-1].ID, ms[0].ID)
 
 			t.Run("on another network", func(t *testing.T) {
```
aeneasr, line 128: Nice test

persistence/sql/migrations/sql/20230104193739000000_courier_list_index.up.sql
aeneasr: As far as I understand, we always have these query components:

* nid, created_at

If the user applies filters, they also add `recipient`, `status`, or both. Thus, this should handle all query cases:

```
CREATE INDEX … (nid ASC, created_at ASC, recipient ASC, status ASC)
CREATE INDEX … (nid ASC, created_at ASC, status ASC)
```

Given that you put recipient first:

```
	q := p.GetConnection(ctx).Where("nid=?", p.NetworkID(ctx))
	if filter.Recipient != "" {
 		q = q.Where("recipient=?", filter.Recipient)
 	}
	if filter.Status != nil {
		q = q.Where("status=?", *filter.Status)
	}
```

Please note that `created_at DESC` has no effect, as we explicitly use `ORDER BY created_at DESC`, which is not using the index for sorting but instead scans all ranges and orders them. Typically, we sort stuff ASC in indicies

jonas-jonas: From my testing, I can now say, that we need to put `recipient` and `status` first, before `created_at` in the index, otherwise the DB refuses to use the index in the query. Also see below on `desc` vs `asc` in the index.

>Please note that created_at DESC has no effect, as we explicitly use ORDER BY created_at DESC, which is not using the index for sorting but instead scans all ranges and orders them. Typically, we sort stuff ASC in indicies

That is not true from my testing on a CRDB, as you can see here: 

- `CREATE INDEX ... (nid, created_at); // this index is not used in the query plan and cost = 17.27`
- `CREATE INDEX ... (nid, created_at DESC); // index is used in the query plan and cost = 14.82`

But indeed, `id` does not seem to be needed. I'll remove it.

aeneasr (approved):

PR #3011, opened 2023-01-10
fix: resilient social sign in

aeneasr:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


selfservice/strategy/oidc/provider_test.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func TestClaimsValidate(t *testing.T) {
+	require.Error(t, new(Claims).Validate())
+	require.Error(t, (&Claims{Issuer: "not-empty"}).Validate())
+	require.Error(t, (&Claims{Issuer: "not-empty"}).Validate())
+	require.Error(t, (&Claims{Subject: "not-empty"}).Validate())
+	require.Error(t, (&Claims{Subject: "not-empty"}).Validate())
+	require.NoError(t, (&Claims{Issuer: "not-empty", Subject: "not-empty"}).Validate())
+	require.NoError(t, (&Claims{Issuer: "not-empty", Subject: "not-empty"}).Validate())
```
CaptainStandby, line 19: Why the duplication?

CaptainStandby (approved):
lgtm :+1:

selfservice/strategy/oidc/provider_test.go
aeneasr: Because I did not pay attention :D :D 

```suggestion
	require.Error(t, (&Claims{Issuer: "not-empty"}).Validate())
	require.Error(t, (&Claims{Issuer: "not-empty"}).Validate())
	require.Error(t, (&Claims{Subject: "not-empty"}).Validate())
	require.Error(t, (&Claims{Subject: "not-empty"}).Validate())
	require.NoError(t, (&Claims{Issuer: "not-empty", Subject: "not-empty"}).Validate())
```

selfservice/strategy/oidc/error.go
```diff
@@ -17,3 +25,17 @@ var (
 	ErrAPIFlowNotSupported = herodot.ErrBadRequest.WithError("API-based flows are not supported for this method").
 				WithReasonf("Social Sign In and OpenID Connect are only supported for flows initiated using the Browser endpoint.")
 )
+
+func logUpstreamError(l *logrusx.Logger, resp *http.Response) error {
+	if resp.StatusCode == http.StatusOK {
```
zepatrik, line 30: Why not
```suggestion
	if resp.StatusCode / 100 == 2 {
```
in case the status code is some other 2XX?

selfservice/strategy/oidc/error.go
```diff
@@ -17,3 +25,17 @@ var (
...
+		return nil
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		body = []byte(err.Error())
```
zepatrik, line 36:
```suggestion
		l = l.WithError(err)
```

selfservice/strategy/oidc/error.go
```diff
@@ -17,3 +25,17 @@ var (
...
+	}
+
+	l.WithField("response_code", resp.StatusCode).WithField("response_body", string(body)).Error("The upstream OIDC provider returned a non 200 status code.")
+	return errors.WithStack(herodot.ErrInternalServerError.WithReasonf("OpenID Connect provider returned a %d status code but 200 is expected.", resp.StatusCode))
```
zepatrik, line 40:
```suggestion
	return errors.WithStack(herodot.ErrInternalServerError.WithReasonf("OpenID Connect provider returned a %d status code but 2XX is expected.", resp.StatusCode))
```

zepatrik (commented):
Looks very reasonable, and much needed :sweat_smile:

selfservice/strategy/oidc/error.go
aeneasr: Good point

selfservice/strategy/oidc/error.go
aeneasr: Good point. I also thought of doing `200 >= x < 300` but decided against it. As far as I could tell from my research, all userinfo endpoints will return 200. Once exception might be a 3xx redirect, however that should be handled be the roundtripper. One problem with accepting all 200 is that we also accept 204 (no content) which would bring back the issue of an empty response body. 204 might be emitted when the service is having a disruption or something. So I think it makes sense to explicitly test for 200. For any service that doesn't return 200 but instead idk 201 (doesn't make sense though imo), we could adjust the logic. WDYT?

PR #3023, opened 2023-01-13
feat: parallelize get identity (spike)

aeneasr:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


persistence/sql/migratest/testdata/20220301102701_testdata.sql
```diff
@@ -1 +1,6 @@
-INSERT INTO identity_credentials (id, nid, config, identity_credential_type_id, identity_id, created_at, updated_at, version) VALUES ('4cefc264-4291-4abc-8f26-cc0217874f14', '884f556e-eb3a-4b9f-bee3-11345642c6c0', '{"hashed_password":"$argon2id$v=19$m=131072,t=2,p=1$lQFPaKxXqPL56/mU7vRi4w$6aldHyBnURt8sP8+xu41Ng"}', '22bff9ae-f5aa-45d7-803b-97ec0b4e7b32', '5ff66179-c240-4703-b0d8-494592cefff5', '2013-10-07 08:23:19', '2013-10-07 08:23:19', 0);
+INSERT INTO identity_credentials (id, nid, config, identity_credential_type_id, identity_id, created_at, updated_at,
+                                  version)
+VALUES ('4cefc264-4291-4abc-8f26-cc0217874f14', '884f556e-eb3a-4b9f-bee3-11345642c6c0',
```
aeneasr, line 3: The code changes actually found that this migration was incorrect. The identity now had two password credentials, which doesn't make sense. By making this a TOTP credential, the new code no longer complained.

persistence/sql/persister_identity.go
```diff
@@ -460,17 +462,81 @@ func (p *Persister) GetIdentity(ctx context.Context, id uuid.UUID, expand identi
...
+	}
+
+	if expand.Has(identity.ExpandFieldCredentials) {
+		eg.Go(func() error {
+			// We use WithContext to get a copy of the connection struct, which solves the race detector
+			// from complaining incorrectly.
+			//
+			// https://github.com/gobuffalo/pop/issues/723
+			if err := con.WithContext(ctx).
+				EagerPreload("IdentityCredentialType", "CredentialIdentifiers").
```
aeneasr, line 519: This could further be optimized. `IdentityCredentialType` is basically a static value that never changes, even across different nids (iirc). It could thus be cached, saving us a whole query (and busy connection).

persistence/sql/persister_session.go
```diff
@@ -267,24 +269,40 @@ func (p *Persister) GetSessionByToken(ctx context.Context, token string, expand
...
-		return nil, sqlcon.HandleError(err)
+	var (
+		i  *identity.Identity
+		sd []session.Device
+	)
+
+	eg, ctx := errgroup.WithContext(ctx)
+	if expand.Has(session.ExpandSessionDevices) {
+		eg.Go(func() error {
+			return sqlcon.HandleError(con.WithContext(ctx).
```
aeneasr, line 285: This is a range query :/

session/manager_http.go
```diff
@@ -214,7 +222,6 @@ func (s *ManagerHTTP) FetchFromRequest(ctx context.Context, r *http.Request) (_
 		return nil, errors.WithStack(NewErrNoActiveSessionFound())
 	}
 
-	se.Identity = se.Identity.CopyWithoutCredentials()
```
aeneasr, line 217: `CopyWithoutCredentials()` is kinda old. We now have a custom JSON marshaller registered on identity which will omit credentials. That's a much safer way than making sure that we delete the credentials properly. It also meant that we could not optimize the whoami endpoint, as it requires the credentials (sometimes) to make a decision whether to request 2FA or not.

session/manager_http.go
```diff
@@ -202,7 +202,15 @@ func (s *ManagerHTTP) FetchFromRequest(ctx context.Context, r *http.Request) (_
 		return nil, errors.WithStack(NewErrNoCredentialsForSession())
 	}
 
-	se, err := s.r.SessionPersister().GetSessionByToken(ctx, token, ExpandEverything, identity.ExpandDefault)
+	expand := identity.ExpandDefault
+	if s.r.Config().SessionWhoAmIAAL(r.Context()) == config.HighestAvailableAAL {
```
aeneasr, line 206: This is a copy of the logic in `DoesSessionSatisfy`. By checking this here, we don't have to use `.Load("Credentials")` in `DoesSessionSatisfy` but can instead load it in parallel (faster) here.

persistence/sql/persister_identity.go
```diff
@@ -460,17 +462,81 @@ func (p *Persister) GetIdentity(ctx context.Context, id uuid.UUID, expand identi
...
+		})
+	}
+
+	if err := eg.Wait(); err != nil {
+		return nil, sqlcon.HandleError(err)
+	}
+
+	i.VerifiableAddresses = verifiableAddresses
+	i.RecoveryAddresses = recoveryAddresses
+	i.InternalCredentials = credentials
```
aeneasr, line 534: Did it like this to prevent the race detector from complaining (incorrectly IMO)

persistence/sql/migratest/testdata/20200402142539_testdata.sql
```diff
@@ -12,6 +12,7 @@ INSERT INTO courier_messages (id, type, status, body, subject, recipient, create
 INSERT INTO courier_messages (id, type, status, body, subject, recipient, created_at, updated_at) VALUES ('a3ea4c30-0c6e-47b1-99ba-8fa69282e166', 1, 2, 'Hi, please verify your account by clicking the following link: <a href="http://127.0.0.1:4455/.ory/kratos/public/self-service/browser/flows/verification/email/confirm/AtsREGbtXu0RlIcwv3RPpxHEZNEcq3R9">http://127.0.0.1:4455/.ory/kratos/public/self-service/browser/flows/verification/email/confirm/AtsREGbtXu0RlIcwv3RPpxHEZNEcq3R9</a>', 'Please verify your email address', 'foo@ory.sh', '2013-10-07 08:23:19', '2013-10-07 08:23:19');
 
 INSERT INTO identity_credential_types (id, name) VALUES ('22bff9ae-f5aa-45d7-803b-97ec0b4e7b32', 'password');
+INSERT INTO identity_credential_types (id, name) VALUES ('8071b37b-0d54-4c6f-8234-72cffb4ce784', 'totp');
```
aeneasr, line 15: Needed for the next file's changes

aeneasr:
Did we crash MySQL? 😅 

https://github.com/ory/kratos/actions/runs/3918003783/jobs/6698161127#step:15:3573

persistence/sql/persister_session.go
```diff
@@ -106,34 +107,27 @@ func (p *Persister) ListSessions(ctx context.Context, active *bool, paginatorOpt
 			return sqlcon.HandleError(err)
 		}
 
-		if expandables.Has(session.ExpandSessionIdentity) {
```
aeneasr, line 109: @kelkarajay @jonas-jonas this would iterate over all sessions found, which is incredibly slow if you ahve large result sets. Instead, we're now using EagerPreload.

persistence/sql/persister_session.go
```diff
@@ -89,9 +91,8 @@ func (p *Persister) ListSessions(ctx context.Context, active *bool, paginatorOpt
 			}
 		}
 
-		// if len(expandables) > 0 {
-		if expandables.Has(session.ExpandSessionDevices) {
-			q = q.Eager(expandables.ToEager()...)
+		if len(expandables) > 0 {
+			q = q.EagerPreload(expandables.ToEager()...)
```
aeneasr, line 95: Please use EagerPreload when loading associations of lists @kelkarajay

persistence/sql/persister_session.go
```diff
@@ -106,34 +107,27 @@ func (p *Persister) ListSessions(ctx context.Context, active *bool, paginatorOpt
...
-		if expandables.Has(session.ExpandSessionIdentity) {
-			for index := range s {
-				sess := &(s[index])
-
-				i, err := p.GetIdentity(ctx, sess.IdentityID, identity.ExpandDefault)
-				if err != nil {
-					return err
-				}
-
-				sess.Active = sess.IsActive()
```
aeneasr, line 118: @kelkarajay here's one of those examples where it doesn't make sense to set IsActive in the persister. By moving it to MarshalJSON (which I have done in another PR), `Active` will always be returned correctly by the JSON API. In our Go code, everyone should instead is `IsActive()`.

aeneasr (commented):
@kelkarajay @jonas-jonas please see my comments

zepatrik (dismissed):
LGTM :+1:

session/handler.go
```diff
@@ -287,12 +287,12 @@ type listSessionsRequest struct {
 	Active bool `json:"active"`
 
 	// ExpandOptions is a query parameter encoded list of all properties that must be expanded in the Session.
-	// Example - ?expand=Identity&expand=Devices
 	// If no value is provided, the expandable properties are skipped.
 	//
 	// required: false
+	// enum: identity,devices
```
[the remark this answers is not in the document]
aeneasr: Very good point, I did not think about that. But I think in this case it's fine because the endpoint was just introduced a few days ago to prod. In the API we use lowercase everywhere, so I think it's good to streamline this!

persistence/sql/persister_session.go
jonas-jonas: Sounds good to me.

jonas-jonas (approved):
LGTM, nice changes. 

We can probably parallelize a few more persister methods using go routines... Stuff for the future.

aeneasr:
> LGTM, nice changes.
> 
> We can probably parallelize a few more persister methods using go routines... Stuff for the future.

Thank you!

It does come with cost such as concurrent connection and code complexity so I think that we’ll use this pattern mostly in critical paths.

PR #3042, opened 2023-01-23
feat: Let the database generate UUIDs on insert

alnr:
Combined with https://github.com/gobuffalo/pop/pull/807, this will allow for fast-path inserts for tables with UUID as primary key.


continuity/container.go
```diff
@@ -48,7 +48,7 @@ func (c Container) TableName(ctx context.Context) string {
 
 func NewContainer(name string, o managerOptions) *Container {
 	return &Container{
-		ID:         x.NewUUID(),
+		ID:         uuid.Nil,
```
jonas-jonas, line 51:
```suggestion
```

persistence/sql/persister_session.go
```diff
@@ -180,9 +180,13 @@ func (p *Persister) UpsertSession(ctx context.Context, s *session.Session) (err
 	s.NID = p.NetworkID(ctx)
 
 	return errors.WithStack(p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) error {
-		exists, err := tx.Where("id = ? AND nid = ?", s.ID, s.NID).Exists(new(session.Session))
-		if err != nil {
-			return sqlcon.HandleError(err)
+		exists := false
+		if !s.ID.IsNil() {
+			var err error
```
jonas-jonas, line 185:
```suggestion
```

Right?

persistence/sql/persister_errorx.go
```diff
@@ -35,7 +33,7 @@ func (p *Persister) Add(ctx context.Context, csrfToken string, errs error) (uuid
 	}
 
 	c := &errorx.ErrorContainer{
-		ID:        x.NewUUID(),
+		ID:        uuid.Nil,
```
jonas-jonas, line 36:
```suggestion
```

jonas-jonas (changes requested):
Just a few nits/cleanup opportunities (while we are at it ;)). Very nice cleanup! I like it! 🎉

alnr:
Regarding the `ID: uuid.Nil` comment on struct initialization: I wanted to make it clear that, when creating a new object, you don't get a valid UUID until you've persisted it. This changes the semantics somewhat, and is not easily possible in some places in Kratos. Still working on it.

examples/go/selfservice/recovery/main_test.go
```diff
@@ -19,7 +19,7 @@ func TestFunc(t *testing.T) {
 	publicURL, _ := testhelpers.StartE2EServer(t, "../../pkg/stub/kratos.yaml", nil)
 	client = pkg.NewSDKForSelfHosted(publicURL)
 
-	flow := performRecovery("dev+" + uuid.New().String() + "@ory.sh")
+	flow := performRecovery("dev+" + uuid.Must(uuid.NewV4()).String() + "@ory.sh")
```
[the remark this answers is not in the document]
alnr: This leads (I think) to import cycles. I'd also prefer to make the examples not depend on kratos internals

examples/go/selfservice/verification/main_test.go
```diff
@@ -19,7 +19,7 @@ func TestFunc(t *testing.T) {
 	publicURL, _ := testhelpers.StartE2EServer(t, "../../pkg/stub/kratos.yaml", nil)
 	client = pkg.NewSDKForSelfHosted(publicURL)
 
-	flow := performVerification("dev+" + uuid.New().String() + "@ory.sh")
+	flow := performVerification("dev+" + uuid.Must(uuid.NewV4()).String() + "@ory.sh")
```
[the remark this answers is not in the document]
alnr: see above

driver/registry_default.go
```diff
@@ -586,71 +586,77 @@ func (m *RegistryDefault) Init(ctx context.Context, ctxer contextx.Contextualize
 	bc := backoff.NewExponentialBackOff()
 	bc.MaxElapsedTime = time.Minute * 5
 	bc.Reset()
-	return errors.WithStack(
```
alnr, line 589: This change is much smaller if ignoring whitespace.

persistence/sql/persister_session.go
alnr: Doesn't matter, really.

session/manager_http.go
```diff
@@ -261,23 +267,28 @@ func (s *ManagerHTTP) DoesSessionSatisfy(r *http.Request, sess *Session, request
...
+		i := sess.Identity
+		if i == nil {
+			i, err = s.r.IdentityPool().GetIdentity(ctx, sess.IdentityID, identity.ExpandCredentials)
+			if err != nil {
+				return err
+			}
+			sess.Identity = i
+		} else if len(i.Credentials) == 0 {
+			// If credentials are not expanded, we load them here.
+			if err := s.r.PrivilegedIdentityPool().HydrateIdentityAssociations(r.Context(), i, identity.ExpandCredentials); err != nil {
```
zepatrik, line 279:
```suggestion
			if err := s.r.PrivilegedIdentityPool().HydrateIdentityAssociations(ctx, i, identity.ExpandCredentials); err != nil {
```

driver/registry.go
```diff
@@ -197,6 +199,12 @@ func ReplaceTracer(f func(*otelx.Tracer) *otelx.Tracer) func(o *options) {
 	}
 }
 
+func Inspect(f func(reg Registry) error) func(o *options) {
```
zepatrik, line 202: What is this used for? Just external usage?

zepatrik (commented):
Nice, I just have a few questions but no blocking issue.

identity/identity.go
```diff
@@ -301,7 +299,10 @@ func (i *Identity) ParseCredentials(t CredentialsType, config interface{}) (*Cre
 }
 
 func (i *Identity) CopyWithoutCredentials() *Identity {
-	var ii = *i
+	i.lock().RLock()
```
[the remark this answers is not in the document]
alnr: correct

driver/registry.go
alnr: Exactly.

continuity/test/persistence.go
```diff
@@ -62,6 +64,7 @@ func TestPersister(ctx context.Context, p interface {
 			expected := createContainer(t)
 
 			require.NoError(t, p.SaveContinuitySession(ctx, &expected))
+			require.NotEqual(t, uuid.Nil, expected.ID)
```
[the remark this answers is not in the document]
alnr: I couldn't find an existing reference to `require.Zero` or `require.NotZero` in the codebase, so I kept with the existing style. We could refactor all instances at one in a later PR if we want to.

driver/registry_default.go
```diff
@@ -586,71 +586,77 @@ func (m *RegistryDefault) Init(ctx context.Context, ctxer contextx.Contextualize
...
+		m.persister = p.WithNetworkID(net.ID)
+		return nil
+	}, bc)
 
-			m.persister = p.WithNetworkID(net.ID)
-			return nil
-		}, bc),
-	)
+	if err != nil {
+		return errors.WithStack(err)
```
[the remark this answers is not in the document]
alnr: fixed

session/manager_http.go
alnr: Good catch, fixed.

driver/registry_default.go
```diff
@@ -586,71 +586,77 @@ func (m *RegistryDefault) Init(ctx context.Context, ctxer contextx.Contextualize
...
 
-			m.persister = p.WithNetworkID(net.ID)
-			return nil
-		}, bc),
-	)
+	if err != nil {
+		return err
+	}
+
+	if o.inspect != nil {
```
zepatrik, line 654: Hm not sure what the use-case is, but I see two code-paths that `return nil` before this is reached. Just FYI.

zepatrik (approved):
LGTM :+1:

session/manager_http_test.go
```diff
@@ -549,6 +549,25 @@ func TestDoesSessionSatisfy(t *testing.T) {
 			} else {
 				require.NoError(t, err)
 			}
+
+			// This should still work even if the session does not have identity data attached yet...
```
aeneasr, line 553: Why? :D Or rather, why was it not an issue before?

aeneasr (approved):
LGTM! Once comments from @zepatrik are addressed this is mergable :)

session/manager_http_test.go
alnr: This change makes our code more resilient against future changes with regards to expanding or not expanding the identity, for example during `whoami` calls. Previously, we would rely on the `whoami` handler to know that it needed the identity credentials later, and "pre-load" them. That's brittle. I played around with trying to enable a whoami call without expanding the identity (`GET `/sessions/whoami?expand=none`, for performance) and the handler would panic.

This change allows us to more aggressively minimize necessary queries without breaking functionality.

driver/registry_default.go
aeneasr: @alnr please double check this

alnr: You mean on lines 637 and 647? Those returns are returning from the closure passed to `backoff.Retry`. We return from `Init` only if `backoff.Retry` returns non-nil. So this codepath is always reached in the happy path.

aeneasr:
Nice, LGTM. Rerunning CI will hopefully resolve the curl timeouts.

Issue #2825, opened 2022-10-19
Email rendering & delivery via API / HTTP


joshbalfour:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [x] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Hiya,
First off: huge fan of what the team has accomplished here, this is by no means a criticism just a feature idea 😄 

I've been setting up email templates (did SMS first) and was thinking:
Would it be more in the project ethos of  "API-only - without templating or theming headaches" to have email/SMS be processed via a user-built API rather than rendered via the templating system and sent via SMTP?
I got to this by running through a requirement that's come up of being able to theme the emails depending on who it's sent to (we have centralised auth but will theme the login screens based on the product you're logging in to).

### Describe your ideal solution

Much like the way you currently configure SMS delivery via the config, you could configure an endpoint that would be sent the message type (e.g. `recovery.valid`), the context variables (e.g. `.RecoveryURL`) and who to deliver it to.

### Workarounds or alternatives

It's incredibly cursed but I could:

1. Change the templates to:
body - render a JSON string with the context variables
subject - the message type

2. Build an SMTP server that would take the to/subject/body from the email, parse the context from the body and message type from the subject

3. Render the new email body and subject based on the message type, context, and who it's going to, then send it on its way

If this feature was built then I could cut out the first two steps here so it'd be a lot less cursed.

### Version

v0.10.1

aeneasr:
Josh, I think this is a great idea! We will probably keep email rendering in for ease of use, but it could make sense to have an API-based courier. This has been requested also in issue https://github.com/ory/kratos/issues/1030

I think it should be relatively straight forward, we can probably re-use the SMS courier as that already is based on a REST API.

hugotiburtino:
joshbalfour, you read my thoughts! 😄

[...]

Issue #3008, opened 2023-01-09
Attachments in E-Mails


SlateDave:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

I'd like to attach some images used within the E-Mail in order to don't rely on linking them to my web-server.

By adding images with links to your web-server you must ensure they stay online and don't change in the future (to preserve the original design for the e-mail).

The benefit of attaching them in the e-mail is also that they don't get automatically blocked from loading by some e-mail clients.

### Describe your ideal solution

The best way is to add an attachment and reference this on the image.

### Workarounds or alternatives

I know you can convert images to base64 and use this as link but many e-mail clients don't support this feature and you end up with a e-mail with non-working images.

### Version

0.11.0

SlateDave:
For example: Gmail doesn't support base64 images: https://www.caniemail.com/features/image-base64/

RobinGasi:
+1

vinckr:
IMO I dont see the problem with storing them on a server and updating that link whenever the template changes. In Ory Network for example you add the base64 to the template, update, and it handles the storing&serving of images.
I think that would also work for your problem?

SlateDave:
Base64 is not supported by all email clients.

Using a link and saving the images on a server will introduce maintenance concerns. It could also lead to older emails being corrupted if, for example, you change your logo in the future and now there is a new logo in the emails from the past that does not match the overall design of the email. So we would have to preserve those old links and switch to a new link.

However, the biggest disadvantage of using links is that almost all email clients nowadays block loading of external images until the user manually approves them, which makes the email ugly until the user agrees to a "possible invasion of privacy".

So the best solution is to attach the images and reference those attachments. This way they will be loaded by any email client and preservation maintenance will be avoided.

kmherrmann:
[...]

Issue #1030, opened 2021-02-01
Support sending emails through CRM transactional email


obukhov:
**Is your feature request related to a problem? Please describe.**

The current implementation only supports sending emails through SMTP using templates bundled with kratos. It would be nice to be able to delegate templating and sending the email to an external service with API call.

**Describe the solution you'd like**

First of all, there should be a configuration parameter to change strategy between built-in templating and sending via SMTP to an API call. The question is how to implement customization - there are a lot of systems/services out there with different APIs. I was thinking of introducing an intermediate service translating kratos events to a target system API. The communication between kratos and "adapter" service can be implemented in several different ways:
- webhooks - HTTP calls with JSON payload standardized and documented
- gRPC call - similar to the one above but more performant
- queueing a message - but then again there are a lot of queueing systems: Kafka, AMPQ, ...

Not sure what would be the best in this case but in any case, it's possible to also provide some boilerplate project to start with to implement a specific system adapter.

**Describe alternatives you've considered**

The only possibility I see is to mock the SMTP server as with MailSlurper but then you would also need to pack and unpack event data into the message body that seems to be a nasty workaround.

**Additional context**

If you can provide some guidance (what you think in general about the idea and which of described options to pick) I may try to prepare a PR with it. I've looked in Sender and Courier services I think the solution can be using sender strategy there.

zepatrik:
It would make sense to then externalize the current SMTP and templating solution as well. This is kind of required anyway (#1024), and having this intermediate API would make it possible to plug in other systems. What do you think @aeneasr ? We can only support the SMTP+template solution officially and then the community can write adapters as needed.

aeneasr:
Yes, thank you for writing down this issue! So generally we target services like SparkPost or Mailgun which have SMTP APIs with the current implementation. This does not allow using their template language however - it would need Mailgun/SparkPost specific API calls for that.

Are you planning on using a CRM to handle your verification and recovery mails? Or do you want to generally sync your users with your CRM?

obukhov:
Hi @aeneasr. Yeah sorry for a bit misleading description. I've meant any email delivery system that allows you to send data (trigger name and set of key-value parameters) and then manage to templates and actual delivery. This can be either services like Mailgun or Sendgrid, but can also be more sophisticated CRM systems. This wasn't about CRM users sync though (also an important but different subject).

[...]

PR #3048, opened 2023-01-26
feat: send emails via http api endpoint instead of smtp

barnarddt:
This change adds a new delivery method to the courier called `mailer`. Similar to SMS functionality it posts a templated Data model to a API endpoint.  This API can then send emails via a CRM or any other mechanism that it wants.

`Mailer` still uses the existing email data models so any new email added will automatically be sent to the API/CRM as well.

## Related issue(s)
Resolves https://github.com/ory/kratos/issues/2825
Also see #1030 and #3008 
Documentation PR https://github.com/ory/docs/pull/1298

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


karaimin:
Nice work @barnarddt. I was looking for this feature for a long time. Hope it will be reviewed and released soon.

courier/stub/request.config.mailer.jsonnet
```diff
@@ -0,0 +1,11 @@
+function(ctx) {
```
aeneasr, line 1: This looks like a good default value for the body payload. Maybe it makes sense to use it as the default?

driver/config/config.go
```diff
@@ -78,6 +78,8 @@ const (
 	ViperKeyCourierTemplatesVerificationValidEmail           = "courier.templates.verification.valid.email"
 	ViperKeyCourierTemplatesVerificationCodeInvalidEmail     = "courier.templates.verification_code.invalid.email"
 	ViperKeyCourierTemplatesVerificationCodeValidEmail       = "courier.templates.verification_code.valid.email"
+	ViperKeyCourierMailerEnabled                             = "courier.mailer.enabled"
```
aeneasr, line 81: I would prefer this layout:

```
courier:
  email:
    strategy: smtp|http
  http:
    url: ...
  smtp:
    connection_uri: ...
```

aeneasr (commented):
Thank you for the PR! This is already going in a good direction. Here are some improvement ideas:

1. Add tests for the new config options (right now the config won't work because it's not in the config JSON Schema)
2. Add an e2e tests to test/e2e which ensures that this works end-to-end
3. Make a few changes to the configuration layout
4. Add documentation :)

courier/stub/request.config.mailer.jsonnet
barnarddt: Not sure how we'd make it the default as the body of the template engine is a file URI not a absolute value?

barnarddt:
> 2\. Add an e2e tests to test/e2e which ensures that this works end-to-end

@aeneasr 

For the end2end test what would be the preferred setup? Configure kratos to use the http email functionality with the url being a localhost port that the end2end test will listen on and ensure that that gets called when a admin call gets issued to do something that will trigger an email to be sent?

SoGoDev:
Guys, any updates on this PR?

courier/courier.go
```diff
@@ -49,11 +49,12 @@ type (
...
 
 	courier struct {
-		smsClient   *smsClient
-		smtpClient  *smtpClient
-		deps        Dependencies
-		failOnError bool
-		backoff     backoff.BackOff
+		smsClient    *smsClient
+		smtpClient   *smtpClient
+		mailerClient *mailerClient
```
jonas-jonas, line 54: Could we name this `httpClient` instead? The `smtpClient` also refers to the protocol it's using, and `mailer` is arguably a bit ambiguous.

courier/mailer.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/ory/kratos/request"
+)
+
+type mailerDataModel struct {
+	Recipient    string
+	TemplateType TemplateType
+	TemplateData EmailTemplate
+}
+
+type mailerClient struct {
```
jonas-jonas, line 22: And similarly here.

jonas-jonas (changes requested):
This already looks fantastic. And sorry for the delays on our side. 

As for the E2E tests, we already have some "mock" services that just provide a target for the webhooks to be tested with, here: https://github.com/ory/kratos/blob/master/test/e2e/mock. That service is started in [test/e2e/modd.conf](https://github.com/ory/kratos/blob/master/test/e2e/modd.conf)

Maybe you could set up some kind of echo server here, that makes all incoming messages available on an endpoint and assert that the expected messages end up in there?

courier/courier.go
kmherrmann: +1 ... I thought the same

kmherrmann:
Thanks @barnarddt - this is a great addition. Would you be up to address the requested changes? It'd be fantastic to get http email delivery support merged!

Resolves #2825

aeneasr (commented):
Great job! :)

Could you please add an end-to-end Testcafe / Cypress test (or some other type of integration test in Go) to prove that this does indeed work as intended? Thank you!

aeneasr:
> > 2. Add an e2e tests to test/e2e which ensures that this works end-to-end
> 
> @aeneasr
> 
> For the end2end test what would be the preferred setup? Configure kratos to use the http email functionality with the url being a localhost port that the end2end test will listen on and ensure that that gets called when a admin call gets issued to do something that will trigger an email to be sent?

Sorry - I missed this comment :(

Yes that could work! We now support PlayWright and there you can set up a mock server easily: https://playwright.dev/docs/test-webserver

So probably the test would be this:

1. set up web server
2. trigger registration flow
3. ensure web server got api request with correct payloads (potentially using a snapshot)

That should be it!

hperl:
@barnarddt thank you so much for your contribution! I did some clean-ups in https://github.com/ory/kratos/pull/3341 to get this across the finish line:

* added the config values to the schema, including a default configuration for the `request_config`
* added an E2E test
* added subject and body to the template directly

hperl:
Closing in favor of https://github.com/ory/kratos/pull/3341. Again, big thanks to @barnarddt for the excellent work!

Issue #2946, opened 2022-12-12
Allow import peppered password hashes


ci42:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

 We need support for importing [peppered (and salted) password hashes](https://en.wikipedia.org/wiki/Pepper_(cryptography)).

### Describe your ideal solution

The PHC string format has no support for pepper values and they shouldn't stored alongside the password hash anyway. The hash comparsion in [hash_comparator.go](https://github.com/ory/kratos/blob/master/hash/hash_comparator.go) uses free functions which can not hold state like the pepper value (secret salt).

I would propose to introduce an interface analogous to the [Hasher](https://github.com/ory/kratos/blob/master/hash/hasher.go) interface, like this:

```
type HashComparator interface {
    // compare password and hash
    Compare(password []byte, hash []byte) error
    // wether the HashComparator understand the hash format
    Understands(hash []byte) bool
}
```

and to rewrite the free functions in [hash_comparator.go](https://github.com/ory/kratos/blob/master/hash/hash_comparator.go) to types which implements the above interface. The types should be configured like `hashers` in the kratos configuration file, so that only configured hash comparator will be checked in [`Compare`](https://github.com/ory/kratos/blob/master/hash/hash_comparator.go#L28) and not all supported hash comparators.

This will make it easier to support other legacy hashing algorithms. Changing [hash_comparator.go](https://github.com/ory/kratos/blob/master/hash/hash_comparator.go) would no longer be necessary to add new hashing algorithms, you could just add a new file with the implementation.

### Workarounds or alternatives

Reset the passwords of all imported users, which is not desirable.

### Version

0.10.1

### Additional Context

Support for other legacy hashing algorithms has been discussed in #2422 and implemented in #2725 and #2741.

[...]

PR #3057, opened 2023-01-30
feat: allow import peppered/salted password hashes #2946

ci42:
## Related issue(s)

#2946 

## Checklist

- [] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [x] I am following the
      the changes.
      works.


embedx/config.schema.json
```diff
@@ -2327,7 +2327,29 @@
       },
       "additionalProperties": false
     },
-    "ciphers": {
+    "passwordhash_comparators": {
```
zepatrik, line 2330: Why is this a separate top level config when we already have the `hashers` config? Can you add it there instead?

hash/hash_comparator.go
```diff
@@ -458,9 +624,6 @@ func decodeSSHAHash(encodedHash string) (hasher string, salt, hash []byte, err e
 		hasher = "sha512"
 		index_of_hash_begin = 9
 		index_of_salt_begin = 64
-
-	default:
-		return "", nil, nil, ErrInvalidHash
```
zepatrik, line 463: I don't think this should be removed, please revert.

hash/hasher_test.go
```diff
@@ -251,19 +253,44 @@ func TestCompare(t *testing.T) {
 	assert.Error(t, hash.Compare(context.Background(), []byte("tesu"), []byte("$scrypt$ln=16384,r=8,p=1$2npRo7P03Mt8keSoMbyD/tKFWyUzjiQf2svUaNDSrhA=$MiCzNcIplSMqSBrm4HckjYqYhaVPPjTARTzwB1cVNYE=")))
 	assert.Error(t, hash.Compare(context.Background(), []byte("tesu"), []byte("$scrypt$ln=abc,r=8,p=1$2npRo7P03Mt8keSoMbyD/tKFWyUzjiQf2svUaNDSrhA=$MiCzNcIplSMqSBrm4HckjYqYhaVPPjTARTzwB1cVNYE=")))
 
+	assert.Nil(t, hash.Compare(context.Background(), []byte("test"), []byte("$md5$CY9rzUYh03PK3k6DJie09g==")))
```
zepatrik, line 256: Moving the block of md5 assertions makes it very hard to review, please move it back so the diff properly shows what changed.

hash/hash_comparator.go
```diff
@@ -28,45 +28,106 @@ import (
...
-		return CompareSSHA(ctx, password, hash)
-	case IsSHAHash(hash):
-		return CompareSHA(ctx, password, hash)
-	case IsFirebaseScryptHash(hash):
-		return CompareFirebaseScrypt(ctx, password, hash)
-	case IsMD5Hash(hash):
-		return CompareMD5(ctx, password, hash)
-	default:
-		return errors.WithStack(ErrUnknownHashAlgorithm)
+var hashComparators []PasswordHashComparator
```
zepatrik, line 31: I don't think we need to refactor the package like this because there is already an established way to pass config to the hash comparators. Regardless of whether the code is nicer or not, in the end it makes it impossible to review with refactoring and new features mixed. Please use the existing way to pass the hash config, and revert all the refactoring. I am open to discuss that, but it should really be a separate PR.

zepatrik (changes requested):
Thanks for putting in the effort. however we need to fix a few things. It is also not a good idea to mix new features with refactoring in one PR, so I would suggest you either split this in two PRs, or we don't do the refactoring at all.
For pepper secrets, I think we need to include the possibility to rotate the secret as well. It is not cheap to do, as we need to hash the password multiple times, but there is no other way to update peppers or increase the secret entropy without a rotation mechanism, so this is a requirement IMO.

aeneasr:
See also feedback from https://github.com/ory/kratos/issues/2946#issuecomment-1451020646

hash/hash_comparator.go
ci42: > there is already an established way to pass config to the hash comparators.

Maybe I missed something, but other than the PHC string, I don't see any way to pass conf data in the existing approach. IMHO the PHC string is not the right place for pepper values (see the discussion [here](https://github.com/ory/kratos/issues/2946#issuecomment-1453139768)). So refactoring is a precondition and not just to make the code nicer.

embedx/config.schema.json
```diff
@@ -2327,7 +2327,29 @@
...
+      "type": "array",
+      "title": "",
+      "description": "",
+      "items": {
+        "required": ["Id"],
+        "type": "object",
+        "properties": {
+          "Id": {
```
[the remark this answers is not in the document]
ci42: done

embedx/config.schema.json
ci42: moved to the `hashers` config

ci42:
@aeneasr @zepatrik
I have completely rewritten the PR. The new hash comparison function is now similar to the existing hash comparison functions. t now supports various old hashing methods (md5, sha1, sha256, sha512) as well as optional salts, peppers and a work factor (number of iterations). The old hash itself is hashed with argon2id for improved security (e.g. offline attacks in case of a database leak).

All parameters are encoded in the hash string. The hash string has the form: 
`$legacy$<legacy hash parameters>$<argon2id in phc format>` - where the legacy hash parameters are encoded as follows  
`alg=<md5|sha1|sha256|sha512>,format=<base64 encoded format, e.g: '{PEPPER}-{PWD}-{SALT}'>,i=<number of iterations>,(salt=<base64 encoded salt>,pepper=<base64 encoded pepper>)`

An example using a salted and peppered md5 in the format `{SALT}::{PWD}::{PEPPER}`with 3 iterations: `$legacy$alg=md5,format=e1NBTFR9Ojp7UFdEfTo6e1BFUFBFUn0,i=3,salt=7aCA,pepper=/v7//w$argon2id$v=19$m=65536,t=2,p=2$YuYcJII0fm8UXm8XiCqimA$xzUihMR5G40U3Gf2CmTfNhXDYTtva6ykbjGnfQnWCac`

hash/hash_comparator.go
```diff
@@ -558,3 +576,78 @@ func decodeMD5Hash(encodedHash string) (pf, salt, hash []byte, err error) {
...
+	parts := strings.SplitN(encodedHash, "$", 4)
+	if len(parts) != 4 {
+		return "", "", "", 0, nil, nil, ErrInvalidHash
+	}
+	phc = "$" + parts[3]
+	legacyParams := strings.Split(parts[2], ",")
+	for _, lp := range legacyParams {
+		switch kv := strings.Split(lp, "="); kv[0] {
+		case "alg":
+			alg = kv[1]
```
zepatrik, line 592: Here is a potential for index out of bounds panics, I'd propose to use this instead:

```suggestion
		k, v, found := strings.Cut(lp, "=")
		if !found {
			return "", "", "", 0, nil, nil, ErrInvalidHash
		}
		switch k {
		case "alg":
			alg = v
```

hash/hash_comparator.go
```diff
@@ -558,3 +576,78 @@ func decodeMD5Hash(encodedHash string) (pf, salt, hash []byte, err error) {
 		return nil, nil, nil, ErrInvalidHash
 	}
 }
+
+// decodeLegacyHash decodes argon2id wrapped legacy password hash.
+// format: $legacy$alg=<md5|sha1|sha256|sha512>,format=<base64 encoded format, e.g: '{PEPPER}-{PWD}-{SALT}'>,i=<number of iterations>,(salt=<base64 encoded salt>,pepper=<base64 encoded pepper>)$<argon2id encoded hash>
+func decodeLegacyHash(encodedHash string) (phc, alg, format string, iterations int, salt, pepper []byte, err error) {
```
zepatrik, line 582: As is, this does not require any of the params to be set. Do we need to set some defaults, e.g. for the format? Or should we return an `ErrInvalidHash` in case some of the required params are missing?

hash/hash_comparator.go
```diff
@@ -558,3 +576,78 @@ func decodeMD5Hash(encodedHash string) (pf, salt, hash []byte, err error) {
...
+			}
+		}
+	}
+	return
+}
+
+func legacyHash(alg, format string, iterations int, salt, pepper, pwd []byte) ([]byte, error) {
+	if len(pwd) < 1 || !strings.Contains(format, "{PWD}") || iterations < 1 || !(alg == "md5" || alg == "sha1" || alg == "sha256" || alg == "sha512") {
+		return nil, ErrInvalidHash
+	}
```
zepatrik, line 622: I guess validation is covered in this part.

hash/hash_comparator.go
```diff
@@ -558,3 +576,78 @@ func decodeMD5Hash(encodedHash string) (pf, salt, hash []byte, err error) {
...
+			hash = sum[:]
+		case "sha512":
+			sum := sha512.Sum512(toHash)
+			hash = sum[:]
+		default:
+			return nil, ErrUnknownHashAlgorithm
+		}
+		toHash = hash
+	}
+	return hash, nil
```
zepatrik, line 652: I think this is a bit easier to understand:

```suggestion
	toHash := []byte(strings.NewReplacer(replacements...).Replace(format))

	var hasher hash.Hash
	switch alg {
	case "md5":
		hasher = md5.New() //#nosec G401 -- compatibility for imported passwords
	case "sha1":
		hasher = sha1.New() //#nosec G401 -- compatibility for imported passwords
	case "sha256":
		hasher = sha256.New()
	case "sha512":
		hasher = sha512.New()
	default:
		return nil, ErrUnknownHashAlgorithm
	}
	for i := 1; i <= iterations; i++ {
		toHash = hasher.Sum(toHash)
	}
	return toHash, nil
```

zepatrik (commented):
Looks good now, thanks for revisiting!
I just have some minor suggestions, nothing critical.

hash/hash_comparator.go
ci42: yep

hash/hash_comparator.go
ci42: `toHash = hasher.Sum(toHash)` does not do what you and I think 😄  Should be:
```
hasher.Write(toHash)
toHash = hasher.Sum([]byte{})
hasher.Reset()
```

ci42:
@zepatrik Thanks for the review. I've included your suggested changes, added some more tests and rebased the PR onto the current master.

By the way, I made a [PR](https://github.com/ory/kratos/pull/3186) with some minor hash_comparator.go refactorings. Maybe you can have a look at it as well.

zepatrik (approved):
Nice :+1:

PR #3059, opened 2023-01-30
feat: webhooks that update identities

aeneasr:
Introducing a new configuration `response.parse` in webhooks. This enables updating of identity data during registration, including admin/public metadata, identity traits, enabling/disabling identity, and modifying verified/recovery addresses.

Closes #2161

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/persister_identity.go
+++ b/persistence/sql/persister_identity.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ory/x/pointerx"
+
 	"golang.org/x/sync/errgroup"
 
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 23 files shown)


persistence/sql/persister_identity.go
```diff
@@ -257,9 +259,25 @@ func (p *Persister) normalizeAllAddressess(ctx context.Context, id *identity.Ide
...
+		v.Via = x.Coalesce(v.Via, identity.AddressTypeEmail)
+		if len(v.Status) == 0 {
+			if v.Verified {
+				v.Status = identity.VerifiableAddressStatusCompleted
+			} else {
+				v.Status = identity.VerifiableAddressStatusPending
+			}
+		}
+
+		if v.Verified && (v.VerifiedAt == nil || time.Time(*v.VerifiedAt).IsZero()) {
```
jonas-jonas, line 276: Does this mean, you can set the email address to `verified` in the webhook but not set the timestamp? Shouldn't that be possible as well? 

And could you add some comments around this to explain what's going on in this (changed) part? I find it a bit difficult to gauge what this is for.

jonas-jonas (changes requested):
Looks great so far, just some questions. I want to test it out locally for a bit, but going to do that on Friday on the train.

selfservice/hook/web_hook_integration_test.go
```diff
@@ -607,6 +609,90 @@ func TestWebHooks(t *testing.T) {
...
+
+			t.Run("case=identity has updated verified addresses", func(t *testing.T) {
+				actual := run(t, expected, http.StatusOK, []byte(`{"identity":{"traits":{"email":"some@other-example.org"},"verifiable_addresses":[{"value":"some@other-example.org","via":"email"}]}}`))
+				snapshotx.SnapshotT(t, &actual)
+			})
+
+			t.Run("case=identity has updated recovery addresses", func(t *testing.T) {
+				actual := run(t, expected, http.StatusOK, []byte(`{"identity":{"traits":{"email":"some@other-example.org"},"recovery_addresses":[{"value":"some@other-example.org","via":"email"}]}}`))
+				snapshotx.SnapshotT(t, &actual)
+			})
```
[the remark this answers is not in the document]
aeneasr: Good point, I'm pretty sure that this would latest break during the registration itself, but validating that the JSON is valid can probably be done prior with an improved error message

persistence/sql/persister_identity.go
aeneasr: You can do both - if you set verified to true, the timestamp is now. If you set both, you control both values.

aeneasr:
```suggestion
		// If verified is true but no timestamp is set, we default to time.Now
		if v.Verified && (v.VerifiedAt == nil || time.Time(*v.VerifiedAt).IsZero()) {
```

selfservice/hook/web_hook_integration_test.go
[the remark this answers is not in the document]
aeneasr: I think this is sufficiently handled in the manager itself, that is later being called. Generally speaking, webhook observability needs to be improved in any case.

PR #3075, opened 2023-02-06
feat: make notification to unknown recipients configurable

### LINKED ISSUE #2345: Disable mail dispatch on attempted account access

jonas-jonas:
```
BREAKING CHANGES: Added the ability to configure whether the system should notify unknown recipients, if some tries to recover their account or verify their address ("anti-account-enumeration measures"). 
By default, Kratos no longer sends out these Emails. 
If you want to keep notifying unknown addresses (keep the current behavior), set `selfservice.flows.recovery.notify_unknown_recipients` to `true` for recovery, or `selfservice.flows.verification.notify_unknown_recipients` for verification flows.

```
## Related issue(s)

Closes #2345 & supersedes #2585

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


driver/config/config.go
```diff
@@ -69,6 +69,7 @@ const (
 	ViperKeyCourierSMTPURL                                   = "courier.smtp.connection_uri"
 	ViperKeyCourierSMTPClientCertPath                        = "courier.smtp.client_cert_path"
 	ViperKeyCourierSMTPClientKeyPath                         = "courier.smtp.client_key_path"
+	ViperKeyCourierEnableInvalidDispatch                     = "courier.enable_invalid_email_dispatch"
```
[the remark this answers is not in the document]
jonas-jonas: Hm, the key already contains `courier`, so that would be duplicated.

Maybe `disable_invalid_message_dispatch`, WDYT?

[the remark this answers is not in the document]
kelkarajay: Yeah, sounds good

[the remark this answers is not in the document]
jonas-jonas: Done.

driver/config/config.go
```diff
@@ -69,6 +69,7 @@ const (
 	ViperKeyCourierSMTPURL                                   = "courier.smtp.connection_uri"
 	ViperKeyCourierSMTPClientCertPath                        = "courier.smtp.client_cert_path"
 	ViperKeyCourierSMTPClientKeyPath                         = "courier.smtp.client_key_path"
+	ViperKeyCourierEnableInvalidMessageDispatch              = "courier.enable_invalid_message_dispatch"
```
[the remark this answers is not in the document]
zepatrik: Or maybe `courier.flows.recovery.enable_invalid_message` as this is really a feature of recovery. I'm just not very happy with this high-level, pretty undescriptive key.

selfservice/strategy/code/code_sender.go
```diff
@@ -90,7 +96,7 @@ func (s *Sender) SendRecoveryCode(ctx context.Context, r *http.Request, f *recov
 		CreateRecoveryCode(ctx, &CreateRecoveryCodeParams{
 			RawCode:         rawCode,
 			CodeType:        RecoveryCodeTypeSelfService,
-			ExpiresIn:       s.deps.Config().SelfServiceCodeMethodLifespan(r.Context()),
+			ExpiresIn:       s.deps.Config().SelfServiceCodeMethodLifespan(ctx),
```
zepatrik, line 99: good catch

selfservice/strategy/code/code_sender.go
```diff
@@ -124,27 +130,27 @@ func (s *Sender) SendRecoveryCodeTo(ctx context.Context, i *identity.Identity, c
...
 		WithSensitiveField("address", to).
 		Debug("Preparing verification code.")
 
 	address, err := s.deps.IdentityPool().FindVerifiableAddressByValue(ctx, via, to)
-	if err != nil {
-		if errors.Is(err, sqlcon.ErrNoRows) {
-			s.deps.Audit().
-				WithField("via", via).
-				WithSensitiveField("email_address", address).
-				Info("Sending out invalid verification via code email because address is unknown.")
```
zepatrik, line 142: I guess we want to keep this audit log, probably regardless of whether the message is enabled.

selfservice/strategy/code/code_sender_test.go
```diff
@@ -160,4 +160,56 @@ func TestSender(t *testing.T) {
...
+
+				t.Cleanup(func() {
+					conf.Set(ctx, config.ViperKeyCourierEnableInvalidMessageDispatch, true)
+				})
+
+				tc.send(t)
+
+				messages, err := reg.CourierPersister().NextMessages(context.Background(), 0)
+
+				require.True(t, errors.Is(err, courier.ErrQueueEmpty))
```
zepatrik, line 209:
```suggestion
				require.ErrorIs(t, err, courier.ErrQueueEmpty)
```

selfservice/strategy/code/code_sender.go
```diff
@@ -124,27 +130,27 @@ func (s *Sender) SendRecoveryCodeTo(ctx context.Context, i *identity.Identity, c
...
-				return err
-			}
-			return errors.Cause(ErrUnknownAddress)
+	if errors.Is(err, sqlcon.ErrNoRows) {
+		if !s.deps.Config().CourierEnableInvalidDispatch(ctx) {
+			// do nothing
+		} else if err := s.send(ctx, string(via), email.NewVerificationCodeInvalid(s.deps, &email.VerificationCodeInvalidModel{To: to})); err != nil {
+			return err
 		}
+		return ErrUnknownAddress
```
zepatrik, line 151:
```suggestion
		return errors.WithStack(ErrUnknownAddress)
```

selfservice/strategy/code/code_sender_test.go
```diff
@@ -160,4 +160,56 @@ func TestSender(t *testing.T) {
...
+					require.ErrorIs(t, err, code.ErrUnknownAddress)
+				},
+			},
+			{
+				flow: "verification",
+				send: func(t *testing.T) {
+					s, err := reg.VerificationStrategies(ctx).Strategy("code")
+					require.NoError(t, err)
+					f, err := verification.NewFlow(conf, time.Hour, "", u, s, flow.TypeBrowser)
+					require.NoError(t, err)
```
zepatrik, line 188: This can be below in the loop and then the flow is passed as a param.

selfservice/strategy/code/code_sender_test.go
```diff
@@ -160,4 +160,56 @@ func TestSender(t *testing.T) {
...
+
+					require.NoError(t, reg.VerificationFlowPersister().CreateVerificationFlow(context.Background(), f))
```
zepatrik, line 190: You can even make `reg.VerificationFlowPersister().CreateVerificationFlow` and `reg.RecoveryFlowPersister().CreateRecoveryFlow` the functions of the test case structs. Then this gets sooo much easier to read.

selfservice/strategy/link/sender.go
```diff
@@ -59,21 +57,28 @@ func NewSender(r senderDependencies) *Sender {
...
 	address, err := s.r.IdentityPool().FindRecoveryAddressByValue(ctx, identity.RecoveryAddressTypeEmail, to)
-	if err != nil {
-		if err := s.send(ctx, string(via), email.NewRecoveryInvalid(s.r, &email.RecoveryInvalidModel{To: to})); err != nil {
+	if errors.Is(err, sqlcon.ErrNoRows) {
+		if !s.r.Config().CourierEnableInvalidDispatch(ctx) {
+			// do nothing
+		} else if err := s.send(ctx, string(via), email.NewRecoveryInvalid(s.r, &email.RecoveryInvalidModel{To: to})); err != nil {
 			return err
 		}
 		return errors.Cause(ErrUnknownAddress)
```
zepatrik, line 78:
```suggestion
		return errors.WithStack(ErrUnknownAddress)
```

selfservice/strategy/link/sender.go
```diff
@@ -94,31 +99,31 @@ func (s *Sender) SendRecoveryLink(ctx context.Context, r *http.Request, f *recov
...
-				return err
-			}
-			return errors.Cause(ErrUnknownAddress)
+	if errors.Is(err, sqlcon.ErrNoRows) {
+		if !s.r.Config().CourierEnableInvalidDispatch(ctx) {
+			// do nothing
+		} else if err := s.send(ctx, string(via), email.NewVerificationInvalid(s.r, &email.VerificationInvalidModel{To: to})); err != nil {
+			return err
 		}
+		return errors.Cause(ErrUnknownAddress)
```
zepatrik, line 120:
```suggestion
		return errors.WithStack(ErrUnknownAddress)
```

selfservice/strategy/link/sender_test.go
```diff
@@ -98,4 +98,56 @@ func TestManager(t *testing.T) {
 		require.NoError(t, err)
 		assert.EqualValues(t, identity.VerifiableAddressStatusSent, address.Status)
 	})
+
+	t.Run("case=should be able to disable invalid email dispatch", func(t *testing.T) {
```
zepatrik, line 102: Same test restructure here.

zepatrik (commented):
Found a few minor issues to improve this

driver/config/config.go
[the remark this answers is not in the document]
jonas-jonas: It's not just recovery, though. Verification as well. 

Ideally, this would be false by default, and an opt-in feature. So this key wouldn't be used except by people really wanting this feature. But that would be breaking.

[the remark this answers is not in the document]
aeneasr: I think the default should be enabled - it prevents sending a ton of mails out when script kiddies play around with rate limiting. We can either make this configurable per flow (recovery, **login**, verification) or globally. My feeling is that someone will want to configure this for magic link login

For the concrete key name (regardless of where) I suggest:

```
notify_unknown_recipients = false|true
send_to_unknown_recipients = false|true
ignore_unknown_recipients = false|true
```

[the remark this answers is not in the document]
jonas-jonas: "enabled" right now means, that the emails **will** be sent to unknown recipients. Is that what you meant? @aeneasr 

I like `courier.notify_unknown_recipients`. Maybe we can do:

```yaml
courier:
  notify_unknown_recipients:
    global: false|true
    verification: false|true
    recovery: false|true
    # login: false|true (in the future)
```

And we can set the default to `false` to prevent the rate-limit issues.

For OS this would be a breaking change, but for Ory Network, we can keep it enabled for existing projects, and set it to `false` for new projects (similar to the way we rolled out `code`), avoiding the breaki-ness. WDYT?

selfservice/strategy/code/code_sender.go
jonas-jonas: Ok, makes sense.

selfservice/strategy/code/code_sender_test.go
jonas-jonas: Good catch.

selfservice/strategy/code/code_sender_test.go
zepatrik: Ah nevermind, I thought it is the same struct.

driver/config/config.go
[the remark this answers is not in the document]
aeneasr: Thinking about it a bit more, this should go to `selfservice.flows.<login|recovery|verificatio>` instead. The courier doesn't know what recovery is, it just sends mail :)

>And we can set the default to false to prevent the rate-limit issues.
>For OS this would be a breaking change, but for Ory Network, we can keep it enabled for existing projects, and set it to false for new projects (similar to the way we rolled out code), avoiding the breaki-ness. WDYT?

SGTM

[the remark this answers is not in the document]
jonas-jonas: > Thinking about it a bit more, this should go to selfservice.flows.<login|recovery|verificatio> instead. The courier doesn't know what recovery is, it just sends mail :)

Makes sense to me.

text/message_recovery.go
```diff
@@ -35,7 +35,7 @@ func NewRecoveryEmailSent() *Message {
 	return &Message{
 		ID:      InfoSelfServiceRecoveryEmailSent,
 		Type:    Info,
-		Text:    "An email containing a recovery link has been sent to the email address you provided.",
+		Text:    "An email containing a recovery link has been sent to the email address you provided. If you have not received an email, check the spelling of the address.",
```
aeneasr, line 38: This is an acceptable workaround. I think that it would be cleaner to have a dedicated message when this feature is enabled though, but I'll leave that up to you.

aeneasr (commented):
Looking great! Should we add an e2e test covering this case? Or does it not make sense? I'm fine with both

text/message_recovery.go
jonas-jonas: Even if we have notifying unknown recipients enabled, and a user types their email address wrong "joas(at)ory.sh" (instead of "jonas(at)ory.sh") for example, they wouldn't notice either, right? 

So this message applies either way, IMO.

aeneasr: ha, good point!

aeneasr: in that case though, the message should be: "if you did not receive a message you either have a typo or the address is unknown"?

jonas-jonas: I guess it could be, but what would be the benefit for a user here? Either way, if they think they have an account, but this message shows up, and they're sure they have no typo, they need to contact the administrator of the application, who'd then check in the admin portal or API. 

> An email containing a recovery link has been sent to the email address you provided. If you have not received an email, check the spelling of the address and make sure to use the address you registered with.

Gets quite long, but would work. WDYT?
image

jonas-jonas:
Tests are failing, and I can't repro locally. Going to take another look tomorrow... 😓

aeneasr:
Maybe the feature is enabled on accident for that e2e test, it looks like the test is expecting an email but not receiving one

jonas-jonas:
> Maybe the feature is enabled on accident for that e2e test, it looks like the test is expecting an email but not receiving one

The config schema didn't match the config for the key. I fixed that.

Just to double-check, the config schema is manually written, right? And I assume, there is no practical way to generate it?

aeneasr:
> Just to double-check, the config schema is manually written, right? And I assume, there is no practical way to generate it?

Correct and correct

jonas-jonas:
> Should we add an e2e test covering this case? 

I thought about this a bit, and IMO it doesn't make sense to test for an email not arriving. As in theory, we'd need to wait an indefinite amount of time to be sure that no email arrives. This would slow down the E2e test pipeline. 

I have updated the existing tests that test for the email arriving to enable this feature.

We already test this config in the go-tests, by checking the DB, which is good enough, IMO.

test/e2e/cypress/integration/profiles/mfa/totp.spec.ts
```diff
@@ -292,7 +292,7 @@ context("2FA TOTP", () => {
 
       // The React app keeps using the same flow. The following scenario used to be broken,
       // because the internal context wasn't populated properly in the flow after settings were saved.
-      it.only("should allow changing other settings and then setting up totp", () => {
+      it("should allow changing other settings and then setting up totp", () => {
```
jonas-jonas, line 295: Looks like this was forgotten [a few weeks ago](https://github.com/ory/kratos/pull/2960/files#diff-154448811eddc216f531ef50d9a613a5c6c2f0bf544f9896f8d45617e8a68ed2R295). These tests are still passing, though. So crisis averted. 

Would be great to have an ESLint rule or some mechanism to prevent `.only`s ending up on master. But AFAICT, there is nothing of that sort for cypress :(

test/e2e/cypress/integration/profiles/verification/verify/errors.spec.ts
```diff
@@ -138,12 +140,15 @@ context("Account Verification Error", () => {
             })
           })
 
-          it("unable to verify non-existent account", async () => {
-            cy.get('input[name="email"]').type(gen.identity().email)
+          it("unable to verify non-existent account", () => {
```
jonas-jonas, line 143: `async` prevented this test from being executed properly. So it never actually tested what it was supposed to.

I checked, and this seemed to be the only instance of an `async` test body.

aeneasr (approved):
LGTM, as discussed on Slack, one test change

jonas-jonas:
>LGTM, as discussed on Slack, one test change

Added the test to the appropriate places.

aeneasr (approved):

Issue #2974, opened 2022-12-20
Errors on json-schema validation for traits always returns 4000001 as message id


eschmittel:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

The message code of traits  with failed  with for example validation minLength or pattern always results in 4000001 = ErrorValidationGeneric. 

According to https://www.ory.sh/docs/kratos/concepts/ui-user-interface#ui-message-codes 
and https://github.com/ory/kratos/blob/master/text/id.go I would expect a 4000003 = ErrorValidationMinLength
if I defined a minLength or 4000004 = ErrorValidationInvalidFormat for a given trait in json schema.

If this is expected behavior it would be still nice to find a way to differentiate the different validation messages by some sort of identifier.

### Reproducing the bug

1. create new ory project
2. change identity to the schema defined down below 
3. try to update the registration flow with invalid values for trait name.last or name.first 

examples:
name.first = "" // invalid minLength requirement
name.last = "invalidname?" // invalid pattern

View Full example: https://github.com/eschmittel/ory_validation_bug

### Relevant log output

```shell
// this is the resulting flow
{
    "id": "fba7c40b-76b1-4582-8ca0-71b8d6699566",
    "oauth2_login_challenge": null,
    "type": "browser",
    "expires_at": "2022-12-20T17:43:56.338554Z",
    "issued_at": "2022-12-20T16:43:56.338554Z",
    "request_url": "http://localhost:4000/self-service/registration/browser",
    "ui": {
        "action": "http://localhost:4000/self-service/registration?flow=fba7c40b-76b1-4582-8ca0-71b8d6699566",
        "method": "POST",
        "nodes": [
            ... // removed invalid fields
            {
                "type": "input",
                "group": "password",
                "attributes": {
                    "name": "traits.name.first",
[...]

PR #3085, opened 2023-02-08
fix: report correct errors for json schema validation

CaptainStandby:
- Implemented the translation of `jsonschema.ValidationError` to errors codes documented [here](https://www.ory.sh/docs/kratos/concepts/ui-user-interface#machine-readable-format)
- Added missing error codes for relevant schema errors
  | Validation         | Name                            | ID      |
  | ------------------ | ------------------------------- | ------- |
  | `maxLength`        | ErrorValidationMaxLength        | 4000017 |
  | `minimum`          | ErrorValidationMinimum.         | 4000018 |
  | `exclusiveMinimum` | ErrorValidationExclusiveMinimum | 4000019 |
  | `maximum`          | ErrorValidationMaximum          | 4000020 |
  | `exclusiveMaximum` | ErrorValidationExclusiveMaximum | 4000021 |
  | `multipleOf`       | ErrorValidationMultipleOf       | 4000022 |
  | `maxItems`         | ErrorValidationMaxItems         | 4000023 |
  | `minItems`         | ErrorValidationMinItems         | 4000024 |
  | `uniqueItems`      | ErrorValidationUniqueItems      | 4000025 |
  | `type`             | ErrorValidationWrongType        | 4000026 |
- Updated e2e tests to check these IDs explicitly

## Related issue(s)

#2974 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


aeneasr:
fyi :) in the open source repositories we have a different changelog strategy - here you can use regular commit messages :)

CaptainStandby:
> fyi :) in the open source repositories we have a different changelog strategy - here you can use regular commit messages :)

good to know :-D

cmd/clidoc/main.go
```diff
@@ -80,8 +80,18 @@ func init() {
...
+		"NewErrorValidationMinimum":                               text.NewErrorValidationMinimum("must be >= 5 but found 3"),
+		"NewErrorValidationExclusiveMinimum":                      text.NewErrorValidationExclusiveMinimum("must be > 5 but found 5"),
+		"NewErrorValidationMaximum":                               text.NewErrorValidationMaximum("must be <= 5 but found 6"),
+		"NewErrorValidationExclusiveMaximum":                      text.NewErrorValidationExclusiveMaximum("must be < 5 but found 5"),
+		"NewErrorValidationMultipleOf":                            text.NewErrorValidationMultipleOf("7 not multipleOf 3"),
+		"NewErrorValidationMaxItems":                              text.NewErrorValidationMaxItems("maximum 3 items allowed, but found 4 items"),
+		"NewErrorValidationMinItems":                              text.NewErrorValidationMinItems("minimum 3 items allowed, but found 2 items"),
+		"NewErrorValidationUniqueItems":                           text.NewErrorValidationUniqueItems("items at index 0 and 2 are equal"),
+		"NewErrorValidationWrongType":                             text.NewErrorValidationWrongType("expected number, but got string"),
```
jonas-jonas, line 94: I guess these need variables, right?

jonas-jonas (commented):
Nice work, just one question about the messages & variables.

cmd/clidoc/main.go
jonas-jonas: ah lol, I see. LGTM then.

jonas-jonas (dismissed):
LGTM.

aeneasr (dismissed):
Epic! Are there any backwards compatibility issues that could happen with this and if so how do we deal with them?

aeneasr:
I see these scenarios:

1. Tests rely on this behavior - this is ok if we break it
2. Custom ui implementation uses translation based on the IDs - it will mean that they have untranslated strings which is ok
3. Custom ui implementation uses the IDs to perform logic - this would break without warning

Did I miss something? What should we do to get (3) over the line smoothly in your opinion?

CaptainStandby:
I don't have a good solution for this.
Maybe we could just issue two errors for the same validation, one with the old value and one with the new. I'm not entirely sure if this is possible and it doesn't sound like the best idea.
WDYT?

aeneasr:
I'm fine with breaking it. I don't think anyone really uses it this way. We can probably check in the community and see if it's a problem before deploying it to production?

aeneasr (approved):

lukaszxion:
I'm using it. Now there is no context information and the messages look like this:
`Länge muss >= {{expected_length}} sein, hat aber {{actual_length}}.`

The use of `context` was proposed by Ory:
https://github.com/ory/kratos/pull/692#issuecomment-688176348

How can I now get the params for the translations?

PR #3088, opened 2023-02-09
fix: implement offline scope in the way google expects

tsearle:
google doesn't implements offline access as a scope but instead as a special uri parameter

this PR adapts the code such that if the offline_access scope is requested, it gets transformed into the parameter
access_type=offline

### DIFF
--- a/selfservice/strategy/oidc/provider_google.go
+++ b/selfservice/strategy/oidc/provider_google.go
@@ -5,9 +5,11 @@ package oidc
 
 import (
 	"context"
+
 	gooidc "github.com/coreos/go-oidc"
-	"github.com/ory/x/stringslice"
 	"golang.org/x/oauth2"
+
+	"github.com/ory/x/stringslice"
 )
 
 type ProviderGoogle struct {
--- a/selfservice/strategy/oidc/provider_google.go
+++ b/selfservice/strategy/oidc/provider_google.go
@@ -44,6 +44,16 @@ func (g *ProviderGoogle) oauth2ConfigFromEndpoint(ctx context.Context, endpoint 
 	}
 }
 
+func (g *ProviderGenericOIDC) OAuth2(ctx context.Context) (*oauth2.Config, error) {
+	p, err := g.provider(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	endpoint := p.Endpoint()
+	return g.oauth2ConfigFromEndpoint(ctx, endpoint), nil
+}
+
 func (g *ProviderGoogle) AuthCodeURLOptions(r ider) []oauth2.AuthCodeOption {
 	var options []oauth2.AuthCodeOption
 	scope := g.config.Scope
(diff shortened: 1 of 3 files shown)


tsearle:
linked issue
https://github.com/ory/network/issues/242

selfservice/strategy/oidc/provider_google.go
```diff
@@ -19,3 +28,46 @@ func NewProviderGoogle(
...
+	if isForced(r) {
+		options = append(options, oauth2.SetAuthURLParam("prompt", "login"))
+	}
+	if len(g.config.RequestedClaims) != 0 {
+		options = append(options, oauth2.SetAuthURLParam("claims", string(g.config.RequestedClaims)))
+	}
+
+	if !stringslice.Has(scope, gooidc.ScopeOfflineAccess) {
+		options = append(options, oauth2.AccessTypeOffline)
+	}
```
aeneasr, line 70: Not sure if I understand this - if the scope is not "offline_access", we automatically add it? Is that really what we want?

selfservice/strategy/oidc/provider_google.go
```diff
@@ -19,3 +28,46 @@ func NewProviderGoogle(
...
+func (g *ProviderGoogle) AuthCodeURLOptions(r ider) []oauth2.AuthCodeOption {
+	var options []oauth2.AuthCodeOption
+	scope := g.config.Scope
+
+	if isForced(r) {
+		options = append(options, oauth2.SetAuthURLParam("prompt", "login"))
+	}
+	if len(g.config.RequestedClaims) != 0 {
+		options = append(options, oauth2.SetAuthURLParam("claims", string(g.config.RequestedClaims)))
+	}
```
aeneasr, line 66: You can simplify this by calling `options := g.ProviderGenericOIDC.AuthCodeURLOptions()` and then add your custom logic afterwards :)

tsearle:
indeed it should be the inverse
I'll fix it

selfservice/strategy/oidc/provider_google.go
```diff
@@ -19,3 +28,46 @@ func NewProviderGoogle(
...
+	scope := g.config.Scope
+	if !stringslice.Has(scope, gooidc.ScopeOpenID) {
+		scope = append(scope, gooidc.ScopeOpenID)
+	}
+
+	return &oauth2.Config{
+		ClientID:     g.config.ClientID,
+		ClientSecret: g.config.ClientSecret,
+		Endpoint:     endpoint,
+		Scopes:       stringslice.Filter(scope, func(s string) bool { return s == gooidc.ScopeOfflineAccess }),
```
aeneasr, line 42: please add below line 34 for better readability :)

aeneasr (changes requested):

aeneasr:
Maybe you can add a small test that shows that the offline access thing is correctly configured when the scope is set? :)

aeneasr:
Thank you for making Ory better :)

PR #3092, opened 2023-02-10
feat: let the database generate session device IDs

alnr:
This is the same as #3042 but for session devices.

Also renamed some tracing span names in hooks.

### DIFF
--- a/selfservice/hook/web_hook.go
+++ b/selfservice/hook/web_hook.go
@@ -103,7 +103,7 @@ func NewWebHook(r webHookDependencies, c json.RawMessage) *WebHook {
 }
 
 func (e *WebHook) ExecuteLoginPreHook(_ http.ResponseWriter, req *http.Request, flow *login.Flow) error {
-	return otelx.WithSpan(req.Context(), "selfservice.hook.ExecuteLoginPreHook", func(ctx context.Context) error {
+	return otelx.WithSpan(req.Context(), "selfservice.hook.WebHook.ExecuteLoginPreHook", func(ctx context.Context) error {
 		return e.execute(ctx, &templateContext{
 			Flow:           flow,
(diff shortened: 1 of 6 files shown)


selfservice/hook/web_hook.go
```diff
@@ -174,26 +174,26 @@ func (e *WebHook) ExecutePostRecoveryHook(_ http.ResponseWriter, req *http.Reque
 			RequestCookies: cookies(req),
 			Identity:       session.Identity,
 		})
-	}, trace.WithAttributes(attribute.String("hook", KeyWebHook)))
+	})
 }
 
 func (e *WebHook) ExecuteRegistrationPreHook(_ http.ResponseWriter, req *http.Request, flow *registration.Flow) error {
-	return otelx.WithSpan(req.Context(), "selfservice.hook.ExecuteRegistrationPreHook", func(ctx context.Context) error {
+	return otelx.WithSpan(req.Context(), "WebHook.ExecuteRegistrationPreHook", func(ctx context.Context) error {
```
aeneasr, line 181: I think we should include the package name in the trace

```suggestion
	return otelx.WithSpan(req.Context(), "selfservice.hook.WebHook.ExecuteRegistrationPreHook", func(ctx context.Context) error {
```

Alternatively, I think we could use the proper go module naming for the traces:

```
github.com/ory/kratos/selfservice/hook:WebHook.ExecuteRegistrationPreHook
```

aeneasr (commented):
The UUID change is 👍 but for the tracing I have a comment / suggestion

selfservice/hook/web_hook.go
alnr: added

aeneasr (approved):

PR #3091, opened 2023-02-10
docs: rename pat to apikey

tomekpapiernik:
This PR renames `oryAccessToken` to `APIKey` to make the API docs in line with Ory Console and ory.sh/docs. 

## Related issue(s)

https://github.com/ory-corp/cloud/issues/3936

## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ ] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [ ] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [ ] I have read the [security policy](../security/policy).
- [ ] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got the approval (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](https://github.com/ory/docs).

PR #3109, opened 2023-02-16
docs: rename pat to api key

tomekpapiernik:
This PR renames oryAccessToken to APIKey to make the API docs in line with Ory Console and ory.sh/docs.

BREAKING CHANGES: Renamed the `oryAccessToken` to `oryApiKey`. If you're using the `oryAccessToken` to interact with the API, please rename the parameter to `oryApiKey` as well.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

## Further Comments

Replaces the previous PR. https://github.com/ory/kratos/pull/3091

### DIFF
--- a/internal/client-go/api_courier.go
+++ b/internal/client-go/api_courier.go
@@ -131,7 +131,7 @@ func (a *CourierApiService) GetCourierMessageExecute(r CourierApiApiGetCourierMe
 	if r.ctx != nil {
 		// API Key Authentication
 		if auth, ok := r.ctx.Value(ContextAPIKeys).(map[string]APIKey); ok {
-			if apiKey, ok := auth["oryAccessToken"]; ok {
+			if apiKey, ok := auth["oryApiKey"]; ok {
 				var key string
 				if apiKey.Prefix != "" {
 					key = apiKey.Prefix + " " + apiKey.Key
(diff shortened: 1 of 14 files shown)


internal/client-go/api_courier.go
```diff
@@ -131,7 +131,7 @@ func (a *CourierApiService) GetCourierMessageExecute(r CourierApiApiGetCourierMe
 	if r.ctx != nil {
 		// API Key Authentication
 		if auth, ok := r.ctx.Value(ContextAPIKeys).(map[string]APIKey); ok {
-			if apiKey, ok := auth["oryAccessToken"]; ok {
+			if apiKey, ok := auth["oryApiKey"]; ok {
```
zepatrik, line 134: Hm this does cause breaking changes :thinking: No idea tbh on how we can patch this to fall back to the old key if it is not provided...
I guess there is no way, breaking changes in the SDK are unavoidable with these sh... generator templates in general, but old SDK versions should still be compatible with the server.

zepatrik (approved):

internal/client-go/api_courier.go
aeneasr: The question is whether we have to break this, just to update the naming of this key? Maybe we can fix this in all services in one go etc

aeneasr: This also means that we have to update all the docs for the Go examples btw

aeneasr (changes requested):
I think more work is required to get this streamlined in also other servers, and then in the technical docs as well

PR #3111, opened 2023-02-16
fix: make async webhooks fully async

zepatrik:
## Related issue(s)

The current approach does not ignore errors from the request building process, only the actual request. This patch moves the request build into the async part of the webhook as well.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

_

### DIFF
--- a/selfservice/hook/web_hook.go
+++ b/selfservice/hook/web_hook.go
@@ -266,63 +266,75 @@ func (e *WebHook) ExecuteSettingsPrePersistHook(_ http.ResponseWriter, req *http
 }
 
 func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
-	builder, err := request.NewBuilder(e.conf, e.deps)
-	if err != nil {
-		return err
-	}
-
-	req, err := builder.BuildRequest(ctx, data)
-	if errors.Is(err, request.ErrCancel) {
-		return nil
-	} else if err != nil {
-		return err
-	}
-
-	attrs := semconv.HTTPClientAttributesFromHTTPRequest(req.Request)
-	if data.Identity != nil {
-		attrs = append(attrs,
-			attribute.String("webhook.identity.id", data.Identity.ID.String()),
-			attribute.String("webhook.identity.nid", data.Identity.NID.String()),
-		)
-	}
-
 	var (
 		httpClient     = e.deps.HTTPClient(ctx)
 		ignoreResponse = gjson.GetBytes(e.conf, "response.ignore").Bool()
 		canInterrupt   = gjson.GetBytes(e.conf, "can_interrupt").Bool()
 		parseResponse  = gjson.GetBytes(e.conf, "response.parse").Bool()
 		tracer         = trace.SpanFromContext(ctx).TracerProvider().Tracer("kratos-webhooks")
-		spanOpts       = []trace.SpanStartOption{trace.WithAttributes(attrs...)}
-		errChan        = make(chan error, 1)
 	)
-
 	if ignoreResponse && (parseResponse || canInterrupt) {
 		return errors.WithStack(herodot.ErrInternalServerError.WithReasonf("A webhook is configured to ignore the response but also to parse the response. This is not possible."))
 	}
 
-	ctx, span := tracer.Start(ctx, "selfservice.webhook", spanOpts...)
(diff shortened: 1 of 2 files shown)


selfservice/hook/web_hook.go
```diff
@@ -332,40 +343,30 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
 			if canInterrupt || parseResponse {
 				if err := parseWebhookResponse(resp, data.Identity); err != nil {
 					span.SetStatus(codes.Error, err.Error())
-					errChan <- err
```
zepatrik, line 335: After this there is no return, so another error would be send.

selfservice/hook/web_hook.go
```diff
@@ -332,40 +343,30 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
...
 			}
-			errChan <- fmt.Errorf("webhook failed with status code %v", resp.StatusCode)
-			return
+			return fmt.Errorf("webhook failed with status code %v", resp.StatusCode)
 		}
 
 		if parseResponse {
 			if err := parseWebhookResponse(resp, data.Identity); err != nil {
 				span.SetStatus(codes.Error, err.Error())
-				errChan <- err
```
zepatrik, line 345: Same here, after the error `nil` is send.

selfservice/hook/web_hook.go
```diff
@@ -332,40 +343,30 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
...
-			if err := <-errChan; err != nil {
-				logger.WithField("duration", time.Since(startTime)).WithError(err).Warning("Webhook request failed but the error was ignored because the configuration indicated that the upstream response should be ignored.")
-			} else {
-				logger.WithField("duration", time.Since(startTime)).Info("Webhook request succeeded")
-			}
-		}()
 		return nil
 	}
 
-	return <-errChan
```
zepatrik, line 368: This always waits for one message on the channel, I don't understand how it could even run async as this is blocking AND racy with the go routine also reading from that channel :sweat_smile:

zepatrik (commented):
~I honestly don't know how tests passed before...~

Kinda buggy before, but the bugs did not surface.

selfservice/hook/web_hook.go
zepatrik: Ahh it would return in the `if` above...

selfservice/hook/web_hook.go
```diff
@@ -332,40 +344,30 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
...
+	if !ignoreResponse {
+		return makeRequest()
+	}
+	go func() {
+		// we cannot handle the error as we are running async, and it is logged anyway
+		_ = makeRequest()
+	}()
+	return nil
```
jonas-jonas, line 370:
```suggestion
	if ignoreResponse {
		go func() {
		    // we cannot handle the error as we are running async, and it is logged anyway
		    _ = makeRequest()
	    }()
	    return nil
	}
	return makeRequest()
```

Just a nit. I like positive branching. But not sure that it improves this much. It's fine either way.

selfservice/hook/web_hook.go
```diff
@@ -266,63 +266,75 @@ func (e *WebHook) ExecuteSettingsPrePersistHook(_ http.ResponseWriter, req *http
...
+		if err != nil {
+			return err
+		}
+
+		req, err := builder.BuildRequest(ctx, data)
+		if errors.Is(err, request.ErrCancel) {
+			return nil
+		} else if err != nil {
+			return err
+		}
```
jonas-jonas, line 320: This prints the `kratos` cli usage help if, the jsonnet is invalid: 
image

I guess it did that before, but can we improve that somehow?

jonas-jonas (approved):
Very nice! Looks a lot cleaner as well, IMO.

selfservice/hook/web_hook.go
zepatrik: Hm I typically go for shorter branches in the if body, as that makes it easier to read.

selfservice/hook/web_hook.go
zepatrik: Yeah, we can do that but separately. I guess it does not hurt though.

Issue #3106, opened 2023-02-15
Support multiple browser clients for a single Kratos instance by using return_to URL of a flow if specified


reginbald:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Context and scope

When two separate browser applications use the same Kratos instance we need to be able to specify a return_to url that will redirect users to the browser application that initiated the login/registration/etc flow.

### Goals and non-goals

Goals
- URLs in return_to query parameter should be validated against the allow list before being stored with a flow
- Kratos should choose the return_to URL from a flow before falling back to the default URLs specified in the Kratos configuration
- When a registration flow replaces a login flow the return_to url should be migrated to the new flow
  - This happens when a user that does not exist in Kratos as an identity uses a social login

### The design

The main goal is to support multiple browser clients hosted on different subdomains to interact with a single Kratos instance. 
This would be done by redirecting browsers to the return_to URL if specified before falling back to the default UI URLs.

### Alternatives considered

- Single Login browser application that would then route users to their final destination
- Running multiple instances of Kratos, one per browser application

aeneasr:
We already have this? https://www.ory.sh/docs/kratos/concepts/browser-redirect-flow-completion

reginbald:
You're correct, however what I've found is that this not true in all scenarios (so this should maybe be a bug fix?).  As an example when a user that has an email + password entity tries to login with social login (that has not been linked) there will be a NewDuplicateCredentialsError, the user is redirected to the default uri and the  return_to query parameter is ignored.

aeneasr:
Correct, the return_to parameter is only used on success, it can not alter how the system behaves on errors. This is intentional because the user needs feedback on why the registration/login/... failed

reginbald:
[...]

PR #3112, opened 2023-02-16
fix: return_to used if specified

reginbald:
The goal of this pull request is to make minimal changes that allow Kratos to support multiple browser applications on separate domains to handle their own login/logout/etc UI. This is done by using the return_to URL if specified to redirect the user and returning an additional flow type (login/registration/...) during error condition in order for the application to inform the user of the error situation. 

This means that there can be two browser applications on separate subdomains, app1.example.com and app2.example.com (white listed in the Kratos allowed_return_urls config). Each application would add its URL with the return_to parameter during flow creation to inform Kratos of which application initiated the flow. After the execution of the flow Kratos will redirect the user to the application that initiated the flow if the ReturnTo property was specified otherwise use the default urls from the Kratos configuration.

If a flow execution ends with an error the user will be redirected back to the application with both the flow id and flow type, allowing the application to query Kratos for the error message to inform the user. 

By returning the flow type Kratos no longer forces the application to have separate routes for each flow.

## Related issue(s)

Based on the discussion from [Issue: 3106](https://github.com/ory/kratos/issues/3106)

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/flow/error_test.go
+++ b/selfservice/flow/error_test.go
@@ -66,8 +66,12 @@ func (t *testFlow) GetRequestURL() string {
 	return t.RequestURL
 }
 
+func (t *testFlow) GetReturnTo() *url.URL {
+	return nil
+}
+
 func (t *testFlow) AppendTo(url *url.URL) *url.URL {
-	return AppendFlowTo(url, t.ID)
+	return AppendFlowTo(url, t.ID, TypeLogin)
 }
 
 func (t *testFlow) GetUI() *container.Container {
--- a/selfservice/flow/flow.go
+++ b/selfservice/flow/flow.go
@@ -19,8 +19,8 @@ import (
 	"github.com/ory/x/urlx"
 )
 
-func AppendFlowTo(src *url.URL, id uuid.UUID) *url.URL {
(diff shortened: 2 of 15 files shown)


reginbald:
This will of course break the expectation that on error the user will be redirected to the url specified in the kratos configuration on error

aeneasr (changes requested):
Thank you for your work on this PR! Unfortunately, I don't think that we can merge it as is. It breaks many current assumptions around how the UI works, and it introduces significant complexity (that is not yet addressed in the PR properly, and a lot of work) because it basically deals with multi-tenant login, which is not supported by Ory Kratos itself.

My recommendation would be to decide this in your UI app instead. There, you can implement redirects in any way you want. Ory Kratos delivers you with all the information required (e.g. the initial requested URL) which may contain URL parameters you can use for further propagation.

Thank you for understanding!

reginbald:
I was expecting that :), Thanks for your feedback. Do you think some solution is on the roadmap or is this something you don't want to support? I could also set a side some time during the weekend to create a proper proposal for you.

reginbald:
Good to know, thank you :)

PR #3141, opened 2023-03-02
fix: respect the after recovery return to URL from config

jonas-jonas:
## Related issue(s)
Fixes https://github.com/ory-corp/cloud/issues/1405

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/x/http_secure_redirect.go
+++ b/x/http_secure_redirect.go
@@ -76,12 +76,14 @@ func SecureRedirectToIsAllowedHost(returnTo *url.URL, allowed url.URL) bool {
 	return strings.EqualFold(allowed.Host, returnTo.Host)
 }
 
-func TakeOverReturnToParameter(from string, to string) (string, error) {
+// TakeOverReturnToParameter carries over the return_to parameter to a new URL
+// If `from` does not contain the `return_to` query parameter, the first non-empty value from `fallback` is used instead.
+func TakeOverReturnToParameter(from string, to string, fallback ...string) (string, error) {
 	fromURL, err := url.Parse(from)
 	if err != nil {
 		return "", err
 	}
-	returnTo := fromURL.Query().Get("return_to")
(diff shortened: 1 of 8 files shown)


x/http_secure_redirect.go
```diff
@@ -76,12 +76,14 @@ func SecureRedirectToIsAllowedHost(returnTo *url.URL, allowed url.URL) bool {
...
-func TakeOverReturnToParameter(from string, to string) (string, error) {
+// TakeOverReturnToParameter carries over the return_to parameter to a new URL
+// If `from` does not contain the `return_to` query parameter, the first non-empty value from `fallback` is used instead.
+func TakeOverReturnToParameter(from string, to string, fallback ...string) (string, error) {
 	fromURL, err := url.Parse(from)
 	if err != nil {
 		return "", err
 	}
-	returnTo := fromURL.Query().Get("return_to")
+	returnTo := stringsx.Coalesce(append([]string{fromURL.Query().Get("return_to")}, fallback...)...)
```
jonas-jonas, line 86: Couldn't find a better way to combine the string value with a []string in a var-args parameter...

zepatrik: I don't think there is a better one.

zepatrik (dismissed):
Very nice :tada:

jonas-jonas:
@zepatrik I realized, that we probably want the settings flow of the recovery flow behave the same way as the normal settings flow, if no `returnTo` was set in the config or the creation URL. With my code before, we would always return the user to the `defaultReturnTo` after submitting the settings, but the settings flow returns to the settings UI in that case. So I added that behavior here too. PTAL.

CaptainStandby (approved):
LGTM

Benehiko (approved):
LGTM :)

PR #3144, opened 2023-03-03
feat: return verification flow ID after registration flow

### LINKED ISSUE #2975: Unable to get verification flow id after registration when using code method.
### Preflight checklist

- [X] This issue affects my [Ory Network](https://www.ory.sh/) project.

### Describe the bug

Greetings,

We are using ory Kratos (v.0.11.0) with the FrontendApi in our React Native app. We were previously using link as verification method and now we have migrated to code. And after calling the `updateRegistrationFlow` function, the verification code is sent automatically, without us creating a Native Verification Flow manually by calling `createNativeVerificationFlow`. Hence, since we don't know the id of the verification flow which was just created, we can't really call the `updateVerificationFlow` function when the code is entered. We tried to use the id from the Registration flow, thinking that since it creates the Verification Flow the id's might be the same. But obviously that was not the case. 

Here the question is, how to access the id of the Verification flow so that we can call the update function and do the verification? 

In the relevant fields of kratos.yml we have :
```
methods:
    password:
      enabled: true
    link:
      enabled: true
    code:
      enabled: true
    ...
flows:
    ...
    verification:
      enabled: true
      use: code
      ...
      
```
We call the registration flow as follows: 
```
const flow = (await kratosApi().createNativeRegistrationFlow()).data;
      const data = await kratosApi().updateRegistrationFlow({
        flow: flow.id,
        updateRegistrationFlowBody: {
          csrf_token: "",
          method: "password",
          password: state.password,
          traits: {
            email,
            ...
          },
        },
      });
```

With kratosApi being 
``` return new FrontendApi(
    new Configuration({
      basePath,
      apiKey: token,
      baseOptions: {
        withCredentials: false,
      },
    }),
    "",
    axios
  );

```
  
When we call the flow like this, we do get an email with the code inside to the email address that's specified in the `traits.email` field. Alongside, we also get the id of the verification flow which was created by the registration flow. But we need to get the id in the code somehow in order to use it in verification for the `flow` field which can be seen in the next code block.
  
  And finally, we are trying to get the verification as follows: 
  ```
  updateVerificationFlow({
        flow: <id of the Registration Flow (to be changed with the actual id of verification flow)>,
        updateVerificationFlowBody: { code, email }
      })
  ```
  
 Thank you and best regards, 

### Reproducing the bug

Adding kratos v0.11.0 to the application and specifying the kratos version in docker-compose.yml & kube-kratos-values.yml

Enabling and adding code as verification method in kratos.yml

Creating a Native Registration flow with FrontendApi and updating the flow with the account parameters needed. 

The verification flow will have been automatically created, mail will be sent to the email address stated in traits with the id, but the id of verification flow is unreachable to the code to call `updateVerificationFlow` with it.

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

0.11.0

Linux

Docker Compose

### Additional Context

_No response_

jonas-jonas:
## Related issue(s)

Closes #2975

## Checklist

- [x] ~Discuss, how an E2E test for this could look?~ See https://github.com/ory/kratos/pull/3144#issuecomment-1466276768 
- [x] ~Better way to pass data from the hook executors back to the handler?~ We attached it to the flow object which is modified by other hooks, anyway.


selfservice/hook/verification.go
```diff
@@ -37,19 +39,19 @@ func NewVerifier(r verifierDependencies) *Verifier {
 	return &Verifier{r: r}
 }
 
-func (e *Verifier) ExecutePostRegistrationPostPersistHook(_ http.ResponseWriter, r *http.Request, f *registration.Flow, s *session.Session) error {
+func (e *Verifier) ExecutePostRegistrationPostPersistHook(w http.ResponseWriter, r *http.Request, f *registration.Flow, s *session.Session, continueWith *flow.ContinueWithContainer) error {
```
jonas-jonas, line 42: I don't like this function parameter. It is used to carry a slice of `ContinueWith` items, which the hook executors can add their actions (`verification_ui` or `set_token`) to, which is then used to return/write into the response.

If you have another idea, please LMK.

selfservice/hook/session_issuer.go
```diff
@@ -39,23 +39,27 @@ func NewSessionIssuer(r sessionIssuerDependencies) *SessionIssuer {
...
+			c.Items = append(c.Items, NewContinueWith(s.Token))
+		}
 		e.r.Writer().Write(w, r, &registration.APIFlowResponse{
-			Session:  s,
-			Token:    s.Token,
-			Identity: s.Identity,
+			Session:      s,
+			Token:        s.Token,
+			Identity:     s.Identity,
+			ContinueWith: c.Items,
```
jonas-jonas, line 62: This is essentially duplicated from https://github.com/ory/kratos/pull/3144/files#diff-2b4e8e6ce993f442aa73df0ab4d28c0dd56b66f3d5819908f96b965d73f8a3d0, but I didn't want to refactor that here, as that would also potentially change the order of these hooks.

selfservice/flow/continue_with.go
```diff
@@ -0,0 +1,77 @@
...
+
+import "github.com/gofrs/uuid"
+
+// ContinueWithContainer is helper struct to pass to Hook executor functions
+type ContinueWithContainer struct {
+	Items []ContinueWith
+}
+
+// swagger:model continue_with
+type ContinueWith any
```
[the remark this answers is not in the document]
zepatrik: Ah `json.Marshal`, nevermind.

zepatrik (commented):
Nice, looking pretty good :+1: 
What other actions do we expect to be added?

selfservice/flow/continue_with.go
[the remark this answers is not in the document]
jonas-jonas: Also, what I had, but then the json marshal methods got really weird, and working with interface and json marshalling seemed a pain as well.

selfservice/flow/registration/session.go
```diff
@@ -38,4 +39,12 @@ type APIFlowResponse struct {
 	//
 	// required: true
 	Identity *identity.Identity `json:"identity"`
+
+	// Contains a list of actions, that could follow this flow
```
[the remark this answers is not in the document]
jonas-jonas: They're set in the hooks, you _can_ but not _must_ have configured. So optional.

And regarding the priority, if the identity schema contains multiple verifiable emails, there will be one `continue_with` item for each address in here, of which all technically have the same priority. I am not sure how we can handle that case best. I would argue, that that is up to the consumer/client to decide, though we don't really give a lot of context here. Maybe we can at least include the `verifiable_address` as well (not sure of the security ramifications, tho).

>So we would always expect the client to respect the actions.

Not necessarily, a client could just choose to ignore the verification flow and let the user click on the link in the email, for example.

selfservice/hook/verification.go
jonas-jonas: The registration flow is not returned after the registration is completed. We return an `registration.APIFlowResponse` which contains session, identity, a token, and now this.

I guess I could add it to the flow and then just carry it over, but we don't want to persist this slice. So this would then be some field that has some meaning sometimes, which just feels wrong to me.

jonas-jonas:
>What other actions do we expect to be added? 

For the recovery flow, we could use this and add `settings_ui`.

For the various `expired` type of errors, we are planning on also using this, so we would add these for all the flows `recovery_ui`, etc.

Then we will also have a special one for OIDC.

selfservice/flow/settings/hook.go
```diff
@@ -40,7 +40,7 @@ type (
 	PostHookPrePersistExecutorFunc func(w http.ResponseWriter, r *http.Request, a *Flow, s *identity.Identity) error
 
 	PostHookPostPersistExecutor interface {
-		ExecuteSettingsPostPersistHook(w http.ResponseWriter, r *http.Request, a *Flow, s *identity.Identity) error
+		ExecuteSettingsPostPersistHook(w http.ResponseWriter, r *http.Request, a *Flow, s *identity.Identity, continueWith *flow.ContinueWithContainer) error
```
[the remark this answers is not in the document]
jonas-jonas: I agree, but we are fetching the flow after all the hooks have executed again, [here](https://github.com/ory/kratos/blob/1da0e40b6d83f4b39fc61308d93ec92cd33a7787/selfservice/flow/settings/hook.go#L284-L287) and return that. Since the `continue_with` items can contain tokens in plaintext, I think we don't want to store this in the DB. There is probably a good reason, that we re-fetch the flow at the end, so changing that could be breaking something non-obvious... WDYT?

selfservice/hook/verification.go
jonas-jonas: I refactored it, to use the flow as the "carrier" for these items. Not super happy with it, but a lot better. Great suggestion!

selfservice/flow/continue_with.go
```diff
@@ -0,0 +1,77 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package flow
+
+import "github.com/gofrs/uuid"
+
+// ContinueWithContainer is helper struct to pass to Hook executor functions
+type ContinueWithContainer struct {
+	Items []ContinueWith
+}
```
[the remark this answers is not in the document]
jonas-jonas: Removed this, in favor of using the flow itself.

selfservice/flow/settings/hook.go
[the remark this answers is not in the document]
jonas-jonas: Replaced with the hook itself, and in the settings case, I am now carrying it over from the old flow.

selfservice/flow/registration/flow.go
```diff
@@ -97,10 +97,19 @@ type Flow struct {
...
-	TransientPayload json.RawMessage `json:"transient_payload,omitempty"  faker:"-" db:"-"`
+	TransientPayload json.RawMessage `json:"transient_payload,omitempty" faker:"-" db:"-"`
+
+	// Contains a list of actions, that could follow this flow
+	//
+	// It can, for example, contain a reference to the verification flow, created as part of the user's
+	// registration.
+	//
+	// required: false
+	// swagger:ignore
```
jonas-jonas, line 111: Since we aren't returning the registration flow itself, once the registration finishes, this is just used internally, and should not be sent to the client.

selfservice/flow/registration/hook.go
```diff
@@ -217,7 +217,10 @@ func (e *HookExecutor) PostRegistrationHook(w http.ResponseWriter, r *http.Reque
 		Debug("Post registration execution hooks completed successfully.")
 
 	if a.Type == flow.TypeAPI || x.IsJSONRequest(r) {
-		e.d.Writer().Write(w, r, &APIFlowResponse{Identity: i})
+		e.d.Writer().Write(w, r, &APIFlowResponse{
+			Identity:     i,
+			ContinueWith: a.ContinueWith(),
+		})
```
jonas-jonas, line 223: One more thing, I just thought about for this situation, is that all of this only applies to the API based interactions with these endpoints (SPA/Native).

We have not defined how this could work for browser based interactions. A few ideas:
1. Break backward compatibility and redirect to the verification flow here, if a flow was created as part of the registration flow.
2. Add a Query parameter to decide this behavior, that only applies to browser flows
3. Add a config option to set this.

Thoughts? cc @aeneasr

selfservice/strategy/code/.snapshots/TestVerification-description=should_set_all_the_correct_verification_payloads.json
```diff
@@ -49,5 +36,18 @@
       }
     },
     "type": "input"
+  },
+  {
+    "attributes": {
+      "disabled": false,
+      "name": "csrf_token",
```
jonas-jonas, line 43: not sure if re-ordering CSRF token "inputs" will be an issue?

selfservice/flow/registration/hook.go
jonas-jonas: Another thought, if the identity schema defines multiple verifiable addresses, we can't know where to redirect here. In that case, IMO, a `continue_with_address` query or even body param hint would be great.

jonas-jonas:
As for E2E tests, right now they (should be) are passing, because the changes are not breaking. We are merely adding new fields to payloads.

I would propose, adding this feature to the reference implementations, once this is merged and adjusting the E2E tests, once that's done. That way we don't need to update the SDK from a branch, etc.

selfservice/flow/registration/session.go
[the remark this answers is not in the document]
zepatrik: We decided that we will have the best-practice to always do all actions, but one can always choose not to do them for whatever reason.

selfservice/hook/session_issuer_test.go
```diff
@@ -63,8 +67,14 @@ func TestSessionIssuer(t *testing.T) {
...
 
 			require.NoError(t, reg.PrivilegedIdentityPool().CreateIdentity(context.Background(), i))
+
 			err := h.ExecutePostRegistrationPostPersistHook(w, &http.Request{Header: http.Header{"Accept": {"application/json"}}}, f, s)
-			require.True(t, errors.Is(err, registration.ErrHookAbortFlow), "%+v", err)
+			require.ErrorIs(t, err, registration.ErrHookAbortFlow, "%+v", err)
+			require.Len(t, f.ContinueWithItems, 1)
+
+			st := f.ContinueWithItems[0]
+			assert.IsType(t, &flow.ContinueWithSetToken{}, st)
```
zepatrik, line 76:
```suggestion
			require.IsType(t, &flow.ContinueWithSetToken{}, st)
```
because then the next assertion cannot panic, but this is more nit-picking.

selfservice/strategy/code/.snapshots/TestVerification-description=should_set_all_the_correct_verification_payloads.json
zepatrik: It should not matter.

zepatrik (dismissed):
Nice :+1:

selfservice/hook/session_issuer_test.go
jonas-jonas: Good point, done!

selfservice/flow/registration/hook.go
zepatrik: We decided to go with a query param added to the `return_to` redirect. This way we can send the verification flow ID to the UI, and it can then decide whether it wants to use it to do the verification, or continue and do the verification later. An existing implementation does not change.

Benehiko (dismissed):
Nice! This looks really good :)

selfservice/hook/verification.go
```diff
@@ -24,9 +24,11 @@ var _ settings.PostHookPostPersistExecutor = new(Verifier)
 type (
 	verifierDependencies interface {
 		config.Provider
-		x.CSRFTokenGeneratorProvider
+		x.CSRFProvider
```
CaptainStandby, line 27: Why was this changed?

selfservice/hook/verification.go
```diff
@@ -63,15 +65,23 @@ func (e *Verifier) do(r *http.Request, i *identity.Identity, f flow.Flow) error
...
 			continue
 		}
+		var csrf string
+		if f.GetType() == flow.TypeBrowser {
+			csrf = e.r.CSRFHandler().RegenerateToken(w, r)
+		}
 		verificationFlow, err := verification.NewPostHookFlow(e.r.Config(),
 			e.r.Config().SelfServiceFlowVerificationRequestLifespan(r.Context()),
-			e.r.GenerateCSRFToken(r), r, strategy, f)
+			csrf, r, strategy, f)
```
CaptainStandby, line 74: Shouldn't there be a fresh csrf token?

CaptainStandby: never mind 🙈

selfservice/flow/continue_with.go
```diff
@@ -0,0 +1,85 @@
+// Copyright © 2023 Ory Corp
+// SPDX-License-Identifier: Apache-2.0
+
+package flow
+
+import "github.com/gofrs/uuid"
+
+// swagger:model continue_with
```
aeneasr, line 8: Models are `upperCase`

selfservice/flow/continue_with.go
```diff
@@ -0,0 +1,85 @@
...
+type ContinueWith any
+
+// swagger:enum ContinueWithAction
+type ContinueWithAction string
+
+const (
+	ContinueWithActionSetToken       ContinueWithAction = "set_token"
```
aeneasr, line 15: Set what token?

```suggestion
	ContinueWithActionSetToken       ContinueWithAction = "set_ory_session_token"
```

```suggestion
	ContinueWithActionSetToken       ContinueWithAction = "set_session_token"
```

selfservice/flow/continue_with.go
```diff
@@ -0,0 +1,85 @@
...
+// swagger:model continue_with_verification_ui
+type ContinueWithVerificationUI struct {
+	// Action will always be `verification_ui`
+	//
+	// required: true
+	Action ContinueWithAction `json:"action"`
+	// Flow contains the ID of the verification flow
+	//
+	// required: true
+	Flow struct {
```
aeneasr, line 55: Please use a dedicated struct for this and give it a `swagger:model` annotation

selfservice/flow/continue_with.go
```diff
@@ -0,0 +1,85 @@
...
+		Action: ContinueWithActionSetToken,
+		Token:  t,
+	}
+}
+
+var _ ContinueWith = new(ContinueWithVerificationUI)
+
+// Indicates, that the UI flow could be continued by showing a verification ui
+//
+// swagger:model continue_with_verification_ui
```
aeneasr, line 46: upperCase - https://intranet.ory.sh/docs/dev/api-guidelines

selfservice/flow/registration/flow.go
aeneasr: Remove the required field, and swagger ignore. `json:"-"` already omits it from swagger

aeneasr (commented):
First sweep

selfservice/flow/continue_with.go
jonas-jonas: Good point, thanks.

selfservice/flow/continue_with.go
jonas-jonas: done.

selfservice/flow/continue_with.go
jonas-jonas: The names get quite long. Would've liked to avoid those.. But fine with me, done.

selfservice/flow/registration/flow.go
jonas-jonas: Done.

selfservice/hook/verification.go
jonas-jonas: Because `CSRFTokenGeneratorProvider` only contains the `GenerateCSRFToken` method, but we want to create a new CSRF token here.

selfservice/flow/continue_with.go
jonas-jonas: Done.

Benehiko (approved):
This looks very nice!

Benehiko (dismissed):
LGTM :)

aeneasr:
There is one thing missing from our discussion, I don't quite recall wahat it was @jonas-jonas

jonas-jonas:
> There is one thing missing from our discussion, I don't quite recall wahat it was @jonas-jonas

Turning the browser redirect into a hook. Working on it as we speak.

.schemastore/config.schema.json
jonas-jonas, line 1: btw. what's the difference between `embedx/config.schema.json` and `.schemastore/config.schema.json`? Do I need to update both manually? Can we consolidate them?

aeneasr: The embedx should IMO move to spec. Schemastore you can ignore.

selfservice/flow/registration/hook_test.go
```diff
@@ -159,6 +160,44 @@ func TestRegistrationExecutor(t *testing.T) {
...
+							"hook": hook.KeyVerificationUI,
+						},
+					})
+					i := testhelpers.SelfServiceHookFakeIdentity(t)
+					i.Traits = identity.Traits(`{"email": "verifiable@ory.sh"}`)
+
+					res, _ := makeRequestPost(t, newServer(t, i, flow.TypeBrowser), false, url.Values{})
+					assert.EqualValues(t, http.StatusOK, res.StatusCode)
+					assert.Contains(t, res.Request.URL.String(), verificationTS.URL)
+					assert.NotEmpty(t, res.Request.URL.Query().Get("flow"))
```
jonas-jonas, line 179: I changed the flow to no longer use some obscure query parameter and instead make this configurable with a hook (similar to the `session` hook), as we discussed @aeneasr.

However, I am unsure if we should do the same thing for the settings flow. WDYT?

aeneasr:
Is this good to review?

jonas-jonas:
> Is this good to review?


If CI is green, yes from my POV.

contrib/quickstart/kratos/email-password/kratos.yml
```diff
@@ -65,6 +65,7 @@ selfservice:
         password:
           hooks:
             - hook: session
+            - hook: show_verification_ui
```
aeneasr, line 68: Maybe continue_with_verification_ui for naming consistency?

selfservice/hook/show_verification_ui.go
```diff
@@ -0,0 +1,61 @@
...
+	var vf *flow.ContinueWithVerificationUI
+	for _, c := range f.ContinueWithItems {
+		if item, ok := c.(*flow.ContinueWithVerificationUI); ok {
+			vf = item
+		}
+	}
+
+	if vf != nil {
+		redir := e.d.Config().SelfServiceFlowVerificationUI(r.Context())
+		x.ContentNegotiationRedirection(w, r, s.Declassified(), e.d.Writer(), vf.AppendTo(redir).String())
```
aeneasr, line 57: Does this work for native apps?

aeneasr (changes requested):
This is looking grand! The only thing I'm missing is some proof that continue_with works as intended. Basically, this means an end to end test with registration + verification on mobile and/or SPA?

.schema/openapi/patches/schema.yaml
```diff
@@ -33,3 +33,17 @@
 
 - op: remove
   path: "#/components/schemas/identityTraits/type"
+
+- op: add
+  path: /components/schemas/continueWith/discriminator
+  value:
+    propertyName: action
+    mapping:
+      verification_ui: "#/components/schemas/continueWithVerificationUI"
```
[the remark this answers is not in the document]
jonas-jonas: Done.

selfservice/flow/continue_with.go
```diff
@@ -0,0 +1,108 @@
...
+// swagger:model continueWith
+type ContinueWith any
+
+// swagger:enum ContinueWithAction
+type ContinueWithAction string
+
+// #nosec G101 -- only a key constant
+const (
+	ContinueWithActionSetOrySessionToken ContinueWithAction = "set_ory_session_token"
+	ContinueWithActionVerificationUI     ContinueWithAction = "verification_ui"
```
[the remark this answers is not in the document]
aeneasr: Ok, then you can scratch the comment https://github.com/ory/kratos/pull/3144#discussion_r1142056526 because then we have consistency

contrib/quickstart/kratos/email-password/kratos.yml
jonas-jonas: Also possible, though the key in a client application's code would be `show_verification_ui` with https://github.com/ory/kratos/pull/3144#discussion_r1142064211

jonas-jonas: :D nice timing. Alright. Gonna disregard this.

selfservice/hook/show_verification_ui.go
jonas-jonas: Good point, I've added a check to only execute this for browsers. Should be fine like that.

jonas-jonas:
>The only thing I'm missing is some proof that continue_with works as intended. Basically, this means an end to end test with registration + verification on mobile and/or SPA?

Yes, that's true. I wanted to merge this first and write some E2E tests then, to avoid the SDK release-dance needed to update both. But I can do that before we merge this PR.

selfservice/hook/show_verification_ui.go
jonas-jonas: It doesn't, that was my point on Friday. This hook can only be used with server rendered browser flows. 

For native apps you get the `continue_with` array in the JSON response.

jonas-jonas:
>do you know how this works?

No :)

aeneasr:
https://github.com/ory/kratos-selfservice-ui-node/blob/7a838e2a2900592817df07067f998c4d6325a7ff/Makefile#L27

Make sure to have the paths set correctly, and you'll also need access to the npm package - if you give me your ID i can add you

.github/workflows/ci.yaml
```diff
@@ -184,17 +184,25 @@ jobs:
         with:
           repository: ory/kratos-selfservice-ui-node
           path: node-ui
+      - name: Install selfservice-ui-react-nextjs
+        uses: actions/checkout@v3
+        with:
+          repository: ory/kratos-selfservice-ui-react-nextjs
+          path: react-ui
+          ref: jonas-jonas/feat/showVerificationAfterRegistration
```
jonas-jonas, line 192: TODO

selfservice/flow/continue_with.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// The address that should be verified in this flow
+	//
+	// required: true
+	VerifiableAddress string `json:"verifiable_address"`
+
+	// The URL of the verification flow
+	//
+	// required: false
+	URL string `json:"url,omitempty"`
```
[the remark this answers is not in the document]
aeneasr: I think that's fine :)

selfservice/hook/show_verification_ui.go
aeneasr: Got it!

aeneasr (approved):

selfservice/hook/verification.go
```diff
@@ -80,6 +90,16 @@ func (e *Verifier) do(r *http.Request, i *identity.Identity, f flow.Flow) error
...
 		}
 
+		flowURL := ""
+		if verificationFlow.Type == flow.TypeBrowser {
+			flowURL = verificationFlow.AppendTo(e.r.Config().SelfServiceFlowVerificationUI(r.Context())).String()
+		}
+
+		f.AddContinueWith(flow.NewContinueWithVerificationUI(verificationFlow, address.Value, flowURL))
+
+		if err := e.r.PrivilegedIdentityPool().UpdateVerifiableAddress(r.Context(), address); err != nil {
```
aeneasr, line 100: Why do we need this? There have been no changes to the address as far as I can tell?

selfservice/hook/verification.go
```diff
@@ -63,15 +65,23 @@ func (e *Verifier) do(r *http.Request, i *identity.Identity, f flow.Flow) error
...
 			e.r.Config().SelfServiceFlowVerificationRequestLifespan(r.Context()),
-			e.r.GenerateCSRFToken(r), r, strategy, f)
+			csrf, r, strategy, f)
 		if err != nil {
 			return err
 		}
 
 		verificationFlow.State = verification.StateEmailSent
 
+		if err := strategy.PopulateVerificationMethod(r, verificationFlow); err != nil {
```
aeneasr, line 81: NewPostHookFlow doing this already, why do we need to do it again?

selfservice/hook/verification_test.go
```diff
@@ -87,10 +86,13 @@ func TestVerifier(t *testing.T) {
...
-			require.NoError(t, err)
-			expectedVerificationFlow, err := verification.NewPostHookFlow(conf, conf.SelfServiceFlowVerificationRequestLifespan(ctx), "", u, s, originalFlow)
-			require.NoError(t, err)
+			assert.Lenf(t, originalFlow.ContinueWith(), 2, "%#ßv", originalFlow.ContinueWith())
+			assertContinueWithAddresses(t, originalFlow.ContinueWith(), []string{"foo@ory.sh", "bar@ory.sh"})
+			vf := originalFlow.ContinueWith()[0]
+			assert.IsType(t, &flow.ContinueWithVerificationUI{}, vf)
+			fView := vf.(*flow.ContinueWithVerificationUI).Flow
+
+			expectedVerificationFlow, err := reg.VerificationFlowPersister().GetVerificationFlow(ctx, fView.ID)
```
aeneasr, line 95: This test assertion changed - previously we synthetically created a new verification URL and compared that with the assertion from the DB. Now we fetch both verification flows from the DB and compare them. Is this still a useful test?

selfservice/hook/verification_test.go
```diff
@@ -119,8 +121,18 @@ func TestVerifier(t *testing.T) {
...
+				originalFlow = &settings.Flow{RequestURL: "http://foo.com/settings?after_verification_return_to=verification_callback"}
+			case "register":
+				originalFlow = &registration.Flow{RequestURL: "http://foo.com/registration?after_verification_return_to=verification_callback"}
+			default:
+				t.FailNow()
+			}
 			require.NoError(t, hf(h, i, originalFlow))
-			expectedVerificationFlow, err = verification.NewPostHookFlow(conf, conf.SelfServiceFlowVerificationRequestLifespan(ctx), "", u, s, originalFlow)
+
+			assert.Emptyf(t, originalFlow.ContinueWith(), "%+v", originalFlow.ContinueWith())
```
aeneasr, line 134: Not sure if I understand what this is doing - you're creating a new original flow and then checking whether it has continue with. That of course will be true, because it's not added in the switch statement.

OR is `hf()` somehow setting that value?

selfservice/hook/verification.go
jonas-jonas: Because the state changed two lines above this and `PopulateVerificationMethod` uses that to decide whether to render the code or email input. Ideally, `NewPostHookFlow` would set that when creating the flow, but `verification.NewFlow` does not allow this, and changing that would require touching quite a few places, which I didn't want to do in this PR. 

I'd like to follow up here, and maybe refactor that part in a future PR, to use a single payload struct as parameter, as that function already takes in a bunch of parameters, making its use quite difficult, IMO. WDYT?

selfservice/hook/verification.go
jonas-jonas: True, removed.

selfservice/hook/verification_test.go
jonas-jonas: Do you mean removing https://github.com/ory/kratos/pull/3144/files/8169ac2945655590e9210777cf26b7766da2c4a5#diff-24ee502b250d67d70d70ba02ad18f303654b9af76552128958425b1c9043cc18R97-R100? I think that would still be fine, indeed.

selfservice/hook/verification.go
aeneasr: ok

selfservice/hook/verification_test.go
jonas-jonas: `hf` is executing the verification hook, which persists the verification flow, and since all verifiable addresses are now verified, we expect no `continue_with` items anymore.

PR #3101, opened 2023-02-14
feat: refactor identity persister

alnr:

PR #3145, opened 2023-03-04
fix: webhook tracing and missing defers

alnr:
Found by @jonas-jonas [here](https://github.com/ory/kratos/pull/3101#discussion_r1124712299) (thanks!)

The webhook tracing was broken in the last refactoring. This change restores the correct passing of span contexts, and also re-enables retries. Previously, all attempts were aborted after a total of 30s. Now, we get 30s for each attempt.


jonas-jonas (approved):
Nice! Thanks for fixing this so quickly! :)

PR #3148, opened 2023-03-06
fix: use type alias instead of type definition

CaptainStandby:
## Related issue(s)

https://github.com/ory/network/issues/251

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/schema/handler.go
+++ b/schema/handler.go
@@ -70,7 +70,7 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
 //
 //nolint:deadcode,unused
 //lint:ignore U1000 Used to generate Swagger and OpenAPI definitions
-type identitySchema json.RawMessage
+type identitySchema = json.RawMessage
 
 // Get Identity JSON Schema Response
 //
--- a/spec/api.json
+++ b/spec/api.json
@@ -971,7 +971,8 @@
             "type": "string"
           },
           "schema": {
-            "$ref": "#/components/schemas/identitySchema"
+            "description": "The actual Identity JSON Schema",
+            "type": "object"
           }
         },
         "type": "object"
--- a/internal/client-go/model_identity_schema_container.go
+++ b/internal/client-go/model_identity_schema_container.go
@@ -19,7 +19,7 @@ import (
 type IdentitySchemaContainer struct {
 	// The ID of the Identity JSON Schema
 	Id *string `json:"id,omitempty"`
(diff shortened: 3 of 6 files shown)


schema/handler.go
```diff
@@ -70,7 +70,7 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
 //
 //nolint:deadcode,unused
 //lint:ignore U1000 Used to generate Swagger and OpenAPI definitions
-type identitySchema json.RawMessage
+type identitySchema = json.RawMessage
```
jonas-jonas, line 73: What the...

jonas-jonas (approved):
LGTM.

Would this be a breaking change in the SDK? I guess since the Go SDK didn't change (apart from the doc comment) no? What does this look like in TS?

schema/handler.go
zepatrik: When you define a new type (without `=`), it removes all of the underlying types methods. Using `=` means that we only have another name for the same type.

zepatrik (approved):

spec/api.json
```diff
@@ -971,7 +971,8 @@
             "type": "string"
           },
           "schema": {
-            "$ref": "#/components/schemas/identitySchema"
```
alnr, line 974: Why this change?

schema/handler.go
zepatrik: See also https://go.dev/play/p/rZhjaze9ZYv

alnr: https://go.googlesource.com/proposal/+/master/design/18130-type-alias.md

CaptainStandby: The problem was that `func (m RawMessage) MarshalJSON() ([]byte, error)` and `func (m *RawMessage) UnmarshalJSON(data []byte) error` is dropped when using type definitions which in turn leads to the json marshaller treating `identitySchema` like a byte array and serializing it as a base64 string

spec/api.json
CaptainStandby: This was generated by `make sdk`. This is a result of using type alias instead of type definition

alnr (approved):

CaptainStandby:
> LGTM.
> 
> Would this be a breaking change in the SDK? I guess since the Go SDK didn't change (apart from the doc comment) no? What does this look like in TS?

I hope not 🙃

PR #3151, opened 2023-03-07
feat: improved oidc flow on duplicate account registration

Benehiko:
This PR improves the OIDC registration flow when a duplicate account error happens. 

Currently the flow looks as follows:

1. User registers with password (or other credentials)
2. User forgot they registered with password and tries to login through an OIDC provider (e.g. Google)
3. Kratos attempts a registration since the OIDC credentials do not exist
4. (optional) User needs to add missing traits (e.g. full name) which could not be retrieved from the OIDC provider
5. User gets a duplicate account error with a "Continue" button.
6. After submitting the "Continue" button the flow continues again to the OIDC provider, back to Kratos and redirects to UI with duplicate error (Steps 3 to 5)

Instead of causing a confusing redirect loop we should show the user the error with a fresh login flow (since the account exists). This also gives the user the option to do a recovery flow.

1. User registers with password (or other credentials)
2. User forgot they registered with password and tries to login through an OIDC provider (e.g. Google)
3. Kratos attempts a registration since the OIDC credentials do not exist
4. (optional) User needs to add missing traits
5. User is returned to a Login flow with the duplication error

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


selfservice/flow/registration/error.go
```diff
@@ -23,6 +23,7 @@ var (
 	ErrHookAbortFlow        = errors.New("aborted registration hook execution")
 	ErrAlreadyLoggedIn      = herodot.ErrBadRequest.WithID(text.ErrIDAlreadyLoggedIn).WithError("you are already logged in").WithReason("A valid session was detected and thus registration is not possible.")
 	ErrRegistrationDisabled = herodot.ErrBadRequest.WithID(text.ErrIDSelfServiceFlowDisabled).WithError("registration flow disabled").WithReason("Registration is not allowed because it was disabled.")
+	ErrDuplicateCredentials = herodot.ErrBadRequest.WithError(text.NewErrorValidationDuplicateCredentials().Text)
```
Benehiko, line 26: This could maybe be specific to OIDC only? 
So this type must be `ErrDuplicateOIDCCredentials`?

A new error could also be created for this specific case of duplicate account on oidc registration to provide better information to the user.

selfservice/flow/registration/hook.go
```diff
@@ -138,7 +137,10 @@ func (e *HookExecutor) PostRegistrationHook(w http.ResponseWriter, r *http.Reque
 		// would imply that the identity has to exist already.
 	} else if err := e.d.IdentityManager().Create(r.Context(), i); err != nil {
 		if errors.Is(err, sqlcon.ErrUniqueViolation) {
-			return schema.NewDuplicateCredentialsError()
+			// TODO: instead of throwing a duplicate error, we should return
+			// the user to a login screen with a message that the account already exists.
+			// we could even pre-fill the email address from the OIDC provider.
```
Benehiko, line 142: TODO: remove

selfservice/strategy/oidc/strategy.go
```diff
@@ -457,6 +457,16 @@ func (s *Strategy) handleError(w http.ResponseWriter, r *http.Request, f flow.Fl
 		// Reset all nodes to not confuse users.
 		// This is kinda hacky and will probably need to be updated at some point.
 
+		if errors.Is(err, registration.ErrDuplicateCredentials) {
+			rf.UI.Messages.Add(text.NewErrorValidationDuplicateCredentials())
+			lf, err := s.registrationToLogin(w, r, rf, provider)
+			if err != nil {
+				return err
```
Benehiko, line 464: not sure what to do here with the `err`, I guess it is okay to return as is?

test/e2e/cypress/integration/profiles/oidc/registration/success.spec.ts
```diff
@@ -211,6 +211,56 @@ context("Social Sign Up Successes", () => {
...
+          app,
+          acceptConsent: true,
+          acceptLogin: true,
+          expectSession: false,
+          email,
+          website,
+          route: registration,
+        })
+
+        cy.get('[data-testid="ui/message/4000007"]').should("be.visible")
```
Benehiko, line 247: checks if the duplicate error message is present on the form

test/e2e/cypress/integration/profiles/oidc/registration/success.spec.ts
```diff
@@ -211,6 +211,56 @@ context("Social Sign Up Successes", () => {
...
+
+        cy.location("href").should("contain", "/login")
+
+        cy.get("[name='provider'][value='hydra']").should("be.visible")
+        cy.get("[name='provider'][value='google']").should("be.visible")
+        cy.get("[name='provider'][value='github']").should("be.visible")
```
Benehiko, line 253: ensure we have form values for logging in

test/e2e/cypress/integration/profiles/oidc/registration/success.spec.ts
```diff
@@ -211,6 +211,56 @@ context("Social Sign Up Successes", () => {
...
+
+        if (app === "express") {
+          cy.get("[data-testid='forgot-password-link']").should("be.visible")
+        }
+
+        cy.get("input[name='identifier']").type(email)
+        cy.get("input[name='password']").type(password)
+        cy.submitPasswordForm()
```
Benehiko, line 261: ensure we can actually log in after getting such an error.

test/e2e/profiles/oidc/.kratos.yml
```diff
@@ -45,10 +42,12 @@ selfservice:
     registration:
       ui_url: http://localhost:4455/registration
       after:
+        password:
+          hooks:
+            - hook: session
```
Benehiko, line 47: In the `spa` profile we always have a session after registration. Just added it here to the `oidc` profile used by the express app.

Benehiko:
Added a new error specific to OIDC linking to an existing account. I think this might help confused users
image

selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -200,26 +225,15 @@ func (s *Strategy) processRegistration(w http.ResponseWriter, r *http.Request, a
...
-		if len(a.ReturnTo) > 0 {
-			opts = append(opts, login.WithFlowReturnTo(a.ReturnTo))
-		}
-
-		// This endpoint only handles browser flow at the moment.
-		ar, _, err := s.d.LoginHandler().NewLoginFlow(w, r, flow.TypeBrowser, opts...)
+		lf, err := s.registrationToLogin(w, r, a, provider.Config().ID)
 		if err != nil {
-			return nil, s.handleError(w, r, a, provider.Config().ID, nil, err)
+			return nil, err
```
jonas-jonas, line 230: This could lead to confusion down the line. Wouldn't it be clearer, if the error is handled here (call `s.handleError`) and just return the error from `registrationToLogin`? Do we have some kind of "guidelines" on this?

I noticed this pattern a lot and have always wondered why we handle these errors on the deepest level, instead of "at the top of the strategy" (e.g. entry point of the strategy - here the `Register` method).

selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -185,6 +185,31 @@ func (s *Strategy) Register(w http.ResponseWriter, r *http.Request, f *registrat
 	return errors.WithStack(flow.ErrCompletedByStrategy)
 }
 
+func (s *Strategy) registrationToLogin(w http.ResponseWriter, r *http.Request, a *registration.Flow, providerID string) (*login.Flow, error) {
```
jonas-jonas, line 188:
```suggestion
func (s *Strategy) registrationToLogin(w http.ResponseWriter, r *http.Request, rf *registration.Flow, providerID string) (*login.Flow, error) {
```

selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -185,6 +185,31 @@ func (s *Strategy) Register(w http.ResponseWriter, r *http.Request, f *registrat
...
+	if len(a.ReturnTo) > 0 {
+		opts = append(opts, login.WithFlowReturnTo(a.ReturnTo))
+	}
+
+	if len(a.UI.Messages) > 0 {
+		opts = append(opts, login.WithFormErrorMessage(a.UI.Messages))
+	}
+
+	// This endpoint only handles browser flow at the moment.
+	ar, _, err := s.d.LoginHandler().NewLoginFlow(w, r, flow.TypeBrowser, opts...)
```
jonas-jonas, line 200:
```suggestion
	lf, _, err := s.d.LoginHandler().NewLoginFlow(w, r, flow.TypeBrowser, opts...)
```

selfservice/strategy/oidc/strategy.go
jonas-jonas: When would this error happen? Only if creating the login flow somehow failed, right?

jonas-jonas (changes requested):
Looking good already. Just a few nits.

selfservice/strategy/oidc/strategy.go
Benehiko: Yeah, also this gets confusing since we call the `handleError` function inside this function.

selfservice/strategy/oidc/strategy_registration.go
Benehiko: Not sure I follow, do you want me to call `s.handleError` here instead of just returning the error from the `registrationToLogin` function? 

In such as case, I don't think it is possible, since the function adds data to the registration flow (such as error messages) and then passes that along to `handleError` which determines redirect strategies.

selfservice/strategy/oidc/strategy.go
jonas-jonas: >Yeah, also this gets confusing since we call the handleError function inside this function.

That's what I meant with my other comment. Maybe it's possible to not call `handleError` in the function?

selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -185,6 +185,32 @@ func (s *Strategy) Register(w http.ResponseWriter, r *http.Request, f *registrat
...
+
+	if len(rf.UI.Messages) > 0 {
+		opts = append(opts, login.WithFormErrorMessage(rf.UI.Messages))
+	}
+
+	// This endpoint only handles browser flow at the moment.
+	lf, _, err := s.d.LoginHandler().NewLoginFlow(w, r, flow.TypeBrowser, opts...)
+
+	if err != nil {
+		return nil, s.handleError(w, r, rf, providerID, nil, err)
```
jonas-jonas, line 203: All parameters seem to be supplied by the parameters to this function. So you should be able to remove this `s.handleError` call in here, and add it to the part that **calls** this function.

Benehiko: Make sense, have updated it :+1:

selfservice/strategy/oidc/strategy.go
jonas-jonas:
```suggestion
				return s.handleError(err ! ...)
```

I think you need to handle the error here, now too.

Benehiko: this is inside of the `handleError` function :)

Should probably not call it again recursively

jonas-jonas: lol
I am blind. Sorry.

jonas-jonas (changes requested):
Thanks for fixing the issues :)

LGTM, apart from the nil return in `handleError`.

selfservice/strategy/oidc/strategy.go
Benehiko: it's all good, I was also tempted at wrapping it at one point before realising it is inside of `handleError` :laughing:

selfservice/strategy/oidc/strategy.go
```diff
@@ -457,6 +457,16 @@ func (s *Strategy) handleError(w http.ResponseWriter, r *http.Request, f flow.Fl
...
 		// This is kinda hacky and will probably need to be updated at some point.
 
+		if errors.Is(err, registration.ErrDuplicateCredentials) {
+			rf.UI.Messages.Add(text.NewErrorValidationDuplicateCredentialsOnOIDCLink())
+			lf, err := s.registrationToLogin(w, r, rf, provider)
+			if err != nil {
+				return err
+			}
+			x.AcceptToRedirectOrJSON(w, r, s.d.Writer(), lf, lf.AppendTo(s.d.Config().SelfServiceFlowLoginUI(r.Context())).String())
+			return nil
```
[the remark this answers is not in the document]
Benehiko: The `x.AcceptToRedirectJSON` doesn't return anything. 

It redirects or writes json back to the browser.

the return nil just prevents the rest of the function from executing. In this case we don't want to propagate the error further since we want to return a login flow with the error message attached to the login flow.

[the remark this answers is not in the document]
jonas-jonas: We are always calling this function as part of the return part of functions, which return `someValue, error`. For example here https://github.com/ory/kratos/pull/3151/files/9b1b977a8bbc43e86687bb631fc46ac4f3d08241.

In that example, if you now return `nil` from `handleError`, the returned values of `processRegistration` will be `nil, nil`. And that is an issue, because the caller of `processRegistration` ([here](https://github.com/ory/kratos/blob/9c3bfe3d1b3fe69923589e96ba55d919efcc3b3c/selfservice/strategy/oidc/strategy_login.go#L122) for example) will check whether `err != nil`, which is `false` (because you returned `nil` here) and then assume that the `*login.Flow` is a valid object and `processRegistration` completed successfully. It will then go on and reference the `nil` pointer (`*login.Flow`), which will panic.

I would really like to see a test that executes this code path. Did you already write one?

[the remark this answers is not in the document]
Benehiko: I have a test for this though.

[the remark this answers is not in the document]
Benehiko: Please see https://github.com/ory/kratos/pull/3151/files#diff-03d9986942bad676c4567865d479f219a54bb633d7ccc9eed5820703e3d296a3

and an e2e test https://github.com/ory/kratos/pull/3151/files#diff-2695d9d779ed70b5b6a2d9cc5111dcac495b41bfc8353c0e0a50154ff787f7a6

Remember this PR changes the behaviour of the registration flow with `oidc` when the account already exists.
Instead of just giving back an error code and message, it responds with a new Login flow and the error message inside the `login` flow.

CaptainStandby (approved):
LGTM

selfservice/strategy/oidc/strategy.go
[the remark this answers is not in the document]
aeneasr: I think you should return `ErrAbortFlow` here. This prevents further calls in case something changes at some point.

aeneasr (dismissed):
Nicely done

aeneasr:
Ups, please address the one comment with abort flow

PR #3157, opened 2023-03-09
feat: add API to batch insert identities

hperl:
This PR does:
* add a batch API for creating multiple identities in one request
* add a generic `batch.Create` function to create multiple records in one INSERT statement
* uses `batch.Create` in creating multiple identities (with up to four individual queries depending on the structure of the identitites)

All new code is tested extensively, because the single-identity create is now a special-case of the multiple-identities create.

## Related issue(s)

Fixes https://github.com/ory/network/issues/266

## Checklist

- [x] Add new API endpoint and definitions
- [x] Implement a generic batch handler
~(Optional:) add `delete` action and `dry_run` paramenter~


identity/identity.go
```diff
@@ -461,3 +460,99 @@ func (i *Identity) WithDeclassifiedCredentialsOIDC(ctx context.Context, c cipher
...
+	// patch.
+	ID uuid.UUID `json:"patch_id"`
+}
+
+// swagger:enum PatchAction
+type PatchAction string
+
+const (
+	// Insert/create this identity.
+	ActionInsert PatchAction = "insert"
```
aeneasr, line 516: Use `create` to conform with `createIdentity`

identity/identity.go
```diff
@@ -461,3 +460,99 @@ func (i *Identity) WithDeclassifiedCredentialsOIDC(ctx context.Context, c cipher
...
+)
+
+// Patch identities response
+//
+// swagger:model patchIdentitiesResponse
+type patchIdentitiesResponse struct {
+	// The patch responses for the individual identities.
+	Identities []*IdentityPatchResponse `json:"identities"`
+
+	Errors []*IdentityPatchError `json:"errors"`
```
aeneasr, line 532: Move errors to the patch response items. Easier to correlate :)

identity/identity.go
```diff
@@ -461,3 +460,93 @@ func (i *Identity) WithDeclassifiedCredentialsOIDC(ctx context.Context, c cipher
...
+	Identities []*IdentityPatch `json:"identities"`
+
+	// Future fields:
+	// RemotePatchesURL string
+	// Async bool
+}
+
+// Payload for patching an identity
+//
+// swagger:model identityPatch
```
aeneasr, line 493: This is a bit misleading, because we also support patching identities individually using JSON Patch. Maybe `batchIdentityPatch`?

identity/identity.go
```diff
@@ -461,3 +460,93 @@ func (i *Identity) WithDeclassifiedCredentialsOIDC(ctx context.Context, c cipher
...
+//nolint:deadcode,unused
+//lint:ignore U1000 Used to generate Swagger and OpenAPI definitions
+type patchIdentitites struct {
+	// in: body
+	Body PatchIdentitiesBody
+}
+
+// Patch Identities Body
+//
+// swagger:model patchIdentitiesBody
```
aeneasr, line 477: Potentially prefix with batch

identity/identity.go
```diff
@@ -461,3 +460,93 @@ func (i *Identity) WithDeclassifiedCredentialsOIDC(ctx context.Context, c cipher
 	ii.Credentials = credsToPublish
 	return &ii, nil
 }
+
+// Patch Identities Parameters
+//
+// swagger:parameters patchIdentities
```
aeneasr, line 466: Potentially prefix with batch

identity/identity.go
```diff
@@ -461,3 +460,93 @@ func (i *Identity) WithDeclassifiedCredentialsOIDC(ctx context.Context, c cipher
...
+//
+//lint:ignore U1000 Used to generate Swagger and OpenAPI definitions
+type patchIdentitiesResponse struct {
+	// The patch responses for the individual identities.
+	Identities []*IdentityPatchResponse `json:"identities"`
+}
+
+// Response for a single identity patch
+//
+// swagger:model identityPatchResponse
```
aeneasr, line 537: Potentially prefix with batch

identity/handler.go
```diff
@@ -425,24 +449,71 @@ func (h *Handler) create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
...
+	if err := h.importCredentials(ctx, i, cr.Credentials); err != nil {
+		return nil, err
 	}
 
-	if err := h.r.IdentityManager().Create(r.Context(), i); err != nil {
-		h.r.Writer().WriteError(w, r, err)
+	return i, nil
+}
+
+// swagger:route PATCH /admin/identities identity patchIdentities
```
aeneasr, line 459: Potentially prefix with batch

identity/handler.go
```diff
@@ -425,24 +449,79 @@ func (h *Handler) create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
...
-	if err := h.importCredentials(r.Context(), i, cr.Credentials); err != nil {
-		h.r.Writer().WriteError(w, r, err)
+	if err := h.importCredentials(ctx, i, cr.Credentials); err != nil {
+		return nil, err
+	}
+
+	return i, nil
+}
+
+// swagger:route PATCH /admin/identities identity batchPatchIdentities
```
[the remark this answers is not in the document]
hperl: Yes, that's the new batch API.

persistence/sql/batch/create.go
```diff
@@ -0,0 +1,139 @@
+// Copyright © 2023 Ory Corp
```
aeneasr, line 1: This is some complex code that has no appropriate proof (except for circumstantial because some other method invokes it somewhere) that it works. It also does not have tests ensuring we won't introduce regressions. Please add tests.

identity/handler.go
```diff
@@ -425,24 +449,79 @@ func (h *Handler) create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
...
+		}
+	}
+
+	if err := h.r.IdentityManager().CreateIdentities(r.Context(), identities); err != nil {
 		h.r.Writer().WriteError(w, r, err)
 		return
 	}
+	for resIdx, identitiesIdx := range indexInIdentities {
+		res.Identities[resIdx].IdentityID = &identities[identitiesIdx].ID
+	}
```
aeneasr, line 522: Why is this so complicated? Is it not enough to return the identities themselves - their positioning in the index should not change?

identity/handler_test.go
```diff
@@ -566,6 +567,33 @@ func TestHandler(t *testing.T) {
 		})
 	})
 
+	t.Run("suite=PATCH identities", func(t *testing.T) {
```
aeneasr, line 570: Please improve this test. We need to know that all fields are imported and processed correctly. Especially things like verified and recovery addresses and credentials.

identity/handler_test.go
```diff
@@ -566,6 +567,33 @@ func TestHandler(t *testing.T) {
...
+			},
+		}
+		res := send(t, adminTS, "PATCH", "/identities", http.StatusOK, req)
+		assert.Len(t, res.Get("identities").Array(), len(req.Identities))
+		assert.Empty(t, res.Get("identities.0.error").Raw)
+		identityID := res.Get("identities.0.identity").String()
+		require.NotEmpty(t, identityID)
+
+		getIdentity := get(t, adminTS, "/identities/"+identityID, http.StatusOK)
+		assert.Equal(t, "batch-import-1@ory.sh", getIdentity.Get("traits.email").String())
```
aeneasr, line 594: I would prefer a snapshotx assertion here to ensure all fields are correctly imported.

identity/identity.go
```diff
@@ -403,6 +402,24 @@ func (i *Identity) Validate() error {
...
+}
+
+func RecoveryAddresses(i []*Identity) (res []RecoveryAddress) {
+	res = make([]RecoveryAddress, 0, len(i))
+	for _, id := range i {
+		res = append(res, id.RecoveryAddresses...)
+	}
+
+	return res
+}
```
aeneasr, line 421: How about `CollectVerifiableAddresses` or something along those lines?

identity/identity.go
```diff
@@ -403,6 +402,24 @@ func (i *Identity) Validate() error {
...
+func VerifiableAddresses(i []*Identity) (res []VerifiableAddress) {
+	res = make([]VerifiableAddress, 0, len(i))
+	for _, id := range i {
+		res = append(res, id.VerifiableAddresses...)
+	}
+
+	return res
+}
+
+func RecoveryAddresses(i []*Identity) (res []RecoveryAddress) {
```
aeneasr, line 414: Code is simple enough, and so should be the tests :)

identity/manager.go
```diff
@@ -88,6 +88,25 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
 	return m.r.PrivilegedIdentityPool().CreateIdentity(ctx, i)
 }
 
+func (m *Manager) CreateIdentities(ctx context.Context, identities []*Identity, opts ...ManagerOption) (err error) {
```
aeneasr, line 91: Tests please. For example, are identities properly validated? Is the Identity JSON Schema properly being applied?

identity/pool.go
```diff
@@ -62,6 +62,10 @@ type (
 		// if identity exists, backend connectivity is broken, or trait validation fails.
 		CreateIdentity(context.Context, *Identity) error
 
+		// CreateIdentities creates multiple identities. It is capable of setting credentials without encoding. Will return an error
+		// if identity exists, backend connectivity is broken, or trait validation fails.
+		CreateIdentities(context.Context, ...*Identity) error
```
aeneasr, line 67: This does not have an appropriate test in contrast to all other persister methods. Please add some.

persistence/sql/identity/persister_identity.go
```diff
@@ -181,82 +188,114 @@ func (p *IdentityPersister) FindByCredentialsIdentifier(ctx context.Context, ct
...
+	}
 
+	for _, cred := range credentials {
 		for _, ids := range cred.Identifiers {
 			// Force case-insensitivity and trimming for identifiers
 			ids = NormalizeIdentifier(cred.Type, ids)
 
-			if len(ids) == 0 {
-				return errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to create identity credentials with missing or empty identifier."))
+			if ids == "" {
```
aeneasr, line 254: Out of curiosity, why change this?

persistence/sql/update/update.go
```diff
@@ -21,13 +23,14 @@ type Model interface {
...
+func Generic(ctx context.Context, c *pop.Connection, tracer trace.Tracer, v Model, columnNames ...string) (err error) {
 	ctx, span := tracer.Start(ctx, "persistence.sql.update")
-	defer span.End()
+	defer otelx.End(span, &err)
 
 	quoter, ok := c.Dialect.(interface{ Quote(key string) string })
 	if !ok {
-		return errors.Errorf("store is not a quoter: %T", c.Store)
+		err = errors.Errorf("store is not a quoter: %T", c.Store)
+		return
```
aeneasr, line 33: This is not needed. By naming the return value in the function parameter, Go automatically assigns that variable name to the return value. You don't need to set `err` explicitly. Please revert

persistence/sql/update/update.go
```diff
@@ -50,7 +53,7 @@ func Generic(ctx context.Context, c *pop.Connection, tracer trace.Tracer, v Mode
 	)
 
 	var count int
-	if err := c.Store.GetContext(ctx, &count, c.Dialect.TranslateSQL(stmt), v.GetID(), v.GetNID()); err != nil {
+	if err = c.Store.GetContext(ctx, &count, c.Dialect.TranslateSQL(stmt), v.GetID(), v.GetNID()); err != nil {
```
aeneasr, line 56: Same here

aeneasr (changes requested):
Thank you! This looks pretty good already, but I had a hard time understanding how the persister creates the batches. I will need to do another pass.

> All new code is tested extensively, because the single-identity create is now a special-case of the multiple-identities create.

Not on my watch ;) This code is essential and has security implications. There can not be confusion of credentials or addresses when creating things in batch. Yet, there is not a single test assuring that we are actually creating things correctly when creating multiple things at once. Please add appropriate tests to prove that your changes work as expected. Thanks!

persistence/sql/batch/create.go
```diff
@@ -0,0 +1,139 @@
...
+	quoter interface {
+		Quote(key string) string
+	}
+	TracerConnection struct {
+		Tracer     *otelx.Tracer
+		Connection *pop.Connection
+	}
+)
+
+func buildInsertQueryArgs[T any](ctx context.Context, quoter quoter, models []*T) insertQueryArgs {
```
aeneasr, line 40: Is this copy/pasted from pop? If so, can we reuse the one from pop? If not, can we test it please?

persistence/sql/batch/create.go
```diff
@@ -0,0 +1,139 @@
...
+		columns        []string
+		quotedColumns  []string
+		placeholders   []string
+		placeholderRow []string
+	)
+
+	for _, col := range model.Columns().Cols {
+		quotedColumns = append(quotedColumns, quoter.Quote(col.Name))
+		columns = append(columns, col.Name)
+		placeholderRow = append(placeholderRow, "?")
```
aeneasr, line 54: Not sure how low level we are here, but some databases use `?` and others `$1`. Will this be translated later correctly?

persistence/sql/batch/create.go
```diff
@@ -0,0 +1,139 @@
...
+				}
+				id, err := uuid.NewV4()
+				if err != nil {
+					return nil, err
+				}
+				field.Set(reflect.ValueOf(id))
+			}
+			values = append(values, field.Interface())
+
+			// Special-handling for *sqlxx.NullTime: mapper.FieldByName sets this to a zero time.Time,
```
aeneasr, line 94: Is this coming from pop? If not, why do we need it now?

persistence/sql/batch/create.go
```diff
@@ -0,0 +1,139 @@
...
+	for _, m := range models {
+		m := reflect.ValueOf(m)
+
+		// Append model fields to args
+		for _, c := range columns {
+			field := mapper.FieldByName(m, c)
+
+			switch c {
+			case "created_at", "updated_at":
+				field.Set(reflect.ValueOf(now))
```
aeneasr, line 80: This behaves differently from pop. In pop, `created_at` is only set to `now` when its empty.

persistence/sql/batch/create.go
```diff
@@ -0,0 +1,139 @@
...
+				if err != nil {
+					return nil, err
+				}
+				field.Set(reflect.ValueOf(id))
+			}
+			values = append(values, field.Interface())
+
+			// Special-handling for *sqlxx.NullTime: mapper.FieldByName sets this to a zero time.Time,
+			// but we want a nil pointer instead.
+			if i, ok := field.Interface().(*sqlxx.NullTime); ok {
```
aeneasr, line 96: Could this be

```
if i, ok := field.Interface().(sqlxx.NullTime); ok {
```

as well? Or will it always be a pointer. Example:

```
type foo struct {
  a *sqlxx.NullTime
  b *sqlxx.NullTime
}
```

persistence/sql/identity/persister_identity.go
```diff
@@ -85,6 +88,10 @@ func (p *IdentityPersister) GetConnection(ctx context.Context) *pop.Connection {
 	return popx.GetConnection(ctx, p.c.WithContext(ctx))
 }
 
+func (p *IdentityPersister) Tracer(ctx context.Context) *otelx.Tracer {
```
aeneasr, line 91: Do we need this?

persistence/sql/identity/persister_identity.go
```diff
@@ -181,82 +188,114 @@ func (p *IdentityPersister) FindByCredentialsIdentifier(ctx context.Context, ct
...
+		nid         = p.NetworkID(ctx)
+		traceConn   = &batch.TracerConnection{Tracer: p.r.Tracer(ctx), Connection: conn}
+		credentials []*identity.Credentials
+		identifiers []*identity.CredentialIdentifier
+	)
 
-	nid := p.NetworkID(ctx)
-	for k := range i.Credentials {
-		cred := i.Credentials[k]
+	for _, ident := range identities {
```
aeneasr, line 220: This may be buggy because your data structure is a slice with pointers. See https://medium.com/swlh/use-pointer-of-for-range-loop-variable-in-go-3d3481f7ffc9

Please use the index to access the variable like you did with `ident.Credentials`

persistence/sql/identity/persister_identity.go
```diff
@@ -363,50 +406,65 @@ func (p *IdentityPersister) CountIdentities(ctx context.Context) (n int64, err e
...
+func (p *IdentityPersister) CreateIdentities(ctx context.Context, identities ...*identity.Identity) (err error) {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.CreateIdentities")
+	defer otelx.End(span, &err)
 
-	stateChangedAt := sqlxx.NullTime(time.Now())
-	i.StateChangedAt = &stateChangedAt
-	if i.State == "" {
-		i.State = identity.StateActive
-	}
+	for _, ident := range identities {
```
aeneasr, line 420: Like before, use the index, Luke

persistence/sql/identity/persister_identity.go
```diff
@@ -181,82 +188,114 @@ func (p *IdentityPersister) FindByCredentialsIdentifier(ctx context.Context, ct
...
-func (p *IdentityPersister) createVerifiableAddresses(ctx context.Context, i *identity.Identity) (err error) {
+func (p *IdentityPersister) createVerifiableAddresses(ctx context.Context, conn *pop.Connection, identities ...*identity.Identity) (err error) {
 	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.createVerifiableAddresses")
 	defer otelx.End(span, &err)
 
-	for k := range i.VerifiableAddresses {
-		if err := p.GetConnection(ctx).Create(&i.VerifiableAddresses[k]); err != nil {
-			return err
+	work := make([]*identity.VerifiableAddress, 0, len(identities))
+	for _, id := range identities {
```
aeneasr, line 285: Use the index

persistence/sql/identity/persister_identity.go
```diff
@@ -340,16 +380,19 @@ func (p *IdentityPersister) normalizeRecoveryAddresses(ctx context.Context, id *
...
-func (p *IdentityPersister) createRecoveryAddresses(ctx context.Context, i *identity.Identity) (err error) {
+func (p *IdentityPersister) createRecoveryAddresses(ctx context.Context, conn *pop.Connection, identities ...*identity.Identity) (err error) {
 	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.createRecoveryAddresses")
 	defer otelx.End(span, &err)
 
-	for k := range i.RecoveryAddresses {
-		if err := p.GetConnection(ctx).Create(&i.RecoveryAddresses[k]); err != nil {
-			return err
+	work := make([]*identity.RecoveryAddress, 0, len(identities))
+	for _, id := range identities {
```
aeneasr, line 388: Index

persistence/sql/identity/persister_identity.go
```diff
@@ -340,16 +380,19 @@ func (p *IdentityPersister) normalizeRecoveryAddresses(ctx context.Context, id *
...
+		for i := range id.RecoveryAddresses {
+			work = append(work, &id.RecoveryAddresses[i])
 		}
 	}
-	return nil
+	err = batch.Create(ctx, &batch.TracerConnection{Tracer: p.Tracer(ctx), Connection: conn}, work)
+
+	return err
```
aeneasr, line 395: No need for assignment, just do `return batch.Create`

aeneasr (changes requested):
Ok, took another look at the persister itself. I think the code falls into the classic Go "for range over a slice of pointers" trap. I think this could have serious implications when processing `2...n` elements

identity/handler.go
hperl: Because the API allows more patch actions than just `Create`, for example we wanted to add `Delete` as well. So we need to copy the IDs of the identities that were created back into the correct location.

persistence/sql/batch/create.go
hperl: Yes, I invoke `TranslateSQL` later.

persistence/sql/batch/create.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return insertQueryArgs{
+		TableName:    quoter.Quote(model.TableName()),
+		ColumnsDecl:  strings.Join(quotedColumns, ", "),
+		Columns:      columns,
+		Placeholders: strings.Join(placeholders, ",\n"),
+	}
+}
+
+func buildInsertQueryValues[T any](mapper *reflectx.Mapper, columns []string, models []*T) (values []any, err error) {
```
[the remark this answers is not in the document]
hperl: Hand-written, will test :)

persistence/sql/identity/persister_identity.go
hperl: removed.

persistence/sql/batch/create.go
hperl: fixed

persistence/sql/batch/create.go
hperl: If it is not a pointer to `sqlxx.NullTime` we won't need to correct anything.

persistence/sql/identity/persister_identity.go
```diff
@@ -181,82 +188,114 @@ func (p *IdentityPersister) FindByCredentialsIdentifier(ctx context.Context, ct
 	return i.CopyWithoutCredentials(), creds, nil
 }
 
+var credentialsTypes = map[identity.CredentialsType]*identity.CredentialsTypeTable{}
+
 func (p *IdentityPersister) findIdentityCredentialsType(ctx context.Context, ct identity.CredentialsType) (_ *identity.CredentialsTypeTable, err error) {
 	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.findIdentityCredentialsType")
 	defer otelx.End(span, &err)
 
+	if v, ok := credentialsTypes[ct]; ok && v != nil {
```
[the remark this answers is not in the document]
hperl: I added a lock. Isolation is not a concern IIRC because the credentials type table has no network ID.

identity/handler_test.go
hperl: I vastly expanded the test code now. PTAL :)

identity/handler_test.go
```diff
@@ -566,6 +569,126 @@ func TestHandler(t *testing.T) {
...
+			// --> we expect the server to succeed all patches in the list.
+			validPatches := []*identity.BatchIdentityPatch{
+				{Create: validCreateIdentityBody("valid-patch", 0)},
+				{Create: validCreateIdentityBody("valid-patch", 1)},
+				{Create: validCreateIdentityBody("valid-patch", 2)},
+				{Create: validCreateIdentityBody("valid-patch", 3)},
+				{Create: validCreateIdentityBody("valid-patch", 4)},
+			}
+
+			//anotherValidBody := validCreateIdentityBody("valid-patch", 42)
```
hperl, line 587: Will remove

aeneasr (approved):
Very nice :) One potential test could be added to ensure the merging of verification addresses works correctly :)

identity/extension_verify.go
```diff
@@ -58,10 +58,21 @@ func (r *SchemaExtensionVerification) Run(ctx jsonschema.ValidationContext, s sc
 }
 
 func (r *SchemaExtensionVerification) Finish() error {
-	r.i.VerifiableAddresses = r.v
+	r.i.VerifiableAddresses = merge(r.v, r.i.VerifiableAddresses)
```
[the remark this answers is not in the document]
hperl: Will do!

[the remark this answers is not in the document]
hperl: I added a dedicated test for `merge`, but the whole logic is also E2E tested here:

https://github.com/ory/kratos/pull/3157/files#diff-a9eda25ac25097b75b80cd3fcdf08ef119e92c2838052445547421bc3dee613cR677-R682

identity/pool.go
hperl: I now added persister tests.

identity/manager.go
hperl: This is now tested through handler and persister tests.

persistence/sql/batch/create.go
hperl: This is now tested through handler and persister tests as well as tests for the individual methods.

identity/handler_test.go
hperl: Snapshots only work for some fields, as the order of multiple (verifiable & recovery) addresses is random.

identity/handler.go
```diff
@@ -36,7 +36,7 @@ const (
 	RouteItem           = RouteCollection + "/:id"
 	RouteCredentialItem = RouteItem + "/credentials/:type"
 
-	BatchPatchIdentitiesLimit = 100
+	BatchPatchIdentitiesLimit = 2000
```
aeneasr, line 39: If you make this configurable we will have less trouble fine tuning this :)

PR #3169, opened 2023-03-15
fix: do not omit last page on identity list

zepatrik:
The problem is that the persister uses `1` as the fist page, while the pagination header helper assumes `0` is the first page.  Therefore, until now, the second-last page was considered the last and no `next` rel was set in the `Link` header.

This is the only way I could think of to make this not a breaking change, but it is super ugly.

### DIFF
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -180,7 +180,7 @@ func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Para
 		isam[i] = WithCredentialsMetadataAndAdminMetadataInJSON(identity)
 	}
 
-	migrationpagination.PaginationHeader(w, urlx.AppendPaths(h.r.Config().SelfAdminURL(r.Context()), RouteCollection), total, page, itemsPerPage)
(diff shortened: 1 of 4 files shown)


identity/handler.go
```diff
@@ -180,7 +180,7 @@ func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Para
 		isam[i] = WithCredentialsMetadataAndAdminMetadataInJSON(identity)
 	}
 
-	migrationpagination.PaginationHeader(w, urlx.AppendPaths(h.r.Config().SelfAdminURL(r.Context()), RouteCollection), total, page, itemsPerPage)
+	migrationpagination.PaginationHeader(w, urlx.AppendPaths(h.r.Config().SelfAdminURL(r.Context()), RouteCollection), total+int64(itemsPerPage), page, itemsPerPage)
```
zepatrik, line 183: This is a super ugly hack, please suggest a better way if you know any. `page-1` does not work because then we get `1` as the next page for `1`.

jonas-jonas: Why is this necessary? Doesn't that value come from the DB?

aeneasr: This breaks the x-total-count unfortunately.

PR #3176, opened 2023-03-17
feat: support OIDC on API flows

jonas-jonas:
Based on changes in https://github.com/ory/kratos/pull/2346

## Related issue(s)

Related to https://github.com/ory-corp/cloud/issues/3496

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

## Further Comments

The `id_token` flow looks as follows

```mermaid
sequenceDiagram
    participant Phone
    participant Kratos
    participant OIDC as OIDC Provider
    Phone->>+Kratos: Initialize API registration flow
    Kratos-->>-Phone: Return flow incl. UI/OIDC Nodes
    Phone->>+OIDC: Request id_token
    OIDC-->>Phone: Obtains consent from user
    OIDC-->>-Phone: Returns id_token
    Phone->>+Kratos: updateRegistrationFlow w/ id_token
    Kratos->>Kratos: Creates identity from data in id_token using JSONNet mapper from config
    Kratos-->>-Phone: Returns session (if session hook is set)
```

### DIFF
--- a/internal/testhelpers/selfservice_login.go
+++ b/internal/testhelpers/selfservice_login.go
@@ -208,6 +208,7 @@ func SubmitLoginForm(
 	expectedURL string,
 	opts ...InitFlowWithOption,
 ) string {
+	t.Helper()
 	if hc == nil {
 		hc = new(http.Client)
 		if !isAPI {
--- a/internal/testhelpers/selfservice_registration.go
+++ b/internal/testhelpers/selfservice_registration.go
@@ -65,12 +65,12 @@ func InitializeRegistrationFlowViaBrowser(t *testing.T, client *http.Client, ts 
 		flowID = gjson.GetBytes(body, "id").String()
 	}
 
-	rs, _, err := NewSDKCustomClient(ts, client).FrontendApi.GetRegistrationFlow(context.Background()).Id(flowID).Execute()
+	rs, res, err := NewSDKCustomClient(ts, client).FrontendApi.GetRegistrationFlow(context.Background()).Id(flowID).Execute()
 	if expectGetError {
 		require.Error(t, err)
 		require.Nil(t, rs)
 	} else {
-		require.NoError(t, err)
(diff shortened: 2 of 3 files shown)


selfservice/flow/flow.go
```diff
@@ -38,3 +41,18 @@ type Flow interface {
 	AppendTo(*url.URL) *url.URL
 	GetUI() *container.Container
 }
+
+func IsWebViewFlow(ctx context.Context, conf *config.Config, f Flow) (bool, error) {
```
hperl, line 45: From a security perspective, any client can become a WebView by adding the correct `return_to` parameter, even a malicious JS script.

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -84,6 +82,11 @@ type UpdateLoginFlowWithOidcMethod struct {
 	//
 	// required: false
 	UpstreamParameters json.RawMessage `json:"upstream_parameters"`
+
+	// Only used in API-type flows, when an id token has been received by mobile app directly from oidc provider.
+	//
+	// required: false
+	IDToken string `json:"id_token"`
```
hperl, line 89: What are the security implications when we get the ID token from the user instead of the OIDC provider?

aeneasr: If I recall correctly, that should be fine. The ID token will be validated on our end and we will check both issuer and, more importantly, audience. By checking audience, we ensure that the ID token was issued on our behalf, and not someone elses behalf. That way we avoid the vulnerability that could be caused by instead accepting access tokens as proof of authentication, as access tokens don't have an audience and they could have been issued for any app.

Still, this allows a eve to trick adam into using eve's ID token. Would be good to know how others deal with that?

jonas-jonas:
Superseded by https://github.com/ory/kratos/pull/3216

Issue #1412, opened 2021-06-11
Reduce identity list amount of sub-queries


aeneasr:
**Describe the bug**

Currently, we are performing one large query (`SELECT * FROM identities`) and several sub-queries per identity (`findVerifiableAddresses`, `findRecoveryAddresses`, `injectTraitsSchemaURL`) which can be quite slow once several identities exist in the database.

**Expected behavior**

We should reduce the amount of queries and check if we are missing any keys to improve query performance.

**Environment**

- Version: master

aeneasr:
Ok, so the problem is not the queries themselves but instead that we run the sub-queries sequentially. Instead it might make sense to either do all of this in one query or alternatively run the subqueries in a parallelized fashion. See the jaeger trace:


https://user-images.githubusercontent.com/3372410/121717615-a9ca6c80-cae1-11eb-925d-ba4c0ad6a448.mov

PR #3183, opened 2023-03-22
feat: refactor credentials fetching

alnr:
This change revamps the way we fetch identity credentials. We no longer need most of the helper fields for gobuffalo/pop inside the `Identity` and `Credentials` structures, and we collect all the credentials in one joined query rather than using pop's `EagerPreload` functionality.

### DIFF
--- a/identity/identity.go
+++ b/identity/identity.go
@@ -13,8 +13,6 @@ import (
 
 	"github.com/samber/lo"
 
-	"github.com/gobuffalo/pop/v6"
-
 	"github.com/tidwall/sjson"
 
 	"github.com/tidwall/gjson"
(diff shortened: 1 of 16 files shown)


aeneasr (approved):
Very nice changes, I think this simplifies the code a lot!

aeneasr:
It looks like you're running into ordering issues: https://github.com/ory/kratos/actions/runs/4492557398/jobs/7902651309?pr=3183#step:15:12872

Every database handles "deafult" sorting differently. Since we don't really want to sort the dataset (as that is slow in RBR tables), my recommendation is to either sort the result set (with Go) around here: https://github.com/ory/kratos/pull/3183/files#diff-41d54e67508bfcb98af6522993d007de47a74ada2cf8426fc871c8a8a8babaeeR561

identity/identity.go
```diff
@@ -132,36 +127,6 @@ type Identity struct {
...
-	if err := i.setCredentials(tx); err != nil {
-		return err
-	}
-
-	if err := i.Validate(); err != nil {
-		return err
-	}
-
-	return UpgradeCredentials(i)
-}
```
aeneasr, line 145: Does it hurt to keep this validation? Do we need to remove it? As far as I can tell, listing identities would no longer validate them nor upgrade the credentials, which might cause regressions down the road.

persistence/sql/identity/persister_identity.go
```diff
@@ -526,6 +605,14 @@ func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.
 		return nil, err
 	}
 
+	if len(params.Expand) > 0 {
```
aeneasr, line 608: I did not realize that we now need to do this. This will have a significant performance impact on the identity list endpoint, because we are iteratively stepping over the result set. This can easily expand to 400-500 waterfall-like SQL queries. I almost fear that this is a blocker because people will immediately complain about it.

aeneasr (changes requested):
Generally speaking i think this looks really good. One issue though is how we now list identities. We had several complains from e.g. Wikia about list performance before and we were able to resolve those with EagerPreload. This patch will regress on those improvements and reintroduce those issues. See: #1412

persistence/sql/identity/persister_identity.go
alnr: Fixed, I think. The code is a little less nice now, but still worth it IMO.

LMKWYT.

identity/identity.go
alnr: Added validation+upgrade credentials to ListIdentities

aeneasr (approved):
Very nice

PR #3200, opened 2023-03-30
fix: improve webhook resilience

hperl:
Changes:

* Decouple post-registration post-persistence context from request context
* Upgrade ory/x to use retrying in Jsonnet VM
* Add better logging for when webhooks fail

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/flow/registration/hook.go
+++ b/selfservice/flow/registration/hook.go
@@ -48,7 +48,7 @@ type (
 	}
 )
 
-func PostHookPostPersistExecutorNames(e []PostHookPostPersistExecutor) []string {
+func ExecutorNames[T any](e []T) []string {
 	names := make([]string, len(e))
 	for k, ee := range e {
 		names[k] = fmt.Sprintf("%T", ee)
@@ -107,24 +107,31 @@ func (e *HookExecutor) PostRegistrationHook(w http.ResponseWriter, r *http.Reque
 					WithRequest(r).
 					WithField("executor", fmt.Sprintf("%T", executor)).
 					WithField("executor_position", k).
-					WithField("executors", PostHookPostPersistExecutorNames(e.d.PostRegistrationPostPersistHooks(r.Context(), ct))).
+					WithField("executors", ExecutorNames(e.d.PostRegistrationPrePersistHooks(r.Context(), ct))).
 					WithField("identity_id", i.ID).
 					WithField("flow_method", ct).
 					Debug("A ExecutePostRegistrationPrePersistHook hook aborted early.")
 				return nil
 			}
 
-			var traits identity.Traits
-			if i != nil {
-				traits = i.Traits
-			}
+			e.d.Logger().
+				WithRequest(r).
+				WithField("executor", fmt.Sprintf("%T", executor)).
+				WithField("executor_position", k).
+				WithField("executors", ExecutorNames(e.d.PostRegistrationPrePersistHooks(r.Context(), ct))).
+				WithField("identity_id", i.ID).
+				WithField("flow_method", ct).
+				WithError(err).
+				Error("ExecutePostRegistrationPostPersistHook hook failed with an error.")
+
+			traits := i.Traits
 			return flow.HandleHookError(w, r, a, traits, ct.ToUiNodeGroup(), err, e.d, e.d)
 		}
 
 		e.d.Logger().WithRequest(r).
 			WithField("executor", fmt.Sprintf("%T", executor)).
 			WithField("executor_position", k).
-			WithField("executors", PostHookPostPersistExecutorNames(e.d.PostRegistrationPostPersistHooks(r.Context(), ct))).
+			WithField("executors", ExecutorNames(e.d.PostRegistrationPrePersistHooks(r.Context(), ct))).
 			WithField("identity_id", i.ID).
 			WithField("flow_method", ct).
 			Debug("ExecutePostRegistrationPrePersistHook completed successfully.")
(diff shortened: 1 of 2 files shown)


alnr (dismissed):

selfservice/hook/web_hook.go
```diff
@@ -209,7 +210,15 @@ func (e *WebHook) ExecutePostRegistrationPostPersistHook(_ http.ResponseWriter,
 	if gjson.GetBytes(e.conf, "can_interrupt").Bool() || gjson.GetBytes(e.conf, "response.parse").Bool() {
 		return nil
 	}
-	return otelx.WithSpan(req.Context(), "selfservice.hook.WebHook.ExecutePostRegistrationPostPersistHook", func(ctx context.Context) error {
+
+	// We want to decouple the request from the hook execution, so that the hooks still execute even
+	// if the request is canceled.
+	var cancel context.CancelFunc
+	ctx := trace.ContextWithSpan(context.Background(), trace.SpanFromContext(req.Context()))
+	ctx, cancel = context.WithTimeout(ctx, time.Second*30)
```
alnr, line 218: This changes the retry behavior. Before, we had three retries of 30 sec each, now we have a total of 30sec including all retries.

selfservice/hook/web_hook.go
```diff
@@ -209,7 +210,15 @@ func (e *WebHook) ExecutePostRegistrationPostPersistHook(_ http.ResponseWriter,
 	if gjson.GetBytes(e.conf, "can_interrupt").Bool() || gjson.GetBytes(e.conf, "response.parse").Bool() {
 		return nil
 	}
-	return otelx.WithSpan(req.Context(), "selfservice.hook.WebHook.ExecutePostRegistrationPostPersistHook", func(ctx context.Context) error {
+
+	// We want to decouple the request from the hook execution, so that the hooks still execute even
+	// if the request is canceled.
```
alnr, line 215: I don't think we actually want this, do we?

hperl: I think we do. Because otherwise the user can decide (through cancelling the request at the appropriate time) whether or not a post persist hook is executed. And if the post-hook is cancelled, the user is still persisted. That would mean that our customers can't put any faith in the execution of the post-persist hooks.

selfservice/hook/web_hook.go
hperl: Ah good catch! The reason I put some timeout here is because the Jsonnet VM will try to spin up a process until the parent process is cancelled, and we prob want some general timeout here. 5 minutes? WDYT?

```suggestion
	ctx, cancel = context.WithTimeout(ctx, 5*time.Minute)
```

selfservice/hook/web_hook.go
```diff
@@ -284,7 +293,10 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
 			// incoming request context is cancelled.
 			//
 			// The webhook will still cancel after 30 seconds as that is the configured timeout for the HTTP client.
+			var cancel context.CancelFunc
 			ctx = trace.ContextWithSpan(context.Background(), trace.SpanFromContext(ctx))
+			ctx, cancel = context.WithTimeout(ctx, time.Second*30)
```
hperl, line 298:
```suggestion
			ctx, cancel = context.WithTimeout(ctx, 5*time.Minute)
```

selfservice/flow/registration/hook.go
```diff
@@ -189,24 +196,31 @@ func (e *HookExecutor) PostRegistrationHook(w http.ResponseWriter, r *http.Reque
...
-					WithField("executors", PostHookPostPersistExecutorNames(e.d.PostRegistrationPostPersistHooks(r.Context(), ct))).
+					WithField("executors", ExecutorNames(e.d.PostRegistrationPostPersistHooks(r.Context(), ct))).
 					WithField("identity_id", i.ID).
 					WithField("flow_method", ct).
 					Debug("A ExecutePostRegistrationPostPersistHook hook aborted early.")
 				return nil
 			}
 
-			var traits identity.Traits
-			if i != nil {
```
zepatrik, line 200: Are you sure this cannot be `nil`?

selfservice/hook/web_hook.go
zepatrik: IMO not for every hook, but only for the "async", non-blocking hooks. The sync hooks need to share context with the request.

zepatrik (commented):
Please correct me, but I don't see new retry logic added, only context timeout added.

zepatrik:
I realized that we already had an update of ory/x in a different PR, so the retry logic there is already on master :+1: 
However, I do not agree that we want to decouple the request context from the webhook, at least not for blocking hooks. Non-blocking hooks already have their own context, so I think we should keep that as is and just pick the log improvements from this PR.

selfservice/hook/web_hook.go
alnr: > I think we do. Because otherwise the user can decide (through cancelling the request at the appropriate time) whether or not a post persist hook is executed. And if the post-hook is cancelled, the user is still persisted. That would mean that our customers can't put any faith in the execution of the post-persist hooks.

That actually makes a lot of sense to me.

selfservice/hook/web_hook.go
```diff
@@ -209,7 +210,15 @@ func (e *WebHook) ExecutePostRegistrationPostPersistHook(_ http.ResponseWriter,
 	if gjson.GetBytes(e.conf, "can_interrupt").Bool() || gjson.GetBytes(e.conf, "response.parse").Bool() {
```
hperl, line 210: @zepatrik PostRegistrationPostPersist hooks are always non-interrupting and the response is not parsed. So I still think it's fine to run these in a new context, as the rest of the flow is untouched by the result of this.

selfservice/hook/web_hook.go
aeneasr: Could we use the same context/cancel logic for all web hook types?

aeneasr (dismissed):
Looks good :)

selfservice/hook/web_hook.go
hperl: Most of the web hooks need to be executed in the request context and we rely on their answer for further processing. For post-persist hooks and async hooks, we don't rely on the answer, so we can execute them in another context.

selfservice/flow/registration/hook.go
alnr: Should be fine I think, since we're  in the Post-registration-post-persist hook.

selfservice/hook/web_hook.go
alnr: Yeah, interrupting post-persist hooks don't actually make a ton of semantic sense. We should probably rework the config to deprecate them.

zepatrik (dismissed):

aeneasr (approved):

Issue #3203, opened 2023-04-01
Support passing session ID webhook context where possible


mfzl:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Right now after login and after registration hooks have no way to link to active session. I haven't found anyway to link and identify the current session the hook is being fired from.  

It also doesn't need to pass any sensitive information, just passing session ID is enough to make it work, other details can then be retrieved from admin API if desired.

### Describe your ideal solution

Making session ID available in the webhook context or possibly other non sensitive information in the context passed to Jsonnet, if session is available.

### Workarounds or alternatives

There are none so far

### Version

0.11.1

mfzl:
I'm willing to contribute if it's acceptable

mfzl:
Now that I think about it, `Session` is already declassified and it can only be passed to the `web_hook` if it's configured to pass the session information. I don't see a downside of making the entire session object available in template context.

aeneasr:
Sure, makes sense :)

PR #3206, opened 2023-04-03
feat: make session object available in web_hook context

mfzl:
Right now it's not possible to link `web_hook` actions with the corresponding user session. This allows more possibilities in `web_hook` actions when a user logins or a session is started for the first time after user registration.

## Related issue(s)

#3203 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/hook/stub/test_body.jsonnet
+++ b/selfservice/hook/stub/test_body.jsonnet
@@ -1,6 +1,7 @@
 function(ctx) std.prune({
   flow_id: ctx.flow.id,
   identity_id: if std.objectHas(ctx, "identity") then ctx.identity.id,
+  session_id: if std.objectHas(ctx, "session") then ctx.session.id,
   headers: ctx.request_headers,
   url: ctx.request_url,
   method: ctx.request_method,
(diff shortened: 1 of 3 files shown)


mfzl:
I'll update the docs if when this is good enough to merge

mfzl:
This is more nuanced than I thought. Including session is not very useful because `after-login` hooks are fired before session is persisted and we get the zero uuid. 

For backwards compatibility reasons and to stay consistent I think it would be best to keep the current hooks as it is and introduce a new hook group similar to registrations. Something like after-login post-persist.

The only other unique identifiers to identify the sessions are `Token` and `LogoutToken`. They are sensitive and I'm not sure it's a good idea.

Any thoughts?

cc @aeneasr @zepatrik

aeneasr:
Oh good find!

But I don't think adding post-persist hooks to login is a worth the effort right now, as we don't have a mechanism to control the "position" of hooks yet. There would need to be a really good use case for session.id to exist to justify the amount of work/scope.

PR #3212, opened 2023-04-04
feat: return to oauth flow after switching from login to other flows

Benehiko:
## Related issue(s)
https://github.com/ory/network/issues/264

With `oauth2_provider.return_to_enabled = true` the following is possible:
1. Login flow switch to Registration flow, return to OAuth2 provider
2. Login flow switch to Recovery flow to Settings flow, return to OAuth2 provider

An end-user following a login flow from a Single Sign On portal to the OAuth2 provider (Hydra) will always be redirected back to the provider (Hydra) after performing a more complex flow, such as switching from `login` to `registration` or `recovery` flow in Kratos. The OAuth2 provider can then complete the flow and redirect the user accordingly.

**Note**: The provider (Hydra) URL will also need to be added to Kratos's `selfservice.allowed_return_urls` configuration. Any `return_to` parameters supplied before the login request is created will be overwritten in Kratos by the OAuth2 provider URL.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/embedx/config.schema.json
+++ b/embedx/config.schema.json
@@ -1901,6 +1901,12 @@
               "Authorization": "Bearer some-token"
             }
           ]
+        },
+        "return_to_enabled": {
+          "title":"Override `return_to` in OAuth2 flows",
+          "type":"boolean",
+          "default":false,
+          "description":"Override the return_to query paramter in OAuth flows back to the Provider Request URL"
         }
       },
       "additionalProperties": false
--- a/embedx/config.schema.json
+++ b/embedx/config.schema.json
@@ -1902,11 +1902,11 @@
             }
           ]
         },
-        "return_to_enabled": {
-          "title":"Override `return_to` in OAuth2 flows",
+        "override_return_to": {
+          "title":"Persist OAuth2 request between flows",
           "type":"boolean",
           "default":false,
-          "description":"Override the return_to query paramter in OAuth flows back to the Provider Request URL"
+          "description":"Override the return_to query parameter to the OAuth2 provider request URL when perfoming an OAuth2 login flow."
         }
       },
       "additionalProperties": false
--- a/embedx/config.schema.json
+++ b/embedx/config.schema.json
@@ -1906,7 +1906,7 @@
           "title":"Persist OAuth2 request between flows",
           "type":"boolean",
           "default":false,
-          "description":"Override the return_to query parameter to the OAuth2 provider request URL when perfoming an OAuth2 login flow."
+          "description":"Override the return_to query parameter with the OAuth2 provider request URL when perfoming an OAuth2 login flow."
         }
       },
       "additionalProperties": false
--- a/selfservice/flow/login/handler.go
+++ b/selfservice/flow/login/handler.go
@@ -432,7 +432,7 @@ func (h *Handler) createBrowserLoginFlow(w http.ResponseWriter, r *http.Request,
 		// different flows, such as login to registration and login to recovery.
 		// After completing a complex flow, such as recovery, we want the user
 		// to be redirected back to the original flow.
-		if hlr.RequestUrl != "" {
+		if hlr.RequestUrl != "" && h.d.Config().OAuth2ProviderReturnToEnabled(r.Context()) {
 			// replace the return_to query parameter
 			q := r.URL.Query()
 			q.Set("return_to", hlr.RequestUrl)
--- a/selfservice/flow/login/handler.go
+++ b/selfservice/flow/login/handler.go
@@ -406,6 +406,7 @@ type createBrowserLoginFlow struct {
 func (h *Handler) createBrowserLoginFlow(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	var hlr *hydraclientgo.OAuth2LoginRequest
 	var hlc uuid.NullUUID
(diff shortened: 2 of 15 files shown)


embedx/config.schema.json
```diff
@@ -1901,6 +1901,12 @@
               "Authorization": "Bearer some-token"
             }
           ]
+        },
+        "return_to_enabled": {
```
Benehiko, line 1905: not sure if this naming is the best here...

maybe I should make it `override_return_to`.

selfservice/flow/login/handler.go
```diff
@@ -407,6 +407,19 @@ func (h *Handler) createBrowserLoginFlow(w http.ResponseWriter, r *http.Request,
...
+		// on OAuth2 flows, we need to use the RequestURL
+		// as the ReturnTo URL.
+		// This is because a user might want to switch between
+		// different flows, such as login to registration and login to recovery.
+		// After completing a complex flow, such as recovery, we want the user
+		// to be redirected back to the original flow.
+		if hlr.RequestUrl != "" && h.d.Config().OAuth2ProviderReturnToEnabled(r.Context()) {
+			// replace the return_to query parameter
+			q := r.URL.Query()
+			q.Set("return_to", hlr.RequestUrl)
```
Benehiko, line 420: I'm setting the URL query parameter here instead of using passing in the URL to the `h.NewLoginFlow` function since we do not persist the return_to URL. So when retrieving the login flow later with the same ID will not give us the `return_to` URL since it is computed from the request URL at runtime. 

This was the cleanest approach I could find.

test/e2e/cypress/integration/profiles/oidc-provider/login.spec.ts
```diff
@@ -146,3 +147,189 @@ context("OpenID Provider", () => {
     cy.getCookie("ory_hydra_session_dev").should("be.null")
   })
 })
+
+context("OpenID Provider - change between flows", () => {
```
Benehiko, line 151: E2E test which verifies if the `return_to` URL is set when switching between flows after initializing an OAuth2 flow. 

The AX takes the current `return_to` query parameter from the login request and passes it along to the recovery flow. So there is some UI logic that makes this possible. The only thing kratos does is take the current request URL and checks if it has a `login_challenge` query parameter. Based on this parameter it overwrites the `return_to` parameter.

test/e2e/cypress/integration/profiles/oidc-provider/login.spec.ts
```diff
@@ -146,3 +147,189 @@ context("OpenID Provider", () => {
...
+
+    // since we aren't skipping the consent, Kratos will ask us to login again
+    cy.get('input[name="password"]').clear().type(newPassword)
+    cy.get('button[value="password"]').click()
+
+    // we should now end up on the consent screen
+    doConsent(["code_recovery", "password"])
+  })
+
+  it("switch to recovery flow with oidc link", () => {
```
Benehiko, line 267: this test case validates the behaviour when a user updates their profile with an OIDC provider. In this test we reuse Hydra + UI consent node app which is linked to our Kratos instance.

test/e2e/profiles/oidc-provider/.kratos.yml
```diff
@@ -4,6 +4,23 @@ clients:
...
+      config:
+        providers:
+          - id: google
+            provider: generic
+            client_id: ${OIDC_GOOGLE_CLIENT_ID}
+            client_secret: ${OIDC_GOOGLE_CLIENT_SECRET}
+            issuer_url: http://localhost:4444/
+            scope:
+              - offline
+            mapper_url: file://test/e2e/profiles/oidc-provider/hydra.jsonnet
```
Benehiko, line 23: enable the OIDC provider for Kratos in this profile

zepatrik (commented):
Is the login flow always the entry point, or do we need to add this logic also to registration? I guess not, but I wanted to bring it up anyway.

Follow up:
- consider always adding the OAuth2 server URL to the allow list for `return_to` URLs
- if I understand correctly, the UI can be simplified now because it does not have to handle the return to the OAuth2 server anymore, we probably need to adjust that somewhere in the docs

jonas-jonas (approved):
Code wise this LGTM, but I am not 100% sure I fully understand all consequences of this. Would like for someone else to confirm.

embedx/config.schema.json
```diff
@@ -1901,6 +1901,12 @@
               "Authorization": "Bearer some-token"
             }
           ]
+        },
+        "override_return_to": {
+          "title":"Persist OAuth2 request between flows",
+          "type":"boolean",
+          "default":false,
+          "description":"Override the return_to query parameter to the OAuth2 provider request URL when perfoming an OAuth2 login flow."
```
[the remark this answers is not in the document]
Benehiko: isn't it `to` though? since the value is set `to` the OAuth2 provider request URL. Not set `with` the OAuth2 provider request URL?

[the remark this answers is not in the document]
jonas-jonas: But you're overriding, not setting, in this sentence. You could "override to be" as well.

It's a nit, but reads better IMO.

jonas-jonas (approved):

zepatrik (approved):

PR #3216, opened 2023-04-06
feat: support OIDC flows for apps

hperl:
## Flow for Kratos AppAuth

1. Call [createNativeLoginFlow](https://www.ory.sh/docs/reference/api#tag/frontend/operation/createNativeLoginFlow) to create a flow of type api, and request a session token exchange code
2. Kratos sends flow ID, the session token exchange code, and the SSO URL to the app
3. App opens the SSO URL in the native browser
4. Browser opens the SSO URL, the authorization endpoint takes the user through authentication and consent
5. The authorization endpoint redirects to the callback url (for example https://nostalgic-cori-rxgved3nrs.projects.oryapis.com/self-service/methods/oidc/callback/google), containing the authorization code
6. The browser opens that URL. Kratos exchanges the authorization code for an identity token, and invokes the redirect (through the post-login redirect) to an URL registered to be handled by the app (Android App links or iOS Universal links)
7a. Upon receiving the URL, the app can retrieve the session token at `/self-service/exchange-code-for-session-token&code=<session token exchange code>`
7b. If the app can't receive a deep link (for example, if we are simulating the app on the web), then the app can also poll `/self-service/exchange-code-for-session-token&code=<session token exchange code>` and wait for a 200 status code with the session.

## Related issue(s)

- https://github.com/ory-corp/cloud/issues/3496
- https://github.com/ory/kratos/issues/707

Docs PR: https://github.com/ory/docs/pull/1379

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Security Considerations

During the authentication process we need to ensure that the following entities are all under the same control:

- A1: The app that started the flow
- B1: The browser that performs the authentication with the OIDC provider
- B2: The browser that performs the OIDC callback with Kratos
- A2: The app that fetches the session token

For example, it should not be possible for an attacker to start a flow, trick a victim into performing the authentication with the OIDC provider, and then be able to fetch the session token in the name of the victim.

We ensure that A1, A2, B1, B2 are under the same control as follows:

- [x] The `initial_session_token_exchange_code` binds A1 and A2 by generating in when the flow starts and requiring it in exchange for the session token
- [x] The OIDC `state` parameter binds B1 and B2 by deriving the `state` from the flow ID and the hashed code, which guarantees that no attacker can forge the `state` for B2
- [x] The `return_to_code` binds B2 and A2 by redirecting the browser to the specified and whitelisted `return_to` URL and including the code in a query parameter.

It then follows that A1, A2, B1, B2 are all under the same control


persistence/sql/migrations/sql/20230405000000000001_create_session_token_exchangers.mysql.up.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE session_token_exchangers (
```
aeneasr, line 1: `session_token_exchanges`? :)

persistence/sql/persister_sessiontokenexchanger.go
```diff
@@ -0,0 +1,62 @@
...
+	e := sessiontokenexchange.Exchanger{
+		NID:    p.NetworkID(ctx),
+		FlowID: flowID,
+		Code:   code,
+	}
+
+	e.ID, err = uuid.NewV4()
+	if err != nil {
+		return err
+	}
```
aeneasr, line 32: This will cause high latency on INSERTS on RBR tables. It's faster if the system (cockroach) generates these IDs

selfservice/flow/login/handler.go
```diff
@@ -250,6 +264,12 @@ type createNativeLoginFlow struct {
 	//
 	// in: header
 	SessionToken string `json:"X-Session-Token"`
+
+	// EnableSessionTokenExchangeCode requests the login flow to include a code that can be used to retrieve the session token
+	// after the login flow has been completed.
+	//
+	// in: query
+	EnableSessionTokenExchangeCode bool `json:"enable_session_token_exchange_code"`
```
aeneasr, line 272: Is this parameter needed?

selfservice/sessiontokenexchange/handler.go
```diff
@@ -0,0 +1,143 @@
...
+	// The Session
+	//
+	// The session contains information about the user, the session device, and so on.
+	// This is only available for API flows, not for Browser flows!
+	//
+	// required: true
+	Session *session.Session `json:"session"`
+}
+
+// swagger:route GET /self-service/login/exchange-session-token frontend exchangeSessionToken
```
aeneasr, line 96: How about

```suggestion
// swagger:route POST /self-service/login/exchange frontend exchangeLoginCodeForSessionToken
```

persistence/sql/migrations/sql/20230405000000000001_create_session_token_exchangers.mysql.up.sql
hperl: done

selfservice/sessiontokenexchange/handler.go
hperl: I renamed this to `/self-service/exchange-code-for-session-token` because it is now flow-agnostic. WDYT?

selfservice/flow/login/handler.go
hperl: Yes, so that the user can opt in to the code-specific flow behaviour.

persistence/sql/persister_sessiontokenexchanger.go
hperl: fixed.

.github/workflows/ci.yaml
```diff
@@ -203,13 +203,23 @@ jobs:
...
+      #      TODO(hperl): Enable this once the React Native app uses the new SDK
+      #      - name: "Run Playwright tests"
+      #        run: |
+      #          cd test/e2e
+      #          npm run playwright
+      #        env:
+      #          DB: ${{ matrix.database }}
+      #          RN_UI_PATH: react-native-ui
+      #          NODE_UI_PATH: node-ui
+      #          REACT_UI_PATH: react-ui
```
hperl, line 222: The Playwright tests will be enabled once the SDK is updated and bumped in the react-native app repo.

selfservice/flow/login/handler.go
```diff
@@ -250,6 +264,17 @@ type createNativeLoginFlow struct {
...
```
aeneasr, line 272: Alternative: `return_session_token_exchange_code`

selfservice/hook/session_issuer.go
```diff
@@ -52,6 +56,19 @@ func (e *SessionIssuer) executePostRegistrationPostPersistHook(w http.ResponseWr
 	}
 
 	if a.Type == flow.TypeAPI {
+		if ok, _ := e.r.SessionTokenExchangePersister().CodeExistsForFlow(r.Context(), a.ID); ok {
+			if err := e.r.SessionTokenExchangePersister().UpdateSessionOnExchanger(r.Context(), a.ID, s.ID); err != nil {
+				return errors.WithStack(err)
+			}
+			returnTo, err := x.SecureRedirectTo(r, e.r.Config().SelfServiceBrowserDefaultReturnTo(r.Context()), a.SecureRedirectToOpts(r.Context(), e.r)...)
```
aeneasr, line 63: The idea here being that this would open the registered URL of the app? So for example `myapp://post-login`?

selfservice/strategy/oidc/strategy.go
```diff
@@ -258,46 +245,74 @@ func (s *Strategy) validateFlow(ctx context.Context, r *http.Request, rid uuid.U
...
+	default:
+		return uuid.Nil, false
+	}
 }
 
-func (s *Strategy) alreadyAuthenticated(w http.ResponseWriter, r *http.Request, req interface{}) bool {
-	// we assume an error means the user has no session
-	if _, err := s.d.SessionManager().FetchFromRequest(r.Context(), r); err == nil {
-		if _, ok := req.(*settings.Flow); ok {
+func (s *Strategy) alreadyAuthenticated(w http.ResponseWriter, r *http.Request, f interface{}) bool {
```
aeneasr, line 296: Why are these changes here needed?

test/e2e/playwright/tests/app_login.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+})
+
+test.describe("Login", () => {
+  test("login twice", async ({ page }) => {
+    await testLogin(page, "login@example.com")
+    await logout(page)
+    await testLogin(page, "login@example.com")
+  })
+
+  test("register, then login", async ({ page }) => {
```
aeneasr, line 93: login, then register?

test/e2e/playwright/tests/app_login.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+async function testRegistrationOrLogin(page: Page, username: string) {
+  const popupPromise = page.waitForEvent("popup")
+  await page.getByText(/sign (up|in) with ory/i).click()
+  const popup = await popupPromise
+
+  await performOidcLogin(popup, "registration@example.com")
+
+  await page.waitForURL("Home")
+  expect(popup.isClosed()).toBeTruthy()
+  await expect(page.getByText("Welcome back")).toBeVisible()
```
aeneasr, line 35: This means we have a session?

test/e2e/playwright/tests/app_login.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    await performOidcLogin(popup, "register-reject-then-accept@example.com")
+
+    await page.waitForURL("Home")
+    expect(popup.isClosed()).toBeTruthy()
+    await expect(page.getByText("Welcome back")).toBeVisible()
+  })
+})
+
+test.describe("Login", () => {
+  test("login twice", async ({ page }) => {
```
aeneasr, line 87: When logging in with an account that does not exist yet, the user is signed up, and then log in is attempted again. Does that work with this flow too? It's similar to "register twice", where on the second registration we actually perform login.

Would be great to test this explicitly

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -109,7 +105,12 @@ func (s *Strategy) processLogin(w http.ResponseWriter, r *http.Request, a *login
 			}
 
 			// This flow only works for browsers anyways.
-			aa, err := s.d.RegistrationHandler().NewRegistrationFlow(w, r, flow.TypeBrowser, opts...)
+			aa, err := s.d.RegistrationHandler().NewRegistrationFlow(w, r, a.Type, opts...)
```
aeneasr, line 108: I would have expected a few tests in Go (strategy_test) that cover this behavior. They're also going to execute faster than the playwright tests. It is a bit of duplication though, but it makes debugging much simpler. WDYT? Worth the effort?

aeneasr (commented):
Very nice & clean, I have a few remarks. One question is whether we should add some tests to the oidc go test suite. It's a bit of duplication because we test the same thing in e2e, but those tests execute faster and are easier to debug / step through if something breaks

selfservice/hook/session_issuer.go
hperl: Yes, exactly. The behaviour of a flow when a code was requested is that we always redirect to the `return_to` URL and then let the client fetch the session token (on success) or refetch the flow (on failure).

selfservice/strategy/oidc/strategy.go
hperl: There are two changes here:

* First, if the user is already authenticated and a code was requested, we want to write the existing session to the code-token exchange, so that the client can fetch the session token in exchange for the code.
* Second, the old behaviour was to always redirect the client to the default return_to URL, not considering the more specific return_to URLs that usually take precedence over the default (return_to in the flow, then flow-specific return_to URL). I treated this as a bug and fixed it for all flows.

test/e2e/playwright/tests/app_login.spec.ts
hperl: Ah yes, copy-pasta :)

test/e2e/playwright/tests/app_login.spec.ts
hperl: Yes

test/e2e/playwright/tests/app_login.spec.ts
hperl: Yes, this flow is exactly testing that the translation between the flows is working correctly, and that the code can still be exchanged.

How would you test this more explicitly?

selfservice/flow/login/handler.go
hperl: No we can't, because the code-specific flow behaviour is different from the non-code behaviour.

When issuing a code, we redirect the user to the `return_to` address so that the app gets back control and can retrieve the session token from the code.

When *not* issuing the code, we return the session token directly as JSON.

persistence/sql/migrations/sql/20230405000000000001_create_session_token_exchanges.mysql.up.sql
```diff
@@ -0,0 +1,15 @@
...
+    session_id CHAR(36) DEFAULT NULL,
+    code VARCHAR(64) NOT NULL,
+
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+-- Relevant query:
+--   SELECT * from session_token_exchanges
+--   WHERE flow_id = ? AND nid = ? AND code = ? AND session_id IS NOT NULL AND code <> '';
```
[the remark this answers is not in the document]
hperl: I need to revisit the indices because we now need to lookup an exchange by (nid, code) as well (without the flow), but in any case I think the combination of nid and flow_id should already uniquely identify a single exchange, so adding more columns to the index does not help with the lookup speed. Correct?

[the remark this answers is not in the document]
hperl: I now added two indices:
* (flow_id, nid) for queries specifying the flow_id
* (code, nid) for queries specifying the code

Both combinations should return at maximum one query, and flow_id and code respectively are more selective than nid.

selfservice/sessiontokenexchange/handler.go
aeneasr: 👍

aeneasr: Or, in that case, it might make more sense to have this under `/sessions/exchange`?

selfservice/strategy/oidc/strategy.go
```diff
@@ -125,8 +125,7 @@ type authCodeContainer struct {
 }
 
 func generateState(flowID string) string {
-	state := x.NewUUID().String()
-	return base64.RawURLEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", flowID, state)))
+	return flowID
```
hperl, line 128: @aeneasr / reviewers: this is potentially security relevant. Previously the state sent to the OIDC IdP was random, now it is "just" the flow ID.

The technical reason is that in the App/Code case, there is no shared state between the code that generates the state (for/in the app) and the code that invokes the OIDC callback (from the browser tab). So in the calllback we use the flow ID from the state to update the exchange with the session token.

In the OAuth flow, the state acts as a CSRF token to make sure that an attacker can't log in the user into his own account. However, here, the attacker would need to guess the victim's flow ID, which has high entropy.

selfservice/strategy/oidc/strategy_login.go
hperl: Definitely, it would have made debugging much simpler during development as well! I just didn't find the right spot before, but now I added the tests. PTAL :)

selfservice/sessiontokenexchange/handler.go
aeneasr: naming-wise the `/self-service/exchange-code-for-session-token` API doesn't really match any other endpoint exposed by ory - lots of words and hyphens. Maybe `/sessions/token-exchange`?

selfservice/strategy/oidc/strategy.go
hperl: fixed

selfservice/sessiontokenexchange/handler.go
hperl: fixed

session/manager_http.go
```diff
@@ -320,3 +323,28 @@ func (s *ManagerHTTP) SessionAddAuthenticationMethods(ctx context.Context, sid u
...
+			if err == nil {
+				returnTo = r
+			}
+		}
+
+		if err = s.r.SessionTokenExchangePersister().UpdateSessionOnExchanger(r.Context(), f.GetID(), sessionID); err != nil {
+			return false, errors.WithStack(err)
+		}
+
+		returnTo.RawQuery = "code=" + code.ReturnToCode
```
aeneasr, line 344: This overrides the original raw query. Use this instead:

```
q = returnTo.Query()
q.Add(code, code.ReturnToCode)
returnTo.RawQuery = q.String()
```

session/handler.go
```diff
@@ -903,3 +908,91 @@ func RespondWitherrorGenericOnAuthenticated(h herodot.Writer, err error) httprou
...
+	//
+	// required: true
+	// in: query
+	InitCode string `json:"init_code"`
+
+	// The part of the code returned by the return_to URL.
+	//
+	// required: true
+	// in: query
+	ReturnToCode string `json:"return_to_code"`
```
aeneasr, line 929: In the future, this endpoint may support other token exchanges as well. Should we somehow prefix / clarify what code you're exchanging here?

session/manager_http.go
```diff
@@ -320,3 +323,28 @@ func (s *ManagerHTTP) SessionAddAuthenticationMethods(ctx context.Context, sid u
 	sess.SetAuthenticatorAssuranceLevel()
 	return s.r.SessionPersister().UpsertSession(ctx, sess)
 }
+
+func (s *ManagerHTTP) MaybeRedirectAPICodeFlow(w http.ResponseWriter, r *http.Request, f flow.Flow, sessionID uuid.UUID) (handled bool, err error) {
+	ctx, span := s.r.Tracer(r.Context()).Tracer().Start(r.Context(), "sessions.ManagerHTTP.MaybeRedirectAPICodeFlow")
+	defer otelx.End(span, &err)
+
+	if code, ok, _ := s.r.SessionTokenExchangePersister().CodeForFlow(ctx, f.GetID()); ok {
+		returnTo := s.r.Config().SelfServiceBrowserDefaultReturnTo(ctx)
```
aeneasr, line 332: No need to change, just a remark. It's often better to read when the short code path is indented and the long one not:

```
if true {
  // short
}
// very
// long
// execution
```

aeneasr (commented):
Very nice, a couple of questions and remarks on the code-side. I will look into the flow description next!

persistence/sql/persister_sessiontokenexchanger.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (p *Persister) MoveToNewFlow(ctx context.Context, oldFlow, newFlow uuid.UUID) (err error) {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.MoveToNewFlow")
+	defer otelx.End(span, &err)
+
+	conn := p.GetConnection(ctx)
+	query := fmt.Sprintf("UPDATE %s SET flow_id = ? WHERE flow_id = ? AND nid = ? %s",
+		conn.Dialect.Quote(new(sessiontokenexchange.Exchanger).TableName()),
+		updateLimitClause(conn),
```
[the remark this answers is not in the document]
hperl: No, but since the database can't know that I'd thought the `LIMIT 1` clause would serve as a hint to CRDB to not scan all regional replicas.

persistence/sql/persister_sessiontokenexchanger.go
```diff
@@ -0,0 +1,108 @@
...
+func (p *Persister) GetExchangerFromCode(ctx context.Context, initCode string, returnToCode string) (e *sessiontokenexchange.Exchanger, err error) {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.GetExchangerFromCode")
+	defer otelx.End(span, &err)
+
+	e = new(sessiontokenexchange.Exchanger)
+	conn := p.GetConnection(ctx)
+	if err = conn.Where(`
+nid = ? AND
+init_code = ? AND init_code <> '' AND
+return_to_code = ? AND return_to_code <> '' AND
```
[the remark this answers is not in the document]
hperl: >Do we need the non-empty assertions when the parameters are provided?

This is only a precaution that we never fetch from empty credentials.

>Is this not missing an index? Have you checked whether querying this scans excessive rows?

The index should be this: [CREATE INDEX session_token_exchanges_nid_code_idx ON session_token_exchanges (init_code, nid);](https://github.com/ory/kratos/pull/3216/files#diff-aff0acaa5f6f5cc800bc3c7e011181c3f327ca8effb94936430ba2e31e7adc9eR17). I didn't verify this with `EXPLAIN` though.

persistence/sql/persister_sessiontokenexchanger.go
```diff
@@ -0,0 +1,108 @@
...
+	return sqlcon.HandleError(conn.RawQuery(query, sessionID, flowID, p.NetworkID(ctx)).Exec())
+}
+
+func (p *Persister) CodeForFlow(ctx context.Context, flowID uuid.UUID) (codes *sessiontokenexchange.Codes, found bool, err error) {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.CodeForFlow")
+	defer otelx.End(span, &err)
+
+	var e sessiontokenexchange.Exchanger
+	switch err = sqlcon.HandleError(p.GetConnection(ctx).
+		Where("flow_id = ? AND nid = ? AND init_code <> '' and return_to_code <> ''", flowID, p.NetworkID(ctx)).
```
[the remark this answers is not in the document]
hperl: The index should be [CREATE INDEX session_token_exchanges_nid_flow_id_idx ON session_token_exchanges (flow_id, nid);](https://github.com/ory/kratos/pull/3216/files#diff-aff0acaa5f6f5cc800bc3c7e011181c3f327ca8effb94936430ba2e31e7adc9eR21)

selfservice/flow/login/handler.go
```diff
@@ -133,6 +135,14 @@ func (h *Handler) NewLoginFlow(w http.ResponseWriter, r *http.Request, ft flow.T
 		return nil, nil, errors.WithStack(herodot.ErrBadRequest.WithReasonf("Unable to parse AuthenticationMethod Assurance Level (AAL): %s", cs.ToUnknownCaseErr()))
 	}
 
+	if ft == flow.TypeAPI && r.URL.Query().Get("return_session_token_exchange_code") == "true" {
+		e, err := h.d.SessionTokenExchangePersister().CreateSessionTokenExchanger(r.Context(), f.ID)
+		if err != nil {
+			return nil, nil, errors.WithStack(herodot.ErrInternalServerError.WithWrap(err))
```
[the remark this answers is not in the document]
hperl: As the user can't do anything about this we should return a 500 code. Is this always the case when we don't wrap? If so, I can remove the wrapping.

selfservice/flow/registration/handler.go
```diff
@@ -119,6 +121,16 @@ func (h *Handler) NewRegistrationFlow(w http.ResponseWriter, r *http.Request, ft
 		o(f)
 	}
 
+	if ft == flow.TypeAPI && r.URL.Query().Get("return_session_token_exchange_code") == "true" {
```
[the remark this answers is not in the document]
hperl: Well, it is used twice, once for the registration flow and once for the login flow. But those flows are different request objects and could change for different reasons.

I'd rather have a general helper to parse the request params into a struct. I found https://github.com/gorilla/schema for this but it is in maintenance mode. WDYT?

session/handler.go
hperl: I'm not sure what other usages you think of here. If there are other types of tokens, wouldn't this be a different end point? It would probably be easier to reason about the security then?

session/handler.go
```diff
@@ -903,3 +908,91 @@ func RespondWitherrorGenericOnAuthenticated(h herodot.Writer, err error) httprou
...
+//
+//	Schemes: http, https
+//
+//	Responses:
+//	  200: successfulNativeLogin
+//	  403: errorGeneric
+//	  404: errorGeneric
+//	  410: errorGeneric
+//	  default: errorGeneric
+func (h *Handler) exchangeCode(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
```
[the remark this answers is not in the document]
hperl: Here: https://github.com/ory/kratos/pull/3216/files/cfbc327b89911795a62b07fc43a704fe6f943152#diff-03d9986942bad676c4567865d479f219a54bb633d7ccc9eed5820703e3d296a3R180-R194

session/manager_http.go
hperl: You're right! I extracted it previously, but it's nicer when the guard clause comes first.

persistence/sql/persister_sessiontokenexchanger.go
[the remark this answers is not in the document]
aeneasr: But that's missing return_to_code, no?

persistence/sql/persister_sessiontokenexchanger.go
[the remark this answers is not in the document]
aeneasr: But here too we are missing return_to_code?

persistence/sql/persister_sessiontokenexchanger.go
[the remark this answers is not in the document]
aeneasr: Ah, that makes sense

selfservice/flow/registration/handler.go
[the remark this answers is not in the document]
aeneasr: Ok

> I found https://github.com/gorilla/schema for this but it is in maintenance mode. WDYT?

Not worth it IMO

aeneasr:
I wanted to execute the tests to check the flow end-to-end, but I was not able to do so, because of:

```
foobar@foobair playwright % npm run playwright

> @ory/kratos-e2e-suite@0.0.1 playwright
> playwright test

ReferenceError: Cypress is not defined

   at cypress/helpers/index.ts:43

  41 |   new DOMParser().parseFromString(html, "text/html")
```

How can I execute the tests?

Additionally I found the code in the react app a bit confusing. I kind of would have expected the code to be more obvious, but it's hiding behind several nested function calls that are relatively deep in the stack trace.

This made it difficult to do a final end-to-end review. It would be really helpful if there was a simple Makefile command (or something similar) to get the playwright tests started!

hperl:
> [...]
> 
> Additionally I found the code in the react app a bit confusing. I kind of would have expected the code to be more obvious, but it's hiding behind several nested function calls that are relatively deep in the stack trace.

The entry point to the OIDC handling in the native app is the `422` error returned by Kratos. @jonas-jonas told me that you discussed this with him as a means to signal to the native app that it needs to get a browser involved. If you have ideas to improve the code lets meet here: https://github.com/ory/kratos-selfservice-ui-react-native/pull/69

> This made it difficult to do a final end-to-end review. It would be really helpful if there was a simple Makefile command (or something similar) to get the playwright tests started!

I added a make target. You can now run the Playwright tests with

```shell
RN_UI_PATH="path/to/kratos-selfservice-ui-react-native" make test-e2e-playwright
```

(env var needed until https://github.com/ory/kratos-selfservice-ui-react-native/pull/69 is merged)

persistence/sql/persister_sessiontokenexchanger.go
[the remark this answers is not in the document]
hperl: see above.

aeneasr:
Some of the tests are failing for me:

Screenshot 2023-04-24 at 18 19 52

Fruthermore, as far as I understand we now have two codes:

1. One code is returned by createNativeFlow
2. Second code is returned as part of return URL

We need both codes to complete the flow - is that correct?

hperl:
> Fruthermore, as far as I understand we now have two codes:
> 
> One code is returned by createNativeFlow
> Second code is returned as part of return URL
> We need both codes to complete the flow - is that correct?

Yes that's correct.

jonas-jonas (commented):
Code-wise this looks great to me. E2e just straight up fail for me locally. But probably an issue on my machine.

selfservice/flow/registration/handler.go
```diff
@@ -119,6 +121,16 @@ func (h *Handler) NewRegistrationFlow(w http.ResponseWriter, r *http.Request, ft
...
+		// Panicing here is ok since it will return a 500 to the user, which is accurate for when
+		// we can't generate a random string.
```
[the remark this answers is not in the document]
hperl: removed :)

test/e2e/playwright/kratos.config.json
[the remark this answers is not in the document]
hperl: fixed

spec/api.json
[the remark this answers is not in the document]
hperl: Your comment lost it's link to the source code. What is the issue here?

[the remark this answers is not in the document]
jonas-jonas: I was just wondering whether the (current) 422 response from the registration handler is documented in the open api doc.

aeneasr (approved):

PR #3218, opened 2023-04-06
fix: error messages in OpenAPI/Swagger / improve error messages from failed webhooks and client timeouts

alnr:
Partial fix for ory-corp/cloud#4354.

https://github.com/ory/herodot/compare/v0.9.13...v0.10.2
https://github.com/ory/x/compare/v0.0.549...v0.0.551

Previously, the Swagger and OpenAPI specs would incorrectly describe
specific errors returned from Kratos' API. Instead of having a nested
"error" key in the returned JSON payload, the API specs would incorrectly
say the contents of the generic error were flattened into the payload.

Example before (incorrect):
```json
{
  "id": "browser_location_change_required",
  "code": 422,
  "status": "Unprocessable Entity",
  "reason": "In order to complete this flow please redirect the browser to: /ui/settings?flow=22b3ad6f-c50a-4c2f-8c94-a16e9dc20083",
  "message": "browser location change required",
  "redirect_browser_to": "/ui/settings?flow=22b3ad6f-c50a-4c2f-8c94-a16e9dc20083"
}

```
Actual Kratos response (correct, unchanged):
```json
{
  "error": {
    "id": "browser_location_change_required",
    "code": 422,
    "status": "Unprocessable Entity",
    "reason": "In order to complete this flow please redirect the browser to: /ui/settings?flow=22b3ad6f-c50a-4c2f-8c94-a16e9dc20083",
    "message": "browser location change required"
  },
  "redirect_browser_to": "/ui/settings?flow=22b3ad6f-c50a-4c2f-8c94-a16e9dc20083"
}

```
This issue occurs because go-swagger does not interpret embedded structs with `json` tags correctly, i.e., it flattens the embedded struct's fields into outer type instead of nesting it under the key specified in the `json` tag.

The least bad fix is to use modified copies of the affected error structs just for Swagger API spec generation.


selfservice/errorx/manager.go
```diff
@@ -64,7 +64,8 @@ func (m *Manager) Forward(ctx context.Context, w http.ResponseWriter, r *http.Re
 	to, errCreate := m.Create(ctx, w, r, err)
 	if errCreate != nil {
 		// Everything failed. Resort to standard error output.
-		m.d.Writer().WriteError(w, r, errCreate)
```
alnr, line 67: This is important. Previously, in case of `ctx.Err() == context.Canceled`, we would attempt to write the error container to the database but that would always fail because the context used for that operation is already canceled. We'd then proceed to write back that second error (failing to persist the error container) to the response writer. What we want to do instead is log the persistence error, and output the actual error to the response writer.

selfservice/errorx/persistence.go
```diff
@@ -12,16 +12,19 @@ import (
 
 type (
 	Persister interface {
-		// Add adds an error to the manager and returns a unique identifier or an error if insertion fails.
-		Add(ctx context.Context, csrfToken string, err error) (uuid.UUID, error)
+		// CreateErrorContainer adds an error to the manager and returns a
+		// unique identifier or an error if insertion fails.
+		CreateErrorContainer(ctx context.Context, csrfToken string, err error) (uuid.UUID, error)
```
alnr, line 17: I've renamed these functions because they look very weird in the persister.

selfservice/hook/web_hook.go
```diff
@@ -352,6 +353,15 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
 
 		resp, err := httpClient.Do(req)
 		if err != nil {
+			if te := interface{ Timeout() bool }(nil); errors.As(err, &te) && te.Timeout() || errors.Is(err, context.DeadlineExceeded) {
```
alnr, line 356: Getting to know if a HTTP request timed out is surprisingly hard in Go. That's an artifact from the days before `context`. If you use the deprecated `Timeout` field in the `http.Client, you get a `Timeout` error. If the request's context times out, you get `Context.DeadlineExceeded`.

x/pointer.go
```diff
@@ -11,10 +11,3 @@ func PointToUUID(id uuid.UUID) *uuid.UUID {
 	}
 	return &id
 }
```
zepatrik, line 13: You should be able to remove the `PointToUUID` function here as well.

zepatrik (dismissed):
Very nice :+1:

x/pointer.go
alnr: `PointToUUID` has different semantics than `pointerx.Ptr`.

selfservice/hook/web_hook.go
CaptainStandby, line 356: The condition could be a function with a nice name to make the code more readable
```go
func isTimeoutError(err error) boolean {
	te := interface{ Timeout() bool }(nil)
	return errors.As(err, &te) && te.Timeout() || errors.Is(err, context.DeadlineExceeded)
}
```

```suggestion
			if isTimeoutError(err) {
```

.schema/openapi/patches/health.yaml
CaptainStandby, line 1: Why was this deleted?

alnr: This file was unused and it's functionality was superseded by [this](https://github.com/ory/kratos/blob/f3123ec28303325d4c9944cc9225a98802cd37db/Makefile#L98).

selfservice/hook/web_hook.go
alnr: Yep, that's fair. Will fix.

jonas-jonas:
E2e tests fail, with 502 in webhooks, so it looks like this might related to these changes. If you haven't looked into it yet, I can try to repro this locally tomorrow (e2e test setup in Kratos is a bit finicky). We can also do a call, if you want. Let me know on slack :)

spec/api.json
```diff
@@ -4564,15 +4371,15 @@
             },
             "description": "Ory Kratos is ready to accept connections."
           },
-          "500": {
+          "default": {
             "content": {
-              "application/json": {
+              "text/plain": {
```
aeneasr, line 4376: This looks phishy

x/swagger/swagger_types_global.go
```diff
@@ -13,13 +13,16 @@ import "github.com/ory/herodot"
...
-type errorGeneric struct {
+type ErrorGeneric struct {
 	// Contains error details
 	//
 	// required: true
-	Error herodot.DefaultError `json:"error"`
+	Error GenericError `json:"error"`
 }
 
+// swagger:model genericError
```
aeneasr, line 23: We now have `errorGeneric` and `genericError` as types in our SDK, I don't think that's desired. How about naming this

```
errorDetails
```

or something like that?

aeneasr (changes requested):

spec/api.json
alnr: This change is related to https://github.com/ory/x/pull/685. The liveness check never returns a structured error. Any error the caller might see (if any) will be from cloudflare or the ingress controller.  Either way, it'll be unstructured stuff in their logs.

It was necessary to remove the structured error from the health/alive endpoint because of this change https://github.com/ory/herodot/pull/106, which was done months ago but never rolled out. The cleanest and simplest solution was just to get rid of the structured error alltogether.

x/swagger/swagger_types_global.go
alnr: Yup, it's weird. But this is exactly how it was before this change https://github.com/ory/herodot/pull/106 (which was done months ago but never rolled out). Any name other than `genericError` for this type will be a much larger change in the generated specs/code.

selfservice/hook/web_hook.go
```diff
@@ -387,6 +400,18 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
...
+		if err != nil {
+			err = herodot.DefaultError{
+				CodeField:     http.StatusBadGateway,
+				StatusField:   http.StatusText(http.StatusBadGateway),
+				GRPCCodeField: grpccodes.Aborted,
+				ReasonField:   "A third-party upstream service responded improperly. Please try again later.",
+				ErrorField:    err.Error(),
+			}.WithWrap(errors.WithStack(err))
+		}
+	}()
```
jonas-jonas, line 413: This is the issue for the failing tests. 

Do we actually always want to return 502 here? If the webhook target responded 400, we interpret that as "a validation error occurred". Reason being, that we can parse the response and show validation errors in the form, for example. 
image

alnr: Yep, makes sense. Fixed. Thanks for your help!

CaptainStandby (approved):
LGTM 🎉

jonas-jonas (approved):

x/swagger/swagger_types_global.go
aeneasr: I see, too bad :(

aeneasr (approved):

Issue #2727, opened 2022-09-12
Add a way to merge identity credentials.


David-Wobrock:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Kratos does not allow automatic account linking when doing social sign-in. This is great for security reasons.
So, for instance, when an account for david@gmail.com exists in Kratos, and I connect through Google SSO, a new identity with david@gmail.com will be created if this Google account has never been used.

In our case, we want to allow our users to prove that they own the existing account by typing their password.
If they can prove this, we want to merge both identities together. More precisely, we want the Google SSO credentials to be linked to the existing identity -  so that next time the user connects, they are connected with the existing identity.

### Describe your ideal solution

One possible solution would be to provide an admin API endpoint in Kratos, that allows merging two identites.
We could identify the current identity through the Kratos cookie, and the target identity by passing their UUID in the URL.
The end result would be that the current identity has been deleted, and the credentials associated to it are now linked to the target identity.

This will allow having one transaction to encapsulate the logic.
The problem with the workaround is that it is subject to race conditions, if the target identity is updated at the same time by two or more requests I believe.

### Workarounds or alternatives

The workaround we plan to use for now is:
1. Fetch the credentials from identity A => using https://www.ory.sh/docs/kratos/reference/api#operation/adminGetIdentity
2. Import the credentials from A into identity B => using https://www.ory.sh/docs/kratos/manage-identities/import-user-accounts-identities#social-sign-in-connections
3. Delete identity A => using https://www.ory.sh/docs/kratos/reference/api#operation/adminDeleteIdentity

### Version

0.10.1

### Additional Context

Discussed in the Ory Slack with Thomas => https://ory-community.slack.com/archives/C012RJ2MQ1H/p1662728891106129

David-Wobrock:
[...]

PR #3222, opened 2023-04-10
feat: link oidc credentials when login

splaunov:
When user tries to login with OIDC for the first time but has already registered before with email/password a credentials identifier conflict may be detected by Kratos. In this case user needs to login with email/password first and then link OIDC credentials on a settings screen.
This PR simplifies UX and allows user to link OIDC credentials to existing account right in the login flow, without
switching to settings flow.

## Related issue(s)
#2727 

## Checklist

      introduces a new feature.
- [ x] I am following the
      the changes.
      works.

### DIFF
--- a/cmd/clidoc/main.go
+++ b/cmd/clidoc/main.go
@@ -132,6 +132,7 @@ func init() {
 		"NewInfoSelfServiceLoginContinue":                         text.NewInfoSelfServiceLoginContinue(),
 		"NewErrorValidationSuchNoWebAuthnUser":                    text.NewErrorValidationSuchNoWebAuthnUser(),
 		"NewInfoSelfServiceLoginLinkCredentials":                  text.NewInfoSelfServiceLoginLinkCredentials(),
+		"NewErrorValidationLoginLinkedCredentialsDoNotMatch":      text.NewErrorValidationLoginLinkedCredentialsDoNotMatch(),
 	}
 }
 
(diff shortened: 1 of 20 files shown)


aeneasr (changes requested):
Thank you! Unfortunately, merging this would introduce some quite serious security vulnerabilities which allows account take over:

1. Adam signs up with adam@example.org in Kratos
2. Eve creates account at e.g. Twitter with adam@example.org
3. Eve uses Twitter account to sign into Kratos, system things it's adam@example.org
4. Account is taken over

There are some ways to prevent this problem by, for example, requiring email verification but not all services support returning these fields.

Emails, in general, can not be used to cross reference authentication between completely separate systems.

To improve UX, the best way would be:

1. Sign up with adam@example.org and password
2. Adam tries to sign in with Google account adam@example.org
3. UI says: "You already have an account. Please sign in with your password, and link your Google account in the next step"

splaunov:
>3. Eve uses Twitter account to sign into Kratos, system things it's [adam@example.org](mailto:adam@example.org)
>4. Account is taken over

Eve will be requested to login with email/password first. So, takeover is not possible.
Security-vice it's the same flow as already implemented in kratos. The only difference is that the settings UI is skipped. Credentials are auto-merged if user logins with both credentials.

aeneasr:
I see! My bad, in that case, this looks like a very nice approach :) I'll try to get it running in the next couple of days locally

hperl:
Thank you for your contribution! I merged the latest changes into your branch and opened a new PR here: https://github.com/ory/kratos/pull/3563. I'll finish off the PR there.

aeneasr:
@splaunov thank you so much for this work! @hperl has done an amazing job of iterating over your work and we are happy to announce that the feature has landed on master!

PR #3235, opened 2023-04-13
feat: emit events for login/logout and registration

CaptainStandby:
## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


x/events/events.go
```diff
@@ -0,0 +1,49 @@
...
+func AttrFlowType(val flow.Type) otelattr.KeyValue {
+	return otelattr.String(attributeKeyFlowType.String(), string(val))
+}
+
+func AttrSessionID(val uuid.UUID) otelattr.KeyValue {
+	return otelattr.String(attributeKeySessionID.String(), val.String())
+}
+
+// Emit adds an event to the current span in the context.
+func Emit(ctx context.Context, event semconv.Event, opt ...otelattr.KeyValue) {
```
alnr, line 39: Please don't :) Just use the OpenTelemetry API.

CaptainStandby: I like the (partial) abstraction from OpenTelemetry which gives us the flexibility to replace the actual implementation.
Also the code gets really noisy and cluttered when using the OpenTelemetry API directly.
```go
	events.Emit(r.Context(), events.IdentityCreated,
		semconv.AttrIdentityID(i.ID),
		events.AttrFlowType(a.Type))
```
vs
```go
	trace.SpanFromContext(r.Context()).AddEvent(
		events.IdentityCreated.String(),
		trace.WithAttributes(
			append(
				semconv.AttributesFromContext(ctx),
				semconv.AttrIdentityID(i.ID),
				events.AttrFlowType(a.Type),
			)...,
		),
	)
```
I'd much prefer the former.

alnr: We already have the "flexibility to replace the actual implementation": see https://github.com/ory-corp/cloud/blob/dfe42f3e344f62495253212b117bde14c24e4020/cloudlib/analytics/tracer.go#L70.

`events.Emit` only introduces a layer of obfuscation around the fact that this is a OpenTelemetry span event, a well-known entity with a widely-used API.

Please don't wrap the OpenTelemetry API.

jonas-jonas: I get what @alnr is saying, but the default syntax is really verbose and IMO hard to read if you're not familiar with the otelx package. So I'd prefer the indirection through the Emit function.

jonas-jonas: But maybe we can move it to ory/x?

aeneasr: I have to side here with @alnr. The syntax is not so much different except for an append and a `semconv.AttributesFromContext` that we save. It would probably make more sense to have a helper for the attributes instead, for example.

In the end this triggers an otel event, so i think it should be clear that this is otel

zepatrik: I disagree with the argumentation because it is not just reducing clutter, but instead does an abstraction. One could, when writing this by hand, easily forget to add all attributes from the context. Nothing prevents you to write

```go
	trace.SpanFromContext(r.Context()).AddEvent(
		events.IdentityCreated.String(),
		trace.WithAttributes(
			semconv.AttrIdentityID(i.ID),
			events.AttrFlowType(a.Type),
		),
	)
```

instead of the code above. And I am pretty sure no one will notice that during review, as it looks totally legit.

In the end everyone will just copy-paste this whole code block and potentially forget to change the event type. That's what happens a lot with span names. If the API would be easy enough to just type it, and not require you to copy-paste because it is too much to remember, less errors will be made.

CaptainStandby:
I want to especially get feedback concerning the placement of the event emitting code.
I'm not 100% confident, that there are no edge cases in which an event could be missed or be emitted twice. The logical flow of the program is relative hard to follow in that regard.
I would love for @aeneasr and @zepatrik as most senior kratos devs to give this specific aspect a thorough look (thanks in advance 😄).

selfservice/flow/logout/handler.go
```diff
@@ -198,6 +202,16 @@ func (h *Handler) performNativeLogout(w http.ResponseWriter, r *http.Request, _
...
+	sess, err := h.d.SessionPersister().GetSessionByToken(r.Context(), p.SessionToken, session.ExpandNothing, identity.ExpandNothing)
+	if err != nil {
+		if errors.Is(err, sqlcon.ErrNoRows) {
+			h.d.Writer().WriteError(w, r, errors.WithStack(herodot.ErrForbidden.WithReason("The provided Ory Session Token could not be found, is invalid, or otherwise malformed.")))
+			return
+		}
+
+		h.d.Writer().WriteError(w, r, err)
+		return
+	}
```
CaptainStandby, line 214: Not 100% sure it's a good idea to load the session just so we can attach some metadata to the event (Line 226). Is there maybe a less invasive way of getting this data here?

selfservice/flow/logout/handler.go
```diff
@@ -209,6 +223,10 @@ func (h *Handler) performNativeLogout(w http.ResponseWriter, r *http.Request, _
 		return
 	}
 
+	events.Emit(r.Context(), events.SessionRevoked,
```
CaptainStandby, line 226: TODO: What about other ways of revoking a session?

selfservice/flow/logout/handler.go
```diff
@@ -286,6 +304,10 @@ func (h *Handler) updateLogoutFlow(w http.ResponseWriter, r *http.Request, ps ht
...
```
CaptainStandby, line 307: TODO: What about other ways of revoking a session?

selfservice/flow/registration/hook.go
```diff
@@ -169,15 +168,10 @@ func (e *HookExecutor) PostRegistrationHook(w http.ResponseWriter, r *http.Reque
...
-		semconv.EventIdentityCreated,
-		trace.WithAttributes(
-			attribute.String(semconv.AttrIdentityID, i.ID.String()),
-			attribute.String(semconv.AttrNID, i.NID.String()),
-			attribute.String(semconv.AttrClientIP, httpx.ClientIP(r)),
-			attribute.String("flow", string(a.Type)),
-		),
-	)
+
+	events.Emit(r.Context(), events.IdentityCreated,
```
CaptainStandby, line 172: TODO: What about other ways of creating an identity?

x/events/events.go
hperl: I wrote the function exactly for this reason, because I copy-pasted the code around and found that there was a common usage pattern that can be extracted into a simple function.

I follow this pattern a lot, actually. When I see a repetitive pattern popping up in (usually around three or more) places, I try to think about an abstraction. A wrapper function is really just about the easiest and most efficient abstraction. There are definitely abstractions that add a layer of obfuscation to the code (callbacks, because you can't step through them as easily in the debugger, or interfaces where you can't statically know which implementation is called), but *functions* are just about as cost-free as it gets. If at all, we should break our code more into smaller functions (and name and document them well).

In my opinion, this code is fine.

selfservice/flow/registration/hook.go
```diff
@@ -169,15 +168,10 @@ func (e *HookExecutor) PostRegistrationHook(w http.ResponseWriter, r *http.Reque
...
+		semconv.AttrIdentityID(i.ID),
+		events.AttrFlowType(a.Type))
 
 	s, err := session.NewActiveSession(r, i, e.d.Config(), time.Now().UTC(), ct, identity.AuthenticatorAssuranceLevel1)
```
CaptainStandby, line 176: Should we emit a `SessionIssued` Event here?

jonas-jonas: I think so, but actually in:
https://github.com/ory/kratos/blob/master/selfservice/hook/session_issuer.go, I think.

The session is created, but only returned if the session hook is set.

x/events/events.go
alnr: The `events.Emit` API has literally no connection to the OpenTelemetry API anymore, beyond using the slightly awkward `otel/attribute` package for no reason. Why use a custom OpenTelemetry tracer at that point at all?

The point of using the OpenTelemetry API for event emission is precisely to avoid implementing a homegrown leaky abstraction/event emission API and instead use an industry standard instead.

If we feel that the OpenTelemetry API is not a good fit after all, that's fine. But then please completely replace the mechanism instead of wrapping the existing API.

alnr: Recommended reading: https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction

selfservice/flow/login/hook.go
```diff
@@ -178,15 +177,11 @@ func (e *HookExecutor) PostLoginHook(w http.ResponseWriter, r *http.Request, g n
...
-			semconv.EventSessionIssued,
-			trace.WithAttributes(
-				attribute.String(semconv.AttrIdentityID, i.ID.String()),
-				attribute.String(semconv.AttrNID, i.NID.String()),
-				attribute.String(semconv.AttrClientIP, httpx.ClientIP(r)),
-				attribute.String("flow", string(flow.TypeAPI)),
-			),
-		)
+
+		events.Emit(r.Context(), events.SessionIssued,
```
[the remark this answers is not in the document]
aeneasr: Not sure if I follow - what webhook logic?

selfservice/flow/registration/hook.go
alnr: I believe all places where `SessionIssued` is required are already instrumented.

selfservice/flow/login/hook.go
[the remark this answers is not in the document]
CaptainStandby: I think the question is more of an observation that this file is problematic in regards to SRP. Improving this is out of scope for this PR, though.

We can probably disregard this question for now.

selfservice/flow/login/hook.go
```diff
@@ -207,15 +202,11 @@ func (e *HookExecutor) PostLoginHook(w http.ResponseWriter, r *http.Request, g n
...
-		semconv.EventSessionIssued,
-		trace.WithAttributes(
-			attribute.String(semconv.AttrIdentityID, i.ID.String()),
-			attribute.String(semconv.AttrNID, i.NID.String()),
-			attribute.String(semconv.AttrClientIP, httpx.ClientIP(r)),
-			attribute.String("flow", string(flow.TypeBrowser)),
-		),
-	)
+
+	events.Emit(r.Context(), events.SessionIssued,
```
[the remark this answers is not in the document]
alnr: The whole call graph is structured improperly in this part of Kratos. That's a problem for another day.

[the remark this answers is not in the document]
CaptainStandby: agreed

x/events/events.go
CaptainStandby: I refactored this a bit and would consider this discussion resolved

selfservice/flow/registration/hook.go
CaptainStandby: Added an event to the session issuer hook

selfservice/flow/registration/hook.go
CaptainStandby: --> #3241

selfservice/flow/logout/handler.go
CaptainStandby: --> #3241

selfservice/flow/logout/handler.go
CaptainStandby: --> #3241

selfservice/flow/logout/handler.go
zepatrik: All databases, except mysql ofc, support returning deleted rows. You could make the delete call return the session and in the case of mysql do a select before the delete.
Who expects performance when using mysql anyway? :sweat_smile:

x/events/events.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"context"
+	"reflect"
+
+	"github.com/gofrs/uuid"
+	otelattr "go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+
+	f "github.com/ory/kratos/selfservice/flow"
```
zepatrik, line 14: Hm, AFAIK the convention is to not have any dependencies to other kratos packages within x, as it is meant to be importable everywhere. I would suggest to move this into the flow package instead.

x/events/events.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/ory/x/otelx/semconv"
+)
+
+const (
+	SessionIssued      semconv.Event = "SessionIssued"
+	SessionRevoked     semconv.Event = "SessionRevoked"
+	RegistrationFailed semconv.Event = "RegistrationFailed"
+	IdentityCreated    semconv.Event = "IdentityCreated"
+	LoginFailed        semconv.Event = "LoginFailed"
```
zepatrik, line 24: Any reason to make these `semconv.Event`s instead of just strings?

selfservice/flow/logout/handler.go
alnr: I'd rather skip the session ID completely. It's not good for much after the session's been revoked anyway. If we find we're missing it in analysis later, we can add it then. But probably YAGNI.

x/events/events.go
alnr: Just inline the necessary interface:
```go
NewSessionIssued(ctx context.Context, sessionID, identityID uuid.UUID, flow interface{GetType() string}) (string, trace.EventOption) {
```

x/events/events.go
```diff
@@ -0,0 +1,112 @@
...
+func attrFlowType(val f.Type) otelattr.KeyValue {
+	return otelattr.String(attributeKeyFlowType.String(), string(val))
+}
+
+func attrSessionID(val uuid.UUID) otelattr.KeyValue {
+	return otelattr.String(attributeKeySessionID.String(), val.String())
+}
+
+func NewSessionIssued(ctx context.Context, sessionID, identityID uuid.UUID, flow f.Flow) (string, trace.EventOption) {
+	var flowType f.Type
```
alnr, line 41: This could be a helper util in this file.

selfservice/flow/registration/hook.go
alnr: Actually I might have lied. There are more instances where we issue sessions. For example during recovery flows. Those don't currently have audit logs either.

Could be covered in https://github.com/ory/kratos/issues/3241

selfservice/flow/logout/handler.go
CaptainStandby: I agree with the session ID, but the IdentityID could be interesting

x/events/events.go
CaptainStandby: I didn't know about this convention.

Inlining the interface doesn't work, because `flow.Flow.GetType()` returns `flow.Type` and not `string`.
I could instead just take the raw string instead of the whole `Flow` object or remove the attribute completely (YAGNI?).

WDYT?

x/events/events.go
CaptainStandby: Helps with type safety [here](https://github.com/ory-corp/cloud/blob/651ca4b708c65e38f71be9aded34a7aeb6ec6691/backoffice/backoffice/projectstats/projectstats.go#L169)

x/events/events.go
alnr: YAGNI that shit.

selfservice/flow/logout/handler.go
aeneasr: `ExpandNothing` on `GetSessionByToken` should be fast enough since it's a global, I think this is fine :) And for logout we should know the user id that is affected!

aeneasr (approved):
design-wise this LGTM

selfservice/flow/logout/handler.go
CaptainStandby: 👍

aeneasr (approved):

PR #3239, opened 2023-04-18
chore: add launch config for VSCode

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


.vscode/launch.json
```diff
@@ -0,0 +1,34 @@
+{
+  // Launch configuration to build and launch Kratos
```
aeneasr, line 2: does this parse?

jonas-jonas: yep, VS Code launch configs allow comments.

Makefile
```diff
@@ -16,7 +16,8 @@ GO_DEPENDENCIES = github.com/ory/go-acc \
 				  github.com/go-swagger/go-swagger/cmd/swagger \
 				  golang.org/x/tools/cmd/goimports \
 				  github.com/mattn/goveralls \
-				  github.com/cortesi/modd/cmd/modd
+				  github.com/cortesi/modd/cmd/modd \
+				  github.com/mailhog/MailHog
```
alnr, line 20: Maybe a good idea to add this to `go_mod_indirect_pins.go`.

alnr (approved):

Makefile
jonas-jonas: Can do, could you explain, what that the point of that file is? I don't quite understand it.

alnr: This pattern is used to manage/pin versions of tools the repo uses (like code generators, coverage calculators, testing utilities like MailHog) using the `go.mod` file of the actual Go module -- kratos in this case.

`go_mod_indirect_pins.go` is excluded from compilation via build tags which are never set. In fact, that file cannot be compiled at all. But it is part of the dependency resolution, and it's dependencies are noted down in `go.mod`. It serves only to import (as `_`) all the tools required in that repo, which are themselves written in Go.

jonas-jonas: ah, I see. Thanks, I'll add Mailhog there as well.

alnr (approved):

Issue #2832, opened 2022-10-24
Recovery flow is not redirecting user to settings page when 2FA is enabled


mmeller-wikia:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

Recovery flow is inconsistent to password users and 2FA users.
Recovery flow is not redirecting user to settings page when 2FA is enabled

### Reproducing the bug

1. Configure AuthenticatorAssuranceLevel to use `HighestAvailableAAL`
2. Add 2FA to your account
3. Logout
4. Request recovery link
5. Click recovery link -> you are redirected to 2-step login page
6. Complete login with 2FA
7. You are redirected to default return url configured in Kratos (expected: redirect to settings flow and recovery message is displayed - like with only password protected accounts)

### Relevant log output

Not relevant

### Relevant configuration

```yml
selfservice:
  flows:
    settings:
      required_aal: highest_available
```

### Version

v0.10.1, current master

### On which operating system are you observing this issue?

Not relevant

### In which environment are you deploying?

Not relevant

### Additional Context

It is not possible to create suitable workaround on frontend (at least recovery message is lost).

mmeller-wikia:
WIP PR https://github.com/ory/kratos/pull/2831

PR #3263, opened 2023-05-02
fix: settings should persist `return_to` after required mfa login flow

Benehiko:
## Related issue(s)

fixes https://github.com/ory/network/issues/222 
fixes https://github.com/ory/kratos/issues/2832

This PR fixes the incorrect behavior after completing the required login (MFA) required by the settings flow. Users would incorrectly be redirected to the `/welcome` page or default UI route after submitting a successful MFA credential instead of being redirected back to the original settings flow.

**possible breaking change**
Observed through the E2E tests, if the host from which the request to Kratos is proxied and the `serve.public.base_url` is set to point to the proxy URL, then Kratos will use the current host which is the Kratos URL. MFA required flows such as settings will attempt to redirect back to the Kratos Host instead of the proxy URL which is not automatically on the allow list for browser redirects - especially since our `serve.public.base_url` now points to the proxy URL instead of Kratos. This will cause the flow to fail and the user to be redirected to the `/error` page.

To avoid this, allow the proxy to forward the host through the `X-Forwarded-Host` header. Kratos will then correctly redirect back.

Before Merge:
- [x] Merge fix for UI https://github.com/ory/kratos-selfservice-ui-node/pull/256
- [x] Merge fix for React UI https://github.com/ory/kratos-selfservice-ui-react-nextjs/pull/58
- [x] Merge fix for UI https://github.com/ory/kratos-selfservice-ui-node/pull/262
- [x] Add more E2E tests 
  - [x] ~navigate to settings immediately after password login with MFA required Identity.~ We have such a test here https://github.com/ory/kratos/blob/master/test/e2e/cypress/integration/profiles/mfa/totp.spec.ts#L48
  

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


test/e2e/proxy/proxy.js
```diff
@@ -1,17 +1,20 @@
...
+import fs from "fs"
 
 const app = express()
 
 const proxy =
   (base, prefix = null) =>
   (req, res, next) => {
     let url = urljoin(base, req.url)
+    // we need to tell Krato we are behind a reverse proxy
+    res.set("X-Forwarded-Host", url)
```
Benehiko, line 17: forward the host to kratos

selfservice/flow/settings/handler.go
```diff
@@ -406,7 +410,13 @@ func (h *Handler) fetchFlow(w http.ResponseWriter, r *http.Request) error {
 		return errors.WithStack(herodot.ErrForbidden.WithID(text.ErrIDInitiatedBySomeoneElse).WithReasonf("The request was made for another identity and has been blocked for security reasons."))
 	}
 
-	if err := h.d.SessionManager().DoesSessionSatisfy(r, sess, h.d.Config().SelfServiceSettingsRequiredAAL(r.Context())); err != nil {
+	// we cannot redirect back to the settings flow
+	// because it would just return a json response which is not what we want
+	requestURL := h.d.Config().SelfServiceFlowSettingsUI(r.Context())
```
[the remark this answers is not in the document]
jonas-jonas: How does this work with JSON requests?

test/e2e/cypress/integration/profiles/recovery/return-to/success.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+context("Recovery with `return_to`", () => {
+  ;[
+    //{
+    //  recovery: react.recovery,
+    //  settings: react.settings,
+    //  base: react.base,
+    //  app: "react" as "react",
+    //  profile: "spa",
+    //},
```
jonas-jonas, line 19: What about the SPAs?

Benehiko: Forgot to re-enable it for spa. But this will most likely need to be implemented in SPAs (from the spa side). The payload will return a Json with a redirect url. It remains up to the client how they want to deal with it.

selfservice/flow/settings/handler.go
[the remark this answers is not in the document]
Benehiko: with JSON (spa's) the logic works a bit different. This shouldn't be a problem though as we have e2e tests for it. 

The logic follows the normal pattern of redirecting based on the response payload  as shown here: https://github.com/ory/kratos-selfservice-ui-react-nextjs/pull/58/files#diff-1394afc6e5c7c0f2f2215859eb127c7d9e4cca731d34faa309b8da648c2782b5R101-R103

test/e2e/profiles/kratos.base.yml
```diff
@@ -25,9 +25,11 @@ selfservice:
 
 serve:
   public:
-    # We are using a proxy to make cypress not fail with different origins which is why we use this URL here.
+    # We are using a proxy to make cypress not fail with different origins.
+    # the proxy sets the X-FORWARDED-HOST header which allows the base_url here to be set to
+    # the kratos instance
     # See ../proxy
-    base_url: http://localhost:4455/
+    base_url: http://localhost:4433/
```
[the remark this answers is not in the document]
Benehiko: nvm, just had a lot of cases where the URL wouldn't be crafted correctly e.g. /schemas would be under the kratos public url (localhost:4433) instead of the proxy url (localhost:4455). Reverted this change in the config but still kept the `x-forwarded-host` headers from the proxy side.

session/manager_http.go
```diff
@@ -313,8 +332,14 @@ func (s *ManagerHTTP) DoesSessionSatisfy(r *http.Request, sess *Session, request
 			return nil
 		}
 
-		return NewErrAALNotSatisfied(
-			urlx.CopyWithQuery(urlx.AppendPaths(s.r.Config().SelfPublicURL(ctx), "/self-service/login/browser"), url.Values{"aal": {"aal2"}}).String())
+		loginURL := urlx.CopyWithQuery(urlx.AppendPaths(s.r.Config().SelfPublicURL(ctx), "/self-service/login/browser"), url.Values{"aal": {"aal2"}})
+
+		// return to the requestURL if it was set
+		if managerOpts.requestURL != "" {
+			loginURL = urlx.CopyWithQuery(loginURL, url.Values{"return_to": {managerOpts.requestURL}})
```
aeneasr, line 339: Consider adding a test in `TestDoesSessionSatisfy` to reduce the chance of regressions

selfservice/strategy/code/strategy_recovery_test.go
```diff
@@ -526,13 +560,17 @@ func TestRecovery(t *testing.T) {
...
-
 					res, err = client.Get(public.URL + session.RouteWhoami)
 					require.NoError(t, err)
 					body = string(x.MustReadAll(res.Body))
 					require.NoError(t, res.Body.Close())
-					assert.Equal(t, "code_recovery", gjson.Get(body, "authentication_methods.0.method").String(), "%s", body)
-					assert.Equal(t, "aal1", gjson.Get(body, "authenticator_assurance_level").String(), "%s", body)
+
+					if tc.expectedAAL == "aal2" {
+						assert.Equal(t, http.StatusUnauthorized, res.StatusCode, "%s", body)
```
aeneasr, line 569: Why are we not getting the session result here but the error instead?

test/e2e/profiles/recovery-mfa/.kratos.yml
```diff
@@ -0,0 +1,47 @@
+selfservice:
```
aeneasr, line 1: For future record: unless you significantly need to alter a profile it's faster to use some default profile and change the config settings you need. Profiles can be slow to start up and introduce unnecessary duplication (like the identity schema)

selfservice/flow/login/handler.go
```diff
@@ -111,6 +111,12 @@ func WithFormErrorMessage(messages []text.Message) FlowOption {
 	}
 }
 
+func WithAuthenticatorAssuranceLevel(auth identity.AuthenticatorAssuranceLevel) FlowOption {
```
aeneasr, line 114: Unused

selfservice/flow/settings/error.go
```diff
@@ -163,9 +166,13 @@ func (s *ErrorHandler) WriteFlowError(
 		if shouldRespondWithJSON {
 			s.d.Writer().WriteError(w, r, aalErr)
 		} else {
-			http.Redirect(w, r, urlx.CopyWithQuery(
-				urlx.AppendPaths(s.d.Config().SelfPublicURL(r.Context()), login.RouteInitBrowserFlow),
-				url.Values{"aal": {string(identity.AuthenticatorAssuranceLevel2)}}).String(), http.StatusSeeOther)
+			if aalErr.RedirectTo == "" {
```
aeneasr, line 169: I could not find a place (except for in some tests) where this value is empty. Do we need this code change?

selfservice/flow/settings/handler.go
```diff
@@ -298,7 +301,13 @@ func (h *Handler) createBrowserSettingsFlow(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	if err := h.d.SessionManager().DoesSessionSatisfy(r, s, h.d.Config().SelfServiceSettingsRequiredAAL(r.Context())); err != nil {
+	var managerOptions []session.ManagerOptions
+	requestURL := x.RequestURL(r)
+	if requestURL.Query().Get("return_to") != "" {
+		managerOptions = append(managerOptions, session.WithRequestURL(requestURL.String()))
```
aeneasr, line 307: This is where the magic is happening right? The URL is basically:

```
/settings/browser?return_to=/login_browser?return_to=https://source"
```

selfservice/flow/settings/handler.go
```diff
@@ -406,7 +415,13 @@ func (h *Handler) fetchFlow(w http.ResponseWriter, r *http.Request) error {
 		return errors.WithStack(herodot.ErrForbidden.WithID(text.ErrIDInitiatedBySomeoneElse).WithReasonf("The request was made for another identity and has been blocked for security reasons."))
 	}
 
-	if err := h.d.SessionManager().DoesSessionSatisfy(r, sess, h.d.Config().SelfServiceSettingsRequiredAAL(r.Context())); err != nil {
+	// we cannot redirect back to the settings flow
```
aeneasr, line 418: I'm not sure if I understand what the comment is saying. Could you clarify? 

If I understand:

1. generate url `/ui/settings?flow=...`
2. make DoesSessionSatisfy include it in the return URL `/ui/login?return_to=/ui/settings?flow=...`

selfservice/flow/settings/handler.go
```diff
@@ -406,7 +415,13 @@ func (h *Handler) fetchFlow(w http.ResponseWriter, r *http.Request) error {
...
+	// because it would just return a json response which is not what we want
+	requestURL := h.d.Config().SelfServiceFlowSettingsUI(r.Context())
+	query := requestURL.Query()
+	query.Set("flow", rid.String())
+	requestURL.RawQuery = query.Encode()
```
aeneasr, line 423: We have helpers for that:

`urlx.CopyWithQuery`

aeneasr (commented):
Nice! Just some comments

selfservice/flow/settings/handler.go
Benehiko: Yes exactly, since the getSettingsFlow always returns JSON, you cannot redirect to it. 

I will adjust the comments to make it more apparent

test/e2e/profiles/recovery-mfa/.kratos.yml
Benehiko: i see, i think we need to actually refactor how we do e2e tests in general. But i will keep this in mind :+1:

selfservice/flow/settings/handler.go
Benehiko: it's in multiple places in the settings handler. 

Since we have a create, fetch and update flow, all three need to pass along a request URL to the `DoesSessionSatisfy` method which then appends the request URL to the `/login?aal=aal2` URL. 

The true magic is here:
https://github.com/ory/kratos/pull/3263/files#diff-a50a45cf2bc8e265082b551b7fb148ba52b47e6e6e1e0266d388bbd7955e20ccR338

selfservice/flow/settings/error.go
Benehiko: nice catch! I think this was code added at some point while testing this and forgot to remove it.

selfservice/flow/settings/error.go
```diff
@@ -163,9 +165,7 @@ func (s *ErrorHandler) WriteFlowError(
 		if shouldRespondWithJSON {
 			s.d.Writer().WriteError(w, r, aalErr)
 		} else {
-			http.Redirect(w, r, urlx.CopyWithQuery(
-				urlx.AppendPaths(s.d.Config().SelfPublicURL(r.Context()), login.RouteInitBrowserFlow),
-				url.Values{"aal": {string(identity.AuthenticatorAssuranceLevel2)}}).String(), http.StatusSeeOther)
+			http.Redirect(w, r, aalErr.RedirectTo, http.StatusSeeOther)
```
zepatrik, line 168: Doesn't this significantly change the current behavior? Before, a user would be taken to the login screen to increase the AAL, now they just get redirected to some other page and the settings flow is not completed.

zepatrik: Ah I see now, the error carries the `return_url`, which is the login URL, right? Is it always the login URL, or are there cases were it is different? Or do we just pipe through the original return URL?

Benehiko: The `RedirectTo` URL is set by the `DoesSessionSatisfy` method which sets it to the login route and if applicable pipes the request_url into the `return_to` query parameter

Benehiko: Please take a look at this part of the code https://github.com/ory/kratos/pull/3263/files#diff-a50a45cf2bc8e265082b551b7fb148ba52b47e6e6e1e0266d388bbd7955e20ccR335

that is where the "magic" happens

aeneasr (approved):
LGTM

PR #3264, opened 2023-05-03
feat: add RecoveryURL to recovery code email template data

kszafran:
In some cases we may want to include a link to the recovery UI page (including the flow ID) in the recovery email. That may be useful if the user has closed the page, for example. (In my particular case recovery is triggered by a backend process, so when the user receives the email, they are not on the recovery page and there is no way to pass on the flow ID to them.)

The link added to the template data as `.RecoveryURL` is the exact same link as the one generated by `POST /admin/recovery/code`.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/courier/email_templates_test.go
+++ b/courier/email_templates_test.go
@@ -43,7 +43,7 @@ func TestNewEmailTemplateFromMessage(t *testing.T) {
 	for tmplType, expectedTmpl := range map[courier.TemplateType]courier.EmailTemplate{
 		courier.TypeRecoveryInvalid:         email.NewRecoveryInvalid(reg, &email.RecoveryInvalidModel{To: "foo"}),
 		courier.TypeRecoveryValid:           email.NewRecoveryValid(reg, &email.RecoveryValidModel{To: "bar", RecoveryURL: "http://foo.bar"}),
-		courier.TypeRecoveryCodeValid:       email.NewRecoveryCodeValid(reg, &email.RecoveryCodeValidModel{To: "bar", RecoveryCode: "12345678"}),
+		courier.TypeRecoveryCodeValid:       email.NewRecoveryCodeValid(reg, &email.RecoveryCodeValidModel{To: "bar", RecoveryURL: "http://foo.bar", RecoveryCode: "12345678"}),
 		courier.TypeRecoveryCodeInvalid:     email.NewRecoveryCodeInvalid(reg, &email.RecoveryCodeInvalidModel{To: "bar"}),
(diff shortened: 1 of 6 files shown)


rostyk110 (approved):

jonas-jonas (changes requested):
Thank you for the contribution. We left out the URL intentionally, because the flow is protected against CSRF attacks using a CSRF-token (in cookies and the form fields). 

If you include the link in the email, the flow might open in a different browser, etc., invalidating the flow entirely. Now the user needs to re-do the whole flow, which is not great UX. The account recovery is a crucial part of the account's security and as such needs to be protected well. So I don't think it's a good idea to send links containing such sensitive information to a user.

kszafran:
Hm... are CSRF cookies used here or just the CSRF token in the form fields? If it's the latter, then why would the flow get invalidated? Perhaps I need to test it once more to make sure that this is not the case.

jonas-jonas:
Both a cookie and token are set as part of the flow. You should be able to test this by copying the URL of a flow after submitting the email address into a different browser, and then try to submit the code from the email.

kszafran:
Right, I tested it and I didn't even get to enter the code, because the recovery page got a CSRF error when calling `/self-service/recovery/flows?id=...`... 🤦🏻 

Our whole problem is that we're using the recovery mechanism to fake email invites. I saw a couple of threads regarding the support of invite flows, do you know if there's anyone actively working on it?

kszafran:
Hm... now I see that `POST /admin/recovery/code` works by explicitly disabling CSRF protection. I wonder if that could be made configurable for the recovery flow in general (I'm not quite sure why we even need CSRF cookies on `GET` requests that don't mutate any state). Or perhaps there could be an admin API that instead of fetching the link and the code triggers the email 🤔

kszafran:
I'm basically following https://www.ory.sh/docs/kratos/manage-identities/invite-users except instead of sending emails myself I'm using Kratos' own mechanism. That use case would also be skipping CSRF checks. (I'm not sure to what extent `flow.DangerousSkipCSRFCheck` skips the checks. Here it seems that we can still verify form tokens, just not require cookie tokens on `GET`.)

kmherrmann:
>  I saw a couple of threads regarding the support of invite flows, do you know if there's anyone actively working on it?
It's on the roadmap for the 2nd half of this year, but we haven't gotten to detailed timeline planning. Contributions welcome :)

Issue #2628, opened 2022-07-29
I'm unable to reset password on mobile devices


Pruspat:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

I'm not able to finish account recovery on mobile devices?

1. I'm calling GET recovery/api
2. then in the POST request I pass the email address and method "link"
3. I receive an email message with a URL like: "http://127.0.0.1:4433/self-service/recovery?flow=91b83275-f306-463f-92b3-9052cae8ec23&token=Z5ehZKrk7sWaFBRxi8aI5xDlgYViXaWO"

In a web application, if you will GET the URL below, you will receive cookies which makes it possible to init settings and then change the password, but I'm unable to receive a session token for mobile devices to init setting and then change my password

### Reproducing the bug

Run steps that are above

### Version

oryd/kratos v0.8.2-alpha.1 and oryd/kratos:v0.10.0

### On which operating system are you observing this issue?

Linux

### In which environment are you deploying?

Docker Compose

aeneasr:
Hi, thank you for the report! That is indeed the case - there are some open PRs for this but in general we have plans to switch to a code which people will enter in their app, making this process mobile friendly!

For now, your users will have to change the password in the browser :/

kmherrmann:
@jonas-jonas JFYI as you're drilling into API flows

sobanieca-redocly:
Hello, I just want to point out that I've run into issue where I'm missing this functionality. I'm creating recovery flow via Native Apps endpoint, but I got stuck when I've validated recovery code, as in result I'm just getting 422 response with cookies for browser redirect to configure new password. It would be awesome if for native apps flow we would receive session token when proper recovery code was sent and using it, we could call settings flow endpoint to set a new password.

kmherrmann:
[...]

PR #3273, opened 2023-05-09
feat: add support for recovery on native flows

jonas-jonas:
Adds the ability to complete the recovery flow properly on API flows. This PR also streamlines the behavior for SPA flows to not return 422 errors anymore. To enable this new behavior, set the `features.use_continue_with_transitions` flag in the config to `true`.

Fixes https://github.com/ory/kratos/issues/2628

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


contrib/quickstart/kratos/email-password/kratos.yml
```diff
@@ -14,6 +14,7 @@ selfservice:
   default_browser_return_url: http://127.0.0.1:4455/
   allowed_return_urls:
     - http://127.0.0.1:4455
+    - http://localhost:4457/Callback
```
jonas-jonas, line 17:
```suggestion
    - http://localhost:19006/Callback
```

selfservice/flow/recovery/flow.go
```diff
@@ -100,6 +100,8 @@ type Flow struct {
 	// This is needed, because we can not enforce these measures, if the flow has been initialized by someone else than
 	// the user.
 	DangerousSkipCSRFCheck bool `json:"-" faker:"-" db:"skip_csrf_check"`
+
+	ContinueWith []flow.ContinueWith `json:"continue_with,omitempty" faker:"-" db:"-"`
```
jonas-jonas, line 104:
```suggestion
    // Contains possible actions that could follow this flow
	ContinueWith []flow.ContinueWith `json:"continue_with,omitempty" faker:"-" db:"-"`
```

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -377,9 +183,17 @@ func (s *Strategy) recoveryIssueSession(w http.ResponseWriter, r *http.Request,
 		return s.retryRecoveryFlowWithError(w, r, f.Type, err)
 	}
 
-	// TODO: How does this work with Mobile?
-	if err := s.deps.SessionManager().UpsertAndIssueCookie(ctx, w, r, sess); err != nil {
-		return s.retryRecoveryFlowWithError(w, r, f.Type, err)
+	switch {
+	case f.Type == flow.TypeBrowser:
+		// TODO: How does this work with Mobile?
```
jonas-jonas, line 188:
```suggestion
```

It does now! lol

jonas-jonas (commented):
This definitely still needs e2e tests (either PW or cypress), for all the cases. Since the react-native PR got merged, that should be quite easy to do.

.github/workflows/ci.yaml
```diff
@@ -175,6 +175,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: ory/kratos-selfservice-ui-react-native
+          ref: jonas-jonas/improveRecovery
```
jonas-jonas, line 178: Remove once https://github.com/ory/kratos-selfservice-ui-react-native/pull/81 is merged

selfservice/flow/recovery/error.go
```diff
@@ -89,33 +89,36 @@ func (s *ErrorHandler) WriteFlowError(
...
+			s.forward(w, r, newFlow, err)
 			return
 		}
 
-		// We need to use the new flow, as that flow will be a browser flow. Bug fix for:
-		//
-		// https://github.com/ory/kratos/issues/2049!!
-		if a.Type == flow.TypeAPI || x.IsJSONRequest(r) {
-			http.Redirect(w, r, urlx.CopyWithQuery(urlx.AppendPaths(s.d.Config().SelfPublicURL(r.Context()),
-				RouteGetFlow), url.Values{"id": {a.ID.String()}}).String(), http.StatusSeeOther)
```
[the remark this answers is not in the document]
jonas-jonas: After some thought, I don't think there is anyone hitting this endpoint in production using API flows, as there is just no way to continue with the session cookie.

selfservice/flow/recovery/handler.go
```diff
@@ -116,9 +116,9 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
 //	Schemes: http, https
 //
 //	Responses:
-//	  200: recoveryFlow
-//	  400: errorGeneric
-//	  default: errorGeneric
+//		200: recoveryFlow
+//		400: errorGeneric
+//		default: errorGeneric
```
Benehiko, line 121: not sure about this indentation. It doesn't look like the others we have :suspect: 

```suggestion
//    200: recoveryFlow
//    400: errorGeneric
//    default: errorGeneric
```

selfservice/strategy/code/strategy_recovery_admin.go
```diff
@@ -0,0 +1,216 @@
...
+//	Schemes: http, https
+//
+//	Security:
+//		oryAccessToken:
+//
+//	Responses:
+//		201: recoveryCodeForIdentity
+//		400: errorGeneric
+//		404: errorGeneric
+//		default: errorGeneric
```
Benehiko, line 128: same here. this indentation doesn't look correct

Benehiko (changes requested):

selfservice/flow/recovery/handler.go
jonas-jonas: Done, but the formatter turns it into tab + 2 spaces, which is what we use on the other places. But mixing indentation.. 👎🏼

selfservice/flow/recovery/handler.go
```diff
@@ -103,7 +103,7 @@ func (h *Handler) RegisterAdminRoutes(admin *x.RouterAdmin) {
 //
 // If a valid provided session cookie or session token is provided, a 400 Bad Request error.
 //
-// To fetch an existing recovery flow call `/self-service/recovery/flows?flow=<flow_id>`.
+// If you already created a recovery, fetch the flow's information using the getRecoveryFlow API endpoint.
```
Benehiko, line 106:
```suggestion
// On an existing recovery flow, use the `getRecoveryFlow` API endpoint.
```

Benehiko (changes requested):

Benehiko:
This looks good and can be merged soon

test/e2e/playwright/tests/app_recovery.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    await test.step("enter correct code", async () => {
+      await page.getByTestId("code").fill(code)
+      await page.getByText("Submit").click()
+      await page.waitForURL(/Settings/)
+      await expect(page.getByTestId("ui/message/1060001").first()).toBeVisible()
+    })
+  })
+
+  test("wrong email address does not get sent", async ({ page, identity }) => {
```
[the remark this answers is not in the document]
jonas-jonas: Yes, that's the default.

Benehiko (approved):
LGTM :)

selfservice/flow/recovery/error.go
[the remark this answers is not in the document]
aeneasr: Wouldn't this affect customers who are using the `link` strategy with api flows?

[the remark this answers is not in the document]
aeneasr: I think I misread what we're doing here. This is actually for expiry correct? In that case I think we would have a breaking change.

Before:

1. Submit recovery flow
2. For some reason expired
3. Create new recovery flow
4. HTTP redirect to that flow - which returns JSON (like submit would too)
5. The UI renders JSON

Now:

1. Submit recovery flow
2. For some reason expired
3. Create new recovery flow
4. Parse continue with
5. Call endpoint, get JSON
6. The UI renders JSON

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -394,19 +208,23 @@ func (s *Strategy) recoveryIssueSession(w http.ResponseWriter, r *http.Request,
...
 	sf.UI.Messages.Set(text.NewRecoverySuccessful(time.Now().Add(config.SelfServiceFlowSettingsPrivilegedSessionMaxAge(ctx))))
 	if err := s.deps.SettingsFlowPersister().UpdateSettingsFlow(r.Context(), sf); err != nil {
-		return s.retryRecoveryFlowWithError(w, r, f.Type, err)
+		return s.retryRecoveryFlow(w, r, f.Type, RetryWithError(err))
 	}
 
-	if x.IsJSONRequest(r) {
+	switch {
+	case f.Type.IsAPI():
+		f.ContinueWith = append(f.ContinueWith, flow.NewContinueWithSettingsUI(sf))
```
aeneasr, line 223: This would also be a breaking change right? Because the payload has changed and thus the client code needs to be adapted

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -444,63 +262,74 @@ func (s *Strategy) recoveryUseCode(w http.ResponseWriter, r *http.Request, body
...
-		http.Redirect(w, r, urlx.CopyWithQuery(urlx.AppendPaths(config.SelfPublicURL(ctx),
-			recovery.RouteGetFlow), url.Values{"id": {f.ID.String()}}).String(), http.StatusSeeOther)
-	} else {
+	switch {
+	case f.Type.IsAPI():
+		rErr := new(herodot.DefaultError)
+		if !errors.As(retryOptions.err, &rErr) {
+			rErr = rErr.WithError(retryOptions.err.Error())
+		}
+		s.deps.Writer().WriteError(w, r, flow.ErrorWithContinueWith(rErr, flow.NewContinueWithRecoveryUI(f)))
```
aeneasr, line 326: Also a breaking change

aeneasr (commented):
Thank you! I think the functionality in this PR is great! I do think that the problem right now is that we have a mix of `continue_with` and `303` redirects. We use `continue_with` in native APIs and 303 in AJAX requests.

I think that we should commit to exactly one method of "go to the next state" and if we need to address that in this PR, consider to just commit to `continue_with`.

That would mean that we probably need a feature flag to govern whether to use the old redirect based approach, or the new `continue_with` approach.

Something like:

```
selfservice:
  flows:
    disable_legacy_transition: true # new state transition model is used
```

and then annouce it:

1. Via the changelog and that we will deprecate the old flow in January 2024
2. Add a page to the docs "Upcoming deprecations" with a timeline

WDYT?

selfservice/strategy/code/strategy_recovery.go
aeneasr, line 223: Breaking

jonas-jonas: >I am guessing, there is actually no one hitting this code path with API flows in prod right now, because there is no way to continue with the settings flow after this.

> We've had a few questions around this, but AFAIR, we always ended on "please use the browser flows" instead, as there is (actually was, since you can tokenize a cookie to a token now) no way to get a session token from this response.

https://github.com/ory/kratos/pull/3273/files/058dd0104e34e9b50b2711b6f595bd6173da8f72#r1333047866

Technically, this is a breaking change, yes. But there is likely no one hitting this path right now. 

However, just to be safe, we should probably stick the new behavior behind a feature flag. Agreed.

Benehiko (changes requested):
Just small nits I could see. I think in general the PR looks great!

selfservice/strategy/code/strategy_recovery_admin.go
```diff
@@ -0,0 +1,216 @@
...
+//
+//	Security:
+//		oryAccessToken:
+//
+//	Responses:
+//	  201: recoveryCodeForIdentity
+//	  400: errorGeneric
+//	  404: errorGeneric
+//	  default: errorGeneric
+func (s *Strategy) createRecoveryCodeForIdentity(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
```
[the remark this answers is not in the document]
jonas-jonas: Yes, but this is the HTTP handler method for creating a code via the admin API.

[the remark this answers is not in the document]
Benehiko: yeah all the more reason to trace it :)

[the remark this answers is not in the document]
Benehiko: I don't think it can hurt to have a span here with the name of the function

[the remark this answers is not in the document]
jonas-jonas: But... we don't have spans on the other handlers? https://github.com/ory/kratos/blob/afed81d6e75f3d5c7060d0938bff9eb528b11215/selfservice/flow/login/handler.go#L327

They get added automatically, by the tracing library, no? 
image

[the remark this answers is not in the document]
alnr: The top-level HTTP handlers such as this one are automatically instrumented. We probably don't need anything else.

Benehiko (approved):
LGTM :)

embedx/config.schema.json
```diff
@@ -2620,6 +2617,12 @@
           "title": "Enable Ory Sessions caching",
           "description": "If enabled allows Ory Sessions to be cached. Only effective in the Ory Network.",
           "default": false
+        },
+        "new_flow_transitions": {
```
hperl, line 2621: Should we call this `use_continue_with_transitions` or similar? `new_flow_transitions` is a bit ambiguous.

hperl (approved):
LGTM! I just have a small suggestion regarding the naming of the feature flag.

Also, IIRC, the DX of the 422 errors is not really great in all SDKs. For Java, it requires explicitly overriding some error factory to extract the error. Is this really the best way? We should probably discuss this elsewhere.

jonas-jonas:
>Also, IIRC, the DX of the 422 errors is not really great in all SDKs. For Java, it requires explicitly overriding some error factory to extract the error. Is this really the best way? We should probably discuss this elsewhere.

Yes, you're right, but I am not sure what error code we could use instead. A 200 OK response would still be okay, but it doesn't convey the "urgency" that there is another step required here.

embedx/config.schema.json
jonas-jonas: Done.

selfservice/strategy/code/strategy_recovery.go
```diff
@@ -394,20 +208,33 @@ func (s *Strategy) recoveryIssueSession(w http.ResponseWriter, r *http.Request,
...
-		s.deps.Writer().WriteError(w, r, flow.NewBrowserLocationChangeRequiredError(sf.AppendTo(s.deps.Config().SelfServiceFlowSettingsUI(r.Context())).String()))
+		return s.retryRecoveryFlow(w, r, f.Type, RetryWithError(err))
+	}
+
+	if s.deps.Config().UseContinueWithTransitions(ctx) {
+		switch {
+		case f.Type.IsAPI():
+			fallthrough
+		case x.IsJSONRequest(r):
+			f.ContinueWith = append(f.ContinueWith, flow.NewContinueWithSettingsUI(sf))
```
jonas-jonas, line 226: @aeneasr since you now need to enable the continue_with explicitely we now return the flow for both API and SPA flows to align their behavior. As @hperl request. LMKWYT.

.github/workflows/ci.yaml
aeneasr: Done

selfservice/flow/recovery/error.go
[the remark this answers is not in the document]
aeneasr: @jonas-jonas has this been addressed?

selfservice/flow/recovery/flow.go
```diff
@@ -147,7 +150,7 @@ func NewFlow(conf *config.Config, exp time.Duration, csrf string, r *http.Reques
 func FromOldFlow(conf *config.Config, exp time.Duration, csrf string, r *http.Request, strategy Strategy, of Flow) (*Flow, error) {
 	f := of.Type
 	// Using the same flow in the recovery/verification context can lead to using API flow in a verification/recovery email
-	if of.Type == flow.TypeAPI {
+	if of.Type == flow.TypeAPI && of.Active.String() == string(RecoveryStrategyLink) {
```
aeneasr, line 153: nit: you probably don't need to convert both to string

jonas-jonas: `of.Active` is a `sqlxx.NullString`: 

image

selfservice/strategy/code/strategy_recovery.go
jonas-jonas: I put this behind the feature flag now, as discussed in the other conversations.

selfservice/strategy/code/strategy_recovery.go
jonas-jonas: Also behind the feature flag now.

aeneasr:
What changes are needed in the docs for this?

test/e2e/playwright/fixtures/index.ts
```diff
@@ -0,0 +1,64 @@
...
+  config: [
+    async ({ request, configOverride }, use) => {
+      const configToWrite = merge(default_config, configOverride)
+
+      const resp = await request.get("http://localhost:4434/health/config")
+
+      const configRevision = await resp.body()
+
+      await writeFile(
+        "playwright/kratos.config.json",
```
aeneasr, line 38: This isn't thread safe - meaning that we can't execute playwright tests in parallel. Is parallelization disabled in playwright?

aeneasr (approved):

test/e2e/playwright/fixtures/index.ts
jonas-jonas: Yes - no parallelization, there is no way to make this thread safe with the current config model of Kratos.

jonas-jonas:
> What changes are needed in the docs for this?

I'll need to figure out how to document this properly. Since we now use the feature flag, and we'll need a UI in the console for it, I'll first try to get this into the network and then do the docs changes, if that's okay with you. @aeneasr

aeneasr:
Can you please add an entry for the changelog here?

PR #3274, opened 2023-05-10
fix: keep loginChallenge when going from login to registration

NilsBarlaug:
Hi, thanks for the amazing service you are developing!

This could have been an issue, but I find it easier to convey together with a potential fix.

The problem I'm facing: I've set up Kratos with Hydra integration. When a user logs in the first time they automatically proceed to the registration flow and become registered. However, despite having set up the session hook, they do not get logged in. #3212 makes sure Kratos redirect back to Hydra, but Hydra just redirects back to Kratos again because the the login request was never accepted.

Looking into the code I see that the registration flow will accept the login request if the loginChallenge is present. Should the login challenge have been passed on from the login flow to the registration flow? If so, this is a suggested fix.

Or am I misunderstanding how I'm supposed to get this integration set up correctly? I appreciate any clearification 🙂 

## Related issue(s)

#3212

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

I tried making a test to prove it's working, but found it a bit difficult - so I would need a few pointers to get started on that 🙂 However, I have built it and checked that it solves my problem.

### DIFF
--- a/selfservice/strategy/oidc/strategy_login.go
+++ b/selfservice/strategy/oidc/strategy_login.go
@@ -103,6 +103,9 @@ func (s *Strategy) processLogin(w http.ResponseWriter, r *http.Request, a *login
 			if len(a.ReturnTo) > 0 {
 				opts = append(opts, registration.WithFlowReturnTo(a.ReturnTo))
 			}
+			if a.OAuth2LoginChallenge.Valid {
+				opts = append(opts, registration.WithFlowOAuth2LoginChallenge(a.OAuth2LoginChallenge))
+			}
 
 			aa, err := s.d.RegistrationHandler().NewRegistrationFlow(w, r, a.Type, opts...)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


NilsBarlaug:
I'm not able to reproduce the failing e2e test on postgres locally 🤔

jonas-jonas (changes requested):
Thank you, sounds reasonable to me. 

It would be great to have at least a go handler test showcasing the issue. `selfservice/flow/registration/handler_test.go` should have some existing examples for this.

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -103,6 +103,9 @@ func (s *Strategy) processLogin(w http.ResponseWriter, r *http.Request, a *login
 			if len(a.ReturnTo) > 0 {
 				opts = append(opts, registration.WithFlowReturnTo(a.ReturnTo))
 			}
+			if a.OAuth2LoginChallenge.Valid {
```
aeneasr, line 106: We should do this for registration too :)

PR #3286, opened 2023-05-19
feat: extend admin API to create/read/update/delete OIDC providers configuration

splaunov:
Corporate customers willing to use OIDC for SSO should be able to configure provider parameters by their own.
This PR extends admin API to make it possible to build admin self-service UI for OIDC providers configuration management.
Configuration data should be stored in db and take precedence over file-based config.

## Checklist

      introduces a new feature.
- [ x] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/oidc/provider_config.go
+++ b/selfservice/strategy/oidc/provider_config.go
@@ -170,5 +170,7 @@ type WithSecretHidden Configuration
 func (c WithSecretHidden) MarshalJSON() ([]byte, error) {
 	type localConfiguration Configuration
 	c.ClientSecret = ""
+	c.PrivateKeyId = ""
+	c.PrivateKey = ""
 	return json.Marshal(localConfiguration(c))
 }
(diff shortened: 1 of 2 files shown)


splaunov:
Marking this as "draft" because only reading of providers configs has been implemented.
But need to know what you think about this initiative as a whole, @aeneasr @zepatrik

aeneasr:
Thank you for the PR! I definitely understand the reason why such a feature should exist. Unfortunately it contradicts our architecture design between configuration which is separated on purpose from the database for many reasons:

- complexity of managing a variety of configuration sources
- circular dependencies (what loads first? config or database?)
- unclear config priority (what takes precedence?)
- inability to do hot reloading without additional database magic such as postgres hooks or CDC
- additional database strain to augment config on every API request

While there probably is an answer to all of these questions (it is just software after all), it does complicate the architecture significantly.

In the Ory Network, we load the configuration file from a dynamic source (which supports hot reloading) to provide a configuration plane which acts independently of the data plane of Ory Kratos.

We could also offer an option to load configs from a websocket connection which would allow for dynamic configs with hotreloading, but I'm not sure how well this would work in practice with reconnects and other things.

PR #3295, opened 2023-05-30
chore: update security scanners

Demonsthere:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

## Further Comments
- Update CVE scanners
- Use proper image for scans, instead of a dummy one
- Update system versions
- Dockerfile good practices

### DIFF
--- a/.github/workflows/cve-scan.yaml
+++ b/.github/workflows/cve-scan.yaml
@@ -15,11 +15,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Setup Env
-        id: vars
-        shell: bash
-        run: |
-          echo "SHA_SHORT=$(git rev-parse --short HEAD)" >> "${GITHUB_ENV}"
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/cve-scan.yaml
```diff
@@ -14,32 +14,35 @@ jobs:
...
-          image: oryd/kratos:${{ steps.vars.outputs.sha_short }}
+          image: oryd/kratos:${{ env.SHA_SHORT }}
           fail-build: true
           severity-cutoff: high
-          debug: false
-          acs-report-enable: true
+          add-cpes-if-none: true
+      - name: Inspect action SARIF report
+        shell: bash
+        if: ${{ always() }}
```
jonas-jonas, line 41: Doesn't a step always run if it's not specified? Any reason to have this?

Demonsthere: This means that the step will run even if a previous failed. This ensures that we run all the scanners even if one of them fails (reports findings higher then some threshold) :)

jonas-jonas: Ah right.

Issue #3291, opened 2023-05-24
Ability to import `crypt(3)` hashes


indietyp:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [x] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [x] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

We're currently using `md5-crypt` (from the [`libcrypt`](https://manpages.debian.org/bullseye/libcrypt-dev/crypt.3.en.html) ) to hash passwords (it is part of the POSIX C library), which are currently not supported by Kratos.

Adding support for them would be very beneficial, as many systems, especially older ones that seek to move Kratos, use `crypt(3)` functions for their hashes.

### Describe your ideal solution

Ability to ideally import any `crypt(3)` hash, but at least have the ability to import [`md5crypt`](https://manpages.debian.org/bullseye/libcrypt-dev/crypt.5.en.html#md5crypt) hashes.

### Workarounds or alternatives

There are currently no workarounds, as the algorithms for `md5crypt` based hashes is not compatible with the `md5` import (it uses `md5` internally).

### Version

v0.13.0

### Additional Context

Having looked around the go ecosystem, it seems that there are quite a lot existing libraries that support `crypt(3)` hashes.

aeneasr:
Sounds good, please use the PHC hashing format for this

PR #3303, opened 2023-06-05
feat: implement `crypt(3)` hashers

indietyp:
Implement `crypt(3)` based hashers.

This PR implements `md5crypt`, `sha256crypt`, `sha512crypt`, which are considered legacy (like `md5`), but are used in legacy systems looking to convert to ory. They use the existing format of `crypt(5)` (which is compliant to PHC).

## Related issue(s)

* #3291

## Checklist

      introduces a new feature.
- [x] I am following the
- x] I confirm that this pull request does not address a security
      the changes.
      works.

### DIFF
--- a/hash/hash_comparator.go
+++ b/hash/hash_comparator.go
@@ -23,6 +23,10 @@ import (
 	"golang.org/x/crypto/pbkdf2"
 	"golang.org/x/crypto/scrypt"
 
+	"github.com/go-crypt/crypt"
+	"github.com/go-crypt/crypt/algorithm/md5crypt"
+	"github.com/go-crypt/crypt/algorithm/shacrypt"
+
 	"github.com/ory/kratos/driver/config"
 )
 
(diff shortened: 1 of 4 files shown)


hash/hash_comparator.go
```diff
@@ -200,7 +240,10 @@ func CompareMD5(_ context.Context, password []byte, hash []byte) error {
 }
 
 var (
+	isMD5CryptHash       = regexp.MustCompile(`^\$1\$`)
```
hperl, line 243: Can we use a more descriptive label like "md5-crypt" here instead of just `1`?

hash/hash_comparator.go
```diff
@@ -69,6 +89,26 @@ func CompareBcrypt(_ context.Context, password []byte, hash []byte) error {
 	return nil
 }
 
+func CompareSHA256Crypt(_ context.Context, password []byte, hash []byte) error {
+	decoder := crypt.NewDecoder()
```
hperl, line 93: Would it make sense to initialize the decoder once with the supported algorithms (md5, sha256, sha512) and then just use it here?

hperl (changes requested):
Thanks for your contribution. This looks really good already, I just have two minor comments below.

hash/hash_comparator.go
indietyp: Sure, I considered doing that (something like `md5-crypt`). Still, I decided against it as `bcrypt` already uses their Modular Crypt Identifier (https://manpages.debian.org/bullseye/libcrypt-dev/crypt.5.en.html#bcrypt), and `$1$` is the standardized one for md5-crypt (as is `$5$` and `$6$` for sha256crypt and sha512crypt respectively.)

indietyp: fixed in [216f1db](https://github.com/ory/kratos/pull/3303/commits/216f1dbbd4a8febf3e3550c6765057df4a6a8d4d)

hperl (dismissed):
Thanks for the changes, LGTM! 🎉

vinckr:
Hello @indietyp @hperl 

Would it be good to have docs for this? At least seems good IMO to mention it [here](https://www.ory.sh/docs/kratos/manage-identities/import-user-accounts-identities#hashed-passwords).

hperl:
Docs are already done :)
https://github.com/ory/docs/pull/1426

Issue #3289, opened 2023-05-24
`identity_verification_codes table` violates foreign key constraint error


adamstrawson:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

Our logging flagged an exception from Kratos for the below:

```
{
  "code": 500,
  "status": "Internal Server Error",
  "message": "named insert: ERROR: insert or update on table \"identity_verification_codes\" violates foreign key constraint \"identity_verification_codes_identity_verifiable_addresses_id_fk\" (SQLSTATE 23503)"
}
```

The user was creating the User Recovery flow, they received the email to recover the password, upon entering the password and submitting it, the process failed with the above error. 

It's unclear why this key constraint violation happened at the time, all interactions were through Kratos so I'm not sure why the duplication occured, or what we can do to ensure it doesn't happen again. 

### Reproducing the bug

Although this isn't something we can reproduce, I'm not confident that we won't see it again in the future. 

The steps the user took:

1. The user was created in our UI
2. The user received the email to setup their password via the recovery flow
3. They clicked the link, which took them to the recover flow
4. Entered password, and submitted
5. Exception thrown

### Relevant log output

```shell
{
  "code": 500,
  "status": "Internal Server Error",
  "message": "named insert: ERROR: insert or update on table \"identity_verification_codes\" violates foreign key constraint \"identity_verification_codes_identity_verifiable_addresses_id_fk\" (SQLSTATE 23503)"
}
```

### Version

v0.13.0

### On which operating system are you observing this issue?

Linux

### In which environment are you deploying?

Kubernetes with Helm

aeneasr:
Thank you for the report! Tagging @jonas-jonas who worked on this. @jonas-jonas I'll add it to the backlog for one of the next iterations

aeneasr:
[...]

Issue #3187, opened 2023-03-24
Identity created with uppercase letter in email cannot be PATCHed


sloosch:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

Trying to PATCH the `verifiable_addresses` `verified` and/or `status` flag is not possible if the identity was created with an uppercase letter in the email.

### Reproducing the bug

- Use an identity schema with a trait having an email and verification via email enabled.
- Create a new identity with an email containing a uppercase letter e.g. `Test@test.com`
- Try to change the verified status of the newly created identity via PATCH using the admin api of kratos: e.g. kratos running locally with exposed admin port 4434 `curl -v -X PATCH --data '@./patch.json' -H 'Content-Type: application/json'  http://localhost:4434/admin/identities/<the-identity-id>'`
where `patch.json` does contain the patch operation:
```
[
  {
        "op": "replace",
        "path": "/verifiable_addresses/0/verified",
        "value": true
  }
]
```
- In the response observe that  `verifiable_addresses` was not updated and does contain a null-id and null-created-at date:
```
[...]

PR #3330, opened 2023-06-19
fix: properly normalize uppercase mail addresses

hperl:
## Related issue(s)

Fixes https://github.com/ory/kratos/issues/3187
Fixes https://github.com/ory/kratos/issues/3289
Fixes https://github.com/ory/network/issues/273

### DIFF
--- a/identity/extension_recovery.go
+++ b/identity/extension_recovery.go
@@ -5,6 +5,7 @@ package identity
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/ory/jsonschema/v3"
@@ -32,7 +33,9 @@ func (r *SchemaExtensionRecovery) Run(ctx jsonschema.ValidationContext, s schema
 			return ctx.Error("format", "%q is not valid %q", value, "email")
 		}
 
-		address := NewRecoveryEmailAddress(fmt.Sprintf("%s", value), r.i.ID)
+		address := NewRecoveryEmailAddress(
+			strings.ToLower(strings.TrimSpace(
+				fmt.Sprintf("%s", value))), r.i.ID)
 
 		if has := r.has(r.i.RecoveryAddresses, address); has != nil {
 			if r.has(r.v, address) == nil {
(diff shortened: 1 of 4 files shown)


aeneasr (commented):
LGTM!

How will this affect existing addresses? Does it make sense to use SQL's ToLower like for credential identifiers?

BrandonNoad:
May also fix https://github.com/ory/network/issues/273

hperl:
> May also fix [ory/network#273](https://github.com/ory/network/issues/273)

Indeed, putting that on the list as well!

hperl:
> LGTM!
> 
> How will this affect existing addresses? Does it make sense to use SQL's ToLower like for credential identifiers?

We already normalize the recovery and verifiable addresses on identity create and update here: https://github.com/ory/kratos/blob/c93b54302cba86d2d907473017f3ac68862415ca/persistence/sql/identity/persister_identity.go#L351-L356. So in the database there should always only be lowercase recovery and verifiable addresses.

We also had a migration that made all addresses lowercase here:

https://github.com/ory/kratos/blob/c93b54302cba86d2d907473017f3ac68862415ca/persistence/sql/migrations/sql/20220610155809000000_identity_address_casing.cockroach.up.sql#L1-L2

So existing addresses should be fine, as all addresses are written lowercase, and we migrated all mixedcase addresses to lowercase at some point.

---

Note, this fix is still necessary (or actually fixes things), because in a patch/update of an identity we get the identity from the database (with a lowercase email address), and then verify the traits to see if the verified address changed (which may be mixed case). Finally, in https://github.com/ory/kratos/blob/c93b54302cba86d2d907473017f3ac68862415ca/persistence/sql/identity/persister_identity.go#L304 we compare the addresses by a hash to see if the value needs to be updated. Since the lowercase value from the verifiable address does not match the mixed case email from the trait, the association gets removed.

aeneasr (approved):

adamstrawson:
Thanks for such a quick turn around on this @hperl 👏

PR #3341, opened 2023-06-22
hperl/send-emails-via-http

hperl:
Original PR by @barnarddt: https://github.com/ory/kratos/pull/3048

---

This change adds a new delivery method to the courier called `mailer`. Similar to SMS functionality it posts a templated Data model to a API endpoint.  This API can then send emails via a CRM or any other mechanism that it wants.

`Mailer` still uses the existing email data models so any new email added will automatically be sent to the API/CRM as well.

## Related issue(s)
Resolves https://github.com/ory/kratos/issues/2825
Also see #1030 and #3008 
Documentation PR https://github.com/ory/docs/pull/1298

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/.schemastore/config.schema.json
+++ b/.schemastore/config.schema.json
@@ -126,60 +126,6 @@
         "config"
       ]
     },
-    "httpRequestConfig": {
-      "type": "object",
-      "properties": {
-        "url": {
-          "title": "HTTP address of API endpoint",
-          "description": "This URL will be used to send the emails to.",
-          "examples": [
-            "https://example.com/api/v1/email"
-          ],
-          "type": "string",
-          "pattern": "^https?:\\/\\/.*"
-        },
-        "method": {
-          "type": "string",
-          "description": "The HTTP method to use (GET, POST, etc). Defaults to POST.",
-          "default": "POST"
-        },
-        "headers": {
-          "type": "object",
-          "description": "The HTTP headers that must be applied to request",
-          "additionalProperties": {
-            "type": "string"
-          }
-        },
-        "body": {
-          "type": "string",
-          "format": "uri",
-          "pattern": "^(http|https|file|base64)://",
-          "description": "URI pointing to the jsonnet template used for payload generation. Only used for those HTTP methods, which support HTTP body payloads",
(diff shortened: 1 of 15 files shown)


.schemastore/config.schema.json
```diff
@@ -126,6 +126,60 @@
         "config"
       ]
     },
+    "httpRequestConfig": {
+      "type": "object",
+      "properties": {
+        "url": {
```
hperl, line 132: Any ideas on how to make the URL required only if the "courier.http" object is set?

.schemastore/config.schema.json
```diff
@@ -2119,7 +2197,7 @@
       "additionalProperties": false
     },
     "tracing": {
-      "$ref": "https://raw.githubusercontent.com/ory/x/v0.0.551/otelx/config.schema.json"
+      "$ref": "ory://tracing-config"
```
aeneasr, line 2200: Where are these changes coming from?

.schemastore/config.schema.json
```diff
@@ -1901,6 +1973,12 @@
               "Authorization": "Bearer some-token"
             }
           ]
+        },
```
aeneasr, line 1976: Is this part of the PR?

courier/http.go
```diff
@@ -0,0 +1,79 @@
...
+		return err
+	}
+
+	defer res.Body.Close()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+	case http.StatusCreated:
+	default:
+		return errors.New(http.StatusText(res.StatusCode))
```
aeneasr, line 68: This for sure needs better observability. First, the error needs to be understandable / readable so something like "Unable to dispatch mail delivery because upstream server replied with status code %d". We also definitely need some logging or eventing here for observability.

.schemastore/config.schema.json
aeneasr: Yes, first make "url" a required field in the object. Then you need to do an if statement in the courier object. i think there's some examples in webhooks how this works.

.schemastore/config.schema.json
```diff
@@ -126,6 +126,60 @@
...
+      "type": "object",
+      "properties": {
+        "url": {
+          "title": "HTTP address of API endpoint",
+          "description": "This URL will be used to send the emails to.",
+          "examples": [
+            "https://example.com/api/v1/email"
+          ],
+          "type": "string",
+          "pattern": "^https?:\\/\\/.*"
```
aeneasr, line 139: I would probably allow http in dev mode

aeneasr (commented):
Very nice! Just a few minor comments and an answer to your question

.schemastore/config.schema.json
hperl: Right now HTTP is always allowed, as it is for other URLs. If we want to always disable HTTP when dev mode is not active, do we want to do it more generally here?

https://github.com/ory/kratos/blob/68cb201d4437f49a559239f70e6d5259aa9fe564/request/builder.go#L45-L62

.schemastore/config.schema.json
hperl: reverted

.schemastore/config.schema.json
hperl: reverted

courier/http.go
hperl: done

.schemastore/config.schema.json
aeneasr: That's also fine! I overlooked the `?` instruction

aeneasr (approved):

PR #3355, opened 2023-06-30
feat: apply branding to courier messages

splaunov:
While kratos does a great job by externalising all the UI, courier messages are still rendered by kratos. So, customisations are limited by the context we have stored in `courier_messages` table. It might be sufficient for most cases but in some situations more data are needed.
One of such cases is when the system provides multiple mobile and web applications styled differently according to some brand or a "white label". Each user can potentially use more than one of such applications and expectations are that validation/recovery messages are sent exactly with the same styling/branding as the application which initiated validation/recovery request.
This PR adds `branding` parameter to registration, validation and recovery flow update requests and stores it into `template_data` field in `courier_messages` table.
Now each application can set it's branding name when posting registration, validation or recovery flow and use it appropriately in messages rendering logic.

## Checklist

      introduces a new feature.
- [x ] I am following the
      the changes.
      works.


jonas-jonas:
Hi! We now have support for [HTTP based email sending](https://github.com/ory/kratos/commit/28b7b04a34eeba2d84de5c543f5ba8b41b38a129) (docs are still missing - sorry), where we send the email's data to your remote target, and you can customize the sending behavior there, similar to a webhook.

Could you try that out, and see if it fits your use case?

splaunov:
No, as there is no info about branding attached to the message data. So, it's not about how to send, over SMTP or HTTP, it's about adding branding parameter to the message's context.

selfservice/strategy/link/strategy_verification.go
```diff
@@ -50,6 +50,8 @@ type verificationSubmitPayload struct {
 	CSRFToken string `json:"csrf_token" form:"csrf_token"`
 	Flow      string `json:"flow" form:"flow"`
 	Email     string `json:"email" form:"email"`
+	// A branding to be applied to the email body
+	Branding string `json:"branding" form:"branding"`
```
jonas-jonas, line 54: We have the concept of [`transient_payload`s](https://github.com/ory/kratos/blob/master/selfservice/strategy/password/registration.go#L53) in the registration flow. It's essentially a JSON container for arbitrary data, you can pass through the flow and use in Webhooks. We have it in the backlog to extend that concept to the other flows as well. 

That might be a good fit here too, instead of using a relatively niche parameter on the top level. The piping through to the email templates should be similar to your current approach.

jonas-jonas: Yes, that would work. I think we don't have any special handling of the `transient_payload` parameter (it's literally just passed through), so should almost just be a rename from `Branding` to`TransientPayload`, and it should be good to go.

aeneasr:
I just wanted to note that this is now possible due to some other PRs:

1. We pass `transient_payload` now to all SMS and email channels - essentially making the `branding` parameter here obsolete
2. We additionally pass `request_url` (the original request URL e.g. `https://ory.exmaple.org/ui/login?brand=foo` to the SMS and email channels - allowing someone to adjust the template based on these query parameters.

Since this is now possible in multiple ways, I'll be closing this!

PR #3357, opened 2023-06-30
feat: expand parameter for /sessions/whoami

alnr:
This allows the caller to opt out of querying for personal data.


alnr:
Note to self @alnr 

This is still relevant. It will allow the caller to check a session ultra-fast if they already know who the user is. Close to check-token perf.

PR #3372, opened 2023-07-06
fix: reduce db lookups in whoami for aal check

aeneasr:
This PR removes queries to `identity_credentials`, `identity_credential_types`, and `identity_credential_identifiers` when performing the whoami check.

It does this by “caching” the maximum available AAL in the identity table. The value is updated whenever an identity is created or updated.

The mechanism can also fall back to generate the AAL value itself if the value is not available in the identity (this is a migration strategy).

One potential issue: If the `available_aal` is not correctly set, it might lead to issues where the user has 2FA enabled, but the system thinks they only have 1FA enabled.


session/manager_http.go
```diff
@@ -295,31 +294,50 @@ func (s *ManagerHTTP) DoesSessionSatisfy(r *http.Request, sess *Session, request
...
-			if err := s.r.PrivilegedIdentityPool().HydrateIdentityAssociations(ctx, i, identity.ExpandCredentials); err != nil {
+		}
+
+		i := sess.Identity
+		available := i.AvailableAAL
+		if len(available) == 0 || available == identity.NoAuthenticatorAssuranceLevel {
+			// Available is 0 if the identity was created before the AAL feature was introduced, or if the identity
+			// was directly created in the persister and not the identity manager.
+			//
+			// aal0 indicates that there AAL state of the identity is probably unknown.
```
Benehiko, line 310:
```suggestion
			// aal0 indicates that the AAL state of the identity is probably unknown.
```

Benehiko (dismissed):
LGTM

persistence/sql/migrations/sql/20230706000000000001_available_aal.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE identities ADD COLUMN available_aal VARCHAR(4) NOT NULL DEFAULT 'aal0';
```
zepatrik, line 1: Doesn't `NOT NULL DEFAULT ...` do a full table update? Wouldn't it be better to use a nullable column here instead?
Like this every identity would have `available_aal` set to 0 after the migration, and only on update we would set the correct value.

session/handler.go
```diff
@@ -205,7 +205,10 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, ps httprouter.P
 	}
 
 	var aalErr *ErrAALNotSatisfied
-	if err := h.r.SessionManager().DoesSessionSatisfy(r, s, c.SessionWhoAmIAAL(r.Context())); errors.As(err, &aalErr) {
+	if err := h.r.SessionManager().DoesSessionSatisfy(r, s, c.SessionWhoAmIAAL(r.Context()),
+		// For the time being we want to update the AAL in the database if it is unset.
+		UpsertAAL,
```
zepatrik, line 210: A nullable column makes this unnecessary, because we can distinguish the "not yet computed" from the "zero" case.

persistence/sql/migratest/fixtures/identity/0149ce5f-76a8-4efe-b2e3-431b8c6cceb6.json
zepatrik, line 1: Why does this lack the new column?

identity/manager.go
```diff
@@ -164,9 +172,34 @@ func (m *Manager) Update(ctx context.Context, updated *Identity, opts ...Manager
...
 	}
 
+	if err := m.setAAL(ctx, updated); err != nil {
+		return err
+	}
+
 	return m.r.PrivilegedIdentityPool().UpdateIdentity(ctx, updated)
 }
 
+func (m *Manager) setAAL(ctx context.Context, i *Identity) (err error) {
```
zepatrik, line 182: I think it makes more sense to make this a function on the identity, and provide the manager as a dependency there (potentially as an interface). The logic is duplicated in the session manager well.

session/manager_http.go
```diff
@@ -295,31 +294,50 @@ func (s *ManagerHTTP) DoesSessionSatisfy(r *http.Request, sess *Session, request
...
 			}
-		}
 
-		available := identity.NoAuthenticatorAssuranceLevel
-		if firstCount, err := s.r.IdentityManager().CountActiveFirstFactorCredentials(ctx, i); err != nil {
-			return err
-		} else if firstCount > 0 {
-			available = identity.AuthenticatorAssuranceLevel1
-		}
+			i.AvailableAAL = available
```
zepatrik, line 333:
```suggestion
			i.SetAvailableAAL(s.r.IdentityManager())
```

session/manager_http.go
```diff
@@ -295,31 +294,50 @@ func (s *ManagerHTTP) DoesSessionSatisfy(r *http.Request, sess *Session, request
...
 
-		if secondCount, err := s.r.IdentityManager().CountActiveMultiFactorCredentials(ctx, i); err != nil {
-			return err
-		} else if secondCount > 0 {
-			available = identity.AuthenticatorAssuranceLevel2
+			// This is the migration strategy for identities that already exist.
+			if managerOpts.upsertAAL {
```
zepatrik, line 336: During the "migration" we just recompute it for everyone. Users that don't log in during the migration period stay at AAL0 forever. I think a nullable column is the way to go instead.
```suggestion
			// This is the migration strategy for identities that already exist.
			if !i.AvailableAAL.Valid {
```

zepatrik (changes requested):
I have some doubts about the migration strategy, see comments.

persistence/sql/migrations/sql/20230706000000000001_available_aal.up.sql
aeneasr: Yep, absolutely. However as I understood the people from CRDB, this should not be an issue due to the way migrations work. We still need to verify that though.

session/handler.go
aeneasr: I think that this state could also happen during an import, which is why I think this fallback is required.

persistence/sql/migratest/fixtures/identity/0149ce5f-76a8-4efe-b2e3-431b8c6cceb6.json
aeneasr: Because it's not exposed in JSON

persistence/sql/migrations/sql/20230706000000000001_available_aal.up.sql
aeneasr: It's now nullable

session/manager_http.go
aeneasr: Yup - it's now nullable. This parameter however is still here because we don't always want to perform an INSERT or UPDATE statement when this method is called, so there has to be a way to control it.

identity/manager.go
aeneasr: done

session/manager_http.go
```diff
@@ -295,31 +294,39 @@ func (s *ManagerHTTP) DoesSessionSatisfy(r *http.Request, sess *Session, request
...
-		}
+		i := sess.Identity
+		available := identity.AuthenticatorAssuranceLevel(i.AvailableAAL.String)
+		if !i.AvailableAAL.Valid {
+			// Available is 0 if the identity was created before the AAL feature was introduced, or if the identity
+			// was directly created in the persister and not the identity manager.
+			//
+			// aal0 indicates that the AAL state of the identity is probably unknown.
+			//
+			// In either case, we need to fetch the credentials from the database to determine the AAL.
```
zepatrik, line 312: This comment is not accurate anymore.

session/manager_http.go
zepatrik: OK, makes sense.

zepatrik (approved):
Looks good now :+1:

Issue #2029, opened 2021-12-07
Passwordless email flows


pradyuman:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

We'd like to fall back to email authentication in situations where users do not want to set up TOTP MFA.

### Describe your ideal solution

Ideal solution would be a code emailed to a user to avoid the issue described in https://github.com/ory/kratos/issues/1451.

### Workarounds or alternatives

As far as I can tell, there's no workaround.

### Version

0.8

aeneasr:
If I reset my password using the recovery flow, which sends an email to my inbox that contains a link to change the password, wouldn't that completely bypass the "multi" in multi-factor authentication?

pradyuman:
Calling this MFA was the wrong way to put it, but my goal is to:
1. Not rely on password authentication if they haven't set up TOTP to protect against accidental password leaks / brute forcing.
2. Ideally collect password during initial registration so it's not overly complicated to set up TOTP MFA in the future.

Since Ory doesn't have a passwordless authentication system today, my solution to #1 would be to require email verification on every login, but I'm open to other suggestions.

In an ideal world everyone would just use TOTP. Unfortunately, not everyone understands how that works so we can't require it.

pradyuman:
Calling this MFA was the wrong way to put it, but my goal is to not rely on password authentication if they haven't set up TOTP to protect against accidental password leaks / brute forcing.

Since Kratos doesn't have a passwordless authentication system today, my solution would be to require email verification on every login as part of the MFA system that was recently built, but I'm open to other suggestions.

In an ideal world everyone would just use TOTP. Unfortunately, not everyone understands how that works so we can't require it.

aeneasr:
[...]

PR #3378, opened 2023-07-12
feat: login & registration via code

Benehiko:
Implements Login and Registration via the code strategy.

## Related issue(s)

This feature adds passwordless email code login. When a user signs up, or signs in, a code is sent to their email address which they can use to complete the authentication process.

This feature is currently only working for browser facing APIs.

Closes https://github.com/ory/kratos/issues/2029
Closes https://github.com/ory-corp/cloud/issues/3573

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


selfservice/strategy/code/strategy.go
```diff
@@ -82,14 +118,319 @@ func NewStrategy(deps strategyDependencies) *Strategy {
...
+
+func (c *CodeStateManager) RegisterVerifyCodeState(fn ValidateCodeState) {
+	c.verifyCodeState = fn
+}
+
+func (c *CodeStateManager) RegisterAlreadyValidatedCodeState(fn AlreadyValidatedCodeState) {
+	c.alreadyValidatedCodeState = fn
+}
+
+func (c *CodeStateManager) validatePayload(ctx context.Context) error {
```
Benehiko, line 357: helper to validate the initial request sent to the code method.

selfservice/strategy/code/strategy.go
```diff
@@ -82,14 +118,319 @@ func NewStrategy(deps strategyDependencies) *Strategy {
...
+		if len(c.payload.Email) == 0 {
+			return errors.WithStack(schema.NewRequiredError("#/email", "email"))
+		}
+	default:
+		return errors.New("received unexpected flow type")
+	}
+	return nil
+}
+
+func (c *CodeStateManager) Run(ctx context.Context) error {
```
Benehiko, line 377: function to help run the code method logic.

works with login and registration. can also be adopted in recovery and verification.

selfservice/strategy/code/strategy_verification.go
```diff
@@ -38,35 +38,7 @@ func (s *Strategy) RegisterAdminVerificationRoutes(admin *x.RouterAdmin) {
...
-		node.NewInputField("method", s.VerificationStrategyID(), node.CodeGroup, node.InputAttributeTypeSubmit).
-			WithMetaLabel(text.NewInfoNodeLabelSubmit()),
-	)
-
-	f.UI.Nodes = nodes
-	if f.Type == flow.TypeBrowser {
-		f.UI.SetCSRF(s.deps.GenerateCSRFToken(r))
-	}
-	return nil
+	return s.PopulateMethod(r, f)
```
Benehiko, line 41: all the logic is contained inside a common function now.

selfservice/strategy/code/code_registration.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+	"database/sql"
+	"time"
+
+	"github.com/gofrs/uuid"
+
+	"github.com/ory/kratos/identity"
+)
+
+type RegistrationCode struct {
```
aeneasr, line 16: Swagger model?

.schema/openapi/patches/selfservice.yaml
```diff
@@ -25,6 +25,12 @@
       password: "#/components/schemas/updateRegistrationFlowWithPasswordMethod"
       oidc: "#/components/schemas/updateRegistrationFlowWithOidcMethod"
       webauthn: "#/components/schemas/updateRegistrationFlowWithWebAuthnMethod"
+- op: add
+  path: /components/schemas/RegistrationFlowState/enum
+  value:
+    - choose_method
+    - sent_email
+    - passed_challenge
```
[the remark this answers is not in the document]
aeneasr: Makes sense - question is though how this new state parameter changes existing login and registration flow behavior? Is it simply unset for them?

driver/registry_default_login.go
```diff
@@ -34,6 +34,10 @@ func (m *RegistryDefault) PostLoginHooks(ctx context.Context, credentialsType id
 		}
 	}
 
+	if credentialsType == identity.CredentialsTypeCodeAuth && m.Config().SelfServiceCodeStrategy(ctx).LoginEnabled {
+		b = append(b, m.HookCodeAddressVerifier())
+	}
```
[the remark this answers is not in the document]
aeneasr: 👍

driver/registry_default_registration.go
```diff
@@ -28,6 +28,11 @@ func (m *RegistryDefault) PostRegistrationPostPersistHooks(ctx context.Context,
 		initialHookCount = 1
 	}
 
+	if credentialsType == identity.CredentialsTypeCodeAuth && m.Config().SelfServiceCodeStrategy(ctx).RegistrationEnabled {
+		b = append(b, m.HookCodeAddressVerifier())
+		initialHookCount += 1
+	}
```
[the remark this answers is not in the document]
aeneasr: 👍

embedx/config.schema.json
```diff
@@ -1441,6 +1293,16 @@
...
+                "login_enabled": {
+                  "type": "boolean",
+                  "title": "Enables Login with Code Method",
+                  "default": false
+                },
+                "registration_enabled": {
+                  "type": "boolean",
+                  "title": "Enables Registration with Code Method",
+                  "default": false
+                },
```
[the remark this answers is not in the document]
aeneasr: Ah, I will also leave a comment about this. I'm not sure whether we should favor simplicity or flexibility here. @kmherrmann thoughts? My gut says - if in doubt leave it out. If we get a customer that needs exactly this we know how to build it?

identity/credentials.go
```diff
@@ -154,6 +166,12 @@ type Credentials struct {
 	// Identifiers represents a list of unique identifiers this credential type matches.
 	Identifiers []string `json:"identifiers" db:"-"`
 
+	// IdentifierAddressType represents the type of the identifiers (e.g. email, phone).
+	// This is used to determine the correct courier to send messages to.
+	// The value is set by the code extension schema and is not persisted.
+	// only applicable on the login, registration with `code` method.
+	IdentifierAddressType CredentialsIdentifierAddressType `json:"-" db:"-"`
```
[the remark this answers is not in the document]
aeneasr: Make it private then maybe and potentially expose it as a function?

Since this is essentially a cache, how can we make sure that the places that needs this always have it (and if not know how to look it up?)

identity/extension_credentials.go
```diff
@@ -40,19 +41,41 @@ func (r *SchemaExtensionCredentials) setIdentifier(ct CredentialsType, value int
...
+	if s.Credentials.Code.Identifier {
+		switch f := stringsx.SwitchExact(s.Credentials.Code.Via); {
+		case f.AddCase(AddressTypeEmail):
+
+			if !jsonschema.Formats["email"](value) {
+				return ctx.Error("format", "%q is not a valid %q", value, s.Credentials.Code.Via)
+			}
+			r.setIdentifier(CredentialsTypeCodeAuth, value, CredentialsIdentifierAddressType(AddressTypeEmail))
+
+		case f.AddCase(AddressTypePhone):
```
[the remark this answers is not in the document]
aeneasr: comment it out in that case

selfservice/flow/login/sort.go
```diff
@@ -15,6 +15,7 @@ func sortNodes(ctx context.Context, n node.Nodes) error {
 			node.OpenIDConnectGroup,
 			node.DefaultGroup,
 			node.WebAuthnGroup,
+			node.CodeGroup,
```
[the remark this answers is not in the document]
aeneasr: yeah thats fine

selfservice/strategy/code/strategy_login.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	codeManager := NewCodeStateManager(f, &CodeStateManagerPayload{
+		Identifier: p.Identifier,
+		Resend:     p.Resend,
+		Code:       p.Code,
+	})
+
+	codeManager.RegisterCreateCodeState(func(ctx context.Context, p *CodeStateManagerPayload) error {
+		s.deps.Logger().
```
aeneasr, line 139: Emit a trace instead :)

selfservice/strategy/code/strategy_login.go
```diff
@@ -0,0 +1,258 @@
...
+		return nil, nil, errors.WithStack(schema.NewNoCodeAuthnCredentials())
+	} else if cred.IdentifierAddressType == "" {
+		return nil, nil, errors.WithStack(schema.NewRequiredError("#/code", "via"))
+	}
+
+	return i, cred, nil
+}
+
+func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow, identityID uuid.UUID) (i *identity.Identity, err error) {
+	if err := flow.MethodEnabledAndAllowedFromRequest(r, f.GetFlowName(), s.ID().String(), s.deps); err != nil {
```
aeneasr, line 110: Add tracing?

selfservice/strategy/code/strategy_login.go
```diff
@@ -0,0 +1,258 @@
...
+			http.Redirect(w, r, f.AppendTo(s.deps.Config().SelfServiceFlowLoginUI(ctx)).String(), http.StatusSeeOther)
+		}
+
+		// we return an error to the flow handler so that it does not continue execution of the hooks.
+		// we are not done with the login flow yet. The user needs to verify the code and then we need to persist the identity.
+		return errors.WithStack(flow.ErrCompletedByStrategy)
+	})
+
+	codeManager.RegisterVerifyCodeState(func(ctx context.Context, p *CodeStateManagerPayload) error {
+		s.deps.Logger().
```
aeneasr, line 207: Add tracing?

selfservice/strategy/code/strategy_registration.go
```diff
@@ -0,0 +1,267 @@
+// Copyright © 2023 Ory Corp
```
aeneasr, line 1: note to self: re-review

selfservice/strategy/code/strategy_login.go
```diff
@@ -0,0 +1,258 @@
+// Copyright © 2023 Ory Corp
```
aeneasr, line 1: note to self: re-review

selfservice/strategy/code/strategy.go
```diff
@@ -82,14 +118,319 @@ func NewStrategy(deps strategyDependencies) *Strategy {
...
+
+			resendNode = node.NewInputField("resend", "code", node.CodeGroup, node.InputAttributeTypeSubmit).
+				WithMetaLabel(text.NewInfoNodeResendOTP())
+		default:
+			return errors.New("received unexpected flow type")
+		}
+
+		// Hidden field Required for the re-send code button
+		// !!important!!: this field must be appended before the code submit button since upsert will replace the first node with the same name
+		freshNodes.Upsert(
```
aeneasr, line 252: I'm not sure if I get this logic?

1. We're adding `<input type="hidden" name="method" value="code">`
2. We're adding `<input type="text" required name="code">`
3. We're changing input from 1 to `<input type="submit" name="method" value="code">`

selfservice/strategy/code/strategy_login.go
```diff
@@ -0,0 +1,258 @@
...
+
+		if err := s.deps.LoginFlowPersister().UpdateLoginFlow(ctx, f); err != nil {
+			return errors.WithStack(err)
+		}
+
+		return nil
+	})
+
+	codeManager.RegisterAlreadyValidatedCodeState(func(ctx context.Context, _ *CodeStateManagerPayload) error {
+		return s.HandleLoginError(w, r, f, &p, errors.WithStack(schema.NewNoLoginStrategyResponsible()))
```
aeneasr, line 246: Isn't there a better state we can leave the user with in this scenario?

selfservice/strategy/code/strategy_login.go
```diff
@@ -0,0 +1,258 @@
...
+
+	if err := codeManager.Run(r.Context()); err != nil {
+		if errors.Is(err, flow.ErrCompletedByStrategy) {
+			return nil, err
+		}
+		// the error is already handled by the registered code states
+		return i, s.HandleLoginError(w, r, f, &p, err)
+	}
+
+	return i, nil
```
aeneasr, line 257: What if something fails in the callbacks above and we end up with an `i == nil`?

aeneasr (commented):
Very nice! I still need to review the login tests, the registration code, and the registration tests, as well as e2e and trying out the flow myself. In summary, this looks great! There are a few things we need to address:

1. Should we enable / disable login / regstration for the code method? If yes, then why not also for verification / recovery?
2. The introduced state parameters are not "backwards" looking - they only apply to the new flows which I find a bit strange?
3. The SQL migrations should be kept in as few files as possible and avoid DEFAULTs on existing large tables
4. The state machine we use in login/registration sets values in the callbacks of outer functions. If a callback is not called, these values might be nil, leading to panics

embedx/config.schema.json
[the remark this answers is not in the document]
Benehiko: i guess we could also make it one key such as `auth_enabled`. But there needs to be a distinction here between the recovery / verification and login / registration.

identity/credentials.go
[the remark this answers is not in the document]
Benehiko: It's populated when the identity credentials are verified against the schema. 

I agree, I will move it to a function that exposes it, rather than a property. This value will by default also be "none".

driver/registry_default_registration.go
```diff
@@ -28,6 +28,11 @@ func (m *RegistryDefault) PostRegistrationPostPersistHooks(ctx context.Context,
 		initialHookCount = 1
 	}
 
+	if credentialsType == identity.CredentialsTypeCodeAuth && m.Config().SelfServiceCodeStrategy(ctx).RegistrationEnabled {
+		b = append(b, m.HookCodeAddressVerifier())
```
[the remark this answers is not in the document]
Benehiko: yes, since you might have multiple identifiers with the code extension. Only one of those emails can be used to register with, and the other will be marked as unverified. Once logging in with the other, this will mark the address as verified.

embedx/identity_extension.schema.json
```diff
@@ -38,6 +38,19 @@
                       "type": "boolean"
                     }
                   }
+                },
+                "code": {
```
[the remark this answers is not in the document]
Benehiko: `otp`?. `code` is also okay since the method you need to enable is `code`.

identity/credentials_code.go
```diff
@@ -0,0 +1,26 @@
...
+type CodeAddressType string
+
+const (
+	CodeAddressTypeEmail CodeAddressType = AddressTypeEmail
+	CodeAddressTypePhone CodeAddressType = AddressTypePhone
+)
+
+// CredentialsOTP represents an OTP code
+//
+// swagger:model identityCredentialsOTP
```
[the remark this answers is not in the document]
Benehiko: ah good catch

selfservice/flow/login/flow.go
```diff
@@ -124,8 +124,19 @@ type Flow struct {
 	// This is only set if the client has requested a session token exchange code, and if the flow is of type "api",
 	// and only on creating the login flow.
 	SessionTokenExchangeCode string `json:"session_token_exchange_code,omitempty" faker:"-" db:"-"`
+
+	// State represents the state of this request:
+	//
+	// - choose_method: ask the user to choose a method (e.g. verify your email)
```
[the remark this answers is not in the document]
Benehiko: only within the code method. not really sure what the best approach would be

selfservice/flow/login/state.go
```diff
@@ -0,0 +1,17 @@
...
+
+// Login Flow State
+//
+// The state represents the state of the login flow.
+//
+// - choose_method: ask the user to choose a method (e.g. login account via email)
+// - sent_email: the email has been sent to the user
+// - passed_challenge: the request was successful and the login challenge was passed.
+//
+// swagger:model LoginFlowState
```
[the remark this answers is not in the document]
Benehiko: this is an enum

.schema/openapi/patches/selfservice.yaml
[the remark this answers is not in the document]
Benehiko: yes it is unset for them. we only use it in the code method. i guess we could add another state here specifying unset? i don't know what the right move is here.

selfservice/flow/recovery/state.go
```diff
@@ -11,34 +13,5 @@ package recovery
...
-}
-
-func NextState(current State) State {
-	if current == StatePassedChallenge {
-		return StatePassedChallenge
-	}
-
-	return states[indexOf(current)+1]
-}
+// swagger:model RecoveryFlowState
```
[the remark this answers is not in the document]
Benehiko: no it's the same name just capitalized. the design doc says enums should be capitalized. Should we keep it lower-case for now?

selfservice/flow/state.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// The state represents the state of the verification flow.
+//
+// - choose_method: ask the user to choose a method (e.g. recover account via email)
+// - sent_email: the email has been sent to the user
+// - passed_challenge: the request was successful and the recovery challenge was passed.
+// - show_form: a form is shown to the user to perform the flow
+// - success: the flow has been completed successfully
+//
+// swagger:enum SelfServiceFlowState
```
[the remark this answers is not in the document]
Benehiko: this needs to be of type "enum" and the ones aliasing this needs to be of type "model" otherwise go-swagger crashes when generating the sdk :shrug: 
we then manually patch the other flows aliasing this type here as type enum.

selfservice/hook/code_address_verifier.go
```diff
@@ -0,0 +1,95 @@
...
+			if err := cv.r.PrivilegedIdentityPool().UpdateVerifiableAddress(r.Context(), &va); err != nil {
+				return errors.WithStack(err)
+			}
+			break
+		}
+	}
+	return nil
+}
+
+func (cv *CodeAddressVerifier) ExecutePostRegistrationPostPersistHook(w http.ResponseWriter, r *http.Request, a *registration.Flow, s *session.Session) error {
```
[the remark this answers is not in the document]
Benehiko: we could move this yes. good point

selfservice/strategy/code/strategy_login_test.go
```diff
@@ -0,0 +1,353 @@
...
+		req, err := http.NewRequestWithContext(ctx, "POST", public.URL+login.RouteSubmitFlow+"?flow="+s.flowID, strings.NewReader(url.Values{
+			"csrf_token": {s.csrfToken},
+			"method":     {"code"},
+			"code":       {s.loginCode},
+			"identifier": {s.identityEmail},
+		}.Encode()))
+		require.NoError(t, err)
+
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		req.Header.Set("Accept", "application/json")
```
aeneasr, line 168: Take a look at the password tests - we are testing various flows there (browser html, browser ajax, native api) and I think we need to do this here and in registration too. There are also some testhelpers you can use to easily submit these flows!

selfservice/strategy/code/strategy_registration.go
```diff
@@ -0,0 +1,267 @@
...
+
+func (s *Strategy) handleIdentityTraits(ctx context.Context, f *registration.Flow, traits json.RawMessage, transientPayload json.RawMessage, i *identity.Identity, opts ...options) error {
+	f.TransientPayload = transientPayload
+	if len(traits) == 0 {
+		traits = json.RawMessage("{}")
+	}
+
+	// we explicitly set the Code credentials type
+	i.Traits = identity.Traits(traits)
+	if err := i.SetCredentialsWithConfig(s.ID(), identity.Credentials{Type: s.ID(), Identifiers: []string{}}, &identity.CredentialsCode{UsedAt: sql.NullTime{}}); err != nil {
```
aeneasr, line 100: Will identifiers be hydrated at a later stage? If not, login will probably fail because the identifier is missing.

selfservice/strategy/code/strategy_registration.go
```diff
@@ -0,0 +1,267 @@
...
+		// this is important since the client could have switched out the identity's traits
+		// this method also returns the credentials for a temporary identity
+		cred, err := s.getCredentialsFromTraits(ctx, f, i, p.Traits, p.TransientPayload)
+		if err != nil {
+			return err
+		}
+
+		// Step 2: Check if the flow traits match the identity traits
+		for _, n := range container.NewFromJSON("", node.DefaultGroup, p.Traits, "traits").Nodes {
+			if !strings.EqualFold(f.GetUI().GetNodes().Find(n.ID()).Attributes.GetValue().(string), n.Attributes.GetValue().(string)) {
```
aeneasr, line 226: Are you sure these are all strings allways?

test/e2e/cypress/integration/profiles/code/login/error.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+import { routes as react } from "../../../../helpers/react"
+
+context("Login error messages with code method", () => {
+  ;[
+    {
+      route: express.login,
+      app: "express" as "express",
+      profile: "code",
+    },
+    // {
```
aeneasr, line 15: We should probably enable this :)

test/e2e/cypress/integration/profiles/code/registration/error.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+
+context("Registration error messages with code method", () => {
+  ;[
+    {
+      route: express.registration,
+      login: express.login,
+      app: "express" as "express",
+      profile: "code",
+    },
+    // {
```
aeneasr, line 15: enable

test/e2e/cypress/integration/profiles/code/registration/success.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+
+context("Registration success with code method", () => {
+  ;[
+    {
+      route: express.registration,
+      login: express.login,
+      app: "express" as "express",
+      profile: "code",
+    },
+    // {
```
aeneasr, line 17: enable

test/e2e/cypress/support/commands.ts
```diff
@@ -1171,30 +1239,44 @@ Cypress.Commands.add("expectSettingsSaved", () => {
...
+        if (email) {
+          mailItem = response.body.mailItems.find((m: any) =>
+            m.toAddresses.includes(email),
+          )
+        } else {
+          mailItem = response.body.mailItems[0]
+        }
+        console.log({ mailItems: response.body.mailItems })
+        console.log({ mailItem })
+        console.log({ email })
```
aeneasr, line 1265: Remove logs?

test/e2e/run.sh
```diff
@@ -136,7 +136,7 @@ prepare() {
   nc -zv localhost 4445 && exit 1
   nc -zv localhost 4446 && exit 1
   nc -zv localhost 4455 && exit 1
-  nc -zv localhost 4456 && exit 1
+  # nc -zv localhost 4456 && exit 1
```
aeneasr, line 139: re-enable?

test/e2e/run.sh
```diff
@@ -218,19 +218,19 @@ prepare() {
...
-        >"${base}/test/e2e/ui-node.e2e.log" 2>&1 &
-    )
-  else
-    (
-      cd "$node_ui_dir"
-      PORT=4456 SECURITY_MODE=cookie npm run start \
-        >"${base}/test/e2e/ui-node.e2e.log" 2>&1 &
-    )
-  fi
+  # if [ -z ${NODE_UI_PATH+x} ]; then
```
aeneasr, line 221: re-enable?

aeneasr (commented):
Nice ! I looked at everything now - see comments

selfservice/strategy/code/strategy_login.go
Benehiko: Not sure, we could maybe re-create the flow? or redirect to a fresh flow?

selfservice/strategy/code/strategy.go
Benehiko: No, the upsert overwrites any input's with the name `method`. So we need to add it to the UI nodes first, but we use upsert here since you could be re-submitting the flow and it could be throwing an error, such as traits do not match or the email is required etc. 

So this upsert will replace the current `<input type="hidden" name="method" value="code">` and then we append the submit button again since it was removed earlier. 

Between submits we essentially always remove the submit types and re-add them. The way we do `Upserts` within nodes are thus where the problem is since it's not a true check of similarity. So this is a work-around to that.

identity/credentials.go
```diff
@@ -141,6 +145,14 @@ func ParseCredentialsType(in string) (CredentialsType, bool) {
 	return "", false
 }
 
+type CredentialsIdentifierAddressType string
```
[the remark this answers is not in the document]
Benehiko: I don't think we should expose this enum to the SDK! this is only used internally within the credentials struct.

selfservice/flow/settings/state.go
```diff
@@ -3,17 +3,14 @@
...
 //     when a flow with invalid (e.g. "please use a valid phone number") data was sent.
 //
-// swagger:model settingsFlowState
-type State string
-
-const (
-	StateShowForm State = "show_form"
-	StateSuccess  State = "success"
-)
+// swagger:model SettingsFlowState
```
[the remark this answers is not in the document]
Benehiko: the settings flow will also contain the same state as login, registration, recovery and verification.

selfservice/strategy/code/strategy.go
```diff
@@ -82,14 +118,319 @@ func NewStrategy(deps strategyDependencies) *Strategy {
...
+
+				if n.Group == node.CodeGroup {
+					freshNodes = append(freshNodes, n)
+				}
+			}
+
+			resendNode = node.NewInputField("resend", "code", node.CodeGroup, node.InputAttributeTypeSubmit).
+				WithMetaLabel(text.NewInfoNodeResendOTP())
+		default:
+			return errors.New("received unexpected flow type")
```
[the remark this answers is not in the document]
Benehiko: Would this be fine?

```go
errors.WithStack(herodot.ErrBadRequest.WithReason("received an unexpected flow state"))
```

persistence/sql/persister_login.go
```diff
@@ -84,3 +88,125 @@ func (p *Persister) DeleteExpiredLoginFlows(ctx context.Context, expiresAt time.
...
+		if err := sqlcon.HandleError(tx.RawQuery(fmt.Sprintf("UPDATE %s SET submit_count = submit_count + 1 WHERE id = ? AND nid = ?", flowTableName), flowID, nid).Exec()); err != nil {
+			return err
+		}
+
+		var submitCount int
+		// Because MySQL does not support "RETURNING" clauses, but we need the updated `submit_count` later on.
+		//#nosec G201 -- TableName is static
+		if err := sqlcon.HandleError(tx.RawQuery(fmt.Sprintf("SELECT submit_count FROM %s WHERE id = ? AND nid = ?", flowTableName), flowID, nid).First(&submitCount)); err != nil {
+			if errors.Is(err, sqlcon.ErrNoRows) {
+				// Return no error, as that would roll back the transaction
```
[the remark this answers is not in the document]
Benehiko: true, but if for some reason it doesn't find it we don't want to lose the update query above when we return a nil. to the user it will return ErrCodeNotFound since the loginCode object will be empty.

persistence/sql/migrations/sql/20230703143600000000_selfservice_login_flows_state.up.sql
```diff
@@ -0,0 +1 @@
+ALTER table selfservice_login_flows ADD state VARCHAR(255) NOT NULL DEFAULT '';
```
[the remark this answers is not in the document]
Benehiko: I've updated this to have the field set to NULL by default. In the codebase we always set the value though.

selfservice/strategy/code/strategy_login.go
Benehiko: This state shouldn't happen. Added a trace and returning back the appropriate error.

jonas-jonas:
1. Where do you need to "redirect"/continue to?
2. That sounds good to me. In recovery/verification we have the `use` key that indicates whether to use link/code, maybe something like that is an option here, too?
3. SGTM

> To get the tests passing on kratos-selfserivce-ui-react-nextjs and kratos-selfservice-ui-node we would need to do an SDK pre-release off of this branch.

Make sure to actually do a pre-release, if you just use a random id it is uploaded as the latest package. (I had that happen a while back.)

Benehiko:
> 1. Where do you need to "redirect"/continue to?
> 2. That sounds good to me. In recovery/verification we have the `use` key that indicates whether to use link/code, maybe something like that is an option here, too?
> 3. SGTM
> 
> > To get the tests passing on kratos-selfserivce-ui-react-nextjs and kratos-selfservice-ui-node we would need to do an SDK pre-release off of this branch.
> 
> Make sure to actually do a pre-release, if you just use a random id it is uploaded as the latest package. (I had that happen a while back.)

1. The SPA needs to submit the login and registration flow twice, like we do with recovery and verification with code. So one submission for the identifier / email & traits and then a second for the code. This isn't a normal flow for the login and verification flow since you usually only show the form again if there were error codes like a trait missing or validation error etc. But in this case the first submission responds with a status 200 and then expects the UI to re-render the form with the returned flow data. 
As shown in the example, I need to specifically check if I did a submit with the 'code' method and then see if this was the first submission with checking the `state` parameter. Since the flow doesn't actually "change" to another one, like with registration -> verification, it kind of feels wrong to use the `continue_with` properties. 

3. I think configuring this on the `code` method makes this more intuitive since it's using the `code` method. In recovery for example you enable it and specify the usage of `link` or `code`. But in registration the model is, you use whichever method you have enabled, `oidc`, `password` etc. Adding a `use: code` there might be a bit strange, it might leave users with a "use code for what?" question.
 
 
 For the SDK pre-release I'll version it like we did last time https://github.com/ory/sdk/commit/7c19b422bc3ad4c06f2026531600ee6b3ea6e581
 
So version v1.1.49-alpha.0

aeneasr:
> For the SDK pre-release I'll version it like we did last time https://github.com/ory/sdk/commit/7c19b422bc3ad4c06f2026531600ee6b3ea6e581
>
>So version v1.1.49-alpha.0

Sounds good!

identity/extension_credentials.go
```diff
@@ -40,19 +41,39 @@ func (r *SchemaExtensionCredentials) setIdentifier(ct CredentialsType, value int
...
-		r.setIdentifier(CredentialsTypePassword, value)
+		r.setIdentifier(CredentialsTypePassword, value, CredentialsIdentifierAddressTypeNone)
 	}
 
 	if s.Credentials.WebAuthn.Identifier {
-		r.setIdentifier(CredentialsTypeWebAuthn, value)
+		r.setIdentifier(CredentialsTypeWebAuthn, value, CredentialsIdentifierAddressTypeNone)
+	}
+
+	if s.Credentials.Code.Identifier {
```
aeneasr, line 60: Missing test

selfservice/strategy/code/strategy.go
```diff
@@ -82,14 +119,348 @@ func NewStrategy(deps strategyDependencies) *Strategy {
...
+		// on Login flow we need to populate the form with the values from the initial form generation
+		for _, n := range container.NewFromJSON("", node.DefaultGroup, data, "").Nodes {
+			f.GetUI().GetNodes().SetValueAttribute(n.ID(), n.Attributes.GetValue())
+		}
+	}
+
+	return nil
+}
+
+type (
```
aeneasr, line 318: I think this is complicating the code too much - see in-depth commentary

selfservice/strategy/code/strategy.go
```diff
@@ -82,14 +119,348 @@ func NewStrategy(deps strategyDependencies) *Strategy {
...
+		}
+	case *updateVerificationFlowWithCodeMethod:
+		if len(v.Email) == 0 {
+			return errors.WithStack(schema.NewRequiredError("#/email", "email"))
+		}
+	default:
+		return errors.WithStack(herodot.ErrBadRequest.WithReason("received unexpected flow payload type"))
+	}
+	return nil
+}
```
aeneasr, line 381: If there are no overlaps in the payload validation, there's not really a reason to move this into its own function with type assertion. It's easier to read and understand if it's "simply" in the code at the right place.

selfservice/strategy/code/strategy.go
```diff
@@ -82,14 +119,348 @@ func NewStrategy(deps strategyDependencies) *Strategy {
...
+	case *updateRegistrationFlowWithCodeMethod:
+		return v.Code
+	case *updateRecoveryFlowWithCodeMethod:
+		return v.Code
+	case *updateVerificationFlowWithCodeMethod:
+		return v.Code
+	}
+	return ""
+}
+
```
aeneasr, line 406: This should be interfaces that implement `ShouldResend` and `GetCode` for example, no need for type assertions. It makes the code more difficult to read and navigate and since Go has duck typing, also not needed.

selfservice/strategy/code/strategy.go
```diff
@@ -82,14 +119,348 @@ func NewStrategy(deps strategyDependencies) *Strategy {
...
+func (c *CodeStateManager[T, P]) Run(ctx context.Context) error {
```
aeneasr, line 407: I think this is reasonably valid, one problem with this though is that we don't know (on the outside) whether the correct callback was indeed called. Did callback "verifyCodeState" really get executed? We can't tell! We only know that we received no error.

Apart from a couple of similar checks (is state empty? is code empty?) we seem to be building more logic for this abstraction.

But is the abstraction even needed? I'm not sure if it really saves us a lot of code (which seems to be the purpose) and it also doesn't have dedicated tests that prove its correctness, in which case we might just rely on the test paths in strategy_login_test etc.

persistence/sql/persister_registration.go
```diff
@@ -64,3 +70,133 @@ func (p *Persister) DeleteExpiredRegistrationFlows(ctx context.Context, expiresA
...
+		// and we thus give no indication whether the supplied code was correct or not. See also https://github.com/ory/kratos/pull/2645#discussion_r984732899
+		if submitCount > 5 {
+			return errors.WithStack(code.ErrCodeSubmittedTooOften)
+		}
+
+		var registrationCodes []code.RegistrationCode
+		if err := sqlcon.HandleError(tx.Where("nid = ? AND selfservice_registration_flow_id = ?", nid, flowID).All(&registrationCodes)); err != nil {
+			if errors.Is(err, sqlcon.ErrNoRows) {
+				// Return no error, as that would roll back the transaction
+				return nil
```
aeneasr, line 134: If this is nil, then registrationCode is nil in the later stages. Why not just return code.ErrCodeNotFound here directly?

persistence/sql/persister_login.go
```diff
@@ -84,3 +88,123 @@ func (p *Persister) DeleteExpiredLoginFlows(ctx context.Context, expiresAt time.
...
+		if submitCount > 5 {
+			return errors.WithStack(code.ErrCodeSubmittedTooOften)
+		}
+
+		var loginCodes []code.LoginCode
+		if err = sqlcon.HandleError(tx.Where("nid = ? AND selfservice_login_flow_id = ? AND identity_id = ?", nid, flowID, identityID).All(&loginCodes)); err != nil {
+			if errors.Is(err, sqlcon.ErrNoRows) {
+				return err
+			}
+			return nil
```
aeneasr, line 150: This does not look correct? In all other code this is inverted (return nil on no rows, return err otherwise)

persistence/sql/persister_registration.go
```diff
@@ -64,3 +70,133 @@ func (p *Persister) DeleteExpiredRegistrationFlows(ctx context.Context, expiresA
...
+	if registrationCode.IsExpired() {
+		return nil, errors.WithStack(flow.NewFlowExpiredError(registrationCode.ExpiresAt))
+	}
+
+	if registrationCode.WasUsed() {
+		return nil, errors.WithStack(code.ErrCodeAlreadyUsed)
+	}
+
+	// ensure that the identifiers extracted from the traits are contained in the registration code
+	if !slice.Contains(addresses, registrationCode.Address) {
```
aeneasr, line 178: Can it happen that `addresses` has `eXaMpLe@uppercase.com` and `code.Address` `example@uppercase.com` where the email is the same but we end up with an error here? Aka missing normalization?

aeneasr (commented):
Thanks Alano! Here's an interim review. There are no actionable things you need to do at the moment. I'm working on restructuring and simplifying the code a bit which I'm also doing as part of the security review (understanding how exactly everything works).

Todos for myself:

* [ ] Write tests for persister (probably extract common logic as it's replicated over 4 persister files)
* [ ] Simplify state machine logic (I feel unsafe to not have clear expectations on the behavior of the state machine in such a security sensitive area)

However, none of these are in my view blocking rolling this out as a preview or continuing the work in the network integration, as I don't think that there'll be any API or behavior changes.

test/e2e/run.sh
```diff
@@ -95,13 +95,13 @@ prepare() {
...
   fi
 
-  if [ -z ${NODE_UI_PATH+x} ]; then
-    node_ui_dir="$(mktemp -d -t ci-XXXXXXXXXX)/kratos-selfservice-ui-node"
-    git clone --depth 1 --branch master https://github.com/ory/kratos-selfservice-ui-node.git "$node_ui_dir"
-    (cd "$node_ui_dir" && npm i --legacy-peer-deps && npm run build)
-  else
-    node_ui_dir="${NODE_UI_PATH}"
-  fi
+  # if [ -z ${NODE_UI_PATH+x} ]; then
```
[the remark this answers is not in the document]
Benehiko: This has already been reverted in a previous commit.

driver/registry_default_registration.go
```diff
@@ -12,6 +12,10 @@ import (
 )
 
 func (m *RegistryDefault) PostRegistrationPrePersistHooks(ctx context.Context, credentialsType identity.CredentialsType) (b []registration.PostHookPrePersistExecutor) {
+	if credentialsType == identity.CredentialsTypeCodeAuth && m.Config().SelfServiceCodeStrategy(ctx).PasswordlessEnabled {
+		b = append(b, m.HookCodeAddressVerifier())
```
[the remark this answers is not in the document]
Benehiko: I guess the if can be moved inside the hook. 

Was just to make sure that before the hook added that the code strategy is enabled and the identity is indeed registering using the code credential. 

If we don't have this if statement, we could accidentally mark any new registration on other methods as verified.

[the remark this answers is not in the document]
aeneasr: Yes that makes sense, I thought I deleted the comment after I understood why it's there.

selfservice/strategy/code/strategy_login.go
```diff
@@ -0,0 +1,286 @@
...
+		if err != nil {
+			return err
+		}
+
+		// Step 2: Delete any previous login codes for this flow ID
+		if err := strategy.deps.LoginCodePersister().DeleteLoginCodesOfFlow(ctx, f.GetID()); err != nil {
+			return errors.WithStack(err)
+		}
+
+		var identifier string
```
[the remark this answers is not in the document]
Benehiko: Yes, since you could have multiple identifiers. We need to send it to the correct identifier (based on what was submitted).

selfservice/strategy/code/strategy_login.go
```diff
@@ -0,0 +1,286 @@
...
+		for _, id := range cred.Identifiers {
+			if strings.EqualFold(p.Identifier, id) {
```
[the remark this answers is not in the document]
Benehiko: yes true, the submitted identifier needs to be parsed

selfservice/strategy/code/strategy_login.go
```diff
@@ -0,0 +1,286 @@
...
+				identifier = id
+			}
+		}
+
+		addresses := []Address{
+			{
+				To:  identifier,
```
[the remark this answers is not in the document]
Benehiko: it won't send out a courier email to this address. I guess a courier entry will be made without an address.

selfservice/strategy/code/strategy.go
Benehiko: I think the purpose is two fold.
- readability inside the actual implementation. In the recovery flow for example, it's tough to know which part is for which state of the flow. Are we handling a first submission (flow state `choose_method`) or are we handling a second submission? Where is it exiting the flow on an error is also not too apparent and where was it successful? 

- Verification, Recovery, Login, Registration and Settings will all have the same states:
 1. choose_method (fist submit with traits / email)
 2. Error? retunr back to choose_method
 3. email_sent (second submit with code)
 4. Error? return back to email_sent 
 5. challenge_passed (last submit)
 6. Resubmit passed challenge? Should error immediately. 
 
 
This code was more to manage these states across all of the flows. And since each flow has its own logic, we do the callback code. I guess we could make this one large method with a switch case with each state and flow, but this would be hard to read as well.

persistence/sql/persister_login.go
Benehiko: yeah this is incorrect. should be 
```suggestion
			if !errors.Is(err, sqlcon.ErrNoRows) {
				return err
			}
			return nil
```

persistence/sql/persister_registration.go
Benehiko: AFAIK the identity validation step normalizes the credentials which is reused as the address at a later point. So the stored address would be a normalized version of the identifier submitted by the user. 

We could enforce a normalization strategy here just to be sure.

persistence/sql/persister_login.go
```diff
@@ -84,3 +88,123 @@ func (p *Persister) DeleteExpiredLoginFlows(ctx context.Context, expiresAt time.
...
+			}
+			return err
+		}
+
+		if submitCount > 5 {
+			return errors.WithStack(code.ErrCodeSubmittedTooOften)
+		}
+
+		var loginCodes []code.LoginCode
+		if err = sqlcon.HandleError(tx.Where("nid = ? AND selfservice_login_flow_id = ? AND identity_id = ?", nid, flowID, identityID).All(&loginCodes)); err != nil {
```
aeneasr, line 146: @Benehiko can you please explain why and if the `identity_id` is important in this query please? Are there security implications?

persistence/sql/persister_login.go
```diff
@@ -84,3 +88,123 @@ func (p *Persister) DeleteExpiredLoginFlows(ctx context.Context, expiresAt time.
...
+			if errors.Is(err, sqlcon.ErrNoRows) {
+				return err
+			}
+			return nil
+		}
+
+	secrets:
+		for _, secret := range p.r.Config().SecretsSession(ctx) {
+			suppliedCode := []byte(p.hmacValueWithSecret(ctx, codeVal, secret))
+			for i := range loginCodes {
```
hperl, line 156: This `for` loop looks a bit concerning. If an attacker can create multiple login codes in quick succession, the probability that one of the codes matches a random guess rises.

From a security perspective we should only ever check against one login code, IMHO the latest.

persistence/sql/persister_registration.go
```diff
@@ -64,3 +70,133 @@ func (p *Persister) DeleteExpiredRegistrationFlows(ctx context.Context, expiresA
...
+			if errors.Is(err, sqlcon.ErrNoRows) {
+				// Return no error, as that would roll back the transaction
+				return nil
+			}
+
+			return err
+		}
+
+	secrets:
+		for _, secret := range p.r.Config().SecretsSession(ctx) {
```
hperl, line 141: Same concern about checking multiple codes applies here.

persistence/sql/persister_registration.go
```diff
@@ -64,3 +70,133 @@ func (p *Persister) DeleteExpiredRegistrationFlows(ctx context.Context, expiresA
...
+		ID:          uuid.Nil,
+	}
+
+	if err := p.GetConnection(ctx).Create(registrationCode); err != nil {
+		return nil, sqlcon.HandleError(err)
+	}
+	return registrationCode, nil
+}
+
+func (p *Persister) UseRegistrationCode(ctx context.Context, flowID uuid.UUID, rawCode string, addresses ...string) (*code.RegistrationCode, error) {
```
hperl, line 97: This function is very similar to `UseLoginCode`. Can we extract this out into a generic function?

persistence/sql/persister_login.go
Benehiko: Since there are two steps in the login process:
1. submit identifier
2. submit code

In the second submit, we still accept the identifier (again) from the user. Although we do a validation step to ensure that the identifiers match between the first submit and the second, I thought it might be best to also have the login flow attached to the identity from the first submit. 

This ensures that the second submit step has an additional check before getting the login _code. I guess since you would need to have the correct code and the flow id and somehow bypass the check before the persistence layer, it is unlikely that you would be able to circumvent it.

hperl (changes requested):
This looks very good! I just had some minor questions and one security concern regarding the code that checks the code.

persistence/sql/persister_login.go
Benehiko: hmm, i added this since you could have multiple identifiers  listed under the code credentials type. 

The user submits one of those identifiers, an email is sent out to all of their identifiers linked to the code credential. Each email contains a unique code. The second submit the user can submit any of the codes sent out.

See this identity schema https://github.com/ory/kratos/blob/f1a0fae19e32014a9151f320ed2ad6a903f129ea/test/e2e/profiles/code/identity.complex.traits.schema.json

selfservice/strategy/code/strategy_login.go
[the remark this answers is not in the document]
aeneasr: Ok, then we can also just use `p.Identifier` (since it was used to find the identifier in the DB).

persistence/sql/persister_registration.go
aeneasr: I did this already on review branch

persistence/sql/persister_login.go
aeneasr: I think you're both right. In the default case, where an identity has only one or two "login emails", I think that Alano is correct. As far as I can tell, multiple codes (related to one flow id) would only ever be sent out if the identity schema has more than one code credentials identifier.

However, if an identity schema allows an array of identifiers (so 1..n emails used for login), this could theoretically be abused.

I guess this can be fixed by sending EVERY address the same code. At the moment we send every address a different code, which makes brute forcing more likely because we reduce the search space.

Does that make sense?

Benehiko: hmm true, i didn't think of that :thinking: 

Yeah i agree, sending each the same code will prevent this.

aeneasr: I had to revert this logic, every address needs a unique code, otherwise we don't know which address got the code - and thus can't verify the correct email.

mayankmkh:
Hi, I am trying to use this flow with native API. I was wondering if this flow is available for Native or if is it browser only.

PR #3421, opened 2023-08-07
fix: false-positives for requiring re-authentication on update

### LINKED ISSUE #1889: Ask to re-authenticate despite only updating non-privileged fields.
**Describe the bug**

When wanting to update non-privileged traits like first/last names of an identity and the ```privileged_session_max_age``` has already passed I get an error stating the the login session needs to be refreshed (error.id === ```session_refresh_required```). From the documentation I see that this should only be the case for sensitive fields like the password, username and email.

image

The request payload I am sending (Note: neither the username, email or tos_accepted fields have changed. I only supplied them in the payload so it wont cause other errors telling me that required fields are missing.) I just want to update the first_name trait.

image

The response I got after waiting for ```privileged_session_max_age``` time to pass.

image

I am using the "Complete Settings Flow" in "profile" mode. Is this the correct approach? Or is updating such traits handled differently?

How can I achieve updating trivial traits without asking the user to refresh their session.

**Reproducing the bug**

Waiting till the ```privileged_session_max_age``` as passed and try to modify any traits that are not security critical.

**Expected behavior**

Updating of non-privileged traits like first/last name does not require re-authentication after the ```privileged_session_max_age``` has passed.

**Environment**

Docker v0.8.0

**Additional context**

None.

zepatrik:
closes #1889 

Somehow I could not reproduce the second case with the verified address, probably it is fixed already by accident. Added a test though.

### DIFF
--- a/identity/credentials.go
+++ b/identity/credentials.go
@@ -237,7 +237,14 @@ func CredentialsEqual(a, b map[CredentialsType]Credentials) bool {
 			return false
 		}
 
-		if !reflect.DeepEqual(expect.Identifiers, actual.Identifiers) {
+		expectIdentifiers, actualIdentifiers := make(map[string]struct{}, len(expect.Identifiers)), make(map[string]struct{}, len(actual.Identifiers))
+		for _, i := range expect.Identifiers {
+			expectIdentifiers[i] = struct{}{}
+		}
+		for _, i := range actual.Identifiers {
+			actualIdentifiers[i] = struct{}{}
+		}
+		if !reflect.DeepEqual(expectIdentifiers, actualIdentifiers) {
 			return false
 		}
 	}
--- a/identity/manager_test.go
+++ b/identity/manager_test.go
@@ -6,9 +6,10 @@ package identity_test
 import (
 	"context"
 	"fmt"
-	"github.com/ory/x/pointerx"
 	"testing"
 	"time"
+
+	"github.com/ory/x/pointerx"
 
 	"github.com/gofrs/uuid"
 
--- a/identity/manager_test.go
+++ b/identity/manager_test.go
@@ -6,6 +6,7 @@ package identity_test
 import (
 	"context"
 	"fmt"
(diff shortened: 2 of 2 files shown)


identity/credentials.go
```diff
@@ -237,7 +237,14 @@ func CredentialsEqual(a, b map[CredentialsType]Credentials) bool {
...
 
-		if !reflect.DeepEqual(expect.Identifiers, actual.Identifiers) {
+		expectIdentifiers, actualIdentifiers := make(map[string]struct{}, len(expect.Identifiers)), make(map[string]struct{}, len(actual.Identifiers))
+		for _, i := range expect.Identifiers {
+			expectIdentifiers[i] = struct{}{}
+		}
+		for _, i := range actual.Identifiers {
+			actualIdentifiers[i] = struct{}{}
+		}
+		if !reflect.DeepEqual(expectIdentifiers, actualIdentifiers) {
```
jonas-jonas, line 247: Doesn't this just check whether both arrays have the same length at the end, since you're just filling it with empty structs? It doesn't compare the actual credentials (anymore?), right?

aeneasr: Looks like it uses i not as the index but the value - so the identifier. Should be good then? I guess this prevents issues with ordering?

aeneasr: It's `for key, value := range []string{"abcde"}` so yes

jonas-jonas: My question is more geared towards whether we can trust that both `a` & `b` contain the same set of credentials, because that is not tested anymore!

A better test would be to order these credentials by some indicator and do a deepequal then.

aeneasr: Not sure if I follow - we tested the identifiers before and are testing them now also? The credential itself (config for example) is tested in the code above :)

jonas-jonas: >The credential itself (config for example) is tested in the code above :)

I see, I think I understand the code now, but it's a bit confusing... Thanks for the explanation.

PR #3423, opened 2023-08-08
feat: hot-reload CORS origins

zepatrik:
needs https://github.com/ory/x/pull/713

With this PR, CORS allowed-origins will be hot-reloadable. Is there a good place to put an integration test? Only e2e I guess?


driver/config/cors_test.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expect, allowOrigin(tc.allowedOrigins, "https://www.ory.sh"))
+
+			assert.Equal(t, tc.expectOther, allowOrigin(tc.allowedOrigins, "https://google.com"))
+
+			// check for consistency with rs/cors
+			assert.Equal(t, tc.expect, cors.New(cors.Options{AllowedOrigins: tc.allowedOrigins}).
+				OriginAllowed(&http.Request{Header: http.Header{"Origin": []string{"https://www.ory.sh"}}}))
```
zepatrik, line 95: We could just use this instead of our own implementation, but feels a bit overkill as the library does a lot of other stuff as well. It is quite a strait-forward implementation anyway.

driver/config/cors.go
```diff
@@ -0,0 +1,53 @@
...
+		prefix, suffix, found := strings.Cut(o, "*")
+		if !found {
+			// not a pattern, check for equality
+			if o == origin {
+				return true
+			}
+			continue
+		}
+		// inspired by https://github.com/rs/cors/blob/066574eebbd0f5f1b6cd1154a160cc292ac1835e/utils.go#L15
+		if len(origin) >= len(prefix)+len(suffix) && strings.HasPrefix(origin, prefix) && strings.HasSuffix(origin, suffix) {
```
zepatrik, line 32: Copying the library's logic to not have breaking changes.

aeneasr:
> Only e2e I guess?

You can start the server and make an HTTP request to an endpoint with `HEAD` and the `Origin:` Header and expect to see correct `Cross-Origin-*` headers.

driver/config/config.go
```diff
@@ -341,6 +343,7 @@ func New(ctx context.Context, l *logrusx.Logger, stdOutOrErr io.Writer, opts ...
 		configx.WithStderrValidationReporter(),
 		configx.OmitKeysFromTracing("dsn", "courier.smtp.connection_uri", "secrets.default", "secrets.cookie", "secrets.cipher", "client_secret"),
 		configx.WithImmutables("serve", "profiling", "log"),
+		configx.WithExceptImmutables("serve.public.cors.allowed_origins", "serve.admin.cors.allowed_origins"),
```
aeneasr, line 346: Does admin even have cors?

Also what about cors enabled? I think that if cors is enabled and allowed origins is empty, it allows all origins. Not sure though.

aeneasr (approved):
LGTM please verify that empty domains != allow all domains.

driver/config/config.go
zepatrik: Yes correct, no origins means all origins.
To hot-reload "enabled", I guess we just always return false in the check function when it is not enabled?

zepatrik:
Manual testing confirmed it works.

zepatrik:
Tests are failing because of ory/x bump which does not work on mysql thanks to https://github.com/gobuffalo/pop/pull/829

aeneasr (approved):

Issue #3420, opened 2023-08-07
`return_to` parameter during registration/login api flows don't work for errors


openscript:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

When a user signs up or in via one of the social login provider a `return_to` parameter can be configured, to tell Kratos redirecting the user back to the native app, after the flow has completed. Unfortunately this only works, if the authentication is successful. In case of an error (like an account already exists with this identifier) the user is redirected to the browser self-service user interface.

### Reproducing the bug

The behavior can be reproduced with: https://silly-easley-16pa1lzvl4.projects.oryapis.com

1. Create a email/password account for a Google account of yours (don't use the social login with Google!) via https://silly-easley-16pa1lzvl4.projects.oryapis.com/ui
2. Verify this account
3. Initialize a registration api flow
    ```
    curl --request GET \
      --url 'https://silly-easley-16pa1lzvl4.projects.oryapis.com/self-service/registration/api?return_session_token_exchange_code=true&return_to=http%3A%2F%2Fopenscript.ch%2Fregistration'
    ```
1. Update the registration api flow
    ```
    curl --request POST \
      --url 'https://silly-easley-16pa1lzvl4.projects.oryapis.com/self-service/registration?flow=08d22e2b-900b-476d-9931-a7f6877999dc' \
      --header 'content-type: application/json' \
      --data '{
      "method": "oidc",
      "provider": "google",
      "traits": {
        "lang": "de",
        "tac": "1234",
        "dp": "4321"
      }
    }'
    ```
1. Open the link in a browser where Kratos wants to redirect you to.
2. Authenticate with the Google account you used before to create a email/password account.
3. Instead of being redirected to the `return_to` address you are redirected to https://silly-easley-16pa1lzvl4.projects.oryapis.com/ui/login?flow=34256857-1984-4e10-bc7c-e8967ba55cbc
[...]

PR #3430, opened 2023-08-14
feat: provide login hints when registration fails due to duplicate credentials/addresses

alnr:
When trying to sign up with email+password when that email has previously been used with GitHub SSO:
image

When trying to sign up with the same email+password twice:
image

### DIFF
--- a/identity/manager.go
+++ b/identity/manager.go
@@ -101,7 +101,7 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
 
 	if err := m.r.PrivilegedIdentityPool().CreateIdentity(ctx, i); err != nil {
 		if errors.Is(err, sqlcon.ErrUniqueViolation) {
-			return m.generateCredentialHint(ctx, err, i)
+			return m.findExistingAuthMethod(ctx, err, i)
 		}
 		return err
 	}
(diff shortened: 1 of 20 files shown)


identity/manager.go
```diff
@@ -95,13 +100,99 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
 	}
 
 	if err := m.r.PrivilegedIdentityPool().CreateIdentity(ctx, i); err != nil {
+		if errors.Is(err, sqlcon.ErrUniqueViolation) {
+			return m.generateCredentialHint(ctx, err, i)
```
zepatrik, line 104: Is this behavior something we should make configurable? It is revealing quite a lot of information, so not sure if this should be enabled by default.

text/message_validation.go
```diff
@@ -164,6 +164,15 @@ func NewErrorValidationDuplicateCredentials() *Message {
 	}
 }
 
+func NewErrorValidationDuplicateCredentialsCustomMessage(reason string) *Message {
+	return &Message{
+		ID:      ErrorValidationDuplicateCredentials,
```
zepatrik, line 169: Consider adding a new ID for this message.

zepatrik (changes requested):
Generally looks good, I just have a few questions and suggestions.

identity/manager.go
```diff
@@ -95,13 +100,99 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
...
 	}
 
 	trace.SpanFromContext(ctx).AddEvent(events.NewIdentityCreated(ctx, i.ID))
 	return nil
 }
 
+func (m *Manager) generateCredentialHint(ctx context.Context, e error, i *Identity) (err error) {
+	var hints []string
+	var message string
+	for _, va := range i.VerifiableAddresses {
```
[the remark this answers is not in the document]
Benehiko: hmm yeah this shouldn't be relied upon. The verifiable addresses can also be nil and if the identity schema does not setup the account verification schema extension, then it won't work. 

A better approach is to get all the credentials that exist on this account through its identifier (which you already have). It might be an expensive task and can lead to exposing sensitive account information (as @zepatrik mentioned).

I would make it configurable.

identity/manager.go
```diff
@@ -95,13 +100,99 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
...
+		conflictingAddress, err := m.r.PrivilegedIdentityPool().FindVerifiableAddressByValue(ctx, va.Via, va.Value)
+		if errors.Is(err, sqlcon.ErrNoRows) {
+			continue
+		}
+		if err != nil {
+			m.r.Logger().WithError(err).Error("Failed to generate credential hint")
+			continue
+		}
+		message = fmt.Sprintf("An account for %s already exists.", va.Value)
+		hintIdentity, err := m.r.PrivilegedIdentityPool().GetIdentity(ctx, conflictingAddress.IdentityID, ExpandCredentials)
```
Benehiko, line 126: ah yeah, you're already querying the credentials here, so I guess move this up and query by the identifier.

identity/manager.go
```diff
@@ -95,13 +100,99 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
...
+	case CredentialsTypeOIDC:
+		var cfg CredentialsOIDC
+		_ = json.Unmarshal(c.Config, &cfg)
+		for _, provider := range cfg.Providers {
+			hints = append(hints, fmt.Sprintf("Please sign in with %s instead.", provider.Provider))
+		}
+		return hints
+	case CredentialsTypeWebAuthn:
+		var cfg CredentialsWebAuthnConfig
+		_ = json.Unmarshal(c.Config, &cfg)
```
Benehiko, line 154: ignoring the error from unmarshal

identity/manager.go
```diff
@@ -95,13 +100,99 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
...
+		hints = append(hints, "Alternatively, please recover your account from the sign-in page.")
+	}
+	return &ErrDuplicateCredentials{e, message, hints}
+}
+
+func toPretty(c *Credentials) (hints []string) {
+	switch c.Type {
+	case CredentialsTypeOIDC:
+		var cfg CredentialsOIDC
+		_ = json.Unmarshal(c.Config, &cfg)
```
Benehiko, line 147: and here

identity/manager.go
```diff
@@ -95,13 +100,99 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
...
+		for _, provider := range cfg.Providers {
+			hints = append(hints, fmt.Sprintf("Please sign in with %s instead.", provider.Provider))
+		}
+		return hints
+	case CredentialsTypeWebAuthn:
+		var cfg CredentialsWebAuthnConfig
+		_ = json.Unmarshal(c.Config, &cfg)
+		for _, webauthn := range cfg.Credentials {
+			if webauthn.IsPasswordless {
+				hints = append(hints, "Please sign passwordless using %s instead.", webauthn.DisplayName)
```
Benehiko, line 157: Error message doesn't make sense. Can display name be empty? maybe we need a fallback?

```suggestion
				hints = append(hints, "Please sign in with passwordless on %s instead.", webauthn.DisplayName)
```

identity/manager.go
```diff
@@ -95,13 +100,99 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
...
+		_ = json.Unmarshal(c.Config, &cfg)
+		for _, provider := range cfg.Providers {
+			hints = append(hints, fmt.Sprintf("Please sign in with %s instead.", provider.Provider))
+		}
+		return hints
+	case CredentialsTypeWebAuthn:
+		var cfg CredentialsWebAuthnConfig
+		_ = json.Unmarshal(c.Config, &cfg)
+		for _, webauthn := range cfg.Credentials {
+			if webauthn.IsPasswordless {
```
Benehiko, line 156: what's the difference between the if and else clause?

selfservice/flow/registration/error.go
```diff
@@ -25,11 +26,9 @@ import (
...
-	ErrAlreadyLoggedIn                = herodot.ErrBadRequest.WithID(text.ErrIDAlreadyLoggedIn).WithError("you are already logged in").WithReason("A valid session was detected and thus registration is not possible.")
-	ErrRegistrationDisabled           = herodot.ErrBadRequest.WithID(text.ErrIDSelfServiceFlowDisabled).WithError("registration flow disabled").WithReason("Registration is not allowed because it was disabled.")
-	ErrDuplicateCredentials           = herodot.ErrBadRequest.WithError(text.NewErrorValidationDuplicateCredentials().Text)
-	ErrDuplicateCredentialsOnOIDCLink = herodot.ErrBadRequest.WithError(text.NewErrorValidationDuplicateCredentialsOnOIDCLink().Text)
+	ErrHookAbortFlow        = errors.New("aborted registration hook execution")
+	ErrAlreadyLoggedIn      = herodot.ErrBadRequest.WithID(text.ErrIDAlreadyLoggedIn).WithError("you are already logged in").WithReason("A valid session was detected and thus registration is not possible.")
+	ErrRegistrationDisabled = herodot.ErrBadRequest.WithID(text.ErrIDSelfServiceFlowDisabled).WithError("registration flow disabled").WithReason("Registration is not allowed because it was disabled.")
```
Benehiko, line 31: I don't think we should delete error messages like this. Can't we rather mark them as deprecated somehow? Some users might be matching these error messages in their code, or using e2e tests to match it.

text/message_validation.go
Benehiko: I agree, this needs to be marked as deprecated somehow and then we need to have this new error message have its own ID. 

we don't know if customers rely on the errors ID in their codebase to make the correct decision.

Benehiko (changes requested):

Benehiko:
Another thought, couldn't we maybe instead of showing an error just return back the nodes the user does have an account with? and if no account, just the error message?

aeneasr:
> Another thought, couldn't we maybe instead of showing an error just return back the nodes the user does have an account with? and if no account, just the error message? How the message is structured feels a bit out of place and there isn't any link or button that forces the user into a different method. I know this might be more work, but just thought I'd share the thought even if it isn't viable right now.

Yes, as a follow-up issue! Showing a text message is an interim quick fix, but this needs way more attention after (but also a proper design proposal on how the process would work exactly).

identity/manager.go
```diff
@@ -110,34 +110,126 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
...
+			}
+			return &ErrDuplicateCredentials{
+				error:                e,
+				availableCredentials: []CredentialsType{cred.Type},
+				identifierHint:       identifierHint,
+			}
+		case CredentialsTypeOIDC:
+			var cfg CredentialsOIDC
+			if err := json.Unmarshal(cred.Config, &cfg); err != nil {
+				return errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to JSON decode identity credentials %s for identity %s.", cred.Type, found.ID))
```
alnr, line 194: Since this is for a hint only, IMO we should degrade gracefully here. That is, log the error and continue.

identity/manager.go
```diff
@@ -110,34 +110,126 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
...
+			return &ErrDuplicateCredentials{
+				error:                  e,
+				availableCredentials:   []CredentialsType{cred.Type},
+				availableOIDCProviders: available,
+				identifierHint:         foundConflictAddress,
+			}
+		case CredentialsTypeWebAuthn:
+			var cfg CredentialsWebAuthnConfig
+			if err := json.Unmarshal(cred.Config, &cfg); err != nil {
+				return errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to JSON decode identity credentials %s for identity %s.", cred.Type, found.ID))
```
alnr, line 211: Same here.

identity/manager.go
aeneasr: Hm but this state should never happen (tm) so it's good if we see an alert for this? It says that the data is corrupt

identity/manager.go
aeneasr: Hm but this state should never happen (tm) so it's good if we see an alert for this? It says that the data is corrupt

Sytten:
Is there a setting or will that always be enabled? This is a another source of account enumeration so just wondering.

hperl:
> Is there a setting or will that always be enabled? This is a another source of account enumeration so just wondering.

This is configurable, and off by default:

https://github.com/ory/kratos/pull/3430/files#diff-4a79b8650ca1bb0046229c6603ab7458e7711fc69812f13bbb8fc73b81b833e6R1197-R1202

alnr:
> Is this related with #3420?

No, I don't believe so. This PR is about giving better error messages when attempting to sign up with an existing credential.

PR #3437, opened 2023-08-16
fix: don't return 500 on conflict for POST /admin/identities

alnr:
This is a quick fix to avoid returning status code 500 when attempting to create a new identity which conflicts with an existing one.

As a follow-up, we should improve the diagnostics in the response.

### DIFF
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -421,7 +421,7 @@ func (h *Handler) create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
 
 	if err := h.r.IdentityManager().Create(r.Context(), i); err != nil {
 		if errors.Is(err, sqlcon.ErrUniqueViolation) {
-			h.r.Writer().WriteError(w, r, herodot.ErrConflict)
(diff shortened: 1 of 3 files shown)


identity/handler.go
```diff
@@ -419,7 +420,11 @@ func (h *Handler) create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
 	}
 
 	if err := h.r.IdentityManager().Create(r.Context(), i); err != nil {
-		h.r.Writer().WriteError(w, r, err)
+		if errors.Is(err, sqlcon.ErrUniqueViolation) {
+			h.r.Writer().WriteError(w, r, herodot.ErrConflict)
```
aeneasr, line 424:
```suggestion
			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrConflict.WithReason("This identity conflicts with another identity that already exists.")))
```

Benehiko (approved):

zepatrik (approved):

PR #3456, opened 2023-08-25
chore: code review for `code` strategy (magic code login)

aeneasr:
* [x] Unify code validation logic in the persister in one method
* [x] Remove state engine and simplifies code structure
* [x] Remove SMS fillers (we will add them when they are needed)
* [x] Write tests for email address normalization (capitalization)
* [x] Add unit tests
* [x] Fix failing e2e tests
* [x] Fix flaking e2e mail processing
* [ ] ~Mobile e2e tests~


persistence/sql/persister_code.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type oneTimeCodeProvider interface {
+	GetID() uuid.UUID
+	Validate() error
+	TableName(ctx context.Context) string
+	GetHMACCode() string
+}
+
+func useOneTimeCode[P any, U interface {
```
aeneasr, line 22: This code was the same for login, verification, recovery, and sign up, so I moved it into a common function. Uses some generic magic to make it work nicely. Tried it with duck typing only at first, but it caused issues and would have required reflection.

identity/extension_credentials_test.go
```diff
@@ -72,6 +72,21 @@ func TestSchemaExtensionCredentials(t *testing.T) {
 			},
 			ct: identity.CredentialsTypeWebAuthn,
 		},
+		{
```
aeneasr, line 75: These did not have tests

selfservice/strategy/code/strategy.go
```diff
@@ -294,169 +297,37 @@ func (s *Strategy) PopulateMethod(r *http.Request, f flow.Flow) error {
...
-		return c.alreadyValidatedCodeState(ctx, c.f, c.strategy, c.payload)
-	default:
-		return errors.WithStack(errors.New("Unknown flow state"))
+	if f.GetState() == "" {
+		f.SetState(flow.StateChooseMethod)
 	}
-	return nil
-}
 
-func (s *Strategy) NextFlowState(f flow.Flow) {
```
aeneasr, line 451: This code already exists in `flow.NextState()`

selfservice/strategy/code/strategy_login.go
```diff
@@ -129,157 +124,142 @@ func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow,
...
+	case flow.StateEmailSent:
+		i, err := s.loginVerifyCode(ctx, r, f, &p)
 		if err != nil {
-			return err
+			return nil, s.HandleLoginError(r, f, &p, err)
 		}
+		return i, nil
+	case flow.StatePassedChallenge:
+		return nil, s.HandleLoginError(r, f, &p, errors.WithStack(schema.NewNoLoginStrategyResponsible()))
+	}
```
aeneasr, line 151: IMO, this is easier to read and understand what's going on. Is state choose_method? Do this. Is state passed_challenge? Do that.

With the callbacks, it wasn't very clear what is executed when, because `CreateCodeHandler` does not indicate that the state is `choose_method`. Plus, we save a lot of code complexity :)

test/e2e/cypress/support/commands.ts
```diff
@@ -1269,20 +1266,20 @@ Cypress.Commands.add(
...
         let mailItem: any
         if (email) {
           mailItem = response.body.mailItems.find((m: any) =>
             m.toAddresses.includes(email),
           )
           if (!mailItem) {
-            return req
+            tries++
+            cy.wait(pollInterval)
+            return req()
```
aeneasr, line 1278: This caused flakes - because no waiting and req returned as a function, not a function call.

aeneasr (commented):
@Benehiko fyi

selfservice/strategy/code/code_sender.go
```diff
@@ -70,9 +70,14 @@ func (s *Sender) SendCode(ctx context.Context, f flow.Flow, id *identity.Identit
 		WithSensitiveField("address", addresses).
 		Debugf("Preparing %s code", f.GetFlowName())
 
+	// We generate the code once and use it for all addresses. This is important because
```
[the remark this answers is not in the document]
aeneasr: My bad, we actually need to generate unique codes per address, or it will not be possible to distinguish which address was verified, when multiple addresses are sent a code. So yeah, we must make sure to send an individual code for every distinct address.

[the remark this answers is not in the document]
aeneasr: My bad, we actually need to generate unique codes per address, or it will not be possible to distinguish which address was verified, when multiple addresses are sent a code. So yeah, we must make sure to send an individual code for every distinct address.

test/e2e/cypress/integration/profiles/code/registration/error.spec.ts
```diff
@@ -36,13 +36,12 @@ context("Registration error messages with code method", () => {
...
         cy.submitCodeForm()
 
-        cy.url().should("contain", "registration")
         cy.get('[data-testid="ui/message/1040005"]').should(
           "contain",
           "An email containing a code has been sent to the email address you provided",
         )
 
-        cy.get(' input[name="code"]').type("invalid-code")
+        cy.get('input[name="code"]').type("invalid-code")
```
aeneasr, line 44: Not sure what the spaces are here, but they're not needed.

test/e2e/cypress/integration/profiles/code/registration/error.spec.ts
```diff
@@ -75,12 +77,14 @@ context("Registration error messages with code method", () => {
 
         cy.get('input[name="traits.email"]').type(email)
         cy.submitCodeForm()
-
-        cy.url().should("contain", "registration")
+        cy.get('[data-testid="ui/message/1040005"]').should(
```
aeneasr, line 80: @Benehiko the most important aspect of e2e tests are wait assertions. Before:

1. Go to `/registration`
2. Submit form
3. Wait for URL to be `registration` (it already is registration though!)
4. Flaking test failure because sometimes the wait condition is done before the submit is done

Instead, use proper wait conditions that wait for e.g. specific elements such as this one.

test/e2e/cypress/support/commands.ts
```diff
@@ -1269,19 +1269,23 @@ Cypress.Commands.add(
...
+          if (filtered.length === 0) {
             tries++
             cy.wait(pollInterval)
             return req()
           }
+
+          expect(filtered.length).to.equal(expectedCount)
+          mailItem = filtered[0]
         } else {
+          expect(count).to.equal(expectedCount)
```
aeneasr, line 1285: Counting did not work correctly, caused flakes.

test/e2e/cypress/integration/profiles/code/registration/success.spec.ts
```diff
@@ -109,29 +114,28 @@ context("Registration success with code method", () => {
...
+        )
 
-        cy.url().should("contain", "registration")
         cy.getRegistrationCodeFromEmail(email).should((code) => {
-          cy.get(` input[name=code]`).type(code)
+          cy.get(`input[name=code]`).type(code)
           cy.get("button[name=method][value=code]").click()
         })
 
-        cy.deleteMail({ atLeast: 1 })
```
aeneasr, line 122: All methods like `cy.getRegistrationCodeFromEmail` remove mail automatically. This condition leads to us waiting for about a minute to delete an email, which then times out (and passes silently) because no email was found.

Issue #466, opened 2020-06-02
Courier messages polling the database


jonaslimads:
Hello,
great software you're doing here!

I've enabled Postgres logging and seen a new query in second:

```sql
SELECT courier_messages.body, courier_messages.created_at, courier_messages.id, courier_messages.recipient, courier_messages.status, courier_messages.subject, courier_messages.type, courier_messages.updated_at FROM courier_messages AS courier_messages WHERE status != $1 ORDER BY created_at ASC LIMIT 10
```
which happens because of the for in:
https://github.com/ory/kratos/blob/8050e33a2a4593307fa32feba6c448b850aed3fc/courier/courier.go#L126

While this works, it can lead to some problems in case the database gets a heavy load.
One solution is making it "reactive" via goroutine or something like that. 
I can give it a try coding a solution for this, at least to learn your code :)

Another question: is it possible to disable SMTP connection altogether?

Thank you!

aeneasr:
Thank you! We'll probably make this configurable in the future :) PRs welcomed

aeneasr:
I am closing this issue as it has not received any engagement from the community or maintainers in a long time. That does not imply that the issue has no merit. If you feel strongly about this issue

- open a PR referencing and resolving the issue;
- leave a comment on it and discuss ideas how you could contribute towards resolving it;
- open a new issue with updated details and a plan on resolving the issue.

We are cleaning up issues every now and then, primarily to keep the 4000+ issues in our backlog in check and to [prevent](https://www.jeffgeerling.com/blog/2020/saying-no-burnout-open-source-maintainer) [maintainer](https://www.jeffgeerling.com/blog/2016/why-i-close-prs-oss-project-maintainer-notes) [burnout](https://docs.brew.sh/Maintainers-Avoiding-Burnout). Burnout in open source maintainership is a [widespread](https://opensource.guide/best-practices/#its-okay-to-hit-pause) and serious issue. It can lead to severe personal and health issues as well as [enabling catastrophic](https://haacked.com/archive/2019/05/28/maintainer-burnout/) [attack vectors](https://www.gradiant.org/en/blog/open-source-maintainer-burnout-as-an-attack-surface/).

Thank you to anyone who participated in the issue! 🙏✌️

frederikhors:
@aeneasr I think we should reopen this because the database is flooded with calls all the time for this problem!

PR #3460, opened 2023-08-29
feat: increase courier watchMessages polling delay from 1 second to 30

frederikhors:
The database is flooded with calls all the time for this!

Related to https://github.com/ory/kratos/issues/466.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/courier/courier.go
+++ b/courier/courier.go
@@ -107,6 +107,6 @@ func (c *courier) watchMessages(ctx context.Context, errChan chan error) {
 			errChan <- err
 			return
 		}
-		time.Sleep(time.Second)
+		time.Sleep(time.Second * 30)
 	}
 }


aeneasr (changes requested):
Thank you for the PR @frederikhors

I don't think this is a good solution though. It will take up to 30 seconds + email send delay, before the email arrives at the user. So the user has to wait half a minute before they receive their account recovery code, or their login code. That's unfortunately not acceptable and we would like to keep the watcher at one second for the time being.

frederikhors:
Ok. I get that but I think 30 seconds is a decent wait time for anyone.

What about 15?

jonas-jonas:
Maybe, instead of using a hard-coded value, we can make this configurable, by adding an optional key `courier.dispatch-interval`, that defaults to 1s?

aeneasr:
Telling the truth, I think this should just stay as is. Adding one QPS to the database should not really have a big effect on overall query performance. I also don't want this to be configurable in Ory Network.

frederikhors:
I think the suggestion to make it configurable is good. Please. Please.

aeneasr:
What exactly is the problem you are trying to solve with this? Is there a negative effect you are experiencing?

If configurable, it should be at most a CLI flag, but not a configuration setting in the YAML config.

frederikhors:
> Is there a negative effect you are experiencing?

Yep. A very small DB (many DBS) and no need to send courier emails.

If you prefer to use a CLI flag ok. Thanks anyway.

aeneasr:
If you don’t need to send emails from Kratos, you could also just not run the courier? 🤔

frederikhors:
> If you don’t need to send emails from Kratos, you could also just not run the courier? 🤔

Sorry I meant not as fast as 1 second delay.

I'm fine waiting 30 seconds or more.

aeneasr:
https://github.com/ory/kratos/pull/3601 will make this configurable

PR #3467, opened 2023-08-30
fix: issue session after verification after registration with OIDC SSO

hperl:
This PR fixes an issue where the ID token claims from the identity are empty because the consent screen cannot fetch the session because it was not persisted yet.

To fix this, we now always persist the session when the post-registration hooks are executed. Note that this does not alter the behaviour of giving the session to the client.

### DIFF
--- a/selfservice/hook/session_issuer.go
+++ b/selfservice/hook/session_issuer.go
@@ -6,7 +6,6 @@ package hook
 import (
 	"context"
 	"net/http"
-	"time"
 
 	"github.com/ory/kratos/identity"
 	"github.com/ory/kratos/ui/node"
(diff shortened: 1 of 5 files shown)


aeneasr (changes requested):
Please persist the session only once during the registration lifecycle. The problem here being that every write to the sessions table incurs a RT latency of about 500-700ms.

The session is also persisted in the `SessionIssuer.ExecutePostRegistrationPostPersistHook` hook.

Please also note that `s.AuthenticatedAt = time.Now().UTC()` might be missing here.

aeneasr:
@hperl I made a mistake here - the persisting is happening across 2 different flows. I think that's fine!

zepatrik (approved):

selfservice/hook/session_issuer.go
```diff
@@ -53,11 +52,6 @@ func (e *SessionIssuer) ExecutePostRegistrationPostPersistHook(w http.ResponseWr
 }
 
 func (e *SessionIssuer) executePostRegistrationPostPersistHook(w http.ResponseWriter, r *http.Request, a *registration.Flow, s *session.Session) error {
-	s.AuthenticatedAt = time.Now().UTC()
```
aeneasr, line 56: @hperl did you check that the session can not be modified by any hooks, in which case we would need to update it here?

aeneasr (approved):

selfservice/hook/session_issuer.go
zepatrik: Yes we checked, the hooks all modify the identity, the session is not modified by any hook.

PR #3472, opened 2023-08-31
feat: add ability to convert session to JWT when calling whoami

### LINKED ISSUE #2487: Consider implementing JSON Web Token Exchange endpoint
### Preflight checklist

### Describe your problem

In some cases, developers want to receive a JSON Web Token to be used with APIs that do not neccessarily need/want to call Ory's APIs.

This is useful in cases where it is acceptable to interpret a session as active even if it has logged out within the last (for example) 5 minutes. Using a JSON Web Token at the backend can significantly reduce latency in the service as no API of Ory has to be called.

Having an endpoint such as `/sessions/tokenize?format=jwt` which generates a JSON Web Token with a customizable payload and expiry time, for example

```
{
  "sub": "1234"
  "traits": {
    "email": "..."
  }
  ...
}
```

Customization of the JSON Web Token would probably be done using a JsonNet snippet.

Setting the expiry time would be done through a global config value. It could either be a fixed amount, or a maximum amount (e.g. `1h`) where the caller indicates their preferred expiry time.

Issue JSON Web Tokens at the ingress - requires work at the implementor to add this capability to their e.g. API Gateway.

### Version

master

### Additional Context

_No response_

aeneasr:
This patch adds a query parameter `tokenize_as` to `/session/whoami` which encodes the session to a JWT. It is possible to customize the JWT claims by using a JsonNet template, and furthermore change the expiry of the token.

The tokenize feature supports multiple templates, which makes it easy to use the resulting JWT in a variety of use cases.

Closes #2487


session/handler.go
```diff
@@ -124,6 +126,13 @@ type toSession struct {
 	//
 	// in: header
 	Cookie string `json:"Cookie"`
+
+	// Return the session additional in token format like a JWT
+	//
+	// The value of this parameter has to be a valid, configured Ory Session token template.
+	//
+	// in: query
+	Tokenize string `json:"tokenize"`
```
jonas-jonas, line 135:
```suggestion
	TokenizeWith string `json:"tokenize_with"`
```

Just `tokenize` sounds like it's a boolean value.

session/tokenizer.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		sessionRaw, err := json.Marshal(session)
+		if err != nil {
+			return errors.WithStack(herodot.ErrInternalServerError.WithWrap(err).WithReasonf("Unable to encode session to JSON."))
+		}
+
+		claimsRaw, err := json.Marshal(&claims)
+		if err != nil {
+			return errors.WithStack(herodot.ErrInternalServerError.WithWrap(err).WithReasonf("Unable to encode session to JSON."))
```
jonas-jonas, line 111:
```suggestion
			return errors.WithStack(herodot.ErrInternalServerError.WithWrap(err).WithReasonf("Unable to encode claims to JSON."))
```

session/tokenizer.go
```diff
@@ -0,0 +1,147 @@
...
+		"sub": session.IdentityID.String(),
+		"sid": session.ID.String(),
+		"nbf": now.Unix(),
+		"iat": now.Unix(),
+	}
+
+	if mapper := tpl.ClaimsMapperURL; len(mapper) > 0 {
+		jn, err := fetch.FetchContext(ctx, mapper)
+		if err != nil {
+			return err
```
jonas-jonas, line 101: no `errors.WithStack`?

session/handler.go
aeneasr: That's true! I'm also not super happy with the name right now!

- tokenize_with
- tokenize_to
- tokenize_template

Which one of these should we use?

session/tokenizer.go
aeneasr: `fetch.FetchContext` already adds stack context afaik since it's in ory/x

session/handler.go
aeneasr: I opted for `tokenize_to`

session/handler.go
```diff
@@ -124,6 +126,13 @@ type toSession struct {
 	//
 	// in: header
 	Cookie string `json:"Cookie"`
+
+	// Return the session additional in token format like a JWT
```
jonas-jonas, line 130:
```suggestion
	// Returns the session additionally as a token (such as a JWT)
```

Also, add a link to the docs? (Once they are available)
And a note about the caching behavior?

session/handler.go
```diff
@@ -124,6 +126,13 @@ type toSession struct {
...
+	//
+	// The value of this parameter has to be a valid, configured Ory Session token template.
+	//
+	// in: query
+	TokenizeTo string `json:"tokenize_as"`
```
jonas-jonas, line 135:
```suggestion
	TokenizeAs string `json:"tokenize_as"`
```

session/handler_test.go
```diff
@@ -211,6 +212,32 @@ func TestSessionWhoAmI(t *testing.T) {
...
+		assert.EqualValues(t, http.StatusOK, res.StatusCode, string(body))
+
+		token := gjson.GetBytes(body, "tokenized").String()
+		require.NotEmpty(t, token)
+		segments := strings.Split(token, ".")
+		require.Len(t, segments, 3, token)
+		decoded, err := base64.RawURLEncoding.DecodeString(segments[1])
+		require.NoError(t, err)
+
+		assert.NotEmpty(t, gjson.GetBytes(decoded, "sub").String(), decoded)
```
zepatrik, line 237:
```suggestion
		assert.NotEmpty(t, gjson.GetBytes(decoded, "sub").Str, decoded)
```
because `.String()` converts all types to string. With `.Str` you can ensure it is a string in JSON, not a number, null, or object.

session/manager.go
```diff
@@ -142,6 +142,9 @@ type Manager interface {
 	// MaybeRedirectAPICodeFlow for API+Code flows redirects the user to the return_to URL and adds the code query parameter.
 	// `handled` is true if the request a redirect was written, false otherwise.
 	MaybeRedirectAPICodeFlow(w http.ResponseWriter, r *http.Request, f flow.Flow, sessionID uuid.UUID, uiNode node.UiNodeGroup) (handled bool, err error)
+
+	// TokenizeSession sets a session's tokenized value.
+	TokenizeSession(ctx context.Context, template string, session *Session) error
```
zepatrik, line 147: Is this even used?

zepatrik (approved):
Very nice 🎉 
I only have a few final remarks, but it would be fine to merge IMO right away.

session/handler.go
```diff
@@ -221,11 +243,19 @@ func (h *Handler) whoami(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
 	// s.Devices = nil
 	s.Identity = s.Identity.CopyWithoutCredentials()
```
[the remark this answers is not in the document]
zepatrik: It might make more sense to later provide a new `/sessions/v2/whoami` API that would introduce `expand` across all APIs in a unique way.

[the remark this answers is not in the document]
aeneasr: I think `/sessions/v2/whoami` is the way to go

embedx/config.schema.json
```diff
@@ -2597,6 +2597,44 @@
...
+              "title": "Tokenizer configuration",
+              "description": "Configure the tokenizer, responsible for converting a session into a token format such as JWT.",
+              "type": "object",
+              "properties": {
+                "templates": {
+                  "title": "Tokenizer templates",
+                  "description": "A list of different templates that govern how a session is converted to a token format.",
+                  "type": "object",
+                  "patternProperties": {
+                    "[a-zA-Z0-9-_.]+": {
```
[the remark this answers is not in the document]
aeneasr: Prevent users from using spaces etc, which then need to be properly encoded in the URL

session/manager.go
aeneasr: nope :) deleted

PR #3476, opened 2023-09-04
feat: support native social sign using apple sdk

jonas-jonas:
How this works:

1. Use the Apple SDK on the device to request authentication, which returns an IDToken
2. Submit that token to Kratos using updateRegistrationFlow, alongside any traits you want the identity to have
3. Kratos verifies the token using Apple's public JWKS (make this configurable for the generic provider?)
4. Kratos extracts the data from the IDToken
5. Kratos processes the login/registration as if it came from an OIDC callback

Expo example (_PR in repo following_):
```ts
import * as AppleAuthentication from "expo-apple-authentication"
import * as Crypto from "expo-crypto"
import * as Random from "expo-random"

async function signInWithApple() {
  const nonce = Random.getRandomBytes(16).toString()

  const digest = await Crypto.digestStringAsync(
    Crypto.CryptoDigestAlgorithm.SHA256,
    nonce,
  )
  const credential = await AppleAuthentication.signInAsync({
    requestedScopes: [
      AppleAuthentication.AppleAuthenticationScope.EMAIL,
      AppleAuthentication.AppleAuthenticationScope.FULL_NAME,
    ],
    nonce: digest,
  })

  return orySdk.updateRegistrationFlow({
    flow: flow.id,
    updateRegistrationFlowBody: {
      provider: "apple",
      id_token: credential.identityToken,
      raw_id_token_nonce: nonce,
      traits: {
        name: {
          first: credential.fullName?.givenName || "given name", // When developing, these values might be empty
          last: credential.fullName?.familyName || "last name", // When developing, these values might be empty
        },
      },
    },
  })
}
```
## Related issue(s)

Docs PR: https://github.com/ory/docs/pull/1528

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


selfservice/strategy/oidc/provider_apple.go
```diff
@@ -146,3 +147,21 @@ func decodeQuery(query url.Values, claims *Claims) {
 		}
 	}
 }
+
+var _ IDTokenVerifier = new(ProviderApple)
+
+func (a *ProviderApple) Verify(ctx context.Context, rawIDToken string) (*Claims, error) {
```
jonas-jonas, line 153: One limitation of this implementation is that we don't request ATs & RTs during this flow, so if users want to use them, on behalf of their users, this wouldn't work. 

I am not sure what use cases there are for using an AT against Apple's API, but I'd suggest we can implement the `authCode` flow at a later time. It should be trivial, but [doesn't work with the way Expo has implemented Apple authentication](https://github.com/expo/expo/issues/16162), as it requests the tokens using the client ID `host.exp.Exponent`.

selfservice/strategy/oidc/provider_apple.go
```diff
@@ -146,3 +147,21 @@ func decodeQuery(query url.Values, claims *Claims) {
...
+	keySet := oidc.NewRemoteKeySet(ctx, "https://appleid.apple.com/auth/keys")
+	verifier := oidc.NewVerifier("https://appleid.apple.com", keySet, &oidc.Config{
+		ClientID: a.config.ClientID,
```
jonas-jonas, line 156: This is the only reason you'd need a configured apple client in the OIDC config, as it checks whether the IDToken was issued to the same `client_id`. As mentioned above, this means that for development using expo, you'd need to set it to `host.exp.Exponent`. 

I am not 100% sure this is secure enough, but since we verify that the token was issued by Apple, it _should_ be fine™️.

In production apps, the token would be issued to `sh.ory.iosapp` on the device, and the Ory project/configuration would need a client with the same `client_id`. `apple_team_id`, `apple_private_key_id` an `apple_private_key` would be ignored in this flow.

selfservice/strategy/oidc/provider_config.go
```diff
@@ -116,61 +117,41 @@ type ConfigurationCollection struct {
 	Providers       []Configuration `json:"providers"`
 }
 
+// !!! WARNING !!!
+//
+// If you add a provider here, please also add a test to
+// provider_private_net_test.go
+var supportedProviders = map[string]func(config *Configuration, reg dependencies) Provider{
```
jonas-jonas, line 124: This refactoring is needed to be able to mock a new provider in that can selectively verify or fail verification of an ID Token.

I've double-checked, that I migrated all providers, but *please* check this again! Thank you.

selfservice/strategy/oidc/provider_apple.go
```diff
@@ -146,3 +147,21 @@ func decodeQuery(query url.Values, claims *Claims) {
...
+	})
+	token, err := verifier.Verify(ctx, rawIDToken)
+	if err != nil {
+		return nil, err
+	}
+	claims := &Claims{}
+	if err := token.Claims(claims); err != nil {
+		return nil, err
+	}
+	return claims, nil
```
jonas-jonas, line 166: Any ideas on how to test this are appreciated. I thought about adding a test that just uses a real `idToken` and verifies that, but that token contains my personal email address and user ID, which I'd rather not leak here. :)

I guess we could create a dummy Apple ID, obtain the token and add that here, however we would need to refresh the token, as Apple rotates their JWKs. Also, not sure how to deal with expiry.

jonas-jonas (commented):
I am not sure, we can really e2e test this, as there is no other provider that supports verifying an ID Token from the payload, for now. 

I guess, we could make the generic provider support it and add a JWKS URL to the config? Then we could obtain an ID token from a Hydra instance and submit that.

selfservice/strategy/oidc/provider_apple.go
aeneasr: Generate a JWT with the same claims as apple, and set up a fake HTTP server with a JWK you control - then verify. That should work!

selfservice/strategy/oidc/provider_apple.go
aeneasr: Please document this as a caveat.

aeneasr (commented):
Looks very good, please add three tests to strategy_test:

- [ ] Test login with an ID token, ensure claims end up properly
- [ ] Test registration with an ID token, ensure claims end up properly
- [ ] Test login/registration with an ID token where the Client ID is not matching (should fail)
- [ ] Test login/registration with an ID token where the nonce is not matching (should fail)

Please also make it possible to submit the nonce and validate the nonce as well. Fail the flow if the nonce is not set.

selfservice/strategy/oidc/provider_apple.go
jonas-jonas: Yup, that works. I needed to make the JWKS URL configurable via a private field, which defaults to the apple URL.

selfservice/strategy/oidc/provider_apple.go
jonas-jonas: Done in the Docs PR.

selfservice/strategy/oidc/provider.go
```diff
@@ -52,6 +57,8 @@ type Claims struct {
 	UpdatedAt           int64                  `json:"updated_at,omitempty"`
 	HD                  string                 `json:"hd,omitempty"`
 	Team                string                 `json:"team,omitempty"`
+	Nonce               string                 `json:"nonce,omitempty"`
+	NonceSupported      bool                   `json:"nonce_supported,omitempty"`
```
aeneasr, line 61: Is this an actual claim returned by Apple/Google?

selfservice/strategy/oidc/strategy.go
```diff
@@ -589,3 +591,38 @@ func (s *Strategy) CompletedAuthenticationMethod(ctx context.Context) session.Au
...
+			return nil, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("No nonce was included in the id_token but is required by the provider"))
+		}
+		sh := sha256.New()
+		sh.Write([]byte(idTokenNonce))
+		hashedNonce := hex.EncodeToString(sh.Sum(nil))
+
+		if hashedNonce != claims.Nonce {
+			return nil, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The supplied nonce does not match the nonce from the id_token"))
+		}
+	}
```
aeneasr, line 625: Why is this being hex encoded? From the OIDC specification:

> **nonce:** String value used to associate a Client session with an ID Token, and to mitigate replay attacks. The value is passed through unmodified from the Authentication Request to the ID Token. If present in the ID Token, Clients MUST verify that the nonce Claim Value is equal to the value of the nonce parameter sent in the Authentication Request. If present in the Authentication Request, Authorization Servers MUST include a nonce Claim in the ID Token with the Claim Value being the nonce value sent in the Authentication Request. Authorization **Servers SHOULD perform no other processing on nonce values used. The nonce value is a case sensitive string.**

https://openid.net/specs/openid-connect-core-1_0.html

https://github.com/coreos/go-oidc/blob/f0117a5f178cb36710dde42ede2154071b20ee94/example/idtoken/app.go#L109-L117

If this check is Apple-specific, it should live in the apple strategy IMO.

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -81,6 +81,24 @@ type UpdateLoginFlowWithOidcMethod struct {
...
+	// Supported providers are
+	// - Apple
+	// required: false
+	IDToken string `json:"id_token,omitempty"`
+
+	// RawIDTokenNonce is the nonce, used when generating the IDToken.
+	// If the provider supports nonce validation, the nonce will be validated against this value and required.
+	//
+	// required: false
+	RawIDTokenNonce string `json:"raw_id_token_nonce,omitempty"`
```
aeneasr, line 101: Why is this prefixed `raw_`? IMO it should be `id_token_nonce`

selfservice/strategy/oidc/strategy_test.go
```diff
@@ -549,7 +552,216 @@ func TestStrategy(t *testing.T) {
...
+				idToken: `{
+					"iss": "https://appleid.apple.com",
+					"sub": "{{sub}}"
+				}`,
+				expect: func(t *testing.T, res *http.Response, body []byte) {
+					require.Equal(t, "No nonce was included in the id_token but is required by the provider", gjson.GetBytes(body, "error.reason").String(), "%s", body)
+				},
+			},
+			{
+				name: "should fail if no nonce is supplied in request",
```
aeneasr, line 632: Missing test: nonce mismatch

selfservice/strategy/oidc/provider_apple_test.go
```diff
@@ -4,11 +4,21 @@
...
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"testing"
+	"time"
 
+	_ "embed"
+
+	"github.com/form3tech-oss/jwt-go"
```
aeneasr, line 18: Please use `golang-jwt/jwt-go/v4`

selfservice/strategy/oidc/provider.go
aeneasr: I see, this is returned by Apple. I don't think this is a standard claim but rather something specific to Apple. In that case, I would suggest to only handle this claim in the apple provider. The problem here is that this bool value will default to "false", which disables the nonce check if the value is not available (which is the case with any other platform). So basically, this is a bypass for the nonce check, built in and enabled per default.

aeneasr (changes requested):
The nonce implementation does not look correct to me. Typically, a nonce is a plain string (like the state parameter) and you do a string equality match to verify if it is valid or not. The `nonce_supported` flag defaults to false, which *may* be OK for apple, but is definitely not OK for any other provider, as this claim is not a standard claim (it is never mentioned here: https://openid.net/specs/openid-connect-core-1_0.html).

Additionally, the Apple documentation states:

> A Boolean value that indicates whether the transaction is on a nonce-supported platform. If you send a nonce in the authorization request, but don’t see the nonce claim in the identity token, check this claim to determine how to proceed. If this claim returns true, treat nonce as mandatory and fail the transaction; otherwise, you can proceed treating the nonce as optional.

However, as far as I can see, we use this claim first to check whether we should validate. But the documentation states:

1. If a nonce was sent (aka if we receive the nonce as part of the registration / login payload
2. And the nonce is not available in the id_token
3. Then (and only then) we check this claim

The current implementation however:

1. Checks if the claim is true
2. Proceeds with validation

My suggestion is to move this custom validation logic into the apple ID token verification:

1. The apple provider validates the `nonce` claim:
  1. If a nonce was sent as part of the payload
  2. And the id_token has a nonce -> we validate
  3. And the id_token has no nonce
    * We check the nonce_supported claim
    * If true: fail
    * If false: proceed
2. The Google provider also validates the nonce (but always, because Google returns this always)
3. The generic provider does not validate the nonce at all, it delegates the ID token validation to the specific provider (we can fix this later if desired).

Please don't merge this without another review from myself!

selfservice/strategy/oidc/strategy.go
jonas-jonas: Fixed. Thanks for the sources.

selfservice/strategy/oidc/strategy_login.go
jonas-jonas: Fixed, misunderstand of how nonces work on my part.

selfservice/strategy/oidc/provider_apple_test.go
jonas-jonas: Done.

selfservice/strategy/oidc/strategy_test.go
jonas-jonas: Fixed, nice catch.

selfservice/strategy/oidc/strategy.go
```diff
@@ -589,3 +589,39 @@ func (s *Strategy) CompletedAuthenticationMethod(ctx context.Context) session.Au
...
+
+	if err := claims.Validate(); err != nil {
+		return nil, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("The id_token claims were invalid").WithError(err.Error()))
+	}
+
+	// First check if the JWT contains the nonce claim.
+	if claims.Nonce == "" {
+		// If it doesn't, check if the provider supports nonces.
+		if verifier.NonceSupported(claims) {
```
jonas-jonas, line 610: We could also just require nonces to be present in the token regardless of the provider. If someone wants to integrate with an Apple device that does not support the nonce, we can still add this in later.

aeneasr (commented):
Much better! Almost there

selfservice/strategy/oidc/strategy.go
jonas-jonas: Done! The check is a bit less readable now, if you have an idea on improvement, LMK.

aeneasr (approved):
Awesome!

hperl (approved):

PR #3483, opened 2023-09-06
chore: speed up e2e tests

jonas-jonas:
A few test changed a lot of config values and waited for the whole config update in beforeEachs which could take a few seconds per test. This added up. This PR changes it, so that these tests update the config just once or twice during their runtime. 

Also fixes the broken tests.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -137,6 +137,7 @@ jobs:
       TEST_DATABASE_MYSQL: "mysql://root:test@(localhost:3306)/mysql?parseTime=true&multiStatements=true"
       TEST_DATABASE_COCKROACHDB: "cockroach://root@localhost:26257/defaultdb?sslmode=disable"
     strategy:
+      fail-fast: false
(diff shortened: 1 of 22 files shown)


.github/workflows/ci.yaml
```diff
@@ -137,6 +137,7 @@ jobs:
       TEST_DATABASE_MYSQL: "mysql://root:test@(localhost:3306)/mysql?parseTime=true&multiStatements=true"
       TEST_DATABASE_COCKROACHDB: "cockroach://root@localhost:26257/defaultdb?sslmode=disable"
     strategy:
+      fail-fast: false
```
jonas-jonas, line 140: It's pretty annoying that the other runs are cancelled if one failed, that way the logs don't contain the end summary, and it's hard to tell if the issue is DB specific or happens on the other runs as well.

aeneasr (approved):

PR #3487, opened 2023-09-11
feat: support auth_type param

hperl:
The Facebook OIDC provider supports an auth_type parameter that
    when set to "reauthenticate" will force the user to
    reauthenticate (similar to `prompt=login` for other Providers).

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


selfservice/strategy/oidc/provider.go
```diff
@@ -85,6 +86,7 @@ func UpstreamParameters(provider Provider, upstreamParameters map[string]string)
 		"login_hint": {},
 		"hd":         {},
 		"prompt":     {},
+		"auth_type":  {},
```
jonas-jonas, line 89: Would it be too confusing to just translate `prompt=login` to `auth_type=reauthenticate` for Facebook? Seems like, we're adding too many provider specific things.

hperl: I'm not sure I like provider-specific magic here. I think it's best to defer this to the user.

mblarsen: I'd say let's defer it to the user as well.

mblarsen (approved):

PR #3520, opened 2023-09-19
fix: error handling on identity import

zepatrik:
When importing identities without any traits, or with malformed traits, 500s are returned. This improves the error handling and messaging.
Per default, missing traits now get rewritten to an empty object. Otherwise sjson would produce invalid JSON, and JSON parse errors would be returned even though the payload was valid.

### DIFF
--- a/identity/validator.go
+++ b/identity/validator.go
@@ -6,6 +6,10 @@ package identity
 import (
 	"context"
 
+	"github.com/pkg/errors"
+
+	"github.com/ory/herodot"
+
 	"github.com/tidwall/sjson"
 
 	"github.com/ory/kratos/driver/config"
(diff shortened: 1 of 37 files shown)


identity/validator.go
```diff
@@ -47,9 +51,13 @@ func (v *Validator) ValidateWithRunner(ctx context.Context, i *Identity, runners
 		return err
 	}
 
+	if len(i.Traits) == 0 {
+		i.Traits = []byte(`{}`)
+	}
```
zepatrik, line 56: This is the fix. Without, `sjson.SetRawBytes` below would produce invalid JSON, although the payload is valid JSON.

aeneasr (approved):

aeneasr (approved):

aeneasr (approved):

PR #3526, opened 2023-09-21
feat: allow fuzzy-search on credential identifiers

zepatrik:
This PR adds the ability to search for sub-strings and similar strings in credential identifiers.

Note that the **postgres** and **CRDB** migrations create special indexes useful for this feature. To use [online schema changes](https://www.cockroachlabs.com/docs/v23.1/online-schema-changes) with cockroach, we recommend to manually copy the index definition and run it before applying migrations. The migration will then be a no-op.
If you run on **mysql** (or **sqlite**), no special index is created. If desired, you can create such an index manually, and it would be highly appreciated if you could contribute its definition :pray:

This feature is a preview and might change in behavior! Similarity search is not expected to return deterministic results anyway, but are only really useful for humans.


persistence/sql/identity/persister_identity.go
```diff
@@ -666,56 +666,91 @@ func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.
...
+		wheres = fmt.Sprintf(`
+AND (ic.nid = ? AND ici.nid = ? AND ici.identifier %s ?)
+AND ict.name IN (?, ?)`, identifierOperator)
+		args = append(args, nid, nid, identifier, identity.CredentialsTypeWebAuthn, identity.CredentialsTypePassword)
+	}
+
+	// Follow up: add page token support here, will be easy.
+	paginator := pop.NewPaginator(params.Page, params.PerPage)
+
+	if err := con.RawQuery(fmt.Sprintf(`SELECT DISTINCT identities.*
```
zepatrik, line 704: This query is now build here because pop does not allow `DISTINCT` queries. We need this, because in many cases the query would return an identity twice, once for the `webauthn` and once for the `password` credential type.

persistence/sql/identity/persister_identity.go
```diff
@@ -666,56 +666,91 @@ func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.
...
+				identitiesByID[addr.IdentityID].VerifiableAddresses = append(identitiesByID[addr.IdentityID].VerifiableAddresses, addr)
+			}
+		case identity.ExpandFieldRecoveryAddresses:
+			addrs := make([]identity.RecoveryAddress, 0)
+			if err := con.Where("nid = ?", nid).Where("identity_id IN (?)", identityIDs).Order("id").All(&addrs); err != nil {
+				return nil, sqlcon.HandleError(err)
+			}
+			for _, addr := range addrs {
+				identitiesByID[addr.IdentityID].RecoveryAddresses = append(identitiesByID[addr.IdentityID].RecoveryAddresses, addr)
+			}
```
aeneasr, line 753: I guess this could probably run in parallel at some point

identity/handler.go
```diff
@@ -133,11 +133,19 @@ type listIdentitiesResponse struct {
...
 	// required: false
 	// in: query
 	CredentialsIdentifier string `json:"credentials_identifier"`
+
+	// CredentialsIdentifierSimilar is the (partial) identifier (username, email) of the credentials to look up using similarity search.
+	// Only one of CredentialsIdentifier and CredentialsIdentifierSimilar can be used.
+	//
+	// required: false
+	// in: query
+	CredentialsIdentifierSimilar string `json:"credentials_identifier_similar"`
```
aeneasr, line 148: WDYT of prefixing this as preview or something? We will probably need a separate search endpoint to deal with different use cases like:

- identifier search
- verifiable /recovery addresses search
- traits search

Alternatively this could also be added to a new endpoint marked as experimental. For me it's primarily about finding *the* way we will be solving search in Ory to get a consistent API experience. That's why I think that this should be somehow marked as: this api will be changed once we have an appropriate search design doc

aeneasr (commented):
Nice!

persistence/sql/migrations/sql/20230920171028000000_identity_search_index.cockroach.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX IF NOT EXISTS identity_credential_identifiers_nid_identifier_gin ON identity_credential_identifiers USING GIN (nid, identifier gin_trgm_ops);
```
zepatrik, line 1: To use [online schema changes](https://www.cockroachlabs.com/docs/v23.1/online-schema-changes) with cockroach, we recommend to manually copy this index definition and run it before applying migrations.

persistence/sql/identity/persister_identity.go
```diff
@@ -666,56 +666,91 @@ func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.
...
+	identifier := params.CredentialsIdentifier
+	identifierOperator := "="
+	if identifier == "" && params.CredentialsIdentifierSimilar != "" {
+		identifier = params.CredentialsIdentifierSimilar
+		identifierOperator = "%"
+		switch con.Dialect.Name() {
+		case "postgres", "cockroach":
+		default:
+			identifier = "%" + identifier + "%"
+			identifierOperator = "LIKE"
```
zepatrik, line 682: This is the mysql and sqlite case. Mysql does support full-text and fuzzy search, and special index types to allow them. However, it was not easy to understand at all, and I found no easy way to have the same behavior as postgres/CRDB. I think this is already an improvement on the current situation, and this feature is open for contributions!

persistence/sql/identity/persister_identity.go
```diff
@@ -699,7 +699,7 @@ AND ict.name IN (?, ?)`, identifierOperator)
 	}
 
 	// Follow up: add page token support here, will be easy.
-	paginator := pop.NewPaginator(params.Page, params.PerPage)
+	paginator := pop.NewPaginator(params.Page+1, params.PerPage)
```
aeneasr, line 702: Sure about this? 😅 Spent some time recently on making sure this works now properly.

zepatrik: Yes, tests were failing without, and it was there previously (see diff).

test/schema/fixtures/config.schema.test.success/selfServiceAfterSettings.full.yaml
zepatrik, line 1: Changes here needed because of test failures introduced in #3519

.github/workflows/ci.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       - sdk-generate
     services:
       postgres:
-        image: postgres:9.6
+        image: postgres:11.8
```
zepatrik, line 31: Postgres 9 is EoL by now, and 11 is currently the oldest supported version.

aeneasr (approved):

PR #3537, opened 2023-09-27
feat: webhook analytic events

alnr:
https://github.com/ory-corp/cloud/issues/4392

This defines two new analytics events: `WebhookDelivered` and `WebhookFailed`.

Semantics are such that we get a `WebhookDelivered` event each time we get an HTTP response from a webhook endpoint, one for each retry. It doesn't say anything about whether Kratos considers that response OK or not. This event contains the full bytes of the HTTP request/response traffic.

The `WebhookFailed` event is emitted when a webhook invocation is considered failed. On the one hand, that's error conditions like:
- connection timeout
- hostname not found
- jsonnet evaluation failed

Such errors will not emit a `WebhookDelivered` event, because we didn't get a HTTP response.

We also emit a `WebhookFailed` event if the endpoint returns a HTTP 404 or some other junk. But since we did get an HTTP response, we also get one or more `WebhookDelivered` events beforehand for the same webhook.

Do those semantics make sense?

Added a config field to the `web_hook` block which allows you to opt-out of emitting these events.


selfservice/hook/web_hook.go
```diff
@@ -296,27 +295,30 @@ func (e *WebHook) execute(ctx context.Context, data *templateContext) error {
...
+			// canceled when the parent context is canceled.
 			//
-			// The webhook will still cancel after 30 seconds as that is the configured timeout for the HTTP client.
-			var cancel context.CancelFunc
-			ctx = trace.ContextWithSpan(context.Background(), trace.SpanFromContext(ctx))
-			ctx, cancel = context.WithTimeout(ctx, 5*time.Minute)
-			defer cancel()
+			// The webhook will still cancel after 30 seconds as that is the
+			// configured timeout for the HTTP client.
+			ctx = context.WithoutCancel(ctx)
```
aeneasr, line 303: 👍

aeneasr (dismissed):

selfservice/hook/web_hook.go
```diff
@@ -499,3 +501,18 @@ func isTimeoutError(err error) bool {
...
+		attempt = 0
+		reqBody []byte
+	)
+	httpClient.RequestLogHook = func(_ retryablehttp.Logger, req *http.Request, retryNumber int) {
+		attempt = retryNumber + 1
+		reqBody, _ = httputil.DumpRequestOut(req, true)
+	}
+	httpClient.ResponseLogHook = func(_ retryablehttp.Logger, res *http.Response) {
+		resBody, _ := httputil.DumpResponse(res, true)
+		trace.SpanFromContext(ctx).AddEvent(events.NewWebhookDelivered(ctx, res.Request.URL, reqBody, res.StatusCode, resBody, attempt))
```
jonas-jonas, line 516: Both `reqBody` and `resBody` can be arbitrarily long. Should we limit this somehow? IIUC, if the event attributes are too large, the event is just dropped. But such events should still show up, right?

jonas-jonas, line 516: Can we add some kind of unique identifier per webhook dispatch? E.g. a login flow triggers a webhook, which is retried a few times and then succeeds. This leads to multiple events in the storage. However, it's hard to tell which events belong together.

alnr: Very good point!

alnr: Also a good idea!

alnr: Added a limit for resBody. For reqBody, I want to add a limit to `jsonnetsecure` instead. That's a DOS vector anyway and needs to be fixed there.

alnr: done

selfservice/hook/web_hook.go
```diff
@@ -499,3 +504,23 @@ func isTimeoutError(err error) bool {
...
+		reqBody   []byte
+	)
+	httpClient.RequestLogHook = func(_ retryablehttp.Logger, req *http.Request, retryNumber int) {
+		attempt = retryNumber + 1
+		requestID = uuid.Must(uuid.NewV4())
+		req.Header.Set("Ory-Webhook-Request-ID", requestID.String())
+		reqBody, _ = httputil.DumpRequestOut(req, true)
+	}
+	httpClient.ResponseLogHook = func(_ retryablehttp.Logger, res *http.Response) {
+		res.Body = io.NopCloser(io.LimitReader(res.Body, 5<<20)) // read at most 5 MB from the response
```
zepatrik, line 521: This modifies the response itself, so it will also limit the body read in the webhook itself. Doesn't it make more sense to add a global limit there instead? Because this is only set when the `emitEvent` config is set correct?

zepatrik (commented):
The event semantics are not the most obvious IMO. What about having three event types instead:

1. webhook success
2. webhook error (network/...)
3. webhook failed (status code/...)

and only emit exactly one of those.
I feel like in the end we will have to distinguish exactly those cases in any observability tooling/UI anyway. With the current approach the number of events and their type in combination tell which case occurred, which makes it hard to e.g. display a `97%` success rate metric.

selfservice/hook/web_hook.go
alnr: I've added the `io.LimitReader` also to the code path where no event is emitted. Please check if that makes sense.

This is the easiest way to force a "global" limit, I think

selfservice/hook/web_hook.go
```diff
@@ -499,3 +504,23 @@ func isTimeoutError(err error) bool {
...
+
+func instrumentHTTPClientForEvents(ctx context.Context, httpClient *retryablehttp.Client) {
+	var (
+		attempt   = 0
+		requestID uuid.UUID
+		reqBody   []byte
+	)
+	httpClient.RequestLogHook = func(_ retryablehttp.Logger, req *http.Request, retryNumber int) {
+		attempt = retryNumber + 1
+		requestID = uuid.Must(uuid.NewV4())
```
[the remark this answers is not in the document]
alnr: `Ory-Network-Request-Id` identifies the incoming request though, doesn't it? The point of this header is to allow the webhook receiver to differentiate between different attempts (retries).

alnr:
> The event semantics are not the most obvious IMO. What about having three event types instead:
> 
> 1. webhook success
> 2. webhook error (network/...)
> 3. webhook failed (status code/...)
> 
> and only emit exactly one of those. I feel like in the end we will have to distinguish exactly those cases in any observability tooling/UI anyway. With the current approach the number of events and their type in combination tell which case occurred, which makes it hard to e.g. display a `97%` success rate metric.

The ambiguity between "error" and "failed" is exactly what I tried to avoid with this approach. Specifically, how do we transport information about retries to observability tooling, when we only get to emit one event? In my mind, we definitely want to be able to display each retry individually.

alnr:
How about:

- `WebhookSucceeded` (when overall invocation returns `err == nil`)
- `WebhookFailed` (when overall invocation returns `err != nil` after all retries)
- `WebhookDelivered` (when response received from webhook endpoint). Could also be called `WebhookResponseReceived` or something?

selfservice/hook/web_hook.go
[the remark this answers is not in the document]
zepatrik: Ah I see, but shouldn't we also add a way to correlate attempts, i.e. the `Ory-Network-Request-Id`? This would allow us to better correlate all attempts into one.

zepatrik (commented):
Ok let's maybe think about it this way:

For each attempt, we can have "Error" and "Success" outcomes. For the overall event, we also can have "Error" (when no attempt succeeded) and "Success" (when one attempt succeeded). As a developer, I care in the first place about the overall event success rate, but also about the average attempts it takes to get there. When debugging, I want to look into both  error conditions, where only individual attempts failed but also where the whole event failed.
How about having these 4 types of tracing events (sorry for this ambiguity). Each webhook event therefore consists of all attempt events, and the overall result event, so at least two tracing events.

alnr:
> Ok let's maybe think about it this way:
> 
> For each attempt, we can have "Error" and "Success" outcomes. For the overall event, we also can have "Error" (when no attempt succeeded) and "Success" (when one attempt succeeded). As a developer, I care in the first place about the overall event success rate, but also about the average attempts it takes to get there. When debugging, I want to look into both error conditions, where only individual attempts failed but also where the whole event failed. How about having these 4 types of tracing events (sorry for this ambiguity). Each webhook event therefore consists of all attempt events, and the overall result event, so at least two tracing events.

That's too complicated. Feels difficult to imagine a real scenario where someone would actually need all this data to diagnose the problem.
I'm also still not really convinced about using this event stream to do alerting. After all, the webhook itself _is_ just another type of event. When you've set up your webhook receiver, it's easy to install monitoring right there. You'd know if you stopped receiving hooks or started responding 500s.

> but also about the average attempts it takes to get there

Webhooks succeeding in the retry after failing the first attempt is likely quite rare and almost always transient. So I don't think this needs particular attention. That is beside the initial debugging, where as a user I explicitly want to see as precisely as possible which bytes where sent over the wire. But that's during debugging and initial setup only.

After thinking about this over lunch, here's what I'm going to do:

- `WebhookSucceeded` (when overall invocation returns err == nil)
- `WebhookFailed` (when overall invocation returns err != nil after all retries)
- `DebugWebhookDelivered` (when response received from webhook endpoint). We emit this event only when "Debug mode" is turned on in the console (turns itself off after 60min).

zepatrik:
> That's too complicated.

It is exactly the same as you have proposed, except that we clearly distinguish states. The one state (no attempt errors but overall failure) is just not relevant for real life scenarios.

> [...] need all this data [...]

IMO, it is always easier to drop data when you don't need them, then to be limited by the lack of data availability.

> [...] using this event stream to do alerting.

We're not really talking about alerting, but more about overall execution stats. If I have a 50% error rate, I should probably look at the integration and fix some bug. Single failures would not be too relevant overall, but relevant for debugging.
What would be an alternative approach we can take to solve this issue?

> Webhooks succeeding in the retry after failing the first attempt is likely quite rare and almost always transient.

Sure, but if this happens a lot I probably want to make my API faster to reduce the likeliness of timeouts, increase resources, ...

> [...] initial debugging, where as a user I explicitly want to see as precisely as possible which bytes where sent over the wire. But that's during debugging and initial setup only.

I agree on this point. From experience with stripe I can tell you that it is great to always have this kind of observability. When something failed in CI I can just take a look at the request & response immediately, and see what was happening. We can easily solve any storage size issues by reducing data retention, and maybe only storing the data for webhooks that failed in the end.

In my opinion, we should not restrict the features we might build later (or not) at this point already, but instead provide all necessary data to build any useful feature we want to. Deleting data we need is way easier than filling missing gaps.

alnr:
> IMO, it is always easier to drop data when you don't need them, then to be limited by the lack of data availability.

For the user: yes. For the implementor (us), it is the other way around.

> In my opinion, we should not restrict the features we might build later (or not) at this point already, but instead provide all necessary data to build any useful feature we want to. Deleting data we need is way easier than filling missing gaps.

Once we've exposed this data, we can never take it away again.

I've now added the `WebhookSucceeded` event. That should allow users to monitor successes against failures. For debugging, there is the additional `WebhookDelivered` event which we'll guard behind a "debug mode" setting.

To finally move this forward and unblock development of a UI, let's please merge this as-is. We'll see what additional features and data we need when we actually play with it. 👍

selfservice/hook/web_hook.go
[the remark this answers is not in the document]
alnr: This would be useful for all events and should go into https://github.com/ory/x/blob/master/otelx/semconv/events.go.

Out of scope for this PR.

zepatrik (approved):

PR #3549, opened 2023-10-04
feat: login with code on any credential type

Benehiko:
Should be able to login with the `code` credential even if the user did not register on the `code` credential. 
Only `identifier` matching is done and validation based on the identity schema.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/driver/config/config.go
+++ b/driver/config/config.go
@@ -228,8 +228,8 @@ type (
 	}
 	SelfServiceStrategyCode struct {
 		*SelfServiceStrategy
-		PasswordlessEnabled                bool `json:"passwordless_enabled"`
-		PasswordlessLoginWithAnyCredential bool `json:"passwordless_login_fallback_enabled"`
+		PasswordlessEnabled              bool `json:"passwordless_enabled"`
+		PasswordlessLoginFallbackEnabled bool `json:"passwordless_login_fallback_enabled"`
 	}
 	Schema struct {
 		ID  string `json:"id" koanf:"id"`
@@ -761,8 +761,8 @@ func (p *Config) SelfServiceCodeStrategy(ctx context.Context) *SelfServiceStrate
 			Enabled: pp.BoolF(basePath+".enabled", true),
 			Config:  config,
 		},
-		PasswordlessEnabled:                pp.BoolF(basePath+".passwordless_enabled", false),
-		PasswordlessLoginWithAnyCredential: pp.BoolF(basePath+".passwordless_login_fallback_enabled", false),
+		PasswordlessEnabled:              pp.BoolF(basePath+".passwordless_enabled", false),
+		PasswordlessLoginFallbackEnabled: pp.BoolF(basePath+".passwordless_login_fallback_enabled", false),
 	}
 }
 
--- a/driver/config/config.go
+++ b/driver/config/config.go
@@ -228,7 +228,8 @@ type (
 	}
 	SelfServiceStrategyCode struct {
 		*SelfServiceStrategy
-		PasswordlessEnabled bool `json:"passwordless_enabled"`
+		PasswordlessEnabled                bool `json:"passwordless_enabled"`
+		PasswordlessLoginWithAnyCredential bool `json:"passwordless_login_with_any_credential"`
 	}
 	Schema struct {
 		ID  string `json:"id" koanf:"id"`
@@ -760,7 +761,8 @@ func (p *Config) SelfServiceCodeStrategy(ctx context.Context) *SelfServiceStrate
 			Enabled: pp.BoolF(basePath+".enabled", true),
 			Config:  config,
 		},
-		PasswordlessEnabled: pp.BoolF(basePath+".passwordless_enabled", false),
+		PasswordlessEnabled:                pp.BoolF(basePath+".passwordless_enabled", false),
+		PasswordlessLoginWithAnyCredential: pp.BoolF(basePath+".passwordless_login_with_any_credential", false),
 	}
 }
 
--- a/driver/config/config.go
+++ b/driver/config/config.go
@@ -229,7 +229,7 @@ type (
 	SelfServiceStrategyCode struct {
 		*SelfServiceStrategy
 		PasswordlessEnabled                bool `json:"passwordless_enabled"`
-		PasswordlessLoginWithAnyCredential bool `json:"passwordless_login_with_any_credential"`
+		PasswordlessLoginWithAnyCredential bool `json:"passwordless_login_fallback_enabled"`
 	}
 	Schema struct {
 		ID  string `json:"id" koanf:"id"`
@@ -762,7 +762,7 @@ func (p *Config) SelfServiceCodeStrategy(ctx context.Context) *SelfServiceStrate
 			Config:  config,
 		},
 		PasswordlessEnabled:                pp.BoolF(basePath+".passwordless_enabled", false),
-		PasswordlessLoginWithAnyCredential: pp.BoolF(basePath+".passwordless_login_with_any_credential", false),
+		PasswordlessLoginWithAnyCredential: pp.BoolF(basePath+".passwordless_login_fallback_enabled", false),
 	}
 }
 
--- a/persistence/sql/identity/persister_identity.go
+++ b/persistence/sql/identity/persister_identity.go
@@ -297,6 +297,38 @@ func (p *IdentityPersister) createVerifiableAddresses(ctx context.Context, conn 
 	return batch.Create(ctx, &batch.TracerConnection{Tracer: p.r.Tracer(ctx), Connection: conn}, work)
 }
 
+func (p *IdentityPersister) FindIdentityByAnyCredentialIdentifier(ctx context.Context, identifier string) (*identity.Identity, error) {
+	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.FindIdentityByAnyCredentialIdentifier")
+	defer otelx.End(span, nil)
+
+	identifier = NormalizeIdentifier(identity.CredentialsTypePassword, identifier)
+
+	joins := `
+			INNER JOIN identity_credentials ic ON ic.identity_id = identities.id
+			INNER JOIN identity_credential_types ict ON ict.id = ic.identity_credential_type_id
+			INNER JOIN identity_credential_identifiers ici ON ici.identity_credential_id = ic.id`
+	wheres := `
+			AND (ic.nid = ? AND ici.nid = ? AND ici.identifier = ?)`
+
+	nid := p.NetworkID(ctx).String()
+	args := []any{nid, nid, identifier}
+	query := fmt.Sprintf(`
+		SELECT DISTINCT identities.*
+		FROM identities AS identities
+		%s
+		WHERE
+		%s
+		ORDER BY identities.id ASC
+		LIMIT 1`,
+		joins, wheres)
+
+	id := new(identity.Identity)
+
+	return id, sqlcon.HandleError(p.Transaction(ctx, func(ctx context.Context, connection *pop.Connection) error {
+		return connection.RawQuery(query, args...).All(&id)
+	}))
+}
+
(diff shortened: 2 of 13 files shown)


selfservice/strategy/code/strategy_login.go
```diff
@@ -97,16 +97,25 @@ func (s *Strategy) getIdentity(ctx context.Context, identifier string) (_ *ident
...
+		Expand:                identity.ExpandCredentials,
+	})
+
+	if err != nil || len(identities) == 0 {
 		return nil, nil, errors.WithStack(schema.NewNoCodeAuthnCredentials())
 	}
 
-	if len(cred.Identifiers) == 0 {
-		return nil, nil, errors.WithStack(schema.NewNoCodeAuthnCredentials())
+	i := identities[0]
```
Benehiko, line 109: I assume there will only be 1 identity with this credential identifier and not multiple.

selfservice/strategy/code/strategy_login.go
```diff
@@ -97,16 +97,25 @@ func (s *Strategy) getIdentity(ctx context.Context, identifier string) (_ *ident
...
+	for _, c := range i.Credentials {
+		for _, id := range c.Identifiers {
+			if strings.EqualFold(id, identifier) {
```
aeneasr, line 112: Missing normalization of email addresses

selfservice/strategy/code/strategy_login.go
```diff
@@ -97,16 +97,25 @@ func (s *Strategy) getIdentity(ctx context.Context, identifier string) (_ *ident
 	ctx, span := s.deps.Tracer(ctx).Tracer().Start(ctx, "selfservice.strategy.code.strategy.getIdentity")
 	defer otelx.End(span, &err)
 
-	i, cred, err := s.deps.PrivilegedIdentityPool().FindByCredentialsIdentifier(ctx, s.ID(), identifier)
-	if err != nil {
+	identities, err := s.deps.PrivilegedIdentityPool().ListIdentities(ctx, identity.ListIdentityParameters{
```
aeneasr, line 100: Yeah, this is definitely going to be slow, `ListIdentities` is a slow operation.

Instead this should probably be a dedicated query (similar to `FindCredentialsByIdentifier()`). Maybe you can change `FindCredentialsByIdentifier` so that the credentials type is optional and add it as a new method `FindAnyCredentialsByIdentifier(ctx, identifier, opts)` which then is used `FindCredentialsByIdentifier` -> `FindAnyCredentialsByIdentifier (ctx, identifier, OnlyCredentialType()`.

selfservice/strategy/code/strategy_login.go
```diff
@@ -97,16 +97,25 @@ func (s *Strategy) getIdentity(ctx context.Context, identifier string) (_ *ident
...
 		return nil, nil, errors.WithStack(schema.NewNoCodeAuthnCredentials())
 	}
 
-	if len(cred.Identifiers) == 0 {
-		return nil, nil, errors.WithStack(schema.NewNoCodeAuthnCredentials())
+	i := identities[0]
+	for _, c := range i.Credentials {
+		for _, id := range c.Identifiers {
+			if strings.EqualFold(id, identifier) {
+				return &i, &c, nil
```
aeneasr, line 113: This will return the first credential type it finds, like WebAuthn, TOTP, or whatever. This can have serious negative consequences later such as incorrectly overwriting the credentials. It should be explicit that we did not find the credential for code and are now in a "fault tolerant" code path that can still deal with the login.

aeneasr (changes requested):
Please gate this feature behind a feature flag along the lines of "I want to allow my users to send a code to their email even if they did not sign up with this method".

kmherrmann:
Unless I'm missing it, we still need the Feature flag / project level setting to enable/disable Code login for everyone.

selfservice/strategy/code/strategy_login.go
Benehiko: will do :+1:

persistence/sql/identity/persister_identity.go
```diff
@@ -297,6 +297,38 @@ func (p *IdentityPersister) createVerifiableAddresses(ctx context.Context, conn
...
+	args := []any{nid, nid, identifier}
+	query := fmt.Sprintf(`
+		SELECT DISTINCT identities.*
+		FROM identities AS identities
+		%s
+		WHERE
+		%s
+		ORDER BY identities.id ASC
+		LIMIT 1`,
+		joins, wheres)
```
jonas-jonas, line 323: Since you're not changing the `joins` and `wheres` variables after their declaration, I'd prefer seeing this all be one string. This way, it's hard to read what's going on.

And you can get rid of the `fmt.Sprintf` for the SQL statement, which can be potentially dangerous, if some one adjusts this code in the future (SQL injections!).

persistence/sql/identity/persister_identity.go
```diff
@@ -297,6 +297,38 @@ func (p *IdentityPersister) createVerifiableAddresses(ctx context.Context, conn
...
+
+	id := new(identity.Identity)
+
+	return id, sqlcon.HandleError(p.Transaction(ctx, func(ctx context.Context, connection *pop.Connection) error {
+		return connection.RawQuery(query, args...).All(&id)
```
jonas-jonas, line 328: Shouldn't this be `.First`? What would happen if there are more than one result?

```suggestion
		return connection.RawQuery(query, args...).First(&id)
```

selfservice/strategy/code/strategy_login.go
```diff
@@ -97,16 +97,32 @@ func (s *Strategy) getIdentity(ctx context.Context, identifier string) (_ *ident
...
+			id, err := s.deps.PrivilegedIdentityPool().FindIdentityByAnyCredentialIdentifier(ctx, identifier)
+			if err != nil {
+				return nil, nil, errors.WithStack(schema.NewNoCodeAuthnCredentials())
+			}
+
+			// create a new code credential
+			cred := identity.Credentials{Type: s.ID(), Identifiers: []string{identifier}}
+			// if err := id.SetCredentialsWithConfig(s.ID(), cred, &identity.CredentialsCode{UsedAt: sql.NullTime{}}); err != nil {
+			// 	return nil, nil, errors.WithStack(err)
+			// }
```
jonas-jonas, line 114: What about this? Isn't this needed?

jonas-jonas (changes requested):

selfservice/strategy/code/strategy_login.go
Benehiko: it's a WIP :)

jonas-jonas (changes requested):

test/e2e/cypress/integration/profiles/code/login/success.spec.ts
```diff
@@ -106,6 +111,62 @@ context("Login success with code method", () => {
...
+        cy.visit(route)
+
+        cy.get(Selectors[app]["identity"]).clear().type(email)
+        cy.submitCodeForm(app)
+
+        cy.getLoginCodeFromEmail(email).should((code) => {
+          cy.get(Selectors[app]["code"]).type(code)
+
+          cy.get(Selectors[app]["submit"]).click()
+        })
```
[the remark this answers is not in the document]
Benehiko: good catch!

aeneasr:
I've added persister tests and improved the config option naming a bit. I also adjusted the `getIdentity` method to be more explicit in the API contract. Should be good to go now!

driver/config/config.go
```diff
@@ -228,7 +228,8 @@ type (
 	}
 	SelfServiceStrategyCode struct {
 		*SelfServiceStrategy
-		PasswordlessEnabled bool `json:"passwordless_enabled"`
+		PasswordlessEnabled                bool `json:"passwordless_enabled"`
+		PasswordlessLoginWithAnyCredential bool `json:"passwordless_login_with_any_credential"`
```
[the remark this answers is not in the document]
Benehiko: i mean right now you still need to register an account, the login flow doesn't auto create accounts (except in oidc flows). We could adopt this strategy in future here as well, but this PR addresses when a user already has an account on a different credential, we could allow users to login with a one-time code even if they only have an account registered through password or oidc. 

making this configurable adds flexibility instead of just forcing one way or the other.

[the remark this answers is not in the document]
jonas-jonas: That's not what I mean. I just don't see a use case for only letting users login after they set up the code method for their account. Either the admin wants to allow login via code or they don't. IMO, this config option requires understanding of how we implemented this and doesn't really add value.

Another config option just adds complexity to the code and for users implementing Ory. We already have 100+ config options, that can be confusing to newcomers. And in this specific case, I just don't see the point of disabling it.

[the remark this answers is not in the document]
jonas-jonas: I also might be misunderstanding this.

IIUC, when enabled, this option allows users to login via code, even if we didn't store a credential in the `identity_credentials` table and use _any_ of the credentials stored in that table, even if it's _not_ for the `code` type. Is that correct?

If so, see above.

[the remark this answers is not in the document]
aeneasr: > IIUC, when enabled, this option allows users to login via code, even if we didn't store a credential in the `identity_credentials` table and use _any_ of the credentials stored in that table, even if it's _not_ for the `code` type. Is that correct?

That's correct and the discussion is warranted. Both sides have - IMO - valid arguments. If this PR is enabled per default, the login with code method is essentially the same as account recovery, because it allows someone to sign in even if they lost access to their credentials.

**Feature flag**
- Pro: By explicitly enabling this feature, the administrator has control over whether they accept a weaker / different type of proving authenticity than was originally set up by the user (e.g. using code method to "work around" a lost passkey). 
- Contra: Another config option and questionable UX

**Enable per default**
- Pro: One config option less, one way of doing it
- Contra: Login with code now essentially is a recovery flow. I can recover access to my account if I lost access to may passkeys.

I think the UX issues can be addressed by showing the user only the credentials they have set up for login.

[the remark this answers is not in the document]
kmherrmann: I'm not sure i follow 100% @jonas-jonas. I think the Admin needs to control if they allow Email OTP logins (and therefore also if it shows as an option in AX), or not; indeed, enabling it makes account recovery via emails kinda obsolete, but 🤷 

IMHO we should have this flag and not force it onto projects, but we should name it more clearly to convey it "enables Email Code login for all users". "any credential type" is somewhat confusing :)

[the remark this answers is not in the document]
jonas-jonas: I just don't get why you'd want to only allow signing in with a code if the user set it up. How would they even do it in this example scenario where this setting is disabled, but login via code is enabled: A user registers with a password. Now they want to log in with the code via email. How, e.g. what steps does the user need to take to do it?

With this option enabled, it's pretty easy: put in the email address and click "login via code".

But with `passwordless_login_with_any_credential` set to `false` on a project level: ???

And sure, it's essentially the recovery flow now, and in an effort to clean up the code base, we should probably re-use a lot of the code. But I don't see the issue with that. The recovery flow still has its own templates, and redirects to settings after. So it still serves its own purpose...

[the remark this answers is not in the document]
aeneasr: > I just don't get why you'd want to only allow signing in with a code if the user set it up. How would they even do it in this example scenario where this setting is disabled, but login via code is enabled: A user registers with a password. Now they want to log in with the code via email. How, e.g. what steps does the user need to take to do it?

You wouldn't be able to do that which is the point of this feature flag. It's true that you currently can't set up login with code after you have signed up. We initially wanted to add this to the credentials section in the settings screen but so far have not.

aeneasr (approved):

jonas-jonas (approved):
Since I am the only one against introducing the config option, I'll approve this now.

aeneasr:
Thanks Jonas! Let's talk about it more in-depth at the Ory Summit in person. In general I think it is a good idea to limit config options and I do understand where this is coming from!

PR #3563, opened 2023-10-09
feat: link oidc credentials when login

hperl:
Original PR: https://github.com/ory/kratos/pull/3222

When user tries to login with OIDC for the first time but has already registered before with email/password a credentials identifier conflict may be detected by Kratos. In this case user needs to login with email/password first and then link OIDC credentials on a settings screen.
This PR simplifies UX and allows user to link OIDC credentials to existing account right in the login flow, without
switching to settings flow.

## Related issue(s)
#2727 

## Checklist

      introduces a new feature.
- [ x] I am following the
      the changes.
      works.


selfservice/flow/registration/flow.go
```diff
@@ -267,3 +262,34 @@ func (f *Flow) GetFlowName() flow.FlowName {
...
+		return nil, err
+	}
+	id, err := uuid.FromString(internalContextOuterFlow.ID)
+	if err != nil {
+		return nil, err
+	}
+	return &id, nil
+}
+
+func (f *Flow) SetOuterLoginFlowID(flowID uuid.UUID) error {
```
hperl, line 288: I think we should make this a database column instead.

hperl:
@splaunov I cleaned up the linking logic a bit. Thanks again for laying the foundation! PTAL if you like :)

aeneasr (commented):
- [ ] Unit / functional tests
- [ ] Normalization working correctly? Test :)
- [ ] Edge case issue with verifiable/recovery addresses
- [ ] Consent step for linking (e.g. checkbox or change of label for submit button)

selfservice/flow/login/hook.go
```diff
@@ -321,3 +328,51 @@ func (e *HookExecutor) PreLoginHook(w http.ResponseWriter, r *http.Request, a *F
...
+// maybeLinkCredentials links the identity with the credentials of the inner context of the login flow.
+func (e *HookExecutor) maybeLinkCredentials(r *http.Request, s *session.Session, i *identity.Identity, f *Flow) error {
+	lc, err := f.DuplicateCredentials()
+	if err != nil {
+		return err
+	} else if lc == nil {
+		return nil
+	}
+
+	if err := e.checkDuplicateCredentialsIdentifierMatch(r.Context(), i.ID, lc.DuplicateIdentifier); err != nil {
```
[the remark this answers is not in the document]
hperl: Added a test.

identity/manager.go
```diff
@@ -110,12 +110,7 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
 	return nil
 }
 
-func (m *Manager) findExistingAuthMethod(ctx context.Context, e error, i *Identity) (err error) {
-	if !m.r.Config().SelfServiceFlowRegistrationLoginHints(ctx) {
-		return &ErrDuplicateCredentials{error: e}
-	}
-	// First we try to find the conflict in the identifiers table. This is most likely to have a conflict.
-	var found *Identity
+func (m *Manager) ConflictingIdentity(ctx context.Context, i *Identity) (found *Identity, foundConflictAddress string, err error) {
```
aeneasr, line 113: This should come with tests, now that it is being exported to prevent future regressions.

selfservice/flow/login/flow.go
```diff
@@ -231,6 +231,10 @@ func (f *Flow) EnsureInternalContext() {
 	}
 }
 
+func (f *Flow) GetInternalContext() *sqlxx.JSONRawMessage {
+	return &f.InternalContext
```
aeneasr, line 235: Why return a pointer?

selfservice/flow/login/hook.go
```diff
@@ -321,3 +328,51 @@ func (e *HookExecutor) PreLoginHook(w http.ResponseWriter, r *http.Request, a *F
...
+		return err
+	}
+
+	method := strategy.CompletedAuthenticationMethod(ctx)
+	sess.CompletedLoginForMethod(method)
+
+	return nil
+}
+
+func (e *HookExecutor) checkDuplicateCredentialsIdentifierMatch(ctx context.Context, identityID uuid.UUID, match string) error {
```
aeneasr, line 365: Test?

selfservice/flow/login/hook.go
```diff
@@ -321,3 +328,51 @@ func (e *HookExecutor) PreLoginHook(w http.ResponseWriter, r *http.Request, a *F
 
 	return nil
 }
+
+// maybeLinkCredentials links the identity with the credentials of the inner context of the login flow.
+func (e *HookExecutor) maybeLinkCredentials(ctx context.Context, sess *session.Session, ident *identity.Identity, loginFlow *Flow) error {
```
aeneasr, line 333: Test?

selfservice/flow/registration/flow.go
```diff
@@ -208,6 +202,10 @@ func (f *Flow) EnsureInternalContext() {
 	}
 }
 
+func (f *Flow) GetInternalContext() *sqlxx.JSONRawMessage {
+	return &f.InternalContext
```
aeneasr, line 206: Reconsider pointer here - sqlxx.JSONRawMessage is a byte slice which means that modifying it will change the mem value and not create a copy (I think).

schema/errors.go
```diff
@@ -349,3 +349,13 @@ func NewLoginCodeInvalid() error {
 		Messages: new(text.Messages).Add(text.NewErrorValidationLoginCodeInvalidOrAlreadyUsed()),
 	})
 }
+
+func NewLinkedCredentialsDoNotMatch() error {
+	return errors.WithStack(&ValidationError{
+		ValidationError: &jsonschema.ValidationError{
+			Message:     `linked credentials do not match`,
```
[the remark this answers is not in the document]
hperl: Nothing much, just restart a new flow. Should we do that for the user in this case?

selfservice/flow/login/flow.go
hperl: So that we can mutate the internal context as well.

selfservice/flow/registration/flow.go
hperl: We do want to change the internal context.

schema/errors.go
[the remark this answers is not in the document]
aeneasr: OK - I think you can also show a message "please try again". but creating a new flow would be even nicer!

selfservice/flow/login/flow.go
aeneasr: OK I see - you're using it here:

```
	*flow.GetInternalContext() = bytes
```

I guess that's fine (it works), just a bit unusual (would expect `SetInternalContext()`)

aeneasr: So what I mean by that is that it's unusual to have pointers to slices. Here's why:

```go
package main

func main() {
	var a []byte
	var b *[]byte
	b = &a
	if b == nil {
		print("nil")
	}
	print((*b)[0]) // panics
}
```

This problem exists here where you check whether the pointer is set, but not whether the internal context is actually set:

```
func SetDuplicateCredentials(flow InternalContexter, creds DuplicateCredentialsData) error {
	if flow.GetInternalContext() == nil {
		flow.EnsureInternalContext()
	}
	bytes, err := sjson.SetBytes(
		*flow.GetInternalContext(),
		internalContextDuplicateCredentialsPath,
		creds,
	)
	if err != nil {
		return err
	}
	*flow.GetInternalContext() = bytes

	return nil
}
```

Since `internalContext` is a []byte per default (and thus nil per default), you'll probably run into some errors here  in `SetBytes`.

identity/manager.go
hperl: fixed

selfservice/flow/login/flow.go
hperl: I am calling `flow.EnsureInternalContext()` if the slice is nil to make sure it is always allocated.

In your code example, the panic is not because the pointer points to nothing (it points to `&a`, which is on the stack), but because it is an empty slice and accessing the first element panics.

---

I can still add a SetInternalContext if you want, just to make the code easier to read. WDYT?

selfservice/flow/login/hook.go
hperl: done :)

selfservice/flow/login/hook.go
hperl: done :)

selfservice/flow/registration/flow.go
hperl: fixed

selfservice/flow/login/flow.go
hperl: fixed

test/e2e/cypress/integration/profiles/oidc/login/success.spec.ts
```diff
@@ -1,24 +1,26 @@
...
-      app: "react" as "react",
-      profile: "spa",
-    },
+    // {
+    //   login: react.login,
+    //   registration: react.registration,
+    //   settings: react.settings,
+    //   app: "react" as "react",
+    //   profile: "spa",
+    // },
```
jonas-jonas, line 15: revert?

jonas-jonas (changes requested):

selfservice/flow/duplicate_credentials.go
```diff
@@ -0,0 +1,61 @@
...
+
+type DuplicateCredentialsData struct {
+	CredentialsType     identity.CredentialsType
+	CredentialsConfig   sqlxx.JSONRawMessage
+	DuplicateIdentifier string
+}
+
+type InternalContexter interface {
+	EnsureInternalContext()
+	GetInternalContext() sqlxx.JSONRawMessage
```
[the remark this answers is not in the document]
hperl: Agreed, but this would conflict with an existing field name. Changing this would change much more code than ignoring the best practices here.

test/e2e/cypress/integration/profiles/oidc/login/success.spec.ts
hperl: Yes, good catch!

aeneasr (dismissed):

selfservice/flow/login/flow.go
aeneasr: Yes that's true, what I meant is this:

```
	if flow.GetInternalContext() == nil { // might evaluate to trfalseue because the pointer is not nil but the memory address is nil
		flow.EnsureInternalContext() // would set context to string "{}" but won't because pointer is not-nil even though value is nil
	}
	
	bytes, err := sjson.SetBytes(
		*flow.GetInternalContext(),
		internalContextDuplicateCredentialsPath,
		creds,
	)
	// err is now non-nil because `*flow.GetInternalContext()` is nil which `SetBytes` can't deal with a nil byte slice
```

However, I realized that `SetBytes()` can deal with a nil slice and won't return an error. So this was indeed fine, at the same time I think that the setter is a clearer pattern as it doesn't have this "hidden" nil-pointer versus nil-slice-value meaning.

PR #3565, opened 2023-10-09
fix: use keyset pagination list identities

alnr:
https://github.com/ory-corp/cloud/issues/5412

I'd consider this a MVP/stop-gap solution to the immediate problem. The code in ListIdentities needs an overhaul. Since removal of the query builder in https://github.com/ory/kratos/pull/3526, the code is constructed via printf and hard to read 😞.

Also, the assortment of pagination packages in ory/x are wholly confusing 😢 

Needs https://github.com/ory/x/pull/729


identity/handler_test.go
```diff
@@ -1492,51 +1488,42 @@ func TestHandler(t *testing.T) {
 				t.Parallel()
 				body, res := getFull(t, ts, fmt.Sprintf("/identities?per_page=%d", perPage), http.StatusOK)
 				assert.Len(t, body.Array(), perPage)
-				assert.Equal(t, strconv.Itoa(count), res.Header.Get("X-Total-Count"))
+				_ = res
```
jonas-jonas, line 1491: Reason for removal is we no longer send `X-Total-Count`?

```suggestion
```

identity/test/pool.go
```diff
@@ -759,7 +756,7 @@ func TestPool(ctx context.Context, conf *config.Config, p persistence.Persister,
 
 			t.Run("not if on another network", func(t *testing.T) {
 				_, on := testhelpers.NewNetwork(t, ctx, p)
-				actual, err := on.ListIdentities(ctx, identity.ListIdentityParameters{
+				actual, _, err := on.ListIdentities(ctx, identity.ListIdentityParameters{
```
jonas-jonas, line 759: Do we have any test that asserts that the `nextPage` returned here is the correct page?

internal/driver.go
```diff
@@ -77,8 +79,7 @@ func NewRegistryDefaultWithDSN(t testing.TB, dsn string) (*config.Config, *drive
 	ctx := context.Background()
 	c := NewConfigurationWithDefaults(t)
 	c.MustSet(ctx, config.ViperKeyDSN, stringsx.Coalesce(dsn, dbal.NewSQLiteTestDatabase(t)))
-
-	reg, err := driver.NewRegistryFromDSN(ctx, c, logrusx.New("", ""))
+	reg, err := driver.NewRegistryFromDSN(ctx, c, logrusx.New("", "", logrusx.ForceLevel(logrus.ErrorLevel)))
```
jonas-jonas, line 82: With this, the migration logs are hidden, correct? I guess this replaces https://github.com/ory/kratos/pull/3562 then?

x/pagination.go
```diff
@@ -18,3 +21,42 @@ func ParsePagination(r *http.Request) (page, itemsPerPage int) {
...
+		q.Del("per_page")
+		r.URL.RawQuery = q.Encode()
+	}
+	// ... and defaul to keyset pagination
+	keyset, err := keysetpagination.Parse(q, keysetpagination.NewStringPageToken)
+	if err != nil {
+		return nil, nil, herodot.ErrBadRequest.WithReason(err.Error())
+	}
+	return keyset, nil, nil
+}
```
jonas-jonas, line 62: Should we move this to ory/x instead?

identity/identity.go
```diff
@@ -132,6 +133,14 @@ type Identity struct {
 	OrganizationID uuid.NullUUID `json:"organization_id,omitempty"  faker:"-" db:"organization_id"`
 }
 
+func (i *Identity) PageToken() keysetpagination.PageToken {
+	return keysetpagination.StringPageToken(i.ID.String())
+}
+
+func DefaultPageToken() keysetpagination.PageToken {
+	return keysetpagination.StringPageToken(uuid.Nil.String())
```
jonas-jonas, line 141: Do we want to support sorting in the future? If so, this should probably be a `MapPageToken`. https://github.com/ory/kratos/blob/master/courier/message.go#L208-L213

https://github.com/ory/x/blob/master/pagination/keysetpagination/paginator.go#L87-L116

jonas-jonas (changes requested):

aeneasr (approved):

identity/handler_test.go
```diff
@@ -1492,51 +1488,42 @@ func TestHandler(t *testing.T) {
...
-						isLast, res = run(t, path, knownIDs)
-						if isLast {
-							return
-						}
-						path = fmt.Sprintf("/identities?page_size=%s&page_token=%s", res.Query().Get("page_size"), res.Query().Get("page_token"))
-					})
+				path := fmt.Sprintf("/identities?page_size=%d", perPage)
+				for {
+					pages++
+					next, _ := run(t, path, knownIDs)
```
aeneasr, line 1522: Add explicit assertion to ensure page and per_page are not set, or we might still use page pagination here.

identity/test/pool.go
```diff
@@ -88,7 +88,6 @@ func TestPool(ctx context.Context, conf *config.Config, p persistence.Persister,
 		})
 
 		t.Run("case=expand", func(t *testing.T) {
```
aeneasr, line 90: Could you add an explicit pagination test here for both approaches?

aeneasr (changes requested):
Two things regarding tests

identity/handler_test.go
alnr: `SELECT COUNT(*)` expensive and deprecated. We don't want to support it in the new pagination API anymore.

x/pagination.go
alnr: I'm back and forth on this. We have 5 pagination packages in ory/x already, and they are still not complete.

I _would_ like to improve and clean up the pagination packages, but that's for another PR.

identity/identity.go
alnr: We can still change that later, right?

identity/handler_test.go
zepatrik: See https://github.com/cockroachdb/cockroach/issues/79579 for more context on why we cannot support this anymore.

jonas-jonas: No, I get it, just wanted to make sure this was intended. 

But the `_ = res` can probably be removed?

Thanks for the explanation, though.

go.mod
```diff
@@ -77,7 +79,8 @@ require (
 	github.com/ory/jsonschema/v3 v3.0.8
 	github.com/ory/mail/v3 v3.0.0
 	github.com/ory/nosurf v1.2.7
-	github.com/ory/x v0.0.590
+	github.com/ory/x v0.0.591-0.20231009141749-f6c175b9bd1a
```
zepatrik, line 82: Bump to official version, PR is merged already.

identity/identity.go
jonas-jonas: Not entirely sure. I think it should be possible, but you never know if we break someone then. And if we're going to use the `MapPageToken` then anyway, we might as well do it now, just with one value. I don't think there are any downsides to that. But I haven't worked with it in a while, so feel free to disregard.

internal/driver.go
alnr: #3562 is kinda dangerous IMO, because it competely drops logs including error logs. This change silences warn+info logs only during testing.

jonas-jonas: Feel free to revert in this PR then, as it's already merged. Sorry.

We do the same thing as #3562 in cloud, so I pretty much copied it over from there.

jonas-jonas: #3562 also only drops the logs in testing, though.

identity/handler_test.go
alnr: Fixed and added assertion that `X-Total-Count` does not appear with token pagination.

persistence/sql/identity/persister_identity.go
```diff
@@ -649,26 +644,38 @@ func QueryForCredentials(con *pop.Connection, where ...Where) (map[uuid.UUID](ma
...
-func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.ListIdentityParameters) (res []identity.Identity, err error) {
+func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.ListIdentityParameters) (_ []identity.Identity, nextPage *keysetpagination.Paginator, err error) {
 	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.ListIdentities")
 	defer otelx.End(span, &err)
 
 	span.SetAttributes(
-		attribute.Int("page", params.Page),
-		attribute.Int("per_page", params.PerPage),
+		// attribute.Int("page", params.Page),
+		// attribute.Int("per_page", params.PerPage),
```
zepatrik, line 653: Should we not add some pagination info to traces? Maybe just what type is used?

identity/handler_test.go
alnr: Fixed.

persistence/sql/identity/persister_identity.go
```diff
@@ -649,26 +644,38 @@ func QueryForCredentials(con *pop.Connection, where ...Where) (map[uuid.UUID](ma
...
 	joins := ""
-	wheres := ""
-	args := []any{nid}
+	wheres := "identities.nid = ? AND identities.id > ?"
+	args := []any{nid, paginator.Token().Encode()}
+	limit := fmt.Sprintf("LIMIT %d", paginator.Size()+1)
+	if params.PagePagination != nil {
+		wheres = "identities.nid = ?"
+		args = []any{nid}
+		limit = fmt.Sprintf("LIMIT %d OFFSET %d", params.PagePagination.ItemsPerPage, params.PagePagination.Page*params.PagePagination.ItemsPerPage)
```
zepatrik, line 677: This might change existing behavior.
Please use `pop.NewPaginator(params.Page+1, params.PerPage)` like we had previously to calculate the same limit and offset.

identity/test/pool.go
alnr: added an assertion

zepatrik (changes requested):
Nice, just some minor details apart from what the others already pointed out :+1:

identity/test/pool.go
alnr: Added an assertion, please confirm that matches what you expected 👍

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/ory/kratos
 
 go 1.19
 
+// replace github.com/ory/x => ../x
```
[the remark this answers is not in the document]
alnr: done

go.mod
alnr: done

persistence/sql/identity/persister_identity.go
alnr: Wait, I thought we had changed to page 0 being the first page?

persistence/sql/identity/persister_identity.go
alnr: Added attributes

x/pagination.go
```diff
@@ -18,3 +21,42 @@ func ParsePagination(r *http.Request) (page, itemsPerPage int) {
...
+			return nil, nil, herodot.ErrBadRequest.WithReason(err.Error())
+		}
+		return keyset, nil, nil
+	}
+	// allow fallback page pagination with upper limit
+	if q.Has("page") {
+		paginator := pagepagination.PagePaginator{MaxItems: 500, DefaultItems: 250}
+		page, perPage := paginator.ParsePagination(r)
+		if page*perPage > PagePaginationLimit {
+			return nil, nil, herodot.ErrBadRequest.WithReasonf("Legacy pagination is not supported for enumerating over %d items. Please switch to using page_token and page_size.", PagePaginationLimit)
```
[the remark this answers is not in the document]
alnr: As you said yourself, effectively there already is a limit in place due to database timeouts. I'm open to increasing this number from 1000. Thoughts @aeneasr ?

persistence/sql/identity/persister_identity.go
zepatrik: Yes, but pop starts at 1, so our first one `0` +1 is pop's first one `1`. Because this was already breaking customers a few weeks back, I'd like to keep the same logic and use the pop helper to calculate limit & offset, just to make sure there is no breaking behavior again.

alnr: OK, I think I got it now 😁

zepatrik (approved):

jonas-jonas (approved):

aeneasr (approved):

PR #3570, opened 2023-10-11
fix: increase connection-level timeouts and shutdown timeouts

alnr:
The admin API is generally expected to require longer timeouts, for example during bulk identity import. 

Previously, we used `ory/graceful`'s default write timeout (10s). That has a couple of problems:

1. 10s is too short for expensive operations such as bulk identity import.
2. The server-level timeouts of `http.Server` (`WriteTimeout` etc) work in surprising ways. See https://github.com/golang/go/issues/59602. After reaching the timeout, the connection is closed and no more writes will be transported to the client. But there is no way to detect this from the handler. Writing to the `http.ResponseWriter` returns no error, and the request context is not canceled. So until now, for long requests, we would just keep truckin' and perform all the work and finally write a `200 OK` response. Logs and tracing would show no problems, even though the downstream had their connection closed on them much earlier and they received no response.

We could consider adding more advanced timeout detection in the future, for example into `ory/graceful`.

### DIFF
--- a/cmd/daemon/serve.go
+++ b/cmd/daemon/serve.go
@@ -69,6 +69,10 @@ func WithContext(ctx stdctx.Context) Option {
 	}
 }
 
+func init() {
+	graceful.DefaultShutdownTimeout = 120 * time.Second
+}
+
 func servePublic(r driver.Registry, cmd *cobra.Command, eg *errgroup.Group, slOpts *servicelocatorx.Options, opts []Option) {
 	modifiers := NewOptions(cmd.Context(), opts)
 	ctx := modifiers.ctx
@@ -132,8 +136,12 @@ func servePublic(r driver.Registry, cmd *cobra.Command, eg *errgroup.Group, slOp
 
 	//#nosec G112 -- the correct settings are set by graceful.WithDefaults
 	server := graceful.WithDefaults(&http.Server{
-		Handler:   handler,
-		TLSConfig: &tls.Config{GetCertificate: certs, MinVersion: tls.VersionTLS12},
+		Handler:           handler,
+		TLSConfig:         &tls.Config{GetCertificate: certs, MinVersion: tls.VersionTLS12},
+		ReadHeaderTimeout: 5 * time.Second,
+		ReadTimeout:       15 * time.Second,
+		WriteTimeout:      60 * time.Second,
+		IdleTimeout:       120 * time.Second,
 	})
 	addr := c.PublicListenOn(ctx)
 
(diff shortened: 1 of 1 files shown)


hperl (approved):
LGTM!

cmd/daemon/serve.go
```diff
@@ -69,6 +69,10 @@ func WithContext(ctx stdctx.Context) Option {
 	}
 }
 
+func init() {
+	graceful.DefaultShutdownTimeout = 120 * time.Second
```
aeneasr, line 73: What is this for?

aeneasr:
Recommended reading (this lead to the defaults we use today):

- https://blog.cloudflare.com/exposing-go-on-the-internet/
- https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

cmd/daemon/serve.go
alnr: When a long-running request (say bulk import) is in flight and the process is terminated (SIGINT for example), we want to wait for the request to complete before shutting down. The process will terminate as soon as all in-flight requests have concluded, but after 120 the latest.

The default here is 5 seconds, which I will hard-kill long-running requests leading to errors downstream.

cmd/daemon/serve.go
```diff
@@ -206,8 +214,12 @@ func serveAdmin(r driver.Registry, cmd *cobra.Command, eg *errgroup.Group, slOpt
...
 	//#nosec G112 -- the correct settings are set by graceful.WithDefaults
 	server := graceful.WithDefaults(&http.Server{
-		Handler:   handler,
-		TLSConfig: &tls.Config{GetCertificate: certs, MinVersion: tls.VersionTLS12},
+		Handler:           handler,
+		TLSConfig:         &tls.Config{GetCertificate: certs, MinVersion: tls.VersionTLS12},
+		ReadHeaderTimeout: 5 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      120 * time.Second,
+		IdleTimeout:       600 * time.Second,
```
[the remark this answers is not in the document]
alnr: This is for the admin API only, which is not exposed publicly. The public API has the default 120sec idle timeout (I just made it explicit).

alnr:
> Recommended reading (this lead to the defaults we use today):
> 
> * https://blog.cloudflare.com/exposing-go-on-the-internet/
> * https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

I know these articles. They are not really applicable to us, though, and arguably misleading. These articles are only concerned with connection management if your service is exposed on the internet. That's not the case for most installations of Kratos. They mostly have proxies in front of them, where connection management mostly concerns detecting network partitions between kratos and its proxy.

Also, these articles don't go into any detail about what the behavior of the request handler is, when one of the `http.Server` timeouts is hit. And that's really a critical part of the whole logic.

As outlined in the PR description, timeouts like `http.Server{}.WriteTimeout` are low-level primitives. They close the underlying connection or stream. But that information is not transported to the HTTP handler in any way. So the handler just keeps working like normal and thinks everything is going well. Everything, including logs and traces, will show `all good`, even though the connection was already closed.

So the CF articles are a good reference to lookup the exact points in time where the timeout deadlines begin and end. That's nice. But they are not a definite guide on how to configure timeouts in Go web servers.

cmd/daemon/serve.go
aeneasr: That makes sense - thank you!

aeneasr: Can this lead to problems where the rollout is in flight and we are waiting for connections to finish, which may cause a downtime of ~2 minutes because of this?

alnr: Why would waiting for connections to finish cause a downtime?

aeneasr: I'm not sure how SQL migratinos behave but I believe that first migrations get executed, then the release is started. If the SQL migration is not online, we run into a downtime.

alnr: But the old pods will still be serving traffic if the new release doesn't come up correctly, don't they?

PR #3271, opened 2023-05-09
fix: accept OIDC login request in browser+JSON login flow

hperl:
## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ ] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [ ] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [ ] I have read the [security policy](../security/policy).
- [ ] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got the approval (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](https://github.com/ory/docs).

PR #3572, opened 2023-10-11
fix: respond with 422 when SPA identity requires AAL2

BrandonNoad:
If you submit a browser login flow with an `Accept` header of `application/json`, but the login flow requires AAL2, then there is no way for the code to know it needs to redirect the user to the 2FA page. Instead of responding with the `Session` in this scenario, this PR changes the behaviour to respond with a `browser_location_change_required` error (status `422`) to indicate that the browser needs to open a specific URL, /self-service/login/browser?aal=aal2. 

## Related issue(s)

N/A - but discussed in Slack (#support-network)

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

Similar to changes in https://github.com/ory/kratos/pull/3271


selfservice/flow/login/hook_test.go
BrandonNoad, line 1: I fixed some of the existing test descriptions since they didn't actually describe what the code is doing

hperl (approved):
Thanks for the contribution. LGTM!

aeneasr (approved):
Nice, LGTM

aeneasr:
I think this is a small breaking change, but looking at the existing implementation the current way of doing 2FA is pretty poor. So I think this change in behavior is acceptable without a feature flag because it re-uses existing tooling and response codes. Thanks! :)

PR #3574, opened 2023-10-12
feat: add option for automatic account linking

hperl:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/embedx/config.schema.json
+++ b/embedx/config.schema.json
@@ -526,6 +526,16 @@
           "description": "The ID of the organization that this provider belongs to. Only effective in the Ory Network.",
           "type": "string",
           "examples": ["12345678-1234-1234-1234-123456789012"]
+        },
+        "automatic_account_linking": {
+          "title": "Configure automatic account linking",
+          "description": "Configures how accounts through this provider should be linked automatically to existing accounts.",
+          "type": "string",
+          "enum": [
+            "never",
+            "replace"
+          ],
+          "default": "never"
         }
       },
       "additionalProperties": false,
--- a/identity/identity.go
+++ b/identity/identity.go
@@ -160,6 +160,24 @@ func (t *Traits) String() string {
 	return string(*t)
 }
 
+// Merge merges the other traits into this one. If a key exists in both, the
+// other traits' value is use
+func (t *Traits) Merge(o Traits) (Traits, error) {
+	this := map[string]any{}
+	other := map[string]any{}
+	if err := json.Unmarshal(*t, &this); err != nil {
+		return nil, err
+	}
+	if err := json.Unmarshal(o, &other); err != nil {
+		return nil, err
+	}
+	for k, v := range other {
+		this[k] = v
+	}
+
+	return json.Marshal(this)
+}
+
 // MarshalJSON returns m as the JSON encoding of m.
 func (t Traits) MarshalJSON() ([]byte, error) {
 	if t == nil {
--- a/identity/manager.go
+++ b/identity/manager.go
@@ -110,12 +110,11 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
 	return nil
 }
 
-func (m *Manager) findExistingAuthMethod(ctx context.Context, e error, i *Identity) (err error) {
-	if !m.r.Config().SelfServiceFlowRegistrationLoginHints(ctx) {
-		return &ErrDuplicateCredentials{error: e}
-	}
+func (m *Manager) FindConflictingIdentity(ctx context.Context, i *Identity) (found *Identity, foundConflictAddress string, err error) {
+	ctx, span := m.r.Tracer(ctx).Tracer().Start(ctx, "identity.Manager.FindConflictingIdentity")
+	defer otelx.End(span, &err)
+
 	// First we try to find the conflict in the identifiers table. This is most likely to have a conflict.
-	var found *Identity
 	for ct, cred := range i.Credentials {
 		for _, id := range cred.Identifiers {
 			found, _, err = m.r.PrivilegedIdentityPool().FindByCredentialsIdentifier(ctx, ct, id)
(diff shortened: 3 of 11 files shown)


selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -349,13 +362,57 @@ func (s *Strategy) processRegistration(w http.ResponseWriter, r *http.Request, r
...
 	if err := s.d.RegistrationExecutor().PostRegistrationHook(w, r, identity.CredentialsTypeOIDC, provider.Config().ID, rf, i); err != nil {
+		dup := new(identity.ErrDuplicateCredentials)
+		if provider.Config().AutomaticAccountLinking == AutomaticAccountLinkingReplace && errors.As(err, &dup) {
+			conflictingIdentity, _, err := s.d.IdentityManager().FindConflictingIdentity(r.Context(), i)
+			if err != nil {
+				return nil, s.handleError(w, r, rf, provider.Config().ID, i.Traits, err)
+			}
+			if conflictingIdentity == nil {
+				return nil, s.handleError(w, r, rf, provider.Config().ID, i.Traits, dup)
+			}
```
jonas-jonas, line 373: Can this actually ever happen? Returning `nil, "", nil` from `FindConflictingIdentity` can lead to panics if you forget to check the value of the identity for nil, which you don't typically do. E.g. you assume that the values returned are valid if no `err` was returned. Maybe we can return an error from `FindConflictingIdentity` instead?

hperl: Yeah that sounds sensible. I just extracted the method and the `nil, "", nil` case was when we could not find a conflicting identity. I guess an error is better in this case.

aeneasr (commented):
Can you please create an E2E test for this and briefly explain what the user will see?

identity/identity.go
```diff
@@ -160,6 +160,24 @@ func (t *Traits) String() string {
 	return string(*t)
 }
 
+// Merge merges the other traits into this one. If a key exists in both, the
+// other traits' value is use
+func (t *Traits) Merge(o Traits) (Traits, error) {
```
aeneasr, line 165: This only merges the first level. Is that what we want? If not, this should be something like mapstructure.Merge or similar.

identity/manager.go
```diff
@@ -110,12 +110,11 @@ func (m *Manager) Create(ctx context.Context, i *Identity, opts ...ManagerOption
 	return nil
 }
 
-func (m *Manager) findExistingAuthMethod(ctx context.Context, e error, i *Identity) (err error) {
-	if !m.r.Config().SelfServiceFlowRegistrationLoginHints(ctx) {
-		return &ErrDuplicateCredentials{error: e}
-	}
+func (m *Manager) FindConflictingIdentity(ctx context.Context, i *Identity) (found *Identity, foundConflictAddress string, err error) {
```
aeneasr, line 113: Test pls

identity/identity.go
aeneasr: Also test please

embedx/config.schema.json
```diff
@@ -603,6 +603,16 @@
           "examples": [
             "12345678-1234-1234-1234-123456789012"
           ]
+        },
+        "automatic_account_linking": {
+          "title": "Configure automatic account linking",
+          "description": "Configures how accounts through this provider should be linked automatically to existing accounts.",
```
aeneasr, line 609: What happens when this is "never" and what if it is "replace"?

selfservice/strategy/oidc/provider_config.go
```diff
@@ -108,8 +108,15 @@ type Configuration struct {
 	// An optional organization ID that this provider belongs to.
 	// This parameter is only effective in the Ory Network.
 	OrganizationID string `json:"organization_id"`
+
+	AutomaticAccountLinking string `json:"automatic_account_linking"`
```
aeneasr, line 112: Docstring please

And potentially use a dedicated type if this is an enum

selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -349,13 +362,62 @@ func (s *Strategy) processRegistration(w http.ResponseWriter, r *http.Request, r
 
 	i.SetCredentials(s.ID(), *creds)
 	if err := s.d.RegistrationExecutor().PostRegistrationHook(w, r, identity.CredentialsTypeOIDC, provider.Config().ID, rf, i); err != nil {
+		dup := new(identity.ErrDuplicateCredentials)
+		if provider.Config().AutomaticAccountLinking == AutomaticAccountLinkingReplace && errors.As(err, &dup) {
```
aeneasr, line 366: fyi, you can also do:

```
		if dup := new(identity.ErrDuplicateCredentials); provider.Config().AutomaticAccountLinking == AutomaticAccountLinkingReplace && errors.As(err, &dup) {
```

selfservice/strategy/oidc/strategy_registration.go
aeneasr: Agree, `nil, "", nil` should not be a valid return type of FindConflictingIdentity

aeneasr (changes requested):
Thank you!

I believe that this PR falls into the trap why we do not have automatic account linking in the first place, which is documented here: https://www.ory.sh/docs/kratos/social-signin/link-multiple-provider-account#automatic-account-linking

Please adjust the PR so that this can not happen, or comment on why the implementation is not an exploitable attack vector. Thanks!

embedx/config.schema.json
```diff
@@ -603,6 +603,16 @@
...
+          "type": "string",
+          "enum": [
+            "never",
+            "replace"
```
alnr, line 613:
```suggestion
            "replace_all_other_credentials"
```

aeneasr:
The fundamental problem of this approach is that we (Ory Kratos) does not know whether identity A and identity B are truly the same person or not. Just because some credential is saying that they are the same user, does not mean that they indeed are the same person. In the end, this is an account takeover vulnerability which essentially:

1. Alice signs up with email and password `alice@doe.com`
2. Eve does account enumeration (currently possible) or otherwise knows that alice has an account
3. Eve signs up at the social provider that is enabled with account linking (e.g. GitHub) with `alice@doe.com`
4. Eve logs into Ory Kratos using GitHub with the linked fake `alice@doe.com` email
5. The original password is deleted, and Eve has a valid session token -> account takeover complete

There are probably some ways of fixing this with verified email addresses and so on. But honestly, let's just stick to how we solve it right now but make it more user friendly:

1. Sign up with email password
2. Sign up with social provider -> conflict
3. Present user login screen "sign in with your password to link your account"
4. Sign in with correct credential is complete
5. Social sign in provider is linked to the user

kmherrmann:
> The fundamental problem of this approach is that we (Ory Kratos) does not know whether identity A and identity B are truly the same person or not. Just because some credential is saying that they are the same user, does not mean that they indeed are the same person. In the end, this is an account takeover vulnerability which essentially:
> 
> 1. Alice signs up with email and password `alice@doe.com`
> 2. Eve does account enumeration (currently possible) or otherwise knows that alice has an account
> 3. Eve signs up at the social provider that is enabled with account linking (e.g. GitHub) with `alice@doe.com`
> 4. Eve logs into Ory Kratos using GitHub with the linked fake `alice@doe.com` email
> 5. The original password is deleted, and Eve has a valid session token -> account takeover complete
> 
> There are probably some ways of fixing this with verified email addresses and so on. But honestly, let's just stick to how we solve it right now but make it more user friendly:

Yes, we can only do this automatically if we trust that we have verified email addresses on both sides. For that, I think we should have a flag on OIDC providers `like trust_email_verification`


> 
> 1. Sign up with email password
> 2. Sign up with social provider -> conflict
> 3. Present user login screen "sign in with your password to link your account"
> 4. Sign in with correct credential is complete
> 5. Social sign in provider is linked to the user

I think this flow is great and should be the default.

In the case that the existing Ory account has a verified email address AND the OIDC login claims to have a verified email address AND the OIDC provider is set to trust email verification, then the steps 3 and 4 can automatically be skipped.

PR #3577, opened 2023-10-13
fix: always return relative URLs in the Link header for pagination

alnr:
This is much less brittle when Kratos is deployed behind a proxy which rewrites the Host header. Example from Ory Network: before this change, when the initial request was made to a CNAME, the Link header of the response contains `<slug>.projects.oryapis.com`.

Another problem with the previous solution was that the `/admin` prefix would occasionally be stripped from the Link header in the response. For example, the first request would go to `/admin/identities?page=0`, and the Link header would contain `</identities?page=1>; rel="next"`.

Now, we always return a modified version of the exact path that was requested. Should solve both of these issues.

### DIFF
--- a/x/pagination.go
+++ b/x/pagination.go
@@ -18,8 +18,8 @@ func ParsePagination(r *http.Request) (page, itemsPerPage int) {
 	return migrationpagination.NewDefaultPaginator().ParsePagination(r)
 }
 
-func PaginationHeader(w http.ResponseWriter, u *url.URL, total int64, page, itemsPerPage int) {
-	migrationpagination.PaginationHeader(w, u, total, page, itemsPerPage)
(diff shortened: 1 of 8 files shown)


x/pagination.go
```diff
@@ -18,8 +18,8 @@ func ParsePagination(r *http.Request) (page, itemsPerPage int) {
 	return migrationpagination.NewDefaultPaginator().ParsePagination(r)
 }
 
-func PaginationHeader(w http.ResponseWriter, u *url.URL, total int64, page, itemsPerPage int) {
```
alnr, line 21: This change is so that we don't inadvertently modify the argument.

zepatrik (approved):

PR #3592, opened 2023-10-23
fix: registration should accept hydra login

Benehiko:
## Related issue(s)

https://github.com/ory-corp/cloud/issues/5493

Using `propt=registration` on an OAuth2 Hydra login flow, the request is routed to the Kratos registration handler. This handler immediately redirects to the logout flow when there is an already existing session, which fails and ends up at the base `/welcome` page on the Account Experience (AX). 

This PR adds the functionality of accepting the login request in the registration handler if the user has a session and if the hydra `skip=true` is in the login request. 

When the `skip=false` is set, the request is redirected to the login flow with the login_challenge, which will complete the flow normally.
 

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/flow/registration/handler.go
+++ b/selfservice/flow/registration/handler.go
@@ -364,11 +364,14 @@ func (h *Handler) createBrowserRegistrationFlow(w http.ResponseWriter, r *http.R
 			// hydra indicates that we cannot skip the login request
 			// so we must perform the login flow.
 			if !hydraLoginRequest.GetSkip() {
-				hydraUrl := hydraLoginRequest.RequestUrl
+				loginUrl := h.d.Config().SelfServiceFlowLoginUI(ctx)
+				q := loginUrl.Query()
+				q.Set("login_challenge", hydraLoginChallenge.String())
+				loginUrl.RawQuery = q.Encode()
 				http.Redirect(
 					w,
 					r,
-					hydraUrl,
+					loginUrl.String(),
 					http.StatusFound,
 				)
 				return
--- a/selfservice/flow/registration/handler.go
+++ b/selfservice/flow/registration/handler.go
@@ -27,10 +27,11 @@ import (
 	"github.com/ory/x/sqlxx"
 	"github.com/ory/x/urlx"
 
+	hydraclientgo "github.com/ory/hydra-client-go/v2"
 	"github.com/ory/kratos/driver/config"
 	"github.com/ory/kratos/selfservice/errorx"
 	"github.com/ory/kratos/selfservice/flow"
-	"github.com/ory/kratos/selfservice/flow/logout"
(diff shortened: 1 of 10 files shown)


selfservice/flow/registration/handler.go
```diff
@@ -318,29 +318,79 @@ type createBrowserRegistrationFlow struct {
...
+			r.URL.RawQuery = q.Encode()
+		}
+	}
+
+	if sess, err := h.d.SessionManager().FetchFromRequest(ctx, r); err == nil {
+
+		if hydraLoginRequest != nil {
+			// hydra indicates that we cannot skip the login request
+			// so we must perform the login flow.
+			if !hydraLoginRequest.GetSkip() {
```
hperl, line 366: I think skip is used purely for the consent. For forcing the login, we need to look for `prompt=login`. Prompt can have space-delimited values, so `prompt=login registration` would be valid.

selfservice/flow/registration/handler.go
```diff
@@ -318,29 +318,79 @@ type createBrowserRegistrationFlow struct {
...
+			if err != nil {
+				h.d.SelfServiceErrorManager().Forward(r.Context(), w, r, err)
+				return
+			}
+			returnTo, err := url.Parse(rt)
+			if err != nil {
+				h.d.SelfServiceErrorManager().Forward(r.Context(), w, r, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to parse URL: %s", rt)))
+				return
+			}
+			x.AcceptToRedirectOrJSON(w, r, h.d.Writer(), err, returnTo.String())
```
hperl, line 393: I don't think we ever need to return JSON here.

hperl (changes requested):
Thanks for looking into this! The approach is sensible, I just think we should look for `prompt=login` instead of `skip`.

selfservice/flow/registration/handler.go
Benehiko: what should we then do if `skip=false`? In this case the user cannot go back to the client, on the Login flow we usually do `refresh=true` here. So this is why I redirect to login and have the login logic handle what should happen to the OAuth flow. 

In this case, the user is re-prompted to login or "confirm it's you".

selfservice/flow/registration/handler.go
Benehiko: We do, in the case of an SPA we return 422 with a browser redirect URL - just like in the Login flow.

Benehiko: https://github.com/ory/kratos/blob/master/selfservice/flow/login/handler.go#L479-L496

hperl (approved):
LGTM!

PR #3594, opened 2023-10-25
fix: return HTTP 400 if key unmarshal fails

nipsufn:
This should correct behavior where on key unmarshal error application returns HTTP 500.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


session/tokenizer.go
```diff
@@ -70,6 +71,8 @@ func (s *Tokenizer) TokenizeSession(ctx context.Context, template string, sessio
 	if err != nil {
 		if errors.Is(err, jwksx.ErrUnableToFindKeyID) {
 			return errors.WithStack(herodot.ErrBadRequest.WithReasonf("Could not find key a suitable key for tokenization in the JWKS url."))
+		} else if strings.Contains(err.Error(), "failed to unmarshal JWK set: ") {
+			return errors.WithStack(herodot.ErrBadRequest.WithReasonf("%v", err.Error()))
```
hperl, line 75: LGTM, but can't we assert on the error type with error.As here?

hperl (changes requested):
Nice fix! Just a small comment.

session/tokenizer.go
nipsufn: I did that because the returned error type is very generic (`errors.withStack` containing `errors.withMessage`). Jumping around code it looks like there's lots of other potential error opportunities other than unmarshaling and I wouldn't like to cover those up.

hperl: But error.As will continuously unwrap the error until it finds the type you specify.

hperl: I looked further and the method that returns that error (`jwk.Parse`) also has other error messages related to failing to parse. We should probably handle these cases as a bad request as well.

I think the best solution is to wrap the error with a JWKParserError here https://github.com/ory/x/blob/e6c61ed41fc506d1a0bc9fc617bc53948b0addf6/jwksx/fetcher_v2.go#L157 and then match on it where you currently match on the error message. WDYT?

nipsufn: I hoped to avoid touching `ory/x`, but that does seem a cleaner solution. I'll get that done and adjust this bit accordingly.

hperl (approved):

PR #3597, opened 2023-10-26
fix: change shebangs and makefile from /bin/bash to /usr/bin/env bash

nxy7:
Signed-off-by: nxy7 <lolnoxy@gmail.com>

Not every linux distribution has /bin/bash. One of examples of such distros is NixOS which only has /bin/sh.
Fix to that is pretty easy, instead of linking to /bin/bash we can use /usr/bin/env bash which will find correct bash path.
Other way to solve it would be using /bin/sh (some distros have it linked to bash anyway).

If I'm not mistaken this is non breaking change and shouldn't affect anyone negatively.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-SHELL=/bin/bash -o pipefail
+SHELL=/usr/bin/env bash -o pipefail
 
 #  EXECUTABLES = docker-compose docker node npm go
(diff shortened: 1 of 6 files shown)


Benehiko (approved):
Thank you for the contribution :partying_face: 

LGTM :)

Issue #2448, opened 2022-05-04
Bulk Identity Fetch Api


nxy7:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Using kratos as single source of truth might lead to some suboptimal workflows. Because you can get identities only one by one or paginated list of all identities, it might be necessary to make N requests instead of one bulk request.
Let's say you store just id's from kratos in main app database and you have comments table {userId, text}. I think it would be for the best to store traits in just one place - kratos db, if that's the case you need one additional request to get user traits for every single comment if you want to use data from traits. 

### Describe your ideal solution

Make new kratos endpoint accepting array of 'id's. 

### Workarounds or alternatives

Duplicate traits in main app database or make too many fetch requests :P

### Version

v0.7.1-alpha.1

RubenWerdmuller:
Any news on this end? 👀

nxy7:
I'm willing to try implementing this myself, one question though. Is it fine if we leave the same endpoint and make it accept /admin/identities/{id} and /admin/identities/{id}&{id}&{id}&{id}?

RubenWerdmuller:
@nxy7 I think a conventional solution for this is desirable. I'm not familiar with all patterns, but I'd suggest sending a body that includes the parameters needed. In our case `ids`.

[Similar to how elastic queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-get.html)

nxy7:
Sounds good to me. Would you accept PR modifying `/admin/identities` endpoint then?

alnr:
What's wrong with firing several requests instead? You can also do it in parallel if latency is the issue.

[...]

PR #3598, opened 2023-10-26
feat: batch list identities

nxy7:
Signed-off-by: nxy7 <lolnoxy@gmail.com>

test for filder ids added

Signed-off-by: nxy7 <lolnoxy@gmail.com>

Right now if you want to get many idendities you have to get them one by one. This change allows to filter `/admin/identities` by id
with the following syntax `/admin/identities?ids=id1&ids=id2&ids=id3`. This way we make less requests, decrease latency and bandwith used by apps
interacting with kratos. 

## Related issue(s)
fixes: #2448

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

Before adding documentation to ory/docs I'd like to get green like on changes I've made.

### DIFF
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -138,6 +138,13 @@ type listIdentitiesResponse struct {
 type listIdentitiesParameters struct {
 	migrationpagination.RequestParameters
 
+	// IdsFilter is list of ids used to filter identities.
+	// If this list is empty, then no filter will be applied.
+	//
+	// required: false
+	// in: query
+	IdsFilter []string `json:"ids_filter"`
+
 	// CredentialsIdentifier is the identifier (username, email) of the credentials to look up using exact match.
 	// Only one of CredentialsIdentifier and CredentialsIdentifierSimilar can be used.
 	//
@@ -180,6 +187,7 @@ func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Para
 		err    error
 		params = ListIdentityParameters{
 			Expand:                       ExpandDefault,
+			IdsFilter:                    r.URL.Query()["ids"],
 			CredentialsIdentifier:        r.URL.Query().Get("credentials_identifier"),
 			CredentialsIdentifierSimilar: r.URL.Query().Get("preview_credentials_identifier_similar"),
 			ConsistencyLevel:             crdbx.ConsistencyLevelFromRequest(r),
--- a/identity/handler_test.go
+++ b/identity/handler_test.go
@@ -339,6 +339,34 @@ func TestHandler(t *testing.T) {
 		}
 	})
 
+	t.Run("suite=create and batch list", func(t *testing.T) {
+		var ids []uuid.UUID
+		identitiesAmount := 5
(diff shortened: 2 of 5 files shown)


identity/handler.go
```diff
@@ -175,11 +175,12 @@ type listIdentitiesParameters struct {
 //	Responses:
 //	  200: listIdentities
 //	  default: errorGeneric
-func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+func (h *Handler) list(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
 	var (
 		err    error
 		params = ListIdentityParameters{
 			Expand:                       ExpandDefault,
+			IdsFilter:                    r.URL.Query()["ids"],
```
jonas-jonas, line 183:
```suggestion
			IdsFilter:                    r.URL.Query().Get("ids"),
```

jonas-jonas (changes requested):
Very nice changes, thank you!

identity/handler.go
nxy7: .Get("ids") will only get first id and we need to populate slice here

aeneasr (commented):
This is grand! Can you please add a test to `handler_test` to verify the API is also behaving correctly?

identity/handler.go
hperl, line 183: This new parameter also needs to be mentioned in `listIdentitiesParameters`, so that we have a proper API documentation. Unfortunately we don't use that struct to actually parse the request query, but it still shows up in the OpenAPI spec.

identity/handler.go
```diff
@@ -175,11 +175,12 @@ type listIdentitiesParameters struct {
 //	Responses:
 //	  200: listIdentities
 //	  default: errorGeneric
-func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+func (h *Handler) list(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
```
hperl, line 178:
```suggestion
func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
```

`p` is unused IIRC.

hperl (changes requested):
Awesome, thanks for the patch! I just have two really minor remarks. Also, as Aeneas mentioned, a E2E handler test for this in `identity/handler_test.go` would be really good.

identity/handler.go
nxy7: Yea, this was accidental change from my first experiments. I've reverted that.

nxy7:
I've added few changes related to Nix (flake.nix, flake.lock and .ignorefile). I'll revert them once You accept the rest of the changes, but local dev on NixOS without them was really hard.

hperl (dismissed):
LGTM! Thank you very much for your contribution!

Please remember to remove the extra nix files :)

jonas-jonas (approved):

aeneasr (approved):

PR #3601, opened 2023-10-30
feat: parametrize courier worker

aeneasr:
Allows one to parametrize how many messages the courier will fetch and how often it will fetch messages.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/driver/config/config.go
+++ b/driver/config/config.go
@@ -77,6 +77,8 @@ const (
 	ViperKeyCourierSMSEnabled                                = "courier.sms.enabled"
 	ViperKeyCourierSMSFrom                                   = "courier.sms.from"
 	ViperKeyCourierMessageRetries                            = "courier.message_retries"
+	ViperKeyCourierWorkerPullCount                           = "courier.worker.pull_count"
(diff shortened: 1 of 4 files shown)


driver/config/config.go
```diff
@@ -77,6 +77,8 @@ const (
 	ViperKeyCourierSMSEnabled                                = "courier.sms.enabled"
 	ViperKeyCourierSMSFrom                                   = "courier.sms.from"
 	ViperKeyCourierMessageRetries                            = "courier.message_retries"
+	ViperKeyCourierWorkerPullCount                           = "courier.worker.pull_count"
+	ViperKeyCourierWorkerPullWait                            = "courier.worker.pull_wait"
```
zepatrik, line 81: Shouldn't this rather be `poll`?

aeneasr: Hm I think polling is more when you ask a server / component whether they're ready periodically, and pulling refers to retrieving data and altering state. But really I don't care either way :)

zepatrik: Yeah, both would work in this case I think...

PR #3603, opened 2023-10-30
chore: simplify courier

zepatrik:
The setup code was simplified.

Also, the `courier.MessageStatusProcessing` was removed.
The processing state was intended as a kind of lock to allow running multiple couriers. However, in practice that did not really work because multiple instance would run into transaction conflicts a lot, because they would always try to select & update the same rows. Further, there was no "lock release" strategy, but once the lock was acquired and the worker died without finishing the message, it would never be picked up again. Removing this state means less database writes, and that the worker would pick up half-processed messages.

As a follow-up, we can implement a proper sharding strategy so multiple workers can work in parallel without conflicts or duplicates.

### DIFF
--- a/courier/courier_dispatcher.go
+++ b/courier/courier_dispatcher.go
@@ -57,13 +57,10 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
 
 	messages, err := c.deps.CourierPersister().NextMessages(ctx, 10)
 	if err != nil {
-		if errors.Is(err, ErrQueueEmpty) {
-			return nil
-		}
 		return err
 	}
 
-	for k, msg := range messages {
+	for _, msg := range messages {
 		if msg.SendCount > maxRetries {
 			if err := c.deps.CourierPersister().SetMessageStatus(ctx, msg.ID, MessageStatusAbandoned); err != nil {
 				c.deps.Logger().
@@ -79,41 +76,33 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
 				WithField("message_id", msg.ID).
 				WithField("message_nid", msg.NID).
 				Warnf(`Message was abandoned because it did not deliver after %d attempts`, msg.SendCount)
-		} else if err := c.DispatchMessage(ctx, msg); err != nil {
+			continue
+		}
+
+		if err := c.DispatchMessage(ctx, msg); err != nil {
 			if err := c.deps.CourierPersister().RecordDispatch(ctx, msg.ID, CourierMessageDispatchStatusFailed, err); err != nil {
 				c.deps.Logger().
 					WithError(err).
 					WithField("message_id", msg.ID).
 					WithField("message_nid", msg.NID).
 					Error(`Unable to record failure log entry.`)
-				if c.failOnDispatchError {
-					return err
-				}
-			}
-
-			for _, replace := range messages[k:] {
-				if err := c.deps.CourierPersister().SetMessageStatus(ctx, replace.ID, MessageStatusQueued); err != nil {
-					c.deps.Logger().
-						WithError(err).
-						WithField("message_id", replace.ID).
-						WithField("message_nid", replace.NID).
-						Error(`Unable to reset the failed message's status to "queued".`)
-					if c.failOnDispatchError {
-						return err
-					}
-				}
+				return err
 			}
 
 			if c.failOnDispatchError {
 				return err
 			}
-		} else if err := c.deps.CourierPersister().RecordDispatch(ctx, msg.ID, CourierMessageDispatchStatusSuccess, nil); err != nil {
+			// an error happened, but we want to ignore it
+			continue
+		}
+
+		if err := c.deps.CourierPersister().RecordDispatch(ctx, msg.ID, CourierMessageDispatchStatusSuccess, nil); err != nil {
 			c.deps.Logger().
 				WithError(err).
 				WithField("message_id", msg.ID).
 				WithField("message_nid", msg.NID).
 				Error(`Unable to record success log entry.`)
-			// continue with execution, as the message was successfully dispatched
+			return err
 		}
 	}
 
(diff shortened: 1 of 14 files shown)


courier/courier_dispatcher.go
```diff
@@ -80,41 +77,33 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
...
-					c.deps.Logger().
-						WithError(err).
-						WithField("message_id", replace.ID).
-						WithField("message_nid", replace.NID).
-						Error(`Unable to reset the failed message's status to "queued".`)
-					if c.failOnDispatchError {
-						return err
-					}
-				}
+				return err
```
zepatrik, line 90: This error is actually a database error, so it should always be returned.

courier/courier_dispatcher.go
```diff
@@ -80,41 +77,33 @@ func (c *courier) DispatchQueue(ctx context.Context) error {
...
+		}
+
+		if err := c.deps.CourierPersister().RecordDispatch(ctx, msg.ID, CourierMessageDispatchStatusSuccess, nil); err != nil {
 			c.deps.Logger().
 				WithError(err).
 				WithField("message_id", msg.ID).
 				WithField("message_nid", msg.NID).
 				Error(`Unable to record success log entry.`)
-			// continue with execution, as the message was successfully dispatched
+			return err
```
zepatrik, line 106: This is also a database error we want to always return.

courier/message.go
```diff
@@ -24,15 +24,14 @@ type MessageStatus int
 const (
 	MessageStatusQueued MessageStatus = iota + 1
 	MessageStatusSent
-	MessageStatusProcessing
+	_
```
zepatrik, line 27: This is the removed message status.

courier/persistence.go
```diff
@@ -7,13 +7,10 @@ import (
 	"context"
 
 	"github.com/gofrs/uuid"
-	"github.com/pkg/errors"
 
 	"github.com/ory/x/pagination/keysetpagination"
 )
 
-var ErrQueueEmpty = errors.New("queue is empty")
```
zepatrik, line 15: Not sure why we had an error here, if we would always gracefully handle it. An empty list is still a valid list, and can be iterated over.

courier/persistence.go
```diff
@@ -22,8 +19,6 @@ type (
 
 		SetMessageStatus(context.Context, uuid.UUID, MessageStatus) error
 
-		LatestQueuedMessage(ctx context.Context) (*Message, error)
```
zepatrik, line 25: This was not used in any production code, and not necessary for any test.

courier/test/persistence.go
```diff
@@ -38,85 +39,54 @@ func TestPersister(ctx context.Context, newNetworkUnlessExisting NetworkWrapper,
...
-			_, err = p.LatestQueuedMessage(ctx)
-			require.ErrorIs(t, err, courier.ErrQueueEmpty)
 		})
 
 		messages := make([]courier.Message, 5)
 		t.Run("case=add messages to the queue", func(t *testing.T) {
+			start := time.Now().UTC()
 			for k := range messages {
+				pop.SetNowFunc(func() time.Time {
+					return start.Add(time.Duration(k) * time.Second)
```
zepatrik, line 51: Pop really sucks sometimes...

persistence/sql/persister_courier.go
```diff
@@ -70,66 +67,18 @@ func (p *Persister) NextMessages(ctx context.Context, limit uint8) (messages []c
...
-		}
-
-		if len(m) == 0 {
-			return sql.ErrNoRows
-		}
-
-		for i := range m {
-			message := &m[i]
-			message.Status = courier.MessageStatusProcessing
-			if err := update.Generic(ctx, p.GetConnection(ctx), p.r.Tracer(ctx).Tracer(), message, "status"); err != nil {
```
zepatrik, line 93: Avoiding all of these writes.

selfservice/hook/verification_test.go
```diff
@@ -99,6 +100,9 @@ func TestVerifier(t *testing.T) {
 			require.NoError(t, err)
 			require.Len(t, messages, 2)
 
+			require.NoError(t, reg.CourierPersister().SetMessageStatus(context.Background(), messages[0].ID, courier.MessageStatusSent))
+			require.NoError(t, reg.CourierPersister().SetMessageStatus(context.Background(), messages[1].ID, courier.MessageStatusSent))
```
zepatrik, line 104: A lot of tests relied on the side-effect update from `NextMessages`, we now have to explicitly update their status.

selfservice/strategy/link/sender.go
```diff
@@ -51,7 +52,7 @@ type (
 	}
 )
 
-var ErrUnknownAddress = errors.New("verification requested for unknown address")
+var ErrUnknownAddress = stderrors.New("verification requested for unknown address")
```
zepatrik, line 55: Avoiding init function stack trace.

selfservice/strategy/profile/strategy_test.go
```diff
@@ -550,13 +555,13 @@ func TestStrategyTraits(t *testing.T) {
...
 		t.Run("type=spa", func(t *testing.T) {
-			newEmail := "update-verify-browser@mail.com"
+			newEmail := "update-verify-browser-1@mail.com"
 			actual := expectSuccess(t, false, true, browserUser1, payload(newEmail))
 			check(t, actual, newEmail)
 		})
 
 		t.Run("type=browser", func(t *testing.T) {
-			newEmail := "update-verify-browser@mail.com"
+			newEmail := "update-verify-browser-2@mail.com"
```
zepatrik, line 564: This case was asserting on the email of the previous one 🤦

zepatrik (commented):
Sorry, the diff got really hard to read in some places...

aeneasr (approved):

PR #3608, opened 2023-11-03
fix: omit irrelevant OIDC providers in forced refresh login flows

Saancreed:
Whenever an user is asked to reauthenticate (e.g. because they wish to execute settings flow touching their credentials and their session is no longer privileged) they are asked to provide their credentials again. The forced-refresh login flow generated for such cases already excludes some strategies that are enabled in Kratos but cannot be used to authenticate as current identity, and for example the form presented to the user will not have a password field if the identity does not have a password credential.

This, however, does not currently apply to OIDC providers; the user will always see the full set even if some of them can't be used to sign in as current identity. This change causes forced refresh login flows to also omit irrelevant OIDC providers in generated form in order to avoid confunding the user about which strategies/providers are valid and can actually be used to reauthenticate.

## Related issue(s)

Not reported separately. To reproduce:

1. Configure Kratos with password credentials and some OIDC providers enabled.
2. Register and login using a password.
3. Wait until session is old enough to no longer be considered privileged (1h by default) and try to update password using settings flow.
4. See that the "Confirm it's you" form offers signing in with OIDC providers that are not linked to current identity and can't actually be used to reauthenticate.

The inverse does not happen: if registering with OIDC, the form does _not_ have a password field because the identity does not have a password credential yet.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

Removing nodes/providers should be safe enough because we are operating in the context of previously authenticated identity. The user already knows which OIDC providers are linked because they can see this info in settings flow itself.

I've tested this by extending Settings flow linking and unlinking when unauthed scenarios but right now the check is a bit fragile (because it extracts login flow ID from URL query params) and probably not as precise as it could be (because I'm only checking that there are no nodes for providers other than our identity's `ory` and `github` but ideally I'd do a set-equals there). If you have any ideas on how to improve testing this, let me know and I'll try to apply them.

Hopefully no E2E tests rely on current behavior :pray:

### DIFF
--- a/selfservice/strategy/oidc/strategy.go
+++ b/selfservice/strategy/oidc/strategy.go
@@ -18,6 +18,7 @@ import (
 	"golang.org/x/oauth2"
 
 	"github.com/ory/kratos/cipher"
+	"github.com/ory/kratos/selfservice/flowhelpers"
 	"github.com/ory/kratos/selfservice/sessiontokenexchange"
 	"github.com/ory/x/jsonnetsecure"
 	"github.com/ory/x/otelx"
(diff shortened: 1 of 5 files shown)


Saancreed:
@aeneasr Done :slightly_smiling_face:

jonas-jonas (approved):
Very nice changes! Thank you for adding the e2e test.

PR #3643, opened 2023-11-26
fix: ignore CSRF middleware on Apple OIDC callback

sidartha:
This change ensures that CSRF middleware is ignored on `/self-service/oidc/callback/apple` (previously it was only exempted).

After a user authenticates on Apple, Apple redirects the user back to Kratos using a form POST. Since this is a POST, the CSRF cookie is not passed (since the cookie's SameSite attribute is set to Lax). This causes Kratos to add a different CSRF cookie.

While this doesn't break the registration flow itself, it causes a problem if we want to send the user back to the original login flow. In some cases, we may want to do this to cancel the registration and allow the user to log in with different credentials. It may be important to maintain the existing login flow in case the login originated from an OAuth2 client via Hydra instead of simply creating a new login flow.

## Related issue(s)

Not reported separately. Steps to reproduce: 

1. Configure Kratos with Apple OIDC provider.
2. Attempt to login with an Apple account that hasn't been registered in Kratos previously.
3. After Apple redirects back to Kratos, you will reach the registration page. 
4. On the registration page, redirect the user back to original login flow. 
5. The login flow cannot be accessed since the CSRF cookie has been overridden.

This issue does not impact other OIDC providers such as Google or Facebook since they do not rely on form POST when redirecting to the Kratos callback endpoint.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/internal/testhelpers/session.go
+++ b/internal/testhelpers/session.go
@@ -251,3 +251,13 @@ func (ct *TransportWithHeader) RoundTrip(req *http.Request) (*http.Response, err
 	}
 	return ct.RoundTripper.RoundTrip(req)
 }
+
+func AssertNoCSRFCookieInResponse(t *testing.T, _ *httptest.Server, _ *http.Client, r *http.Response) {
+	found := false
+	for _, c := range r.Cookies() {
+		if strings.HasPrefix(c.Name, "csrf_token") {
+			found = true
+		}
+	}
+	require.False(t, found)
+}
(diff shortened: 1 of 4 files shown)


jonas-jonas (approved):
This LGTM. Thank you. 

Test failures seem unrelated. I re-ran. Let's see.

hperl (approved):
LGTM, thanks for the fix!

Issue #3559, opened 2023-10-07
SMTP Connection URI is required on HTTP delivery strategy


albertilagan:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [x] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [x] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

`smtp.connection_uri` is still required even though `delivery_strategy` is set to `http`

### Reproducing the bug

Config
```
courier:
  delivery_strategy: http
  http:
    request_config:
      url: http://localhost:3000/api/webhook/kratos
      method: POST
      auth:
        type: api_key
        config:
          name: "token"
          value: "abcd1234"
          in: header
```

### Relevant log output

```shell
`courier.smtp.connection_uri: <nil>`
                                 ^-- one or more required properties is missing
```

### Relevant configuration

```yml
courier:
  delivery_strategy: http
  http:
    request_config:
      url: http://localhost:3000/api/webhook/kratos
      method: POST
      auth:
        type: api_key
        config:
          name: "token"
          value: "abcd1234"
          in: header
```

### Version

master

### On which operating system are you observing this issue?

macOS

### In which environment are you deploying?

Other

### Additional Context

workaround, adding dummy smtp connection uri.

```
courier:
  delivery_strategy: http
  http:
    request_config:
      url: http://localhost:3000/api/webhook/kratos
      method: POST
      auth:
        type: api_key
        config:
          name: "token"
          value: "abcd1234"
          in: header
  smtp:
    connection_uri: smtps://foo:bar@my-mailserver:1234/?skip_ssl_verify=false
```

reginbald:
Encountered this as well

matthiastz:
the issue seems not to be fixed, still. would be cool to re-open this issue

when I try to run kratos with docker-compose and kratos v1.1.0

`kratos.yml`
```
...
courier:
  delivery_strategy: http
  http:
    request_config:
      url: http://localhost:8080/api/someUri
```

I still get this error:

```
[...]

PR #3649, opened 2023-11-30
feat: add sms verification for phone numbers

jonas-jonas:
## Related issue(s)
Fixes https://github.com/ory/kratos/issues/3559

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


contrib/quickstart/kratos/phone-password/kratos.yml
```diff
@@ -0,0 +1,113 @@
...
+        url: https://api.twilio.com/2010-04-01/Accounts/AXXXXXXXXXXXXXX/Messages.json
+        method: POST
+        body: base64://ZnVuY3Rpb24oY3R4KSB7CkJvZHk6IGN0eC5ib2R5LApUbzogY3R4LnRvLEZyb206IGN0eC5mcm9tCn0=
+        headers:
+          Content-Type: application/x-www-form-urlencoded
+        auth:
+          type: basic_auth
+          config:
+            user: AXXXXXXX
+            password: XXXX
```
jonas-jonas, line 110: This is a dummy config to use with twilio. The Auth stuff needs to be replaced with real values, but other than that this should be all that's needed to enable this.

contrib/quickstart/kratos/phone-password/kratos.yml
```diff
@@ -0,0 +1,113 @@
...
+
+identity:
+  default_schema_id: default
+  schemas:
+    - id: default
+      url: file:///etc/config/kratos/identity.schema.json
+
+courier:
+  channels:
+    - id: phone
```
jonas-jonas, line 99: This value corresponds to the value from the identity schema.

courier/courier.go
```diff
@@ -51,26 +46,35 @@ type (
...
 	if err != nil {
 		return nil, err
 	}
+	return NewCourierWithCustomEmailChannel(ctx, deps, emailChannel)
+}
+
+func NewCourierWithCustomEmailChannel(ctx context.Context, deps Dependencies, channel *EmailChannel) (Courier, error) {
+	channels := make(map[string]Channel, len(deps.CourierConfig().CourierChannels(ctx)))
+	for _, c := range deps.CourierConfig().CourierChannels(ctx) {
+		// TODO: add support for more channel types (e.g. SMTP, ..?)
```
aeneasr, line 67: TODO?

jonas-jonas: Yea, @zepatrik suggested that we replace the current config structure with a "channel" based system, that allows you to specify which channel the verification should go through in the identity schema (the `"via"` key). And to still support SMTP we'll need to add a "channel" type config key, but for now only the phone channel is supported this way. I'd work on that, once SMS is fully done.

alnr:
Nice work! Two thoughts:
1. We might want SSRF protection on the SMTP+HTTP endpoints configurable by the user.
2. The distiction between HTTP email and SMS is not quite clear to me. I like the use of `via` to distinguish which template and address to select, but exactly is this extendable to other "channels"? For example, native notifications?

jonas-jonas:
>We might want SSRF protection on the SMTP+HTTP endpoints configurable by the user.

Good point, maybe we can do this in a follow-up, depending on demand? Not sure if this is a requirement for anyone right now.

>The distiction between HTTP email and SMS is not quite clear to me. I like the use of via to distinguish which template and address to select, but exactly is this extendable to other "channels"? For example, native notifications?

It's not quite complete yet. The long term plan is to deprecate the `smtp:` and `http:` sections in the config and do everything via channels. You'd then be able to set the channel to use for verification, recovery, login via code, etc. in the identity schema (the `via` key there). But that change seems rather breaking to me, so I'd like to do this in a follow-up, once we verified that this works with the SMS channel.

alnr (dismissed):

aeneasr:
> We might want SSRF protection on the SMTP+HTTP endpoints configurable by the user.

SSRF is currently governed globally, I would keep it that way and not add more configuration options for the courier's SSRF protections.

embedx/config.schema.json
```diff
@@ -1963,9 +1963,29 @@
...
+          "items": {
+            "title": "Courier channel configuration",
+            "type": "object",
+            "properties": {
+              "id": {
+                "type": "string",
+                "title": "Channel id",
+                "description": "The channel id. Corresponds to the .via property of the identity schema for recovery, verification, etc. Currently only phone is supported.",
+                "maxLength": 255,
+                "enum": ["phone"]
```
aeneasr, line 1978: Shouldn't this be `sms` given that we're not calling anyone?

embedx/config.schema.json
```diff
@@ -1963,9 +1963,29 @@
...
+              "request_config": {
+                "$ref": "#/definitions/httpRequestConfig"
+              }
+            },
+            "required": ["id", "request_config"],
+            "additionalProperties": false
+          }
         }
       },
-      "required": ["smtp"],
```
aeneasr, line 1968: From now on it's no longer required to set up an SMTP server in the config. What will happen if we don't define an SMTP server here?

persistence/sql/migrations/sql/20231130094628000000_courier_message_channel.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE
+  courier_messages
+ADD
+  column channel VARCHAR(255) NULL;
```
aeneasr, line 4: Does this need to be 255 long?

aeneasr (commented):
Very nice, just a few minor comments!

aeneasr:
Thinking about the channels for a bit, isn't this a bit too powerful & complicated? Do we need separate sender configs for verification and recovery emails? I imagine this to be quite complex to set up in the UI and to make people understand how to use this feature, especially because you need to understand how it maps to the identity schema.

In the end what most people want is "send an SMS code" and "send an email code". Push notifications maybe at some point too, but that is a whole different beast we can tackle when it becomes a requirement. I would favor a more simplistic configuartion approach, especially for the time being as we only have SMS sending.

embedx/config.schema.json
jonas-jonas: It's essentially an HTTP based sender, and you can use e.g. Twilio to dispatch phone calls instead of SMS here already, AFAICT (have only tried SMS so far). So `phone` makes most sense to me.

embedx/config.schema.json
jonas-jonas: People won't complain about having to set an SMTP URL, even if you only use the HTTP based sender ;)

I can't think of any implications. If you don't set the connection URL, instantiating the courier will return an error here: https://github.com/ory/kratos/blob/c0e528f3408e001cd29e09406094723e09a48147/courier/smtp.go#L30

And if you don't set any other delivery config, Kratos won't start, because we fall back to the SMTP "channel" in that case. 

So essentially, instead of catching this "misconfiguration" in the config validation is caught in the bootstrapping logic of Kratos, which doesn't lead to any change in the behavior of Kratos as it won't start either way.

persistence/sql/migrations/sql/20231130094628000000_courier_message_channel.up.sql
jonas-jonas: Nope, true. I guess 32 would be enough.

embedx/config.schema.json
aeneasr: The bootstraping logic is different in Ory Network, will that have negative consequences there?

embedx/config.schema.json
aeneasr: It looks like phone calls are a bit more difficult as they have more states (calling, ringing, answered, done) and more instructions including a callback for when the call is done (see: https://www.twilio.com/docs/voice/make-calls#initiate-an-outbound-call-with-twilio), as well as a more complex set up for what twilio should do when making the call (https://www.twilio.com/docs/voice/twiml). I'm not sure if this would work with the way we do it here.

I think phone and sms are interchangeable too. SMS is probably easier to understand what it does for people who need to configure/implement this, than `phone`. We could also call it `http_callback` (what it really is) but I think that would be even more difficult to understand. Up to you

embedx/config.schema.json
jonas-jonas: Only for the backoffice courier. Which I have adapted already.

aeneasr: Awesome!

jonas-jonas:
As discussed directly, the channels approach gives great flexibility, and we want to keep it. However, in the Ory Network we're not going to expose this in the UI (for now) and only offer smtp, http, and sms channels. That way, first time users do not get lost in these dependent config values (the via key in the identity schema corresponds to the channel ID, etc.). (The first release of this, is going to be CLI configuration only, anyway.)

Once we have an identity schema editor, this might be easier to describe in the UI and we can re-think this approach.

embedx/config.schema.json
jonas-jonas: I changed the value for the `verification.via` key to `"sms"`, because we also call this message type `"sms"` in the code. That should make it easier to understand for now. We'll have to figure out, if this is an issue, if users want to use whatsapp or other messengers here.

jonas-jonas: Actually nvm. We have a few places in the code that already make assumptions about the verifiable address if it's via value is "phone". Untangling that, is pretty difficult. 

E.g. https://github.com/ory/kratos/blob/c1406fb79fce56f6da9917d50b41065ae5a1c9bf/identity/extension_verification.go#L45

PR #3650, opened 2023-12-01
feat: extend recovery flow by adding the FlowID

matthiastz:
- the pw recovery flow has been extended
- specifically the `RecoveryCodeValidModel` now also holds the FlowID
- context: the FlowID is both persisted and associated to the `RecoveryCode` data model already
- 🚀 when a Courier sends a new email for the pw recovery flow (valid), the `Message.TemplateData` now includes all the props as before + the FlowID
  - this was implemented and asserted in `code_sender_test.go`
  - having this `FlowID` parameter gives more flexible ways to design the recovery flow with a custom UI: when a user first triggers a pw recovery flow in the UI (a FlowID is already created), the user can close this Browser Tab and later on opens the email with the link that includes both the original FlowID and the recovery code _(instead of keeping the Browser tab open and copy over the code manually)_, which all match to one Flow of the user 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/courier/template/email/recovery_code_valid.go
+++ b/courier/template/email/recovery_code_valid.go
@@ -19,6 +19,7 @@ type (
 	}
 	RecoveryCodeValidModel struct {
 		To           string
+		FlowID       string
 		RecoveryCode string
 		Identity     map[string]interface{}
 	}
--- a/selfservice/strategy/code/code_sender.go
+++ b/selfservice/strategy/code/code_sender.go
@@ -228,6 +228,7 @@ func (s *Sender) SendRecoveryCodeTo(ctx context.Context, i *identity.Identity, c
 
 	emailModel := email.RecoveryCodeValidModel{
 		To:           code.RecoveryAddress.Value,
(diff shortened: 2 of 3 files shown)


jonas-jonas:
It was a deliberate choice to not include the link to the recovery form in the email body. The reason is, that the flow is protected against CSRF attacks, and if a violation occurs, the flow is voided. A violation could occur if the user opens the link in a different browser than they started the flow in. This would then just show a CSRF violation error message, and the user would need to re-start the flow. And to not even make it possible to implement these potentially broken flows, we omitted the link from the template data.

Is this the use-case you're trying to cover?

matthiastz:
@jonas-jonas thx for providing the context. Yeah thats pretty much the use-case.

So I assume you would deny/remove this PR, right?

jonas-jonas:
Yes, I don't think we can accept this PR. 

Still, thank you, for your time and contribution! :)

PR #3656, opened 2023-12-11
feat: passkey strategy with resident keys

hperl:
# TODO

* [x] Two distinct strategies, how to prevent passkey + webauthn passkey with too high AAL
	* Passkey is currently always AAL1 and not a second factor.
	* We could disallow using the same authenticator ID (AAGUID) for both webauthn and passkey. However, this could hinder migration from webauthn+passwordless to passkey.
* [x] PassKeys what happens to non-resident keys? For example when migrating WebAuthn passwordless credentials to passkey?
  * This work does not yet consider converting passwordless credentials to a passkey. I think the best way would be to do this with user interaction and insert a "nag screen" after login to prompt the user to create a passkey (and optionally delete/deactivate the webauth passwordless credential). The same technique could be used to onboard users to passkeys regardless of their primary login strategy.
* [x] 2x PassKeys if submission fails
  * In some cases it can be that the passkey is created in the browser but not stored in the database. Ultimately, there is nothing that we can do to completely prevent this, but is should be uncommon, because we store and validate the traits first before we create the passkey.
* [x] WebAuthn.js - use onload for autocomplete
  * Done
* [x] How will the deprecation work of the webauthn strategy?
  * This will need further design and code.
* [x] Clean up `webauthnIdentifierNode`

## Related issue(s)

Docs PR: https://github.com/ory/docs/pull/1626

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


embedx/config.schema.json
```diff
@@ -819,6 +819,9 @@
         "webauthn": {
           "$ref": "#/definitions/selfServiceAfterSettingsAuthMethod"
         },
+        "passkey": {
```
jonas-jonas, line 822: @aeneas passkey or passkeys? your call.

jonas-jonas, line 822: @aeneasr passkey or passkeys? your call.

aeneasr: PassKeys is the official name so I would stick to that :)

selfservice/strategy/passkey/passkey_registration.go
```diff
@@ -0,0 +1,358 @@
...
+		Credentials: identity.CredentialsWebAuthn{*credentialWebAuthn},
+		UserHandle:  webAuthnSess.UserID,
+	})
+	if err != nil {
+		return s.handleRegistrationError(w, r, regFlow, params, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to encode identity credentials.").WithDebug(err.Error())))
+	}
+
+	ident.UpsertCredentialsConfig(s.ID(), credentialsConfig, 1)
+	passkeyCred, _ := ident.GetCredentials(s.ID())
+	passkeyCred.Identifiers = []string{string(webAuthnSess.UserID)}
```
aeneasr, line 182: I think we should  return an easy to understand error if `webAuthnSess.UserID` is empty. If it is empty, the current validation will pass it as valid

```
	c := i.GetCredentialsOr(identity.CredentialsTypePasskey, &identity.Credentials{})
	if len(c.Identifiers) == 0 {
		return schema.NewMissingIdentifierError()
	}
```

and will lead to an empty string identifier. We had this problem once before in the WebAuthn code.

selfservice/strategy/passkey/passkey_login.go
```diff
@@ -0,0 +1,280 @@
...
+
+func (s *Strategy) RegisterLoginRoutes(r *x.RouterPublic) {
+	webauthnx.RegisterWebauthnRoute(r)
+}
+
+func (s *Strategy) PopulateLoginMethod(r *http.Request, _ identity.AuthenticatorAssuranceLevel, sr *login.Flow) error {
+	if sr.Type != flow.TypeBrowser {
+		return nil
+	}
+
```
aeneasr, line 43: Shouldn't this only populate for AAL1?

selfservice/strategy/passkey/passkey_login.go
```diff
@@ -0,0 +1,280 @@
...
+	var webAuthnSess webauthn.SessionData
+	if err := json.Unmarshal([]byte(gjson.GetBytes(f.InternalContext, flow.PrefixInternalContextKey(s.ID(), InternalContextKeySessionData)).Raw), &webAuthnSess); err != nil {
+		return nil, s.handleLoginError(r, f, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Expected WebAuthN in internal context to be an object but got: %s", err)))
+	}
+	webAuthnSess.UserID = nil
+
+	i, _, err := s.d.PrivilegedIdentityPool().FindByCredentialsIdentifier(r.Context(), s.ID(), string(webAuthnResponse.Response.UserHandle))
+	if err != nil {
+		time.Sleep(x.RandomDelay(s.d.Config().HasherArgon2(r.Context()).ExpectedDuration, s.d.Config().HasherArgon2(r.Context()).ExpectedDeviation))
```
aeneasr, line 234: I don't think this is needed here. We're looking for the identity based on the user handle which is a randomly generated string. If this was an email, this could make sense here.

selfservice/strategy/passkey/passkey_login.go
```diff
@@ -0,0 +1,280 @@
...
+		return nil, s.handleLoginError(r, f, errors.WithStack(schema.NewNoWebAuthnCredentials()))
+	}
+	err = s.d.PrivilegedIdentityPool().HydrateIdentityAssociations(ctx, i, identity.ExpandCredentials)
+	if err != nil {
+		return nil, s.handleLoginError(r, f, errors.WithStack(schema.NewNoWebAuthnCredentials()))
```
aeneasr, line 239: This isn't the correct error? We have a DB error above

aeneasr (commented):
Preliminary review - also pushing some code changes now. Can you please add e2e tests for the various flows?

ui/node/attributes.go
```diff
@@ -98,6 +98,8 @@ type InputAttributes struct {
 	//
 	// required: true
 	NodeType string `json:"node_type"`
+
+	DataWebauthnIdentifier bool `json:"-" faker:"-"`
```
aeneasr, line 102: Carries state that is used in another context. Leaky?

ui/node/attributes_input.go
```diff
@@ -152,6 +147,11 @@ func NewInputFieldFromSchema(name string, group UiNodeGroup, p jsonschemax.Path,
 		Type:     toFormType(p.Name, p.Type),
 		Required: p.Required,
 	}
+	if config, ok := p.CustomProperties[schema.ExtensionName]; ok {
+		if config, ok := config.(*schema.ExtensionConfig); ok {
+			attr.DataWebauthnIdentifier = config.Credentials.WebAuthn.Identifier || config.Credentials.Passkey.Identifier
```
aeneasr, line 152: Question this

ui/node/identifiers.go
```diff
@@ -27,4 +27,12 @@ const (
 	WebAuthnRegisterDisplayName = "webauthn_register_displayname"
 	WebAuthnRemove              = "webauthn_remove"
 	WebAuthnScript              = "webauthn_script"
+	PasskeyRegisterTrigger      = "passkey_register_trigger"
+	PasskeyRegister             = "passkey_register"
+	PasskeySettingsRegister     = "passkey_settings_register"
```
aeneasr, line 32: What is `settingsRegister`'s responsibility?

ui/node/identifiers.go
```diff
@@ -27,4 +27,12 @@ const (
...
+	PasskeyCreateData           = "create_passkey_data"
```
aeneasr, line 33: All other names are prefixed with the credential type - please keep consistency?

ui/node/identifiers.go
```diff
@@ -27,4 +27,12 @@ const (
...
+	PasskeyLogin                = "passkey_login"
+	PasskeyChallenge            = "passkey_challenge"
+	PasskeyLoginTrigger         = "login_with_passkey"
```
aeneasr, line 36: All other names are prefixed with the credential type - please keep consistency?

x/webauthnx/aaguid/aaguid.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func init() {
+	err := json.Unmarshal(rawAAGUIDs, &aaguids)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func Lookup(id []byte) *AAGUID {
```
aeneasr, line 32: One test maybe? :)

x/webauthnx/js/webauthn.js
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  async function __oryPasskeyLoginAutocomplete() {
+    const dataEl = document.getElementsByName("passkey_challenge")[0]
+    const resultEl = document.getElementsByName("passkey_login")[0]
+    const identifierEl = document.getElementsByName("identifier")[0]
+
+    if (!dataEl || !resultEl || !identifierEl) {
+      console.debug("__oryPasskeyLoginAutocomplete: mandatory fields not found")
+      return
```
aeneasr, line 135: What happens here? Is the flow simply aborted? Does the user receive some indication why it did not work?

aeneasr (changes requested):

identity/credentials_webauthn.go
```diff
@@ -19,17 +21,24 @@ type CredentialsWebAuthnConfig struct {
...
+		AddedAt:         time.Now().UTC().Round(time.Second),
 		Authenticator: AuthenticatorWebAuthn{
 			AAGUID:       credential.Authenticator.AAGUID,
 			SignCount:    credential.Authenticator.SignCount,
 			CloneWarning: credential.Authenticator.CloneWarning,
 		},
 	}
+	id := aaguid.Lookup(credential.Authenticator.AAGUID)
+	if id != nil {
+		cred.DisplayName = id.Name
```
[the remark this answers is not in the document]
aeneasr: I see now what aaguid is doing!

x/webauthnx/js/webauthn.js
```diff
@@ -0,0 +1,377 @@
...
+              credential.response.clientDataJSON,
+            ),
+          },
+        })
+
+        resultEl.closest("form").submit()
+
+        document
+          .querySelector('*[type="submit"][name="method"][value="passkey"]')
+          .click()
```
aeneasr, line 353: Double submit?

aeneasr (changes requested):
This is in a pretty good

x/webauthnx/js/webauthn.js
hperl: This is now only called when the correct node is there. Should the inputs not be there (because of a bug in the UI most likely), the code will simply not execute.

ui/node/identifiers.go
hperl: The corresponding input field holds the browser response to the passkey challenge.

See: https://github.com/ory/kratos/blob/60a1817c6660d1eb668a9994026da385416dcf44/x/webauthnx/js/webauthn.js#L316

persistence/sql/migrations/sql/20231108111100000000_credential_types_passkey.up.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO identity_credential_types (id, name)
+SELECT '8d0ca544-9bf6-45d3-bd75-0bbb3aeba3c7', 'passkey'
```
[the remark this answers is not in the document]
hperl: I don't understand how this is important, given that we check the value into git. What is the security consideration here?

selfservice/strategy/passkey/passkey_login.go
hperl: fixed

selfservice/strategy/passkey/passkey_registration.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+
+	regFlow.TransientPayload = params.TransientPayload
+
+	if err := flow.EnsureCSRF(s.d, r, regFlow.Type, s.d.Config().DisableAPIFlowEnforcement(ctx), s.d.GenerateCSRFToken, params.CSRFToken); err != nil {
+		return s.handleRegistrationError(w, r, regFlow, params, err)
+	}
+
+	if len(params.Register) == 0 {
+		return s.createPasskey(r, w, regFlow, params)
```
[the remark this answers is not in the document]
hperl: Because we need the identifier to properly name the passkey.

selfservice/strategy/passkey/passkey_registration.go
```diff
@@ -0,0 +1,358 @@
...
+
+	if regFlow.Type != flow.TypeBrowser {
+		return flow.ErrStrategyNotResponsible
+	}
+
+	params, err := s.decode(r)
+	if err != nil {
+		return s.handleRegistrationError(w, r, regFlow, params, err)
+	}
+	if params.Register == "" && params.Method != "passkey" {
```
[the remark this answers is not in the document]
hperl: Not in a benign case, because the register param is set right before submission.

As for an attacker, what would the gain be?

selfservice/strategy/passkey/passkey_registration.go
```diff
@@ -0,0 +1,358 @@
...
+		if devErr := new(protocol.Error); errors.As(err, &devErr) {
+			s.d.Logger().WithError(err).WithField("error_devinfo", devErr.DevInfo).Error("Failed to create WebAuthn credential")
+		}
+		return s.handleRegistrationError(w, r, regFlow, params, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to create WebAuthn credential: %s", err)))
+	}
+
+	credentialWebAuthn := identity.CredentialFromWebAuthn(credential, true)
+	credentialsConfig, err := json.Marshal(identity.CredentialsWebAuthnConfig{
+		Credentials: identity.CredentialsWebAuthn{*credentialWebAuthn},
+		UserHandle:  webAuthnSess.UserID,
```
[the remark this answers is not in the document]
hperl: The display name of the key will be derived from the AAGUID, see here: https://github.com/ory/kratos/blob/60a1817c6660d1eb668a9994026da385416dcf44/identity/credentials_webauthn.go#L36-L39

selfservice/strategy/passkey/passkey_registration.go
hperl: In this case, `webAuthnSess.UserID` is not user-supplied. It comes from the internal context, which is set here: https://github.com/ory/kratos/blob/60a1817c6660d1eb668a9994026da385416dcf44/selfservice/strategy/passkey/passkey_registration.go#L295-L303

I can add another check to make sure it really is not empty, but it would be an internal server error, as the user can't do anything against it.

selfservice/strategy/passkey/passkey_login.go
hperl: makes sense, fixed :)

selfservice/strategy/passkey/passkey_login.go
hperl: fixed.

embedx/config.schema.json
```diff
@@ -1630,6 +1639,67 @@
                 "required": ["config"]
               }
             },
+            "passkey": {
```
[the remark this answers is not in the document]
hperl: Agreed, but the webauthn config also had the legacy single origin, which I did not copy over.

I can also unify this. Your call.

identity/extension_credentials.go
```diff
@@ -56,6 +56,12 @@ func (r *SchemaExtensionCredentials) Run(ctx jsonschema.ValidationContext, s sch
 		r.setIdentifier(CredentialsTypeWebAuthn, value, CredentialsIdentifierAddressTypeNone)
 	}
 
+	if s.Credentials.Passkey.Identifier {
+		// The passkey identifier is a random user ID, not an email.
+		return nil
+		//r.setIdentifier(CredentialsTypePasskey, value, CredentialsIdentifierAddressTypeNone)
+	}
+
```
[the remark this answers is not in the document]
hperl: fixed.

ui/node/attributes.go
hperl: fixed

ui/node/attributes_input.go
hperl: fixed.

selfservice/strategy/passkey/passkey_registration.go
```diff
@@ -0,0 +1,346 @@
...
+	})
+	if err != nil {
+		return s.handleRegistrationError(w, r, regFlow, params, errors.WithStack(
+			herodot.ErrInternalServerError.WithReasonf("Unable to encode identity credentials.").WithDebug(err.Error())))
+	}
+
+	ident.UpsertCredentialsConfig(s.ID(), credentialWebAuthnConfig, 1)
+	passkeyCred, _ := ident.GetCredentials(s.ID())
+	passkeyCred.Identifiers = []string{string(webAuthnSess.UserID)}
```
jonas-jonas, line 186: This can panic, no? shouldn't we check that it's actually a non-nil value?

selfservice/strategy/passkey/passkey_schema_extension.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (e *SchemaExtension) Finish() error { return nil }
+
+// PasskeyIdentifierFromIdentity returns the passkey identifier from the
+// identity. It is usually the email address and used to name the passkey in the
+// browser.
+func (s *Strategy) PasskeyIdentifierFromIdentity(ctx context.Context, id *identity.Identity) string {
+	e := new(SchemaExtension)
+	_ = s.d.IdentityValidator().ValidateWithRunner(ctx, id, e)
```
jonas-jonas, line 40: Why do we ignore the error here?

selfservice/strategy/passkey/passkey_strategy.go
```diff
@@ -0,0 +1,111 @@
...
+	session.HandlerProvider
+	session.ManagementProvider
+}
+
+type Strategy struct {
+	d  strategyDependencies
+	hd *decoderx.HTTP
+}
+
+func NewStrategy(d any) *Strategy {
```
jonas-jonas, line 70: Why not `strategyDependencies`?

test/e2e/cypress/integration/profiles/passwordless/flows.spec.ts
```diff
@@ -193,7 +193,7 @@ context("Passwordless registration", () => {
         cy.get('[value="password"]').should("exist")
       })
 
-      it("should be able to refresh", () => {
+      it.only("should be able to refresh", () => {
```
jonas-jonas, line 196: remove

jonas-jonas (changes requested):
LGTM, just a few minor remarks.

test/e2e/cypress/integration/profiles/passwordless/flows.spec.ts
hperl: fixed.

selfservice/strategy/passkey/passkey_schema_extension.go
hperl: Proper validation already happens later. We just want the identifier node, and if it is empty, handle it properly on the call site.

selfservice/strategy/passkey/passkey_registration.go
hperl: We now check it here: https://github.com/ory/kratos/blob/62c6118bf2e6b3cd984bdda94c084848678e2c6a/selfservice/strategy/passkey/passkey_registration.go#L145

selfservice/strategy/passkey/passkey_strategy.go
hperl: This is the standard factory signature, so that we can pass an array of replacement strategies (which we use in Cloud). Otherwise, each factory would have a different signature.

See: https://github.com/ory/kratos/blob/62c6118bf2e6b3cd984bdda94c084848678e2c6a/driver/registry.go#L211-L220

aeneasr (approved):

aeneasr:
@jonas-jonas are you good with the current state?

jonas-jonas:
@aeneasr I'll re-review now.

selfservice/strategy/passkey/passkey_strategy.go
jonas-jonas: ah I see.

jonas-jonas (approved):
Except for https://github.com/ory/kratos/pull/3656#discussion_r1447420664 this LGTM.

selfservice/strategy/passkey/passkey_registration.go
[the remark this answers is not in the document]
hperl: Requiring the method to be set will further complicate submitting the passkey, which is done through a form submit in JS, see: https://github.com/ory/kratos/blob/96c2164e2ad5befa1b05be14b6011556b88ec136/x/webauthnx/js/webauthn.js#L121

We don't require the method to be set for webauthn register requests either, btw: https://github.com/ory/kratos/blob/96c2164e2ad5befa1b05be14b6011556b88ec136/selfservice/strategy/webauthn/registration.go#L94-L114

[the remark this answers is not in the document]
hperl: Specifically, I could add a hidden input with name=method and value=passkey in JS. Is it worth it?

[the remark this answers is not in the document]
jonas-jonas: I'd say it's worth the effort also submitting the method, since we do have the (sometimes loose) convention to submit the method as well. Since this would just be another edge case, potentially causing issues down the road.

But, in the interest of velocity, I'll leave the decision up to you.

hperl:
@jonas-jonas, @aeneasr, two points remain open:

- As I started writing docs, I stumbled over the fact that users need to modify the identity schema to add a `passkey: { identifier: true }` to their schema, most likely to the same identifier as used for webauthn. Do you think there is even a use case for using different identifiers for webauthn and passkey?
- I can't get the coverage up further. Some errors only trigger when the database misbehaves, which would require extensive mocking. How should I proceed? The remaining tests would look more and more artifical.

jonas-jonas (dismissed):

hperl:
I'd like to have one more pair of eyes on the security side of this before merging. @zepatrik maybe?

aeneasr:
I found a few bugs when enabling both webauthn and passkey strategies, probably the triggers get confused. Wrote those things in Slack - would be good to get those fixed in my view :)

hperl:
included in https://github.com/ory/kratos/pull/3748

PR #3658, opened 2023-12-11
fix: incorrect login accept challenge

aeneasr:


aeneasr:
@hperl can you please add tests to cover these verfication cases? I looked into recovery and it will have the same problem if Hydra code is added there.

@jonas-jonas can you refactor the code to return an error instead of `nil`. In general we always return errors even on validation errors. The error handles should/are capable of handling validation errors properly. The `nil` return on invalid code use is true for both verification and recovery and has a high risk of introducing regressions in the future.

Thanks both!

PR #3669, opened 2023-12-21
fix: check whoami aal before accepting hydra login request

K3das:
For JSON browser flows, the AAL check only happens after accepting Hydra login challenge, causing login challenges to be accepted without AAL2 requirement. This moves the AAL check to before the login challenge fulfillment, and adds tests for both normal and JSON browser flows.

## Related issue(s)

Previously unknown bug, steps to reproduce:
1. Start a hydra login flow as a logged out user
2. Login with a user with AAL2 available using a JSON request (and Kratos configured to use `max_available` AAL for whoami)
3. Kratos will accept the Hydra login request despite a failing whoami

Normal form browser flows do not have this issue and correctly redirect to a login flow that requests a higher AAL.

## Checklist

      introduces a new feature.
- [X] I am following the
      the changes.
      works.


aeneasr (changes requested):
Nice find! I think the new implementation is not yet fully correct (see comment) but in general this is a bug!

selfservice/flow/login/hook.go
```diff
@@ -257,6 +257,16 @@ func (e *HookExecutor) PostLoginHook(
 		// Browser flows rely on cookies. Adding tokens in the mix will confuse consumers.
 		s.Token = ""
 
+		// If we detect that whoami would require a higher AAL, we redirect!
+		if _, err := e.requiresAAL2(r, s, a); err != nil {
+			if aalErr := new(session.ErrAALNotSatisfied); errors.As(err, &aalErr) {
+				span.SetAttributes(attribute.String("return_to", aalErr.RedirectTo), attribute.String("redirect_reason", "requires aal2"))
+				e.d.Writer().WriteError(w, r, flow.NewBrowserLocationChangeRequiredError(aalErr.RedirectTo))
```
aeneasr, line 264: This would loose the OAuth2 URL though, would it not? So the user would end up in the "Please enter your second factor" but after submitting they would end up at the default return URL. Instead however, they should complete the OAuth2 flow.

K3das: Nope, user ends up at correct endpoint because of login_challenge also getting passed down - this uses the same, working(?), logic as the form body response, and after testing it works fine...

aeneasr: I see, that's great! Would it be possible for you to add a test case to make sure this doesn't break in the future? :)

K3das: Made my tests more verbose to check for login_challenge anyway

aeneasr (approved):
Thank you very much!

PR #3680, opened 2024-01-03
fix: rename "phone" courier channel to "sms"

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


aeneasr (approved):

hperl (dismissed):

courier/sms.go
```diff
@@ -29,7 +29,7 @@ func (c *courier) QueueSMS(ctx context.Context, t SMSTemplate) (uuid.UUID, error
 	message := &Message{
 		Status:       MessageStatusQueued,
 		Type:         MessageTypeSMS,
-		Channel:      "sms",
+		Channel:      "phone",
```
zepatrik, line 32: Why use `"phone"`? IMO, this is not ideal as it is not clear whether the message is delivered as an SMS or phone call.

zepatrik (changes requested):

courier/sms.go
aeneasr: I kinda agree with this, see discussion here: https://github.com/ory/kratos/pull/3649#discussion_r1436955299

zepatrik: So you agree with sms or phone? I just thought that it makes more sense to make this as precise as possible, because it would be confusing to have later "phone" and "phone_call".

In the end no one forces you to use this for sms, but that's why we have the channels. One can add "phone_call", "whatsapp", "my_native_app_push_notification", or any other channel (once we fully expose that).

I am also not sure if we really need to mirror all the intermediary states of phone calls, all that Kratos cares about is whether the correct code was submitted, regardless how the user got the code.

aeneasr: I thought that `sms` made the most sense

jonas-jonas: Okay, I refactored the whole thing to use "sms" now.

identity/extension_verification.go
```diff
@@ -24,40 +32,58 @@ func NewSchemaExtensionVerification(i *Identity, lifespan time.Duration) *Schema
...
+			formatString = "email"
+			formatter, ok := jsonschema.Formats[formatString]
+			if !ok {
+				supportedKeys := maps.Keys(jsonschema.Formats)
+				return ctx.Error("format", "format %q is not supported. Supported formats are [%s]", formatString, strings.Join(supportedKeys, ", "))
+			}
+
+			if !formatter(value) {
+				return ctx.Error("format", "%q is not valid %q", value, formatString)
+			}
```
zepatrik, line 70: This block needs to be moved to after the if, right?

identity/identity_verification.go
```diff
@@ -57,7 +57,7 @@ type VerifiableAddress struct {
 	// enum: ["email"]
```
zepatrik, line 57:
```suggestion
	// enum: ["email", "sms"]
```
or remove this.
The generated SDK probably still respects this comment.

identity/extension_verification.go
```diff
@@ -24,40 +32,58 @@ func NewSchemaExtensionVerification(i *Identity, lifespan time.Duration) *Schema
 	return &SchemaExtensionVerification{i: i, lifespan: lifespan}
 }
 
+const (
+	ChannelTypeEmail = "email"
+	ChannelTypeSMS   = "sms"
+	ChannelTypePhone = "phone"
```
zepatrik, line 38:
```suggestion
```

identity/identity_verification.go
```diff
@@ -25,7 +25,7 @@ const (
 // VerifiableAddressType must not exceed 16 characters as that is the limitation in the SQL Schema
 //
 // swagger:model identityVerifiableAddressType
-type VerifiableAddressType string
+type VerifiableAddressType = string
```
zepatrik, line 28: Can we remove this type completely? Or should we keep it for documentation purposes? If yes, I'd suggest to revert the changes where you replace this type by string in other places.

zepatrik (commented):
Just some minor things. For recovery we'll do the same, but we don't yet support SMS recovery, right?

identity/extension_verification.go
zepatrik: Ah I guess not, because the library already did the check and we don't want to double-validate.

identity/identity_verification.go
jonas-jonas: It's still used in a lot of places. I think keeping it doesn't hurt.

identity/extension_verification.go
jonas-jonas: yup, exactly, the library doesn't de-duplicate the error messages, so this would lead to ugly errors.

zepatrik (approved):

PR #3682, opened 2024-01-06
feat: support MFA via SMS

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

## Further Comments

image


hperl (changes requested):
This looks very good! I just had a question regarding the enable logic for the strategies. What I'd like to prevent is cases where with a new update suddenly some strategy is disabled that was previously enabled. The rest looks good!

alnr (commented):
Found just two nits, otherwise this looks OK to me.

courier/template/sms/login_code_valid.go
```diff
@@ -0,0 +1,52 @@
...
+func (t *LoginCodeValid) PhoneNumber() (string, error) {
+	return t.model.To, nil
+}
+
+func (t *LoginCodeValid) SMSBody(ctx context.Context) (string, error) {
+	return template.LoadText(
+		ctx,
+		t.deps,
+		os.DirFS(t.deps.CourierConfig().CourierTemplatesRoot(ctx)),
+		"login_code/valid/sms.body.gotmpl", // TODO:!!!
```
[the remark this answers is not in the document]
jonas-jonas: The config keys for these templates. Thanks for catching this...

driver/registry_default.go
```diff
@@ -344,21 +344,11 @@ func (m *RegistryDefault) selfServiceStrategies() []any {
 }
 
 func (m *RegistryDefault) strategyRegistrationEnabled(ctx context.Context, id string) bool {
-	switch id {
-	case identity.CredentialsTypeCodeAuth.String():
-		return m.Config().SelfServiceCodeStrategy(ctx).PasswordlessEnabled
-	default:
-		return m.Config().SelfServiceStrategy(ctx, id).Enabled
-	}
+	return m.Config().SelfServiceStrategy(ctx, id).Enabled
```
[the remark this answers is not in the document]
jonas-jonas: Hm, that's true... 

I think the decision to deviate from the standard "one `enabled` property to enable a strategy" was a mistake, hence this change. 

WDYT about this migration strategy for the Ory Network: If a project has `code.passwordless_enabled: true` set to true, but `code.enabled: false`, we set `code.enabled: true`. 

I doubt this would affect many projects, since `code.enabled` is `true` by default.

Makefile
```diff
@@ -212,3 +212,7 @@ licenses: .bin/licenses node_modules  # checks open-source licenses
 node_modules: package-lock.json
 	npm ci
 	touch node_modules
+
+.PHONY: kratos-config-e2e
```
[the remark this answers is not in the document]
jonas-jonas: I added it

courier/template/sms/login_code_valid.go
```diff
@@ -0,0 +1,52 @@
...
+		model *LoginCodeValidModel
+	}
+	LoginCodeValidModel struct {
+		To        string
+		LoginCode string
+		Identity  map[string]interface{}
+	}
+)
+
+func NewLoginCodeValid(d template.Dependencies, m *LoginCodeValidModel) *LoginCodeValid {
```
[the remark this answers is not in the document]
jonas-jonas: Forgot the test for this. fixed.

driver/registry_default.go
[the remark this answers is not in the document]
hperl: sounds good. let's sync on that quickly.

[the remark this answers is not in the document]
jonas-jonas: As discussed personally, I added a switch to the configuration layer, to force-enable the code method, if `passwordless_enabled` is true: https://github.com/ory/kratos/pull/3682/commits/523811886eef62cbf699d06c85eb83be2d5d14db

hperl (dismissed):
LGTM!

driver/config/config.go
```diff
@@ -782,6 +793,7 @@ func (p *Config) SelfServiceCodeStrategy(ctx context.Context) *SelfServiceStrate
 			Config:  config,
 		},
 		PasswordlessEnabled: pp.BoolF(basePath+".passwordless_enabled", false),
+		MFAEnabled:          pp.BoolF(basePath+".mfa_enabled", false),
```
aeneasr, line 796: Can the code strategy be now enabled for both passwordless AND MFA at the same time?

identity/extension_credentials.go
```diff
@@ -70,6 +70,8 @@ func (r *SchemaExtensionCredentials) Run(ctx jsonschema.ValidationContext, s sch
 		// 	}
 
 		// 	r.setIdentifier(CredentialsTypeCodeAuth, value, CredentialsIdentifierAddressTypePhone)
+		case f.AddCase(""):
```
aeneasr, line 73: What is this for?

selfservice/flow/request.go
```diff
@@ -106,19 +105,7 @@ func MethodEnabledAndAllowed(ctx context.Context, flowName FlowName, expected, a
 		return errors.WithStack(ErrStrategyNotResponsible)
 	}
 
-	var ok bool
```
aeneasr, line 109: Are you certain that removing this won't cause regressions?

selfservice/flow/request_test.go
```diff
@@ -101,74 +101,3 @@ func TestMethodEnabledAndAllowed(t *testing.T) {
 		assert.Contains(t, string(body), "The requested resource could not be found")
 	})
 }
-
-func TestMethodCodeEnabledAndAllowed(t *testing.T) {
```
aeneasr, line 105: Has this test moved somewhere else? I think that would be important to ensure we don't have regressions.

identity/extension_credentials.go
jonas-jonas: ah, true. removed.

selfservice/strategy/code/strategy_login.go
```diff
@@ -194,22 +197,37 @@ func (s *Strategy) loginSendEmail(ctx context.Context, w http.ResponseWriter, r
...
 	p.Identifier = maybeNormalizeEmail(p.Identifier)
 
-	// Step 1: Get the identity
-	i, _, err := s.findIdentityByIdentifier(ctx, p.Identifier)
-	if err != nil {
-		return err
+	var addresses []Address
+	var i *identity.Identity
+	if f.RequestedAAL > identity.AuthenticatorAssuranceLevel1 {
+		address, found := lo.Find(sess.Identity.VerifiableAddresses, func(va identity.VerifiableAddress) bool {
```
aeneasr, line 203: Why aren't we sending it to the value of `p.Identifier` similar to how we do it for login/registration? I fear that if verified address is not set correctly, this flow won't work and it will not be clear why (just says "unknown address").

driver/config/config.go
jonas-jonas: yes, but a session that already used code for passwordless, does not result in an AAL2 session: https://github.com/ory/kratos/blob/f47c55c8142a087cf92f33ea451429e790da1967/selfservice/strategy/code/strategy_login.go#L64-L78

I think we should probably add a check to login flow creation if `aal=aal2`, to prevent someone from creating an MFA flow if the session used code for passwordless. WDYT?

aeneasr (commented):
Looks pretty good! A few more questions from my side

driver/config/config.go
aeneasr: I think that the administrator has to decide whether this method can be used for login or for MFA and we should not allow that this method can be used for both

jonas-jonas: But that limits the feature too much, I fear. Code via Email + Code via SMS is a legitimate use case, and should result in an aal2 session, right? If not, I am fine with adding a config validation to make these mutually exclusive.

selfservice/flow/request.go
jonas-jonas: I have added a fallback to enable the code strategy, if passwordless is enabled, but the code strategy is not: https://github.com/ory/kratos/blob/8821e6b0c51f7fbec05e05712ec4d7b53647aba4/driver/config/config.go#L764-L772

But, since code is enabled by default, there shouldn't be many cases where this case (`code.enabled: false && code.passwordless_enabled: true`) is relevant, in the first place.

selfservice/flow/request_test.go
jonas-jonas: This test only tested the special handling of `passwordless_enabled` in the method, which I removed. The `MethodEnabledAndAllowed` method is tested in `TestMethodEnabledAndAllowed` in this file. 

The code configuration is covered in the registry tests: https://github.com/ory/kratos/blob/8821e6b0c51f7fbec05e05712ec4d7b53647aba4/driver/registry_default_test.go#L618

driver/config/config.go
aeneasr: I see, that's a good point. I will take another look at this.

selfservice/strategy/code/strategy_login.go
jonas-jonas: We need to find out how to send the code to the user, which we only know through the verifiable address (`.Via`). 

But I don't really understand why we didn't do this for login and registration as well, in the first place.

identity/credentials_code.go
```diff
@@ -7,7 +7,7 @@ import (
 	"database/sql"
 )
 
-type CodeAddressType string
+type CodeAddressType = string
```
aeneasr, line 10: This is now a type alias - is that what we want here?

jonas-jonas: Yes, because these types don't really make sense anymore, with the change to the "channel" based approach. The longer term goal is to make this a string entirely and configure how to get the code to the user using the `"via"` argument in the identity schema.

selfservice/strategy/code/strategy_login.go
```diff
@@ -136,18 +143,14 @@ func (s *Strategy) findIdentityByIdentifier(ctx context.Context, identifier stri
...
-func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow, _ uuid.UUID) (_ *identity.Identity, err error) {
+func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow, sess *session.Session) (_ *identity.Identity, err error) {
 	ctx, span := s.deps.Tracer(r.Context()).Tracer().Start(r.Context(), "selfservice.strategy.code.strategy.Login")
 	defer otelx.End(span, &err)
 
 	if err := flow.MethodEnabledAndAllowedFromRequest(r, f.GetFlowName(), s.ID().String(), s.deps); err != nil {
 		return nil, err
 	}
 
-	if err := login.CheckAAL(f, identity.AuthenticatorAssuranceLevel1); err != nil {
```
aeneasr, line 147: I don't think this should be removed, but rather conditional on whether the code method can be used for 1fa or 2fa. Otherwise I think that we'll have problems with incorrect "you wanted 2FA but you have only 1FA" error handling

selfservice/strategy/code/strategy_login.go
jonas-jonas: I thought about this some more, and came to the conclusion that the best way to do this is using verifiable addresses, and that we should've done that for the normal login via code as well: With verifiable addresses we already established a pattern of "an OOB address, that can receive _something_". Having the code credential define this path in the identity schema again, essentially duplicates this for no (or very little) benefit. (Side note: this would've allowed us to build configs like `"require verified address for login via code"` way easier.)

However, repeating this mistake with MFA isn't necessary, IMO. If we were to use the code credential defined way of delivering the OTP for MFA, we're just delegating this config to the code login (which MFA is sort of, but maybe not obviously so). Verifiable addresses fit this criterion better, IMO.

test/e2e/cypress/integration/profiles/mfa/code.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+
+context("2FA code", () => {
+  ;[
+    // {
+    //   login: react.login,
+    //   settings: react.settings,
+    //   base: react.base,
+    //   app: "react" as "react",
+    //   profile: "spa",
+    // },
```
jonas-jonas, line 16: To get these to pass, we need a release of the SDK with the changes from this PR. I'd rather do that after merging and releasing this change.

selfservice/strategy/code/strategy_login.go
jonas-jonas: Reverted.

internal/testhelpers/selfservice_login.go
```diff
@@ -118,6 +122,12 @@ func InitFlowWithOAuth2LoginChallenge(hlc string) InitFlowWithOption {
 	}
 }
 
+func InitFlowWithVia(via string) InitFlowWithOption {
```
hperl, line 125: Some docstring would be nice here to explain "via".

selfservice/flow/login/extension_identifier_label.go
```diff
@@ -6,32 +6,57 @@ package login
...
 
 type identifierLabelExtension struct {
+	field                     string
 	identifierLabelCandidates []string
 }
 
-var _ schema.CompileExtension = new(identifierLabelExtension)
+var (
+	_               schema.CompileExtension = new(identifierLabelExtension)
+	ErrUnknownTrait                         = herodot.ErrBadRequest.WithReasonf("Trait does not exist in identity schema")
```
hperl, line 26: Is that something the requester (user) can fix, or is this more a server misconfiguration (500)?

selfservice/strategy/code/strategy.go
```diff
@@ -383,3 +408,11 @@ const CodeLength = 6
 func GenerateCode() string {
 	return randx.MustString(CodeLength, randx.Numeric)
 }
+
+func maskAddress(input string) string {
+	if strings.Contains(input, "@") {
+		parts := strings.Split(input, "@")
```
hperl, line 414: Use strings.Cut here to split into exactly two parts.

selfservice/strategy/code/strategy.go
```diff
@@ -383,3 +408,11 @@ const CodeLength = 6
...
+		return parts[0][:2] + strings.Repeat("*", 4) + "@" + parts[1]
+	}
+	return input[:3] + strings.Repeat("*", 4) + input[len(input)-3:]
```
hperl, line 417: Missing checks for minimum length?

hperl (dismissed):
LGTM with small comments.

selfservice/strategy/code/strategy.go
jonas-jonas: Yup, good point. I added some tests to cover all possible (?) variations.

selfservice/strategy/code/strategy.go
jonas-jonas: done.

zepatrik (dismissed):

zepatrik (approved):

driver/config/config.go
wewelll: Hello @aeneasr and @jonas-jonas, I'm currently building an app with Ory, using the code via email for login.
I'd like to add MFA with code via SMS, but I understand that it's currently not possible.

Do you think it's a legitimate use-case (code via email + code via sms), and are you planning to support it soon ? 
Also, if you plan to keep this limitation, what is the reason behind ? is it a security concern ?

PR #3701, opened 2024-01-18
feat: jsonnet caching for OIDC claims mapper, webhooks, JWT session tokenizer

alnr:
Requires https://github.com/ory/x/pull/753


session/tokenizer.go
```diff
@@ -32,35 +31,31 @@ type (
...
 func (s *Tokenizer) SetNowFunc(t func() time.Time) {
 	s.nowFunc = t
 }
 
-var cache, _ = ristretto.NewCache(&ristretto.Config{
-	NumCounters:        100000000,
-	MaxCost:            10000000,
-	BufferItems:        64,
-	IgnoreInternalCost: true,
-	Cost: func(value interface{}) int64 {
```
aeneasr, line 59: Same here - please use a cost function and follow best practices for setting counters / cost functions

selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -10,39 +10,36 @@ import (
...
 	"github.com/ory/kratos/x"
+	"github.com/ory/x/decoderx"
+	"github.com/ory/x/fetcher"
+	"github.com/ory/x/otelx"
+	"github.com/ory/x/sqlxx"
 )
 
 var _ registration.Strategy = new(Strategy)
 
+var jsonnetCache, _ = ristretto.NewCache(&ristretto.Config{
```
aeneasr, line 37: Use best practices for cache cost etc

aeneasr (commented):
LGTM, please just adjust the cache configs. Maybe it also makes sense to use a global cache instead of several individual caches at some point, to improve memory management.

selfservice/hook/web_hook.go
```diff
@@ -60,6 +61,12 @@ var _ interface {
 	settings.PostHookPostPersistExecutor
 } = (*WebHook)(nil)
 
+var jsonnetCache, _ = ristretto.NewCache(&ristretto.Config{
+	MaxCost:     100 << 20, // 100MB,
+	NumCounters: 1_000_000, // 1kB per snippet -> 100k snippets -> 1M counters
+	BufferItems: 64,
```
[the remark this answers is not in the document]
alnr: I specifically read through the docs when deciding on these parameters. Is something off here?

The cost is passed correctly to `SetWithTTL`. I suppose a byte slice s is actually `cap(s) + 3*sizeof uintptr`, is that what you mean?

PR #3705, opened 2024-01-19
revert: "fix: do not generate CSRF token for api flows"

jonas-jonas:
Reverts ory/kratos#3704

The CSRF token is now generated too late, and submitting a flow after an error has occurred can fail with a CSRF violation now.


jonas-jonas:
replaced by github.com/ory/kratos/pull/3706

PR #3714, opened 2024-01-25
docs: improve enum handling and completeness

zepatrik:
Some enum definitions were outdated and had wrong values. I just fixed them by using `// swagger:enum` instead of relying on commenting all cases.


jonas-jonas (approved):
Should this be considered a breaking SDK change, since were removing stuff?

aeneasr (approved):

aeneasr:
> Should this be considered a breaking SDK change, since were removing stuff?

Yeah!

PR #3722, opened 2024-01-30
fix: get additional user claims from Microsoft graph API

alnr:
#3609 Introduced a bug where the subject could not be established from the GraphAPI response. This change fixes that, and also keeps the old way of querying the API as a fallback if anything goes wrong.


alnr:
stale

PR #3727, opened 2024-02-01
fix: sdk generation

alnr:


.schema/openapi/patches/schema.yaml
```diff
@@ -45,7 +45,7 @@
       show_recovery_ui: "#/components/schemas/continueWithRecoveryUi"
 
 - op: add
-  path: /components/schemas/continueWith/oneOf
+  path: /components/schemas/continueWith/anyOf
```
aeneasr, line 48: Correct would be oneOf though no?

.schema/openapi/templates/go/model_anyof.mustache
```diff
@@ -20,7 +20,7 @@ func (dst *{{classname}}) UnmarshalJSON(data []byte) error {
 	{{#-first}}
 	// use discriminator value to speed up the lookup
 	var jsonDict map[string]interface{}
-	err := json.Unmarshal(data, &jsonDict)
+	err = json.Unmarshal(data, &jsonDict)
```
aeneasr, line 23: I actually don't think we need these templates any more. Instead, we should just use the templates from openapi-generator. These templates are very old and probably buggy

.schema/openapi/patches/schema.yaml
aeneasr: You're right!

aeneasr (approved):

aeneasr (dismissed):

.schema/openapi/templates/go/model_anyof.mustache
alnr: done

spec/api.json
```diff
@@ -491,20 +491,15 @@
         "description": "Indicates, that the UI flow could be continued by showing a recovery ui",
         "properties": {
           "action": {
-            "description": "Action will always be `show_recovery_ui`\nshow_recovery_ui ContinueWithActionShowRecoveryUIString",
-            "enum": [
-              "show_recovery_ui"
```
aeneasr, line 496: The enums are gone!

alnr: Correct. However, the generated code works in Go and Typescript:

```typescript
export type ContinueWith =
  | ({ action: "set_ory_session_token" } & ContinueWithSetOrySessionToken)
  | ({ action: "show_recovery_ui" } & ContinueWithRecoveryUi)
  | ({ action: "show_settings_ui" } & ContinueWithSettingsUi)
  | ({ action: "show_verification_ui" } & ContinueWithVerificationUi)

export function ContinueWithFromJSON(json: any): ContinueWith {
  return ContinueWithFromJSONTyped(json, false)
}

export function ContinueWithFromJSONTyped(
  json: any,
  ignoreDiscriminator: boolean,
): ContinueWith {
  if (json === undefined || json === null) {
    return json
  }
  switch (json["action"]) {
    case "set_ory_session_token":
      return {
        ...ContinueWithSetOrySessionTokenFromJSONTyped(json, true),
        action: "set_ory_session_token",
      }
    case "show_recovery_ui":
      return {
        ...ContinueWithRecoveryUiFromJSONTyped(json, true),
        action: "show_recovery_ui",
      }
    case "show_settings_ui":
      return {
        ...ContinueWithSettingsUiFromJSONTyped(json, true),
        action: "show_settings_ui",
      }
    case "show_verification_ui":
      return {
        ...ContinueWithVerificationUiFromJSONTyped(json, true),
        action: "show_verification_ui",
      }
    default:
      throw new Error(
        `No variant of ContinueWith exists with 'action=${json["action"]}'`,
      )
  }
}
```

Usage in Go:
```go
	flow, res, err := client.FrontendApi.CreateNativeRecoveryFlow(ctx).Execute()
	pkg.SDKExitOnError(err, res)

	for _, continueWith := range flow.ContinueWith {
		switch action := continueWith.GetActualInstance().(type) {
		case ory.ContinueWithSetOrySessionToken:
			fmt.Println("ContinueWithSetOrySessionToken")
			sessionToken := action.GetOrySessionToken()
			_ = sessionToken
		case ory.ContinueWithSettingsUi:
			fmt.Println("ContinueWithSettingsUi")
			pkg.PrintJSONPretty(action.GetFlow())

		default:
			fmt.Println("default")
			pkg.PrintJSONPretty(action)
		}
	}
```

I coulnd't figure out a "better" way to make this work otherwise. This is also how the `uiNodeAttributes` have worked since forever.

alnr: actually I might have to twiddle with this some more... fuck

alnr: OK I think I got it. Golang now also correctly uses the `action` field to decide among the elements of the type union.

aeneasr: OK - this is a breaking change then right? If so, we'll need to document it in some sense so that we can publish an upgrade advisory for Ory Network

.schema/openapi/gen.go.yml
```diff
@@ -4,3 +4,4 @@ generateInterfaces: true
 isGoSubmodule: false
 structPrefix: true
 enumClassPrefix: true
+useOneOfDiscriminatorLookup: true
```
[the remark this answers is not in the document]
bweston92: I am waiting on this, would you accept a PR for this against `ory/sdk`?

spec/api.json
aeneasr: @alnr what exactly is the point of removing these enums? I could not really figure it out.

alnr: I was under the impression that the generated code in Go would be invalid without changing the enums, but that may be incorrect.

aeneasr:
merged in another pr

WoodyWoodsta:
Hi @aeneasr - Which PR was this merged in? We're tracking this fix.

PR #3739, opened 2024-02-05
fix: add login succeeded event to post registration hook

sashatalalasha:
This PR adds NewLoginSucceeded event to post registration hook 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/flow/registration/hook.go
+++ b/selfservice/flow/registration/hook.go
@@ -219,20 +219,6 @@ func (e *HookExecutor) PostRegistrationHook(w http.ResponseWriter, r *http.Reque
 		return err
 	}
 
-	e.d.Audit().
-		WithRequest(r).
-		WithField("session_id", s.ID).
-		WithField("identity_id", i.ID).
-		Info("Identity authenticated successfully and was issued an Ory Kratos Session Token.")
-
-	span.AddEvent(events.NewLoginSucceeded(r.Context(), &events.LoginSucceededOpts{
-		SessionID:   s.ID,
-		IdentityID:  i.ID,
-		FlowType:    string(registrationFlow.Type),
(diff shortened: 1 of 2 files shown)


selfservice/flow/registration/hook.go
```diff
@@ -219,6 +219,20 @@ func (e *HookExecutor) PostRegistrationHook(w http.ResponseWriter, r *http.Reque
 		return err
 	}
 
+	e.d.Audit().
+		WithRequest(r).
+		WithField("session_id", s.ID).
+		WithField("identity_id", i.ID).
+		Info("Identity authenticated successfully and was issued an Ory Kratos Session Token.")
+
+	span.AddEvent(events.NewLoginSucceeded(r.Context(), &events.LoginSucceededOpts{
```
aeneasr, line 228: The code that was here before your PR is a bit misleading. While it is true that we create a session, we do not actually issue it to the user. This happens here: https://github.com/ory/kratos/blob/master/selfservice/hook/session_issuer.go#L77

I think the event should move there, because otherwise every registration event is automatically a log in event (which is not true).

Thanks!

selfservice/hook/session_issuer.go
```diff
@@ -78,6 +79,13 @@ func (e *SessionIssuer) executePostRegistrationPostPersistHook(w http.ResponseWr
 		return err
 	}
 
+	trace.SpanFromContext(r.Context()).AddEvent(events.NewLoginSucceeded(r.Context(), &events.LoginSucceededOpts{
```
aeneasr, line 82: This now only triggers for the browser flow, for the API flow the event is missing

aeneasr (approved):

PR #3746, opened 2024-02-07
feat: improved webhook tracing

alnr:
https://github.com/ory/x/pull/761


driver/registry_default.go
```diff
@@ -815,16 +816,13 @@ func (m *RegistryDefault) PrometheusManager() *prometheus.MetricsManager {
...
 		httpx.ResilientClientWithLogger(m.Logger()),
 		httpx.ResilientClientWithMaxRetry(2),
-		httpx.ResilientClientWithConnectionTimeout(30*time.Second))
-
-	tracer := m.Tracer(ctx)
-	if tracer.IsLoaded() {
-		opts = append(opts, httpx.ResilientClientWithTracer(tracer.Tracer()))
-	}
+		httpx.ResilientClientWithConnectionTimeout(30*time.Second),
+		httpx.ResilientClientWithTracer(noop.NewTracerProvider().Tracer("Ory Kratos")), // will use the tracer from a context if available
```
zepatrik, line 824: How can this use the tracer form the context if the context is not even passed to it?

selfservice/hook/web_hook.go
```diff
@@ -525,12 +525,16 @@ func instrumentHTTPClientForEvents(ctx context.Context, httpClient *retryablehtt
...
 	}
 	httpClient.ResponseLogHook = func(_ retryablehttp.Logger, res *http.Response) {
-		res.Body = io.NopCloser(io.LimitReader(res.Body, 5<<20)) // read at most 5 MB from the response
-		resBody, _ := httputil.DumpResponse(res, true)
-		resBody = resBody[:min(len(resBody), 2<<10)] // truncate response body to 2 kB for event
+		// res.Body = io.NopCloser(io.LimitReader(res.Body, 5<<20)) // read at most 5 MB from the response
+		// resBody, _ := httputil.DumpResponse(res, true)
+		// resBody = resBody[:min(len(resBody), 2<<10)] // truncate response body to 2 kB for event
+		// TODO(@alnr): redact sensitive data
+		resBody := []byte("<redacted>")
```
zepatrik, line 537: Todo?

driver/registry_default.go
alnr: The tracer is retrieved from the request context:
```
req, err := http.NewRequest(ctx /* HERE */, GET, "http://localhost", nil)
```

selfservice/hook/web_hook.go
alnr: I haven't figured out how to do this cleanly. That's for another PR.

driver/registry_default.go
zepatrik: Ah so here we instantiate the client, and the tracer does not matter. When the request is made, a context with the correct tracer is passed.
Correct?

alnr: Precisely. The `otelhttp` instrumentation checks for an existing span in the context. If it finds one, it uses the same tracer to start its own spans. If it doesn't find one, it uses the one passed here. Finally, it falls back to the global tracer (which we don't typicall configure).

zepatrik (approved):

PR #3753, opened 2024-02-13
fix: don't require code credential for MFA flows

jonas-jonas:
Removes the requirement to set the phone number as a code credential in the identity schema for MFA via SMS, as we already have the identity from the session in MFA flows and can use it from there.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/code/strategy_login.go
+++ b/selfservice/strategy/code/strategy_login.go
@@ -293,10 +293,21 @@ func (s *Strategy) loginVerifyCode(ctx context.Context, r *http.Request, f *logi
 
 	p.Identifier = maybeNormalizeEmail(p.Identifier)
 
-	// Step 1: Get the identity
-	i, isFallback, err := s.findIdentityByIdentifier(ctx, p.Identifier)
-	if err != nil {
-		return nil, err
+	isFallback := false
+	var i *identity.Identity
+	if f.RequestedAAL > identity.AuthenticatorAssuranceLevel1 {
+		// Don't require the code credential if the user already has a session (e.g. this is an MFA flow)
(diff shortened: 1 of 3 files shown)


selfservice/strategy/code/strategy_login.go
```diff
@@ -293,10 +293,21 @@ func (s *Strategy) loginVerifyCode(ctx context.Context, r *http.Request, f *logi
 
 	p.Identifier = maybeNormalizeEmail(p.Identifier)
 
-	// Step 1: Get the identity
-	i, isFallback, err := s.findIdentityByIdentifier(ctx, p.Identifier)
-	if err != nil {
-		return nil, err
+	isFallback := false
+	var i *identity.Identity
+	if f.RequestedAAL > identity.AuthenticatorAssuranceLevel1 {
```
zepatrik, line 298: What happens when the requested AAL is 2, but the user has no session yet? Then we would run into an error here, right?
Maybe first try to look up the session regardless of the AAL, and fall back to the lookup if it is not found?

jonas-jonas: If we do that, the flow succeeds, but the resulting session is not aal2, because there is no other credential present. I think erroring here is appropriate.

jonas-jonas: Also we already fail when populating the flow, if there is no session available https://github.com/ory/kratos/blob/ac4cd80d89d8b8b27ae4a2e0b8e9c6b659e2377c/selfservice/strategy/code/strategy.go#L206-L210

zepatrik (approved):

aeneasr (approved):

Issue #517, opened 2020-06-16
Social sign in with Twitter


romcok:
**Is your feature request related to a problem? Please describe.**

Hi,
we want to use the sign in with Twitter, but I don't know how to configure it.

This is written in the documentation:
`The oidc method uses OpenID Connect, or OAuth2 where OpenID Connect is not supported, to authenticate identities using a third-party identity provider, such as Google, Microsoft, GitHub - or any other OAuth2 / OpenID Connect provider..`

Twitter is the OAuth2 provider, so I hope it will work somehow 🙏

aeneasr:
Twitter is not OAuth2 but the older OAuth1, so we will need a dedicated strategy to deal with Twitter! By searching on Google it appears that there's at least one library able to deal with Twitter's OAuth (1.0) protocol: https://github.com/kkdai/twitter

romcok:
Oh, now I see. They have OAuth2 only for application-only authentication. 
https://developer.twitter.com/en/docs/basics/authentication/oauth-2-0

Twitter is quite a big player and a lot of people come to us from there, it would be good to have it as an option.

marcinkoziej:
Hello @aeneasr !
What is the hour estimate to add this functionality to Kratos?
Could we sponsor such work to be done by You or someone form Ory team?

aeneasr:
Someone from out team will reach out to you in the next days!

lanphan:
https://github.com/dghubble/oauth1 is a better lib to use, imho.

mfzl:
> Twitter is not OAuth2 but the older OAuth1, so we will need a dedicated strategy to deal with Twitter! By searching on Google it appears that there's at least one library able to deal with Twitter's OAuth (1.0) protocol: https://github.com/kkdai/twitter

Are you suggesting another strategy along with `oidc` and `password`. What would it be called for example `oauth1` or `twitter`?

aeneasr:
No, this can be dealt with in the normal oidc strategy. You just need to use a different library to complete it, and not our generic OIDC provider!

mfzl:
@aeneasr Is this being developed? If not I'd like to take this up

aeneasr:
Noone is working on it afaik :) Please do pick it up! :)

vinckr:
And let me know if I can support you in any way with this :-).

8thblock:
[...]

PR #3778, opened 2024-02-23
feat: add twitter SSO

aeneasr:
Closes https://github.com/ory/kratos/issues/517 https://github.com/ory/kratos/pull/2117

https://github.com/ory/docs/pull/1658


go.mod
```diff
@@ -327,6 +327,7 @@ require (
 
 require (
 	github.com/coreos/go-oidc/v3 v3.9.0
+	github.com/dghubble/oauth1 v0.7.2
```
zepatrik, line 330: > To implement "Login with X", use the [gologin](https://github.com/dghubble/gologin) packages which provide login handlers for OAuth1 and OAuth2 providers.
>
> To call the Twitter, Digits, or Tumblr OAuth1 APIs, use the higher level Go API clients.

Did you check that note in the repo's readme?

selfservice/strategy/oidc/provider_x.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return oauth1.NewToken(accessToken, accessSecret), nil
+}
+
+func (p *ProviderX) AuthURL(ctx context.Context, state string) (string, error) {
+	c := p.OAuth1(ctx)
+
+	// We need to cheat so that callback validates on return
+	c.CallbackURL = c.CallbackURL + fmt.Sprintf("?state=%s&code=unused", state)
```
zepatrik, line 61: What exactly is the cheat referred to here? Do we need to set a code, but it gets ignored by twitter? Will it still be in the URL when being redirected back?

zepatrik (approved):
Looks good, just some minor questions.

go.mod
zepatrik: The usage later looks valid IMO.

aeneasr: I think this should be fine!

selfservice/strategy/oidc/provider_x.go
aeneasr: Good question - I took over some of the code from PR #517 where this originally was at. I believe that OAuth1 doesn't return a code parameter like OAuth2, and that would break the provider logic. So the added this workaround? I'm not sure to be honest.

PR #2428, opened 2022-04-27
feat: update token on login

david972:
Currently Kratos record the token during the registration and the setting.

If the access_token recorded has expired in the meanwhile or tokens has been revoked it would be useful to update them.
I know in case of expiration we can use the refresh token but not in case of revocation.

## Related issue(s)

#1912 

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [x] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](docs/docs).

PR #3788, opened 2024-02-28
feat: update identity data based on id token claims on each login

splaunov:
When new identity is registered with OIDC, any attribute from ID Token can be saved into identity data fields like `traits`, `metadata_public` and `metadata_admin`.
But there is no way to change that data on user login. One common use case is to capture `groups` into `metadata_admin` on registration and then update list of groups on each next login.

This PR adds logic to update identity data based on ID Token content.

## Checklist

      introduces a new feature.
- [x ] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/oidc/provider_config.go
+++ b/selfservice/strategy/oidc/provider_config.go
@@ -99,6 +99,12 @@ type Configuration struct {
 	// It can be either a URL (file://, http(s)://, base64://) or an inline JSONNet code snippet.
 	Mapper string `json:"mapper_url"`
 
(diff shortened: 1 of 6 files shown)


luflow:
@splaunov Isn't this the same idea like https://github.com/ory/kratos/pull/2428?

splaunov:
I would not say it's the same.
#2428 updates the token but not traits and metadata.

kghost:
Are we still working on this one ?

fmiqbal:
This will be beneficial for ory that is used only for the security feature, user management still fully managed on the backing IdP, and ory act as session authority/cookie authority. we are currently transitioning to Ory from our old oauth2 based implementation, and missing this feature means we have to use webhook that will access admin api to update the metadata

PR #3800, opened 2024-03-06
fix: missing indices and foreign keys

alnr:


persistence/sql/identity/persister_identity.go
```diff
@@ -960,8 +960,12 @@ func (p *IdentityPersister) DeleteIdentity(ctx context.Context, id uuid.UUID) (e
 			attribute.Stringer("network.id", p.NetworkID(ctx))))
 	defer otelx.End(span, &err)
 
+	tableName := new(identity.Identity).TableName(ctx)
+	if p.c.Dialect.Name() == "cockroach" {
+		tableName += "@primary"
```
zepatrik, line 965: This is forcing the primary key index, right?

alnr: exactly

persistence/sql/migrations/sql/20230705000000000002_cookie_flow_request_url.cockroach.down.sql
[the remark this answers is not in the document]
alnr: These migrations have no corresponding `up` migration and are, from what I understand, vestigial. 

We didn't notice before because of a bug fixes here https://github.com/ory/x/pull/762

aeneasr (approved):

zepatrik (approved):

PR #3827, opened 2024-03-18
fix: drop trigram index on identifiers

zepatrik:
The trigram GIN index takes up a significant amount of database storage. Fuzzy search should instead be build using a dedicated external service. This will allow indexing other traits as well, based on needs.

### DIFF
--- a/persistence/sql/identity/persister_identity.go
+++ b/persistence/sql/identity/persister_identity.go
@@ -824,7 +824,7 @@ func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.
 		identifier := params.CredentialsIdentifier
 		identifierOperator := "="
 		if identifier == "" && params.CredentialsIdentifierSimilar != "" {
-			identifier = strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(params.CredentialsIdentifierSimilar, "\\", "\\\\"), "%", "\\%"), "_", "\\_") + "%"
+			identifier = x.EscapeLikePattern(params.CredentialsIdentifierSimilar) + "%"
 			identifierOperator = "LIKE"
 		}
 
--- a/persistence/sql/identity/persister_identity.go
+++ b/persistence/sql/identity/persister_identity.go
@@ -265,7 +265,7 @@ INNER JOIN identity_credentials
     AND identity_credentials.identity_credential_type_id = (
         SELECT id
         FROM identity_credential_types
-        WHERE name = ? 
+        WHERE name = ?
      )
 WHERE identity_credentials.config ->> '%s' = ?
   AND identities.nid = ?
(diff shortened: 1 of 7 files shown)


persistence/sql/migrations/sql/20240318143139000000_drop_identity_search_index.postgres.down.sql
```diff
@@ -0,0 +1,4 @@
+CREATE EXTENSION IF NOT EXISTS pg_trgm;
+CREATE EXTENSION IF NOT EXISTS btree_gin;
+
+CREATE INDEX identity_credential_identifiers_nid_identifier_gin ON identity_credential_identifiers USING GIN (nid, identifier gin_trgm_ops);
```
aeneasr, line 4: Isn't cock,roach missing here?

persistence/sql/identity/persister_identity.go
```diff
@@ -824,14 +824,8 @@ func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.
...
 		if identifier == "" && params.CredentialsIdentifierSimilar != "" {
-			identifier = params.CredentialsIdentifierSimilar
-			identifierOperator = "%"
-			switch con.Dialect.Name() {
-			case "postgres", "cockroach":
-			default:
-				identifier = "%" + identifier + "%"
-				identifierOperator = "LIKE"
-			}
+			identifier = strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(params.CredentialsIdentifierSimilar, "\\", "\\\\"), "%", "\\%"), "_", "\\_") + "%"
```
aeneasr, line 827: What is this for? Wouldn't this be enough?

```
identifier = identifier + "%"
```

aeneasr (changes requested):

persistence/sql/identity/persister_identity.go
zepatrik: Well, if you put `%foo` as the parameter, you will be able to search for `%foo%`, which is not covered by an index. So we have to escape all pattern characters, and the escape character.

persistence/sql/migrations/sql/20240318143139000000_drop_identity_search_index.postgres.down.sql
zepatrik: This is just a copy of the previous migration where the index was created. It works as it should on cockroach as well.

aeneasr: But the migration specifically only applies to postgres (see file name)

aeneasr: Oh I see - maybe we applied it manually on prod?

persistence/sql/identity/persister_identity.go
aeneasr: Can you please add one test to ensure this escaping works correctly?

aeneasr (approved):

PR #3834, opened 2024-03-20
fix: recovery with OAuth2 and 2FA

hperl:
## Related issue(s)

https://github.com/ory-corp/cloud/issues/6066

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/session/manager_http.go
+++ b/session/manager_http.go
@@ -341,7 +341,7 @@ func (s *ManagerHTTP) DoesSessionSatisfy(r *http.Request, sess *Session, request
 			return nil
 		}
 
-		loginURL := urlx.CopyWithQuery(urlx.AppendPaths(s.r.Config().SelfPublicURL(ctx), "/self-service/login/browser"), url.Values{"aal": {"aal2"}})
+		loginURL := &url.URL{Path: "/self-service/login/browser", RawQuery: "aal=aal2"}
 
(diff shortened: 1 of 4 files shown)


session/manager_http.go
```diff
@@ -341,7 +341,7 @@ func (s *ManagerHTTP) DoesSessionSatisfy(r *http.Request, sess *Session, request
 			return nil
 		}
 
-		loginURL := urlx.CopyWithQuery(urlx.AppendPaths(s.r.Config().SelfPublicURL(ctx), "/self-service/login/browser"), url.Values{"aal": {"aal2"}})
+		loginURL := &url.URL{Path: "/self-service/login/browser", RawQuery: "aal=aal2"}
```
aeneasr, line 344: I don't see the difference to be honest?

hperl: The bug I want to fix is that if Kratos runs on multiple domains / CNAMEs, redirecting to the public URL will cause the user to end up on a different domain than he started on.

If we redirect without a host, the current host will continue to be used.

hperl: >If the idea is to not use the domain, then this won't work because the public url might have a path prefix

OK, so then we can just extract the path prefix from the public URL?

hperl:
This is not the fix I was looking for.

PR #3862, opened 2024-04-04
chore: don't require method to be  passkey in settings schema

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/passkey/.schema/settings.schema.json
+++ b/selfservice/strategy/passkey/.schema/settings.schema.json
@@ -7,7 +7,7 @@
       "type": "string"
     },
     "method": {
-      "const": "passkey"
+      "type": "string"
     },
     "passkey_settings_register": {
       "type": "string"
--- a/selfservice/strategy/passkey/passkey_settings.go
+++ b/selfservice/strategy/passkey/passkey_settings.go
@@ -125,7 +125,8 @@ func (s *Strategy) PopulateSettingsMethod(r *http.Request, id *identity.Identity
 		Attributes: &node.InputAttributes{
 			Name: node.PasskeySettingsRegister,
 			Type: node.InputAttributeTypeHidden,
-		}})
+		},
+	})
 
 	f.UI.Nodes.Upsert(&node.Node{
 		Type:  node.Input,
@@ -135,7 +136,8 @@ func (s *Strategy) PopulateSettingsMethod(r *http.Request, id *identity.Identity
 			Name:       node.PasskeyCreateData,
 			Type:       node.InputAttributeTypeHidden,
 			FieldValue: string(injectWebAuthnOptions),
-		}})
+		},
+	})
 
 	return nil
 }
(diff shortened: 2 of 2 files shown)


selfservice/strategy/passkey/.schema/settings.schema.json
```diff
@@ -7,7 +7,7 @@
       "type": "string"
     },
     "method": {
-      "const": "passkey"
+      "type": "string"
```
aeneasr, line 10: I think this was correct, could you please explain why this would need to change?

aeneasr: I'm quite confident that's not true, or a majority of our e2e tests would fail. I checked the slack thread but could not really find what the bug exactly is other than getting a success status without updating. If it would indeed fail here, it would fail with a validation error and not with a success state.

aeneasr: OK I see in the thread now - we do apparently return a 400 error but with a success state? Not sure how that can happen

selfservice/strategy/passkey/passkey_settings.go
```diff
@@ -156,7 +158,7 @@ func (s *Strategy) identityListWebAuthn(id *identity.Identity) (*identity.Creden
 
 func (s *Strategy) Settings(w http.ResponseWriter, r *http.Request, f *settings.Flow, ss *session.Session) (*settings.UpdateContext, error) {
 	if f.Type != flow.TypeBrowser {
-		return nil, flow.ErrStrategyNotResponsible
+		return nil, errors.WithStack(flow.ErrStrategyNotResponsible)
```
aeneasr, line 161: Do we need this? I'm not sure if we correctly assert with `errors.Is` everywhere

aeneasr: I checked, looks like we do.

aeneasr (approved):

selfservice/strategy/passkey/.schema/settings.schema.json
aeneasr: > We "swallow" the error as part of the settings flow somewhere. Hence, the 400 with a success state of the flow?

That's quite scary and we need to investigate it.

So what is the problem exactly and why does it not happen for our e2e pipeline which includes react and nodejs?

jonas-jonas: AFAICT, the lookup secrets implementation in the Ory Elements doesn't submit the `method` (which is incorrect, and we should fix this!). And I guess this is the main problem. 

Ideally, Kratos would require the method parameter and only execute the strategy that is sent via `method` (and only try all if no method is sent), but that would be quite a substantial change.

hperl (approved):
Thank you! 🙏

aeneasr:
Merging because the test failure was a timing issue

PR #3867, opened 2024-04-11
fix: claims.update_at is float64

kghost:
## Related issue(s)

https://community.ory.sh/t/failed-to-unmarshal-updated-at-for-generic-oidc-provider-auth0/2004

When using one ory+hydra as OIDC provider to authenticated another kratos, I got an error that it is unable to unmarshal Claims.updated_at, because kratos-selfservice-ui-node returns the session and convert updated_at to a float64, because javascript/typescript doesn't have int64 type.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

The proposed fix makes it competible with multiple types of updated_at field.

### DIFF
--- a/selfservice/strategy/oidc/provider.go
+++ b/selfservice/strategy/oidc/provider.go
@@ -6,9 +6,13 @@ package oidc
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"math"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/dghubble/oauth1"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


alnr:
Shouldn't this rather be fixed in ory/kratos-selfservice-ui-node?

kghost:
> Shouldn't this rather be fixed in ory/kratos-selfservice-ui-node?

Sure, this problem can be fix there, by this PR makes our Kratos more robust without any downside, so can we consider it an improvement ?

alnr:
> > Shouldn't this rather be fixed in ory/kratos-selfservice-ui-node?
> 
> Sure, this problem can be fix there, by this PR makes our Kratos more robust without any downside, so can we consider it an improvement ?

The reason I'm hesitant is that Kratos would accept payloads which are arguably malformed. A unix timestamp is always an integer. If we start accepting floats too, we have to decide on how to handle rounding etc. I'd rather not do that and be more strict. Especially since we could never go back to only accepting integers because that would break existing usage.

alnr:
This issue is fixed in newer versions of ory/kratos-selfservice-ui-node and ory/hydra. Please create an issue if this still occurs. Thanks 👍

PR #3868, opened 2024-04-11
feat: log external latency

zepatrik:
With this change, we measure and log some external latency separately.

Right now, we consider this external:
- OAuth2 provider APIs
- webhooks

Other calls, like loading jsonnet or courier templates, are not considered external because they are tightly coupled to the server Kratos is running on.


selfservice/hook/web_hook.go
```diff
@@ -234,192 +218,186 @@ func (e *WebHook) ExecutePostRegistrationPostPersistHook(_ http.ResponseWriter,
...
-			// configured timeout for the HTTP client.
-			ctx = context.WithoutCancel(ctx)
-		}
-		ctx, span := tracer.Start(ctx, "selfservice.webhook")
-		defer otelx.End(span, &finalErr)
-
-		if emitEvent {
-			instrumentHTTPClientForEvents(ctx, httpClient)
+func (e *WebHook) execute(ctx context.Context, data *templateContext, hookName string) error {
+	return otelx.WithSpan(ctx, "selfservice.hook.WebHook.Execute"+hookName, func(ctx context.Context) error {
```
alnr, line 270: Sorry for bikeshedding, but please don't do this. Right now, the `selfservice.hook.WebHook.Execute SettingsPrePersistHook` literals are quickly greppable, which this change destroys.

alnr (changes requested):
Looks good, just one comment.

PR #3880, opened 2024-04-17
fix: allow updating just the verified_at timestamp of addresses

alnr:


identity/handler_test.go
```diff
@@ -928,6 +928,73 @@ func TestHandler(t *testing.T) {
...
+				cr.SchemaID = "employee"
+				cr.Traits = []byte(`{"email":"` + email + `"}`)
+				res := send(t, ts, "POST", "/identities", http.StatusCreated, &cr)
+				assert.EqualValues(t, email, res.Get("recovery_addresses.0.value").String(), "%s", res.Raw)
+				assert.EqualValues(t, email, res.Get("verifiable_addresses.0.value").String(), "%s", res.Raw)
+				assert.Falsef(t, res.Get("verifiable_addresses.0.verified").Bool(), "%s", res.Raw)
+				assert.Falsef(t, res.Get("verifiable_addresses.0.verified_at").Exists(), "%s", res.Raw)
+				identityID := res.Get("id").String()
+
+				// set to verified, should also update verified_at timestamp
```
jonas-jonas, line 945: I don't think it should, but I guess this is already the current behavior?

alnr: If you don't provide a timestamp in your patch, we will set it to the current time. IMO that's the only useful semantic.

alnr:
Added more tests

jonas-jonas (approved):

aeneasr (approved):

PR #3893, opened 2024-04-23
fix: respect return_to in OIDC API flow error case

hperl:
This fix ensures that we redirect the user to the return_to URL when an error occurs during the OIDC login for native flows.

Native flows are initialized through the API, and the browser URL is retrieved from a 422 response after a POST to submit the login flow. Successful OIDC flows already returned the `code` to the `return_to` URL. Now, unsuccessful flows return the `flow` with the current flow ID (which might have changed), so that the caller can retrieve the full flow and act accordingly.

## Related issue(s)

https://github.com/ory/network/issues/344

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


selfservice/strategy/oidc/strategy.go
```diff
@@ -370,7 +370,7 @@ func (s *Strategy) alreadyAuthenticated(w http.ResponseWriter, r *http.Request,
 			returnTo := s.d.Config().SelfServiceBrowserDefaultReturnTo(ctx)
 			if redirecter, ok := f.(flow.FlowWithRedirect); ok {
 				r, err := x.SecureRedirectTo(r, returnTo, redirecter.SecureRedirectToOpts(ctx, s.d)...)
-				if err != nil {
+				if err == nil {
 					returnTo = r
```
jonas-jonas, line 374: We stumbled upon this, when checking the code paths, and it seems that the current (master) behavior is wrong:

If `x.SecureRedirectTo` returns an err, we set `returnTo = nil` because `x.SecureRedirectTo` always returns `nil, err` or `something, nil`. [We then call `String()` on `returnTo`, which would be `nil.String()`](https://github.com/ory/kratos/pull/3893/files#diff-6651d6164306e51e2d1ea3d1474c1069c69619dc37c7b3a18842c1b250403b22R377), which obviously would panic. I don't think there is a code path that would trigger this, at the moment. The missing test coverage confirms that suspicion

jonas-jonas (approved):

jonas-jonas (approved):

PR #3896, opened 2024-04-25
fix: db index and duplicate credentials error

hperl:
This fixes two minor bugs

- uses an improved index for getting the credential identifier from the user handle
- checks if the password hash is set before suggesting it as a credential identifier on duplicate credential errors

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/identity/manager.go
+++ b/identity/manager.go
@@ -205,8 +205,19 @@ func (m *Manager) findExistingAuthMethod(ctx context.Context, e error, i *Identi
 			if len(cred.Identifiers) > 0 {
 				identifierHint = cred.Identifiers[0]
 			}
-			duplicateCredErr.AddCredentialsType(cred.Type)
 			duplicateCredErr.SetIdentifierHint(identifierHint)
+
+			var cfg CredentialsPassword
+			if err := json.Unmarshal(cred.Config, &cfg); err != nil {
+				// just ignore this credential if the config is invalid
+				continue
+			}
+			if cfg.HashedPassword == "" {
+				// just ignore this credential if the hashed password is empty
+				continue
+			}
+
+			duplicateCredErr.AddCredentialsType(cred.Type)
 		case CredentialsTypeOIDC:
 			var cfg CredentialsOIDC
 			if err := json.Unmarshal(cred.Config, &cfg); err != nil {
--- a/identity/manager.go
+++ b/identity/manager.go
@@ -250,6 +250,24 @@ func (m *Manager) findExistingAuthMethod(ctx context.Context, e error, i *Identi
 					break
 				}
 			}
+		case CredentialsTypePasskey:
+			var cfg CredentialsWebAuthnConfig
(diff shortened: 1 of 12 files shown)


identity/manager.go
```diff
@@ -205,8 +205,19 @@ func (m *Manager) findExistingAuthMethod(ctx context.Context, e error, i *Identi
 			if len(cred.Identifiers) > 0 {
 				identifierHint = cred.Identifiers[0]
 			}
-			duplicateCredErr.AddCredentialsType(cred.Type)
 			duplicateCredErr.SetIdentifierHint(identifierHint)
+
+			var cfg CredentialsPassword
+			if err := json.Unmarshal(cred.Config, &cfg); err != nil {
```
piotrmsc, line 211: @aeneasr do we set empty json object as credential config when oidc is used?

selfservice/strategy/passkey/passkey_registration_test.go
```diff
@@ -372,7 +372,7 @@ func TestRegistration(t *testing.T) {
 					assert.Contains(t, gjson.Get(actual, "ui.action").String(), fix.publicTS.URL+registration.RouteSubmitFlow, "%s", actual)
 					registrationhelpers.CheckFormContent(t, []byte(actual), "csrf_token", "traits.username")
 					assert.Equal(t,
-						"You tried signing in with "+email+" which is already in use by another account. You can sign in using your password.",
```
piotrmsc, line 375: if we hint with webuathn strategy shouldn't we also do it with passkey instead of removing it?

hperl: fixed :)

persistence/sql/migrations/sql/20240425095000000000_identity_credentials_fix_user_handle_index.cockroach.up.sql
```diff
@@ -0,0 +1,6 @@
+DROP INDEX identity_credentials_user_handle_idx;
```
aeneasr, line 1: I think it would be better to invert this, so first add the index and then remove it. Because then you never have a time where no index exists which leads to very slow passkey logins.

aeneasr (changes requested):
Just one minor thing regarding migration

persistence/sql/migrations/sql/20240425095000000000_identity_credentials_fix_user_handle_index.cockroach.up.sql
hperl: Sure, I can do this. Just FYI, we only need this index when logging in with the passkey method when the user only has a webauthn credential, so in the migration case. Usual passkey logins won't even need this index, because there the user handle is the credential identifier.

hperl: fixed

aeneasr (approved):

Issue #2898, opened 2022-11-16
Refresh identity traits on social / SSO login


aeneasr:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Currently, social sign in only executes the JsonNet during registration. However, multiple questions have come up in several issues how this can be solved for linking sign ins, or when the user sign ins again - to keep the data fresh.

Example of such a discussion: https://github.com/ory/network/issues/164#issuecomment-1285989251 (there are more!)

### Describe your ideal solution

The jsonnet should be executed on sign in and on linking social sign in as well. It will have to respect the existing user data as well (so it needs to be available in the context).

### Workarounds or alternatives

There are no workarounds for this.

### Version

master

netthier:
Already posted this in one of the other issues, but I'm wondering what would happen to the user data if the Jsonnet code and schema are changed between logins.
Would the user be effectively recreated with the most recent schema and Jsonnet, or would some old version be run?

aeneasr:
We would add the current identity data to the jsonnet context, and then it's your decision

netthier:
Has  there been any progress on this?

kmherrmann:
We've added the ability to update identities from webhooks, see here: https://www.ory.sh/docs/guides/integrate-with-ory-cloud-through-webhooks#update-identity-traits

Would that solve the problem for you?

netthier:
> We've added the ability to update identities from webhooks, see here: https://www.ory.sh/docs/guides/integrate-with-ory-cloud-through-webhooks#update-identity-traits
> 
> Would that solve the problem for you?

No, as I need the update to happen during sign-in and the docs state the following:
> Modifying the identity is currently only possible during the registration and settings flows.

[...]

Issue #3816, opened 2024-03-11
Metadata and trait not update on link with other google account via oidc


HoangViet144:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

When linking with other google account, the trait and metadata are not updated to with information from new google account.

My kratos supports both password and oidc.
I register an account using oidc with email A. 
Then I set a password for my account
After that, I unlink and link with other email B. But the trait and metadata are not updated with information from email B. In the setting hook, I dont have any info from new link account to update manually using hook

"identity.default.schema.json": |
      {
        "$id": "https://schemas.ory.sh/presets/kratos/identity.email.schema.json",
        "$schema": "http://json-schema.org/draft-07/schema#",
        "title": "Person",
        "type": "object",
        "properties": {
          "traits": {
            "type": "object",
            "properties": {
              "first_name": {
                "type": "string",
                "title": "First name",
                "maxLength": 256
              },
              "last_name": {
                "type": "string",
                "title": "Last name",
                "maxLength": 256
              },
              "email": {
                "type": "string",
                "format": "email",
                "title": "E-Mail",
                "maxLength": 256,
                "ory.sh/kratos": {
                  "credentials": {
                    "password": {
                      "identifier": true
                    },
                    "webauthn": {
                      "identifier": true
                    },
                    "totp": {
                      "account_name": true
                    },
                    "code": {
                      "identifier": true,
                      "via": "email"
                    }
                  },
                  "recovery": {
                    "via": "email"
                  },
                  "verification": {
                    "via": "email"
                  }
                }
              }
            },
            "required": [
              "first_name",
              "last_name",
              "email"
            ],
            "additionalProperties": false
          },
          "metadata_public": {
            "type": "object",
            "properties": {
              "email": {
                "type": "string"
              }
            }
          }
        }
      }
    "oidc.google.jsonnet": |
      local claims = {
[...]

PR #3922, opened 2024-05-14
feat: pass OIDC claims into post-login flow to include in web hook context

fenech:
The login flow doesn't trigger a refresh of the identity when the OIDC claims have changed. By passing the claims through to the web hook context, this means that an external handler can be configured to update the identity as appropriate, when there are changes.

## Related issue(s)

#2898

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

The PR looks a lot bigger due to the fact that the `Claims` type has been moved into its own separate package, to resolve a circular dependency. I suspect that it is not in the right place and should be moved elsewhere.

The linked issue refers to refreshing the identity directly as part of the login flow, to handle the case where the OIDC Claims have changed.

However, my follow-up proposal is different: if instead of directly updating the identity, we just pass the OIDC Claims through to the web hook context, we allow users to write their own web hook handler to update the identity as they would like. I think it's a much less invasive change to Kratos (just passing an extra object through the login/post-login flow).

## Summary by CodeRabbit

* **New Features**
  * Post-login hooks and post-login webhooks now receive OIDC claims, enabling claim-based logic.
  * Webhook payloads can include enriched claim fields (for example, nickname and group membership when available).

* **Refactor**
  * OIDC claims handling is unified across providers with stricter validation (issuer and subject required) and improved locale parsing (language/country object + fallback).

* **Bug Fixes**
  * Correctly threads claims through the post-login hook execution path and updates related hook handling to match the new contract.

### DIFF
--- a/selfservice/flow/login/hook.go
+++ b/selfservice/flow/login/hook.go
@@ -22,6 +22,7 @@ import (
 	"github.com/ory/kratos/schema"
 	"github.com/ory/kratos/selfservice/flow"
 	"github.com/ory/kratos/selfservice/sessiontokenexchange"
+	"github.com/ory/kratos/selfservice/strategy/oidc/claims"
 	"github.com/ory/kratos/session"
 	"github.com/ory/kratos/ui/container"
 	"github.com/ory/kratos/ui/node"
@@ -38,7 +39,7 @@ type (
 	}
 
 	PostHookExecutor interface {
-		ExecuteLoginPostHook(w http.ResponseWriter, r *http.Request, g node.UiNodeGroup, a *Flow, s *session.Session) error
+		ExecuteLoginPostHook(w http.ResponseWriter, r *http.Request, g node.UiNodeGroup, a *Flow, s *session.Session, c *claims.Claims) error
 	}
 
 	HooksProvider interface {
@@ -132,6 +133,7 @@ func (e *HookExecutor) PostLoginHook(
 	f *Flow,
 	i *identity.Identity,
 	s *session.Session,
+	c *claims.Claims,
 	provider string,
 ) (err error) {
 	ctx := r.Context()
@@ -151,15 +153,15 @@ func (e *HookExecutor) PostLoginHook(
 		return err
 	}
 
-	c := e.d.Config()
+	cfg := e.d.Config()
 	// Verify the redirect URL before we do any other processing.
 	returnTo, err := redir.SecureRedirectTo(r,
-		c.SelfServiceBrowserDefaultReturnTo(ctx),
+		cfg.SelfServiceBrowserDefaultReturnTo(ctx),
 		redir.SecureRedirectReturnTo(f.ReturnTo),
 		redir.SecureRedirectUseSourceURL(f.RequestURL),
-		redir.SecureRedirectAllowURLs(c.SelfServiceBrowserAllowedReturnToDomains(ctx)),
-		redir.SecureRedirectAllowSelfServiceURLs(c.SelfPublicURL(ctx)),
-		redir.SecureRedirectOverrideDefaultReturnTo(c.SelfServiceFlowLoginReturnTo(ctx, f.Active.String())),
+		redir.SecureRedirectAllowURLs(cfg.SelfServiceBrowserAllowedReturnToDomains(ctx)),
+		redir.SecureRedirectAllowSelfServiceURLs(cfg.SelfPublicURL(ctx)),
+		redir.SecureRedirectOverrideDefaultReturnTo(cfg.SelfServiceFlowLoginReturnTo(ctx, f.Active.String())),
 	)
 	if err != nil {
 		return err
@@ -187,7 +189,7 @@ func (e *HookExecutor) PostLoginHook(
 		return err
 	}
 	for k, executor := range hooks {
-		if err := executor.ExecuteLoginPostHook(w, r, g, f, s); err != nil {
+		if err := executor.ExecuteLoginPostHook(w, r, g, f, s, c); err != nil {
 			if errors.Is(err, ErrHookAbortFlow) {
 				e.d.Logger().
 					WithRequest(r).
--- a/oryx/dbal/testhelpers.go
+++ b/oryx/dbal/testhelpers.go
@@ -31,7 +31,7 @@ func RestoreFromSchemaDump(t testing.TB, c *pop.Connection, migrations fs.FS, wr
 
 	updateDump := func(t testing.TB) {
 		dump := dockertest.DumpSchema(t, c)
-		f, err := os.OpenFile(dumpFilename, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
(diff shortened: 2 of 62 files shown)


fenech:
I'm periodically coming back to check the status of this PR and clicking "update" to rebase it, but I don't know if you'd prefer for me to just leave it as it is :smile: 

I'm happy to contribute to some docs as well, just wanted to make sure the approach was good before starting to do that.

kghost:
Yes, this is crutial, otherwise we can't keep SSO information up to date.

selfservice/flow/login/hook.go
```diff
@@ -125,6 +126,7 @@ func (e *HookExecutor) PostLoginHook(
 	f *Flow,
 	i *identity.Identity,
 	s *session.Session,
+	c *claims.Claims,
```
jonas-jonas, line 129: It looks like you're passing nil here a lot. I'd prefer to see the functional option pattern ([Option 3: Functional Options Pattern](https://golang.cafe/blog/golang-functional-options-pattern.html)) in this case.

Just for the claims should be fine for now, as the other parameters are required and passed in, it seems.

jonas-jonas (changes requested):
Thanks for your contribution, this is going into the right direction.

We're going to need docs as well.

selfservice/flow/login/hook.go
fenech: I pushed a change, does it look right to you?

jonas-jonas (dismissed):
This is looking great. AFAICT, the missing test coverage is because we didn't have tests for the lines in the first place, and the changes are just renames.

LGTM, @zepatrik PTAL.

zepatrik (dismissed):
Great stuff 🎉

zepatrik:
I just checked and the overall issue with this approach is that the webhook target cannot update the identity on login:

> Modifying the identity is currently only possible during the registration and settings flows.

However, I think that adding the claims to the webhook is an OK solution, but not necessarily for the root problem. It is useful to notify other systems, but there will always be some delay between the login and update of the identity through another API call. This will be especially a problem for integrations using session to JWT or some kind of short-term cache.

IMO we should still merge this PR. I would however consider either allowing the webhook to update the identity in the response (adds latency to login), and/or add some other "quick" path for cases where one only wants to mirror claims into the identity (i.e. jsonnet).

robinknaapen:
@zepatrik 

Another PR is open that implements the jsonnet flow: https://github.com/ory/kratos/pull/3788

Not sure if that has been a consideration down the line

aeneasr (changes requested):
Blocking merge as I want to review this before merge :)

fenech:
Anything I can do to help move this along, @aeneasr (aside from pinging you :smile: )?

spijs:
I did not look into the implementation in depth, but it seems like the settings flow has not been considered? This would probably solve the issue mentioned here https://github.com/ory/kratos/issues/3816

As it is possible to set up new OIDC connections via the settings flow as well, would it make sense to also include the claims in the webhook there? For us it makes sense to have similar information as with registration to perform for example some sort of validation. While the login flow definitely makes sense to keep everything up to date, it does not fix the need for validation BEFORE accepting the new OIDC provider.

 If that would already be the case, I would be very happy.  Otherwise I feel like it could be a nice enhancement to this PR.

fenech:
> As it is possible to set up new OIDC connections via the settings flow as well, would it make sense to also include the claims in the webhook there? For us it makes sense to have similar information as with registration to perform for example some sort of validation. While the login flow definitely makes sense to keep everything up to date, it does not fix the need for validation BEFORE accepting the new OIDC provider.

This isn't something that I'd considered, as our use case only involves self-registration and then login. Since we are already using a version of Kratos based on this branch, I'd quite like to keep the scope minimal to increase its chances of getting merged, but if the project members see it as a useful addition then I'm happy to make an update.

To be honest, this was always intended as a workaround for our use case, and for us it would be sufficient to just do the same Jsonnet mapping onto the identity at both registration and login time, as @zepatrik mentioned earlier in the thread.

fenech:
I've rebased this onto latest master and resolved the conflicts. There are some e2e tests that are failing, but as far as I can tell, they're not related to these changes.

aeneasr:
Totally my bad for the lack of response here. 2024 was a very busy year. I will revisit this now with the intention of merging it upstream.

fenech:
@aeneasr would you like me to rebase and fix the conflicts, or were you planning on making changes yourself?

fenech:
> Totally my bad for the lack of response here. 2024 was a very busy year. I will revisit this now with the intention of merging it upstream.

looks like 2025 is shaping up to be a busy year too :smile: anyway let me know if you want me to resolve the conflicts!

fmiqbal:
any news ?

fenech:
Just rebased onto latest master and resolved the merge conflicts. Over a year after opening the PR and receiving some initial positive feedback, I'm wondering if change is still relevant or there might be a better way to do this?

selfservice/flow/login/hook.go
```diff
@@ -185,7 +200,7 @@ func (e *HookExecutor) PostLoginHook(
 		return err
 	}
 	for k, executor := range hooks {
-		if err := executor.ExecuteLoginPostHook(w, r, g, f, s); err != nil {
+		if err := executor.ExecuteLoginPostHook(w, r, g, f, s, e.c); err != nil {
```
aeneasr: Should not pass e.c here.

selfservice/flow/login/hook.go
```diff
@@ -175,6 +186,10 @@ func (e *HookExecutor) PostLoginHook(
 	classified := s
 	s = s.Declassified()
 
+	for _, o := range opts {
+		o(e)
```
aeneasr: Not thread safe because every thread modifies the same memory structure...

aeneasr (changes requested):
Hey there, better at some point than never - unfortunately the PR can't be merged in the current state because the code is not thread-safe. I understand the problem is extracting the claims in the "id token handler" and then in a complete different part putting them in the webhook context. This is a challenging problem to solve to be honest, and it has to be done correctly by "bubbling up" the state - versus relying on "globals" to store it.

Since this has recently come up as a requirements from some prospects, I think we'll finally put work into this topic.

fenech:
Hello, nice to hear from you! Happy to make some changes here but I just want to confirm what you mean exactly:
> it has to be done correctly by "bubbling up" the state - versus relying on "globals" to store it.

Are you referring to passing the claims as an extra parameter through all the layers, accepting the fact that `nil` will be passed in many cases? Or perhaps they could be stored on the request context?

fenech:
I've rebased the branch and essentially reverted it to one of its earlier incarnations (before I addressed this feedback from @jonas-jonas  https://github.com/ory/kratos/pull/3922#pullrequestreview-2112756704), where the `claims` are passed through as an explicit parameter rather than storing them on the executor service. I'm not sure if there was something else that you had in mind @aeneasr?

rogersindreu:
dear team, this is a very important feature for us... do we have any insight on when it will be released?

PR #3948, opened 2024-06-06
feat: improve session extend performance

aeneasr:
This patch improves the performance for extending session lifespans. Lifespan extension is tricky as it usually is part of the middleware of Ory Kratos consumers. As such, it is prone to transaction contention when we read and write to the same session row at the same time (and potentially multiple times).

To address this, we:

1. Introduce a locking mechanism on the row to reduce transaction contention;
2. Add a new feature flag which toggles returning 204 no content instead of 200 + session.

Be aware that all reads on the session table will have to wait for the transaction to commit before they return a value. This may cause long(er) response times on `/session/whoami` for sessions that are being extended at the same time.

BREAKING CHANGES: Going forward, the `/admin/session/.../extend` endpoint will return 204 no content for new Ory Network projects. We will deprecate returning 200 + session body in the future.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

* [ ] Add tests

### DIFF
--- a/persistence/sql/persister_session.go
+++ b/persistence/sql/persister_session.go
@@ -8,6 +8,9 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ory/herodot"
+	"github.com/ory/x/dbal"
+
 	"github.com/gobuffalo/pop/v6"
 	"github.com/gofrs/uuid"
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


aeneasr:
You can play around yourself with table locking in this guide using `cockroach demo` as your DB: https://www.cockroachlabs.com/docs/stable/select-for-update#examples

jonas-jonas (approved):

persistence/sql/persister_session.go
```diff
@@ -176,6 +179,50 @@ func (p *Persister) ListSessionsByIdentity(
...
+		if !s.CanBeRefreshed(ctx, p.r.Config()) {
+			return nil
+		}
+
+		s = s.Refresh(ctx, p.r.Config())
+		if _, err := tx.Where("id = ? AND nid = ?", sessionID, nid).UpdateQuery(s, "expires_at"); err != nil {
+			return sqlcon.HandleError(err)
+		}
+
+		trace.SpanFromContext(ctx).AddEvent(events.NewSessionLifespanExtended(ctx, s.ID, s.IdentityID, s.ExpiresAt))
```
alnr, line 221: This must be moved outside of the transaction into a deferred function call in order not to send duplicate events when the TX automatically retries (see `UpsertSession`).

persistence/sql/persister_session.go
```diff
@@ -176,6 +179,50 @@ func (p *Persister) ListSessionsByIdentity(
...
+			// This is a special case for CockroachDB. If the row is locked, we return a specific error message.
+			if errors.Is(err, sqlcon.ErrNoRows) && tx.Dialect.Name() == dbal.DriverCockroachDB {
+				return errors.WithStack(herodot.ErrNotFound.WithReason("The session you are trying to extend is already being extended by another request or does not exist."))
+			}
+
+			return sqlcon.HandleError(err)
+		}
+
+		if !s.CanBeRefreshed(ctx, p.r.Config()) {
+			return nil
```
alnr, line 213: return `400`?

aeneasr: Would be a breaking change

persistence/sql/persister_session.go
aeneasr: Nice catch - thanks!

PR #3953, opened 2024-06-14
test: deflake and parallelize persister tests

zepatrik:
Moving more tests to the context-based test configs, and enabling parallelism everywhere 🎉

### DIFF
--- a/driver/config/test_config.go
+++ b/driver/config/test_config.go
@@ -5,9 +5,6 @@ package config
 
 import (
 	"context"
-	"strings"
-
-	"github.com/knadh/koanf/maps"
 
 	"github.com/ory/kratos/embedx"
 	"github.com/ory/x/configx"
@@ -19,8 +16,7 @@ type (
 		contextx.Contextualizer
 		Options []configx.OptionModifier
 	}
-	contextKey  int
-	mapProvider map[string]any
+	contextKey int
 )
 
 func (t *TestConfigProvider) NewProvider(ctx context.Context, opts ...configx.OptionModifier) (*configx.Provider, error) {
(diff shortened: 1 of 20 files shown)


driver/config/test_config.go
```diff
@@ -51,25 +51,14 @@ func WithConfigValue(ctx context.Context, key string, value any) context.Context
...
-	for _, v := range expandedValues {
-		maps.Merge(v, values)
+		values = make([]map[string]any, 0)
 	}
+	newValues := make([]map[string]any, len(values), len(values)+1)
+	copy(newValues, values)
+	newValues = append(newValues, setValues)
 
-	return context.WithValue(ctx, contextConfigKey, values)
+	return context.WithValue(ctx, contextConfigKey, newValues)
```
zepatrik, line 63: Kinda copied from your branch, but adjusted slightly to be a slice of maps. The benefit of that is that overrides still work in the correct order deterministically:
```go
ctx = WithConfigValues(ctx, map[string]any{"serve": map[string]any{"public": x}})
ctx = WithConfigValues(ctx, map[string]any{"serve.public": x})
```
The keys in this case would not be the same, so both would be part of the map after merge, but only one of the values would be used at random.

persistence/sql/persister_errorx.go
```diff
@@ -55,14 +54,19 @@ func (p *Persister) ReadErrorContainer(ctx context.Context, id uuid.UUID) (_ *er
...
-		"UPDATE selfservice_errors SET was_seen = true, seen_at = ? WHERE id = ? AND nid = ?",
-		time.Now().UTC(), id, p.NetworkID(ctx)).Exec(); err != nil {
-		return nil, sqlcon.HandleError(err)
+	if err := p.Transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
+		if err := c.Where("id = ? AND nid = ?", id, p.NetworkID(ctx)).First(&ec); err != nil {
+			return sqlcon.HandleError(err)
+		}
+
+		if err := c.RawQuery(
+			"UPDATE selfservice_errors SET was_seen = true, seen_at = ? WHERE id = ? AND nid = ?",
```
zepatrik, line 63: Ideally we could use a `RETURNING` clause here, but obviously mysql does not support that...

persistence/sql/persister_test.go
```diff
@@ -94,7 +94,7 @@ func pl(t testing.TB) func(lvl logging.Level, s string, args ...interface{}) {
 
 func createCleanDatabases(t testing.TB) map[string]*driver.RegistryDefault {
 	conns := map[string]string{
-		"sqlite": "sqlite://file:" + t.TempDir() + "/db.sqlite?_fk=true",
+		"sqlite": "sqlite://file:" + t.TempDir() + "/db.sqlite?_fk=true&max_conns=1&lock=false",
```
zepatrik, line 97: This allows to run parallel tests with sqlite, because a connection pool of 1 will never execute parallel queries 🤯

PR #3954, opened 2024-06-17
test: enable server-side config from context

zepatrik:
With this change we can use the context-based test configs server side, e.g. in integration tests.

### DIFF
--- a/driver/config/test_config.go
+++ b/driver/config/test_config.go
@@ -5,6 +5,10 @@ package config
 
 import (
 	"context"
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/gofrs/uuid"
 
 	"github.com/ory/kratos/embedx"
 	"github.com/ory/x/configx"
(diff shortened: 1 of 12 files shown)


driver/config/test_config.go
```diff
@@ -51,14 +55,98 @@ func WithConfigValue(ctx context.Context, key string, value any) context.Context
...
+		configs: make(map[uuid.UUID][]map[string]any),
+		handler: h,
+	}
+}
+
+func (t *ConfigurableTestHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	cID := r.Header.Get("Test-Config-Id")
+	if config, ok := t.configs[uuid.FromStringOrNil(cID)]; ok {
+		r = r.WithContext(WithConfigValues(r.Context(), config...))
+	}
```
zepatrik, line 86: The handler looks up the config ID from the request header and adds the config to the request context server-side if it found it.

driver/config/test_config.go
```diff
@@ -51,14 +55,98 @@ func WithConfigValue(ctx context.Context, key string, value any) context.Context
...
+	return id
+}
+
+func (t *ConfigurableTestHandler) UseConfig(r *http.Request, id uuid.UUID) *http.Request {
+	r.Header.Set("Test-Config-Id", id.String())
+	return r
+}
+
+func (t *ConfigurableTestHandler) UseConfigValues(r *http.Request, values ...map[string]any) *http.Request {
+	return t.UseConfig(r, t.RegisterConfig(values...))
```
zepatrik, line 102: Registering the config values and setting the config header on the request (client side).

PR #3968, opened 2024-06-24
feat: allow deletion of an individual OIDC credential

alnr:
This extends the existing `DELETE /admin/identities/{id}/credentials/{type}` API to accept an `?identifier=foobar` query parameter for `{type}==oidc` like such:

`DELETE /admin/identities/{id}/credentials/oidc?identifier=github%3A012345`

This will delete the GitHub OIDC credential with the identifier `github:012345` (`012345` is the subject as returned by GitHub).

To find out which OIDC credentials exist, call `GET /admin/identities/{id}?include_credential=oidc` beforehand.

This will allow you to delete individual OIDC credentials for users, even if they have several set up.

### DIFF
--- a/identity/identity.go
+++ b/identity/identity.go
@@ -507,6 +507,80 @@ func (i *Identity) WithDeclassifiedCredentials(ctx context.Context, c cipher.Pro
 	return &ii, nil
 }
 
+func (i *Identity) deleteCredentialWebAuthFromIdentity() error {
+	cred, ok := i.GetCredentials(CredentialsTypeWebAuthn)
+	if !ok {
+		// This should never happend as it's checked earlier in the code;
+		// But we never know...
+		return errors.WithStack(herodot.ErrNotFound.WithReasonf("You tried to remove a WebAuthn credential but this user has no such credential set up."))
+	}
+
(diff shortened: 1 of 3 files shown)


identity/identity.go
```diff
@@ -507,6 +507,80 @@ func (i *Identity) WithDeclassifiedCredentials(ctx context.Context, c cipher.Pro
 	return &ii, nil
 }
 
+func (i *Identity) deleteCredentialWebAuthFromIdentity() error {
```
alnr, line 510: This function was moved here from `identity/handler.go` where it had the signature `func deletCredentialWebAuthFromIdentity(identity *Identity) (*Identity, error)`

aeneasr (commented):
This is looking great already!

What I'm most worried about is that we accidentally delete the wrong credential. While it doesn't appear to be the case from reading the code, I think it would be good to have test(s) that verify that.

To make it a bit easier to test, you can probably set up a unit test for the new identity methods?

alnr:
I've added a test `TestDeleteCredentialOIDCFromIdentity`. The handler test is also quite comprehensive.

aeneasr (approved):

PR #3978, opened 2024-07-01
feat: password migration hook

hperl:
This PR adds a password migration hook to easily migrate passwords for which we do not have the hash.

- For each user that needs to be migrated to Ory Network, a new identity is created with a credential of type `password` with a config of `{"use_password_migration_hook": true}` .
- When a user logs in, the credential identifier and password will be sent to the `password_migration` web hook if all of these are true:
    - The user’s identity’s password credential is `{"use_password_migration_hook": true}`
    - The `password_migration` hook is configured
- After calling the `password_migration` hook, the HTTP status code will be inspected:
    - On 200, we parse the response as JSON and look for `{"status": "password_match"}`. The password credential config will be replaced with the hash of the actual password.
    - On any other status code, we assume that the password is not valid.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/hook/password_migration_hook.go
+++ b/selfservice/hook/password_migration_hook.go
@@ -1,3 +1,6 @@
+// Copyright © 2024 Ory Corp
+// SPDX-License-Identifier: Apache-2.0
+
 package hook
 
 import (
--- a/selfservice/hook/password_migration_hook.go
+++ b/selfservice/hook/password_migration_hook.go
@@ -75,15 +75,6 @@ func (p *PasswordMigration) Execute(ctx context.Context, data *PasswordMigration
 
 	resp, err := httpClient.Do(req)
 	if err != nil {
-		if isTimeoutError(err) {
-			return herodot.DefaultError{
-				CodeField:     http.StatusGatewayTimeout,
-				StatusField:   http.StatusText(http.StatusGatewayTimeout),
-				GRPCCodeField: grpccodes.DeadlineExceeded,
-				ReasonField:   "A third-party upstream service could not be reached. Please try again later.",
-				ErrorField:    "calling the password migration hook failed",
-			}.WithWrap(errors.WithStack(err))
-		}
 		return herodot.DefaultError{
 			CodeField:     http.StatusBadGateway,
 			StatusField:   http.StatusText(http.StatusBadGateway),
--- a/selfservice/hook/password_migration_hook.go
+++ b/selfservice/hook/password_migration_hook.go
@@ -1,0 +1,114 @@
+package hook
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/pkg/errors"
+	"github.com/tidwall/gjson"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	semconv "go.opentelemetry.io/otel/semconv/v1.11.0"
+	"go.opentelemetry.io/otel/trace"
+	grpccodes "google.golang.org/grpc/codes"
+
+	"github.com/ory/herodot"
+	"github.com/ory/kratos/request"
+	"github.com/ory/kratos/schema"
+	"github.com/ory/x/otelx"
+)
+
+type (
+	PasswordMigration struct {
+		deps webHookDependencies
+		conf json.RawMessage
+	}
+	PasswordMigrationData struct {
+		Identifier string `json:"identifier"`
+		Password   string `json:"password"`
+	}
+)
+
+var passwordMigrationJsonnetTemplate string
+
+func init() {
+	snippet := []byte(`function(ctx) { password: ctx.password, identifier: ctx.identifier }`)
+	passwordMigrationJsonnetTemplate = "base64://" + base64.StdEncoding.EncodeToString(snippet)
+}
+
+func NewPasswordMigrationHook(deps webHookDependencies, conf json.RawMessage) *PasswordMigration {
+	return &PasswordMigration{deps: deps, conf: conf}
+}
+
+func (p *PasswordMigration) Execute(ctx context.Context, data *PasswordMigrationData) (err error) {
+	var (
+		httpClient = p.deps.HTTPClient(ctx)
+		emitEvent  = gjson.GetBytes(p.conf, "emit_analytics_event").Bool() || !gjson.GetBytes(p.conf, "emit_analytics_event").Exists() // default true
+		tracer     = trace.SpanFromContext(ctx).TracerProvider().Tracer("kratos-webhooks")
+	)
+
+	ctx, span := tracer.Start(ctx, "selfservice.login.password_migration")
+	defer otelx.End(span, &err)
+
+	if emitEvent {
+		instrumentHTTPClientForEvents(ctx, httpClient)
+	}
+	builder, err := request.NewBuilder(ctx, p.conf, p.deps, jsonnetCache)
+	if err != nil {
+		return err
+	}
+
+	builder.Config.TemplateURI = passwordMigrationJsonnetTemplate
+
+	req, err := builder.BuildRequest(ctx, data)
+	if errors.Is(err, request.ErrCancel) {
+		span.SetAttributes(attribute.Bool("password_migration.jsonnet.canceled", true))
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	p.deps.Logger().WithRequest(req.Request).Info("Dispatching password migration hook")
+	req = req.WithContext(ctx)
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		if isTimeoutError(err) {
+			return herodot.DefaultError{
+				CodeField:     http.StatusGatewayTimeout,
+				StatusField:   http.StatusText(http.StatusGatewayTimeout),
+				GRPCCodeField: grpccodes.DeadlineExceeded,
+				ErrorField:    err.Error(),
+				ReasonField:   "A third-party upstream service could not be reached. Please try again later.",
+			}.WithWrap(errors.WithStack(err))
+		}
+		return herodot.DefaultError{
+			CodeField:     http.StatusBadGateway,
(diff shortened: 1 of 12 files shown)


selfservice/strategy/password/login.go
```diff
@@ -81,6 +83,32 @@ func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow,
...
+		err = migrationHook.Execute(r.Context(), &hook.PasswordMigrationData{Identifier: identifier, Password: p.Password})
+		if err != nil {
+			return nil, s.handleLoginError(w, r, f, &p, err)
+		}
+
+		if err := s.migratePasswordHash(r.Context(), i.ID, []byte(p.Password)); err != nil {
+			return nil, s.handleLoginError(w, r, f, &p, err)
+		}
+
+		f.Active = identity.CredentialsTypePassword
```
jonas-jonas, line 103:
```suggestion
```

selfservice/hook/password_migration_hook.go
```diff
@@ -0,0 +1,117 @@
...
+			ReasonField:   "A third-party upstream service could not be reached. Please try again later.",
+			ErrorField:    "calling the password migration hook failed",
+		}.WithWrap(errors.WithStack(err))
+	}
+	defer resp.Body.Close()
+	span.SetAttributes(semconv.HTTPAttributesFromHTTPStatusCode(resp.StatusCode)...)
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return nil
```
jonas-jonas, line 104: As it stands, it's pretty easy to misconfigure the hook's target by returning 200 unintentionally. IIRC some web servers return 200 by default, so in that case, if you forget to add an explicit status code for a specific error case, you return 200 erroneously. 

Maybe we could change this to expect the target to also return some JSON response, that is intentional:

```json
{
	"status": "password_match"
}
```
or something like that. 

This would also allow us to implement additional enhancing of the metadata or allow changing the error message that is shown to the user, in the future.

jonas-jonas (changes requested):

selfservice/hook/password_migration_hook.go
hperl: That is a good idea!

jonas-jonas (dismissed):

selfservice/hook/password_migration_hook.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+func NewPasswordMigrationHook(deps webHookDependencies, conf json.RawMessage) *PasswordMigration {
+	return &PasswordMigration{deps: deps, conf: conf}
+}
+
+func (p *PasswordMigration) Execute(ctx context.Context, data *PasswordMigrationRequest) (err error) {
+	var (
+		httpClient = p.deps.HTTPClient(ctx)
+		emitEvent  = gjson.GetBytes(p.conf, "emit_analytics_event").Bool() || !gjson.GetBytes(p.conf, "emit_analytics_event").Exists() // default true
```
zepatrik, line 47: The default is already set to true in the config schema, I don't think that it is the best idea to re-implement that here. Just adds an additional (hidden) point where this is set.

selfservice/strategy/password/login.go
```diff
@@ -81,17 +83,33 @@ func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow,
...
+		}
+
+		if !s.d.Hasher(r.Context()).Understands([]byte(o.HashedPassword)) {
+			if err := s.migratePasswordHash(r.Context(), i.ID, []byte(p.Password)); err != nil {
+				return nil, s.handleLoginError(w, r, f, &p, err)
+			}
+		}
 	}
 
-	f.Active = identity.CredentialsTypePassword
```
zepatrik, line 94: 🤔 what was this?

selfservice/strategy/password/login_test.go
```diff
@@ -864,4 +860,168 @@ func TestCompleteLogin(t *testing.T) {
...
+			addPassword:       func(identifier, password string) { passwordDB[identifier] = password },
+			expectSuccess:     false,
+		}, {
+			name:              "should not call migration hook if disabled",
+			credentialsConfig: `{"use_password_migration_hook": true}`,
+			addPassword:       func(identifier, password string) { passwordDB[identifier] = password },
+			setupFn: func() func() {
+				require.NoError(t, reg.Config().Set(ctx, config.ViperKeyPasswordMigrationHook+".enabled", false))
+				return func() {
+					require.NoError(t, reg.Config().Set(ctx, config.ViperKeyPasswordMigrationHook+".enabled", true))
```
zepatrik, line 951: Please use the context-based test config instead 😉

selfservice/strategy/password/login_test.go
```diff
@@ -864,4 +860,168 @@ func TestCompleteLogin(t *testing.T) {
...
+			default:
+				w.WriteHeader(http.StatusForbidden)
+			}
+		}))
+		t.Cleanup(ts.Close)
+
+		require.NoError(t, reg.Config().Set(ctx, config.ViperKeyPasswordMigrationHook, &config.PasswordMigrationHook{
+			Enabled: true,
+			Config:  json.RawMessage(fmt.Sprintf(`{"URL":"%s"}`, ts.URL)),
+		}))
```
zepatrik, line 906: Please use the context-based test config instead 😉

selfservice/strategy/password/login_test.go
```diff
@@ -864,4 +860,168 @@ func TestCompleteLogin(t *testing.T) {
 			false, true, http.StatusOK, redirTS.URL)
 		assert.Equal(t, identifier, gjson.Get(body, "identity.traits.subject").String(), "%s", body)
 	})
+
+	t.Run("suite=password migration hook", func(t *testing.T) {
+		ctx := context.Background()
+
+		passwordDB := make(map[string]string)
```
zepatrik, line 867: What I would like instead is to have a channel here that you send a response (function) to instead of this custom logic. Then the server is super simple, it just receives the response from the channel and writes it, and the test setup is also very easy to understand without reading the server code.

```suggestion
		passwordDB := make(chan func(context.Context, http.ResponseWriter), 1)
		writeStatus := func(status int) func(context.Context, http.ResponseWriter) {
			return func(_ context.Context, w http.ResponseWriter) {
				w.WriteHeader(status)
			}
		}
		expectedResponseBody := `{"status":"password_match"}`
		writeBodyOK := func(body string) func(context.Context, http.ResponseWriter) 
			return func(_ context.Context, w http.ResponseWriter) {
				w.WriteHeader(http.StatusOK)
				_, _ = io.WriteString(w, body)
			}
		}
```

selfservice/strategy/password/login_test.go
```diff
@@ -864,4 +860,168 @@ func TestCompleteLogin(t *testing.T) {
...
+			b, err := io.ReadAll(r.Body)
+			require.NoError(t, err)
+			_ = r.Body.Close()
+			var payload struct {
+				Identifier string `json:"identifier"`
+				Password   string `json:"password"`
+			}
+			require.NoError(t, json.Unmarshal(b, &payload))
+
+			switch pw := passwordDB[payload.Identifier]; pw {
```
zepatrik, line 879: Looks like we're missing a case for the server responding with a 200 status code, but unexpected body.

zepatrik (changes requested):
Code looks very good, I just have some improvement suggestions for the test.

selfservice/strategy/password/login.go
hperl: `f.Active` was written twice, just some spring cleaning.

selfservice/hook/password_migration_hook.go
zepatrik: I am pretty sure that they do get applied, did you try? You can then also remove the other one.

hperl: Yes, I tried, and it is indeed false when the config is empty.

hperl: The reasons is that we do not parse the hook config as part of the koanf loading.

zepatrik (approved):

jonas-jonas (approved):

selfservice/hook/password_migration_hook.go
```diff
@@ -0,0 +1,111 @@
...
+	defer resp.Body.Close()
+	span.SetAttributes(semconv.HTTPAttributesFromHTTPStatusCode(resp.StatusCode)...)
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		// We now check if the response matches `{"status": "password_match" }`.
+		dec := json.NewDecoder(io.LimitReader(resp.Body, 1024)) // limit the response body to 1KB
+		var response PasswordMigrationResponse
+		if err := dec.Decode(&response); err != nil || response.Status != "password_match" {
+			return errors.WithStack(schema.NewInvalidCredentialsError())
```
aeneasr, line 95: This probably needs some tracing or something, because it will be difficult to debug this hook otherwise

selfservice/strategy/password/login.go
```diff
@@ -81,17 +83,33 @@ func (s *Strategy) Login(w http.ResponseWriter, r *http.Request, f *login.Flow,
 		return nil, herodot.ErrInternalServerError.WithReason("The password credentials could not be decoded properly").WithDebug(err.Error()).WithWrap(err)
 	}
 
-	if err := hash.Compare(r.Context(), []byte(p.Password), []byte(o.HashedPassword)); err != nil {
-		return nil, s.handleLoginError(w, r, f, &p, errors.WithStack(schema.NewInvalidCredentialsError()))
-	}
+	if o.ShouldUsePasswordMigrationHook() {
```
aeneasr, line 86: Who sets this to false once this is executed successfully? Is it `s.migratePasswordHash`?

aeneasr: Looks like it is

hperl: Yes, if the password hash is not empty, this returns false.

PR #3993, opened 2024-07-11
chore(docs): remove unsupported credential types in deleteIdentityCredential operation

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -910,6 +910,18 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
 	h.r.Writer().Write(w, r, WithCredentialsMetadataAndAdminMetadataInJSON(updatedIdentity))
 }
 
+// DeletableCredentialsType contains all credentials that can be deleted using the deleteIdentityCredentials operation
+//
+// swagger:enum DeletableCredentialsType
+type DeletableCredentialsType string
+
+const (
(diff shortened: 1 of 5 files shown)


identity/handler.go
```diff
@@ -910,6 +910,18 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
 	h.r.Writer().Write(w, r, WithCredentialsMetadataAndAdminMetadataInJSON(updatedIdentity))
 }
 
+// DeletableCredentialsType contains all credentials that can be deleted using the deleteIdentityCredentials operation
+//
+// swagger:enum DeletableCredentialsType
+type DeletableCredentialsType string
+
+const (
+	DeletableCredentialsTypeOIDC     DeletableCredentialsType = "oidc"
```
aeneasr, line 919: This is a breaking change for the SDK due to the renaming - is it really needed, assuming that we'll allow removing all credentials type in the future?

jonas-jonas: I guess that's fair. 

If we are going to allow the other credentials to be removed (which I am in favor of), I am fine with leaving as is. 

If not, this would remove another point of friction here (and potential support requests).

aeneasr: OK, we will add support for deleting other types as well, so I'm closing this PR

PR #4021, opened 2024-08-02
fix: validate page tokens for better error codes

zepatrik:
Currently the error returned is similar to `unable to fetch records: ERROR: error in argument for $2: could not parse "1" as type uuid: could not parse "1" as type uuid: uuid: incorrect UUID length: 1 (SQLSTATE 22P02)`.

### DIFF
--- a/persistence/sql/identity/persister_identity.go
+++ b/persistence/sql/identity/persister_identity.go
@@ -799,6 +799,10 @@ func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.
 		attribute.Stringer("network.id", p.NetworkID(ctx)))...))
 	defer otelx.End(span, &err)
 
+	if _, err := uuid.FromString(paginator.Token().Encode()); err != nil {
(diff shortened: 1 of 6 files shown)


persistence/sql/identity/persister_identity.go
```diff
@@ -799,6 +799,10 @@ func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.
 		attribute.Stringer("network.id", p.NetworkID(ctx)))...))
 	defer otelx.End(span, &err)
 
+	if _, err := uuid.FromString(paginator.Token().Encode()); err != nil {
+		return nil, nil, errors.WithStack(herodot.ErrBadRequest.WithReason("The page token is invalid, do not craft your own page tokens"))
```
jonas-jonas, line 803: Could we extract this to a shared `var` somewhere?

jonas-jonas (approved):

aeneasr (approved):

PR #4036, opened 2024-08-13
fix: improve OIDC account linking UI

jonas-jonas:
## Related issue(s)

Fixes [#5983](https://github.com/ory-corp/cloud/issues/5983)

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


selfservice/strategy/oidc/strategy.go
```diff
@@ -692,6 +667,70 @@
...
+		case text.InfoSelfServiceLogin:
+			lf.UI.Nodes[i].Meta.Label = text.NewInfoLoginAndLink()
+		case text.InfoSelfServiceLoginWith:
+			p := gjson.GetBytes(n.Meta.Label.Context, "provider").String()
+			lf.UI.Nodes[i].Meta.Label = text.NewInfoLoginWithAndLink(p)
+		}
+
+		// This can happen, if login hints are disabled. In that case, we need to make sure to show all credential options.
+		// It could in theory also happen due to a mis-configuration, and in that case, we should make sure to not delete the entire flow.
+		if len(availableCredentials) == 0 {
```
hperl, line 705: I think this is more readable if you put this check as a precondition at the top, and then dont touch `lf.UI.Nodes` at all and just clear the messages. WDYT?

selfservice/strategy/oidc/strategy.go
```diff
@@ -692,6 +667,70 @@ func (s *Strategy) handleError(w http.ResponseWriter, r *http.Request, f flow.Fl
 	return err
 }
 
+func (s *Strategy) populateAccountLinkingUI(ctx context.Context, lf *login.Flow, usedProviderID string, duplicateIdentifier string, availableCredentials []string) {
```
hperl, line 670: Can we also have a button to cancel the account linking? This could just redirect to the login UI, which starts a fresh flow.

aeneasr: Makes sense to me!

selfservice/strategy/oidc/strategy.go
jonas-jonas: We need to adjust the labels just above this statement though. If I move it to the top, that wouldn't happen anymore.

selfservice/strategy/oidc/strategy.go
jonas-jonas: We can, but it's a bit more annoying to get looking good.

Either button hell:

image

or (here the "cancel" link just kinda disappears in the form, IMO.)

image

text/message_login.go
```diff
@@ -61,7 +61,7 @@ func NewInfoLoginLinkMessage(dupIdentifier, provider, newLoginURL string) *Messa
 		ID:   InfoSelfServiceLoginLink,
 		Type: Info,
 		Text: fmt.Sprintf(
-			"Signing in will link your account to %q at provider %q. If you do not wish to link that account, please start a new login flow.",
+			"You tried signing in with %q, but that email is already linked to another account. Please confirm your account by signing in with one of the options below. This will add your account %[1]q at %q as another way to sign in to your account.",
```
aeneasr, line 64: You tried signing in with %q, but that email is linked to another account. Please sign in with one of the options below to add %[1]q at %q as another sign-in method.

aeneasr, line 64:
```suggestion
			"You tried to sign in with %q, but that email is already used. Sign in with one of the options below to add %q at %q as another sign-in method to the existing account.",
```

selfservice/strategy/oidc/strategy.go
aeneasr: The first variant looks good enough. May I suggest changes to the copy:

```patch
-Sign in with GitHub and link credential
+Link with GitHub account
```

text/message_login.go
jonas-jonas: This is similar to one of my first attempts, too. I revised that to the current proposal (using chatgpt).

> but that email is already used

This is not obvious enough, IMO. E.g. used by what? I'd propose, "Used by another account" makes this clear.

> as another sign-in method to the existing account.

I felt "method" might not be clear enough (and is pretty technical in our context). "Another way to sign in to your account" is pretty clear, although it's a bit longer.

```suggestion
			"You tried to sign in with %q, but that email is already used by another account. Sign in to your account with one of the options below to add your account %[1]q at %q as another way to sign in.",
```

WDYT?

selfservice/strategy/oidc/strategy.go
jonas-jonas: "Link" is so technical. 

"Confirm with GitHub", "Confirm with code" (though that's also non-obvious) would be better, IMO.

jonas-jonas: The button is problematic, because typically a button does a form submit, which would mean that we would need to cancel the linking on the server side and redirect from there, causing an unnecessary redirect. 

I think the link is good, if it's ordered correctly.

jonas-jonas: And actually rendering the cancel button is super complicated in Ory Elements, because if a user has both the password, and code method (or passkey/webauthn) configured we would show the cancel button for each group.

Untangling that, would be complicated.

--

Adding it to the card footer in ory elements if the form is an account linking form directly is much easier: 

image

text/message_login.go
aeneasr: I find the sentence a bit confusing (`Sign in to your account with one of the options below to add your account %[1]q at %q as another way to sign in`) but I also don't have a suggestion how to improve it. So it's fine by me.

aeneasr (approved):

hperl (approved):
LGTM!

aeneasr:
Needs a merge, then good to go!

PR #4044, opened 2024-08-19
chore: don't return allowed return URLs

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/x/http_secure_redirect.go
+++ b/x/http_secure_redirect.go
@@ -17,8 +17,6 @@ import (
 	"github.com/ory/x/stringsx"
 	"github.com/ory/x/urlx"
 
-	"github.com/samber/lo"
-
 	"github.com/ory/kratos/driver/config"
 )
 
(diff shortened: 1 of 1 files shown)


hperl (approved):

x/http_secure_redirect.go
```diff
@@ -145,10 +143,8 @@ func SecureRedirectTo(r *http.Request, defaultReturnTo *url.URL, opts ...SecureR
 
 	return nil, errors.WithStack(herodot.ErrBadRequest.
 		WithID(text.ErrIDRedirectURLNotAllowed).
-		WithReasonf("Requested return_to URL %q is not allowed.", returnTo).
-		WithDebugf("Allowed domains are: %v", strings.Join(lo.Map(o.allowlist, func(u url.URL, _ int) string {
-			return u.String()
-		}), ", ")))
+		WithReasonf("Requested return_to URL %q is not allowed.", returnTo),
```
aeneasr, line 146: Log this instead then? Or trace?

aeneasr (changes requested):

x/http_secure_redirect.go
jonas-jonas: I don't think that adds any value, because you can always check the project configuration.

I added this a while back mainly to make live for a developer implementing Ory a little easier, but in this configuration we're leaking this information to end users if they deliberately trigger this error (e.g. by providing an illegal `return_to`).

aeneasr: ok

aeneasr (approved):

Issue #4045, opened 2024-08-19
Unable to sign up via passkeys but can log in with Webauthn -> Passkeys 


eadwu:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

Trying to sign up using passkeys fails but passwordless webauthn works.

However trying to sign in using the passkey method with the passwordless webauthn also works.

There shouldn't be a configuration error as `config.rp` for both `passkey` and `webauthn` are the same. The error I get from trying to sign up using passkeys is
```
your device can't be used with this site 
localhost may require a newer or different kind of device
```

### Reproducing the bug

`docker-compose -f quickstart.yml -f quickstart-standalone.yml up --build --force-recreate` after applying the following patch on top of 68693a43e4e1e3028f17789e72d0b79f6298d139
```diff
diff --git a/contrib/quickstart/kratos/passkey/kratos.yml b/contrib/quickstart/kratos/passkey/kratos.yml
index 776b17e3c..55a558664 100644
--- a/contrib/quickstart/kratos/passkey/kratos.yml
+++ b/contrib/quickstart/kratos/passkey/kratos.yml
@@ -100,7 +100,7 @@ hashers:
 identity:
   schemas:
     - id: default
-      url: file://contrib/quickstart/kratos/passkey/identity.schema.json
+      url: file:///etc/config/kratos/identity.schema.json
 
 courier:
   smtp:
diff --git a/quickstart-standalone.yml b/quickstart-standalone.yml
index 8c7992842..a9878bab5 100644
--- a/quickstart-standalone.yml
+++ b/quickstart-standalone.yml
@@ -7,4 +7,4 @@ services:
     environment:
       - PORT=4455
       - SECURITY_MODE=
-      - KRATOS_BROWSER_URL=http://127.0.0.1:4433/
+      - KRATOS_BROWSER_URL=http://localhost:4433/
[...]

PR #4047, opened 2024-08-20
fix: allow cross-platform devices for passkeys

eadwu:
## Related issue(s)
https://github.com/ory/kratos/issues/4045

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

This aligns more with expected behavior. If one can sign in using cross platform devices, it doesn't make sense to restrict it to only devices within the same platform.

This is probably considered less secure when compared to the same-device restriction.

### DIFF
--- a/driver/config/config.go
+++ b/driver/config/config.go
@@ -1494,7 +1494,7 @@ func (p *Config) PasskeyConfig(ctx context.Context) *webauthn.Config {
 		RPID:          id,
 		RPOrigins:     origins,
 		AuthenticatorSelection: protocol.AuthenticatorSelection{
-			AuthenticatorAttachment: "platform",
+			AuthenticatorAttachment: "cross-platform",
 			RequireResidentKey:      pointerx.Ptr(true),
 			UserVerification:        protocol.VerificationPreferred,
 		},


aeneasr:
Would it be possible to add an e2e test for this in cypress or playwright?

@hperl do passkeys generally support this config value?

eadwu:
I'm not exactly sure how to write tests in this repo.

I checked https://github.com/ory/kratos/pull/3748/files since I should really just be able to copy WebAuthn tests as passkeys only really put it in the device instead of storing it on the server (from what I know). If the tests for passkeys and webauthn don't have anything in relation to the authentication modality, I don't think I could write a test for it.

I'm not intimately familiar with the spec so I'm not sure about whether or not I can test it via cypress or playwright.

The difference is in what devices are allowed to connect. In the browser `cross-platform` is the QR-code while `platform` is only directly connected devices (unset allows both).

aeneasr:
I‘m not confident that cross-platform is a valid method for „true“ passkeys. @hperl probably knows

eadwu:
I found this https://passkeys.dev/docs/reference/terms/#cross-device-authentication-cda but yeah I'm not entirely sure either.

aeneasr:
It does indeed look like this is OK:

https://passkeys.dev/docs/use-cases/bootstrapping/#cross-device-authentication

marcporr:
Hi,

first - I really appreciate your work here! Thank you for this cool piece of software.

Are there any news on when this fix will become available in a release build? We would like to use kratos with Yubikeys but can currently not use the release builds because of this issue.

Thank you!

PR #4055, opened 2024-08-23
fix: concurrent map update for webhook header

hperl:


selfservice/hook/web_hook.go
```diff
@@ -448,11 +449,12 @@ var RequestHeaderAllowList = map[string]struct{}{
 }
 
 func removeDisallowedHeaders(data *templateContext) {
-	for key := range data.RequestHeaders {
-		if _, ok := RequestHeaderAllowList[textproto.CanonicalMIMEHeaderKey(key)]; !ok {
-			data.RequestHeaders.Del(key)
-		}
-	}
+	headers := maps.Clone(data.RequestHeaders)
```
hperl, line 452: `data.RequestHeaders` might be read by the logger concurrently.

aeneasr: Use a sync.Map instead?

aeneasr: Clone will also have a concurrent read?

hperl: `maps.Clone` allocates a new map and shallow-copies the value, but since that is a string it is actually a full independent copy.

aeneasr (approved):

Issue #4009, opened 2024-07-19
Support for PKCE in OIDC social providers


OskarsPakers:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Some identity providers require Proof Key for Code Exchange (PKCE) code_challenge query parameter in authorization endpoint and code_verifier in token exchange endpoint

### Describe your ideal solution

Generic provider had configuration to enable and generate code_challenge upon redirect to identity provider and sends verifier value when echanging code for token.
oauth2 go package already supports generating verifier and the challenge https://pkg.go.dev/golang.org/x/oauth2#S256ChallengeFromVerifier

### Workarounds or alternatives

Implementing custom provider might be an option, but it seems that oidc strategy must be change too to pass AuthCodeOption to token exchange https://github.com/ory/kratos/blob/ff90216f03e7b9c8112f07474fa34d960eb9ceb9/selfservice/strategy/oidc/strategy.go#L526

### Version

2.2.0

### Additional Context

I`d like to submit pull request if this feature makes sense. Any hints appreciated

IchordeDionysos:
Would be amazing to have, we've also ran into this twice now (in our case we could just disable PKCE).
Had this with some customers identity provider and recently I've experienced it while enabling support for Salesforce.

aeneasr:
Yeah, that makes sense. There’s probably a library for Go that does that and that can easily be added to e.g. Salesforce

IchordeDionysos:
Note: In Salesforce, you can disable PKCE :) See the documentation here ory/docs#1797

It would be nice to have that possibility in generic providers, though.

OskarsPakers:
We plan to submit pull request for this in upcoming days.
[...]

PR #4033, opened 2024-08-13
feat: add support for Proof Key For Code Exchange (PKCE) in OIDC social providers

OskarsPakers:
Adds support for Proof Key For Code Exchange (PKCS) in OIDC social providers according to rfc7636 https://datatracker.ietf.org/doc/html/rfc7636

## Related issue(s)

#4009 

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [x] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got the approval (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [x] I have added or changed [the documentation](https://github.com/ory/docs).

## Further Comments

Supports methods s256 and plain.
AuthCodeURLOptions method signature does not return err, therefore can`t return an error. Should we change the signature of the method and lift the functionality up to the strategy code?

PR #4059, opened 2024-08-26
feat: client-side PKCE support during OIDC and generic redirect_uri

alnr:
This change introduces a new configuration for OIDC providers: `pkce` with values `auto` (default), `never`, `force`.

When `auto` is specified or the field is omitted, Kratos will perform autodiscovery and perform PKCE when the server advertises support for it. This requires the `issuer_url` to be set for the provider.

`never` completely disables PKCE support. This is only theoretically useful: when a provider advertises PKCE support but doesn't actually implement it.

`force` always sends a PKCE challenge in the initial redirect URL, regardless of what the provider advertises. This setting is useful when the provider offers PKCE but doesn't advertise it in his `./well-known/openid-configuration`.

**Important**: When setting `pkce: force`, you must whitelist a different return URL for your OAuth2 client in the provider's configuration. Instead of `<base-url>/self-service/methods/oidc/callback/<provider>`, you must use `<base-url>/self-service/methods/oidc/callback` (note missing last path segment). This is to enable the use of the same OAuth client ID+secret when configuring several Kratos OIDC providers, without having to whitelist individual `redirect_uri`s for each Kratos provider config.

Supersedes https://github.com/ory/kratos/pull/4033
Closes https://github.com/ory/kratos/issues/4009
Ref https://github.com/ory-corp/cloud/issues/6613


aeneasr:
> Important: When setting pkce: force, you must whitelist a different return URL for your OAuth2 client in the provider's configuration. Instead of <base-url>/self-service/methods/oidc/callback/<provider>, you must use <base-url>/self-service/methods/oidc/callback (note missing last path segment). This is to enable the use of the same OAuth client ID+secret when configuring several Kratos OIDC providers, without having to whitelist individual redirect_uris for each Kratos provider config.

Why is that?

selfservice/strategy/oidc/pkce_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/ory/kratos/internal"
+	"github.com/ory/kratos/selfservice/strategy/oidc"
+	"github.com/ory/x/sqlxx"
+)
+
+func TestPKCESupport(t *testing.T) {
+	supported := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, `{"issuer": "http://%s", "code_challenge_methods_supported":["S256"]}`, r.Host)
```
aeneasr, line 22: This does not need an additional test to check invalid values, because we check for existence of `S256` and nothing else.

selfservice/strategy/oidc/provider_config.go
```diff
@@ -119,9 +119,23 @@ type Configuration struct {
...
+
+	// PKCE controls if the OpenID Connect OAuth2 flow should use PKCE (Proof Key for Code Exchange).
+	// Possible values are: `auto` (default), `never`, `force`.
+	// - `auto`: PKCE is used if the provider supports it. Requires setting `issuer_url`.
+	// - `never`: Disable PKCE entirely for this provider, even if the provider advertises support for it.
+	// - `force`: Always use PKCE, even if the provider does not advertise support for it. OAuth2 flows will fail if the provider does not support PKCE.
+	// IMPORTANT: If you set this to `force`, you must whitelist a different return URL for your OAuth2 client in the provider's configuration.
+	// Instead of <base-url>/self-service/methods/oidc/callback/<provider>, you must use <base-url>/self-service/methods/oidc/callback
+	// (Note the missing <provider> path segment and no trailing slash).
+	PKCE string `json:"pkce"`
```
aeneasr, line 131: You could make this an enum, in which case this would also end up correctly in swagger API definitions and so on.

selfservice/strategy/oidc/state.go
```diff
@@ -0,0 +1,80 @@
...
+	} else {
+		return &State{FlowID: string(id), Data: data}, nil
+	}
+}
+
+func storeProvider(f flow.InternalContexter, provider string) error {
+	f.EnsureInternalContext()
+	bytes, err := sjson.SetBytes(
+		f.GetInternalContext(),
+		"oidc_provider",
```
aeneasr, line 61: The internal context is prone to problems when retrying flows. I believe this is/was a problem for several MFA methods. Relying on the internal context implies that we probably need to test if multiple oidc flows (same provider / different provider) work when this feature is used. It's quite possible that we already have these tests.

selfservice/strategy/oidc/strategy.go
```diff
@@ -291,29 +246,47 @@ func (s *Strategy) validateFlow(ctx context.Context, r *http.Request, rid uuid.U
...
-	f, err := s.validateFlow(r.Context(), r, x.ParseUUID(state.FlowID))
+	f, err = s.validateFlow(r.Context(), r, x.ParseUUID(state.FlowID))
 	if err != nil {
-		return nil, nil, err
+		return nil, "", nil, err
+	}
+
+	// Determine the provider from the flow context or the URL.
+	providerID = providerFromFlow(f)
+	urlParams, _ := r.Context().Value(httprouter.ParamsKey).(httprouter.Params)
```
aeneasr, line 271: Please pass this as an argument instead

selfservice/strategy/oidc/strategy.go
```diff
@@ -390,81 +363,80 @@ func (s *Strategy) alreadyAuthenticated(w http.ResponseWriter, r *http.Request,
...
 		err  error
 	)
 
-	ctx := context.WithValue(r.Context(), httprouter.ParamsKey, ps)
 	ctx, span := s.d.Tracer(ctx).Tracer().Start(ctx, "strategy.oidc.ExchangeCode")
 	defer otelx.End(span, &err)
 	r = r.WithContext(ctx)
 
-	req, cntnr, err := s.ValidateCallback(w, r)
+	f, pid, cntnr, err := s.ValidateCallback(w, r)
```
aeneasr, line 375: Pass `ps` here

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -173,8 +173,11 @@ func (s *Strategy) processLogin(w http.ResponseWriter, r *http.Request, loginFlo
 	}
 
 	sess := session.NewInactiveSession()
-	sess.CompletedLoginForWithProvider(s.ID(), identity.AuthenticatorAssuranceLevel1, provider.Config().ID,
-		httprouter.ParamsFromContext(r.Context()).ByName("organization"))
+	orgID := ""
```
aeneasr, line 176: This change looks incorrect? Previously we fetched this from the http route and now we use it from the loginFlow (but do we know if it is correctly set there?)

selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -245,7 +259,7 @@ func (s *Strategy) Register(w http.ResponseWriter, r *http.Request, f *registrat
 	return errors.WithStack(flow.ErrCompletedByStrategy)
 }
 
-func (s *Strategy) registrationToLogin(w http.ResponseWriter, r *http.Request, rf *registration.Flow, providerID string) (*login.Flow, error) {
+func (s *Strategy) registrationToLogin(w http.ResponseWriter, r *http.Request, rf *registration.Flow) (*login.Flow, error) {
```
aeneasr, line 262: Does this method keep working with PKCE or do we somehow lose track of PKCE? To test it, you could create an e2e test:

1. For an PKCE oauth2 provider with pkce force
2. Sign in with account A which does not exit
3. Kratos asks to sign up
4. Sign up with account A
5. Be logged in
6. Sign out
7. Perform sign up with account A
8. Expect log in
9. Log in with account A
10. Be signed in

If you want to be extra sure add validation errors in between (e.g. incomplete profile data from oidc)

selfservice/strategy/oidc/strategy_registration.go
```diff
@@ -381,9 +395,7 @@ func (s *Strategy) createIdentity(w http.ResponseWriter, r *http.Request, a *reg
 		return nil, nil, s.handleError(w, r, a, provider.Config().ID, i.Traits, err)
 	}
 
-	if orgID := httprouter.ParamsFromContext(r.Context()).ByName("organization"); orgID != "" {
-		i.OrganizationID = uuid.NullUUID{UUID: x.ParseUUID(orgID), Valid: true}
-	}
+	i.OrganizationID = a.OrganizationID
```
aeneasr, line 398: How are you certain that this is not a regression?

selfservice/strategy/oidc/strategy_settings.go
```diff
@@ -390,7 +396,21 @@ func (s *Strategy) initLinkProvider(w http.ResponseWriter, r *http.Request, ctxU
...
+
+	if err := storeProvider(ctxUpdate.Flow, p.Link); err != nil {
+		return s.handleError(w, r, ctxUpdate.Flow, p.Link, nil, errors.WithStack(herodot.ErrInternalServerError.WithReason("Could not update flow context").WithDebug(err.Error())))
+	}
+
+	if err := s.d.SettingsFlowPersister().UpdateSettingsFlow(ctx, ctxUpdate.Flow); err != nil {
+		return s.handleError(w, r, ctxUpdate.Flow, p.Link, nil, errors.WithStack(herodot.ErrInternalServerError.WithReason("Could not update flow").WithDebug(err.Error())))
+	}
+
+	codeURL, err := getAuthRedirectURL(ctx, provider, ctxUpdate.Flow, state, up)
```
aeneasr, line 413: `ctxUpdate.Flow` does not represent the most up-to-date version of the flow. Use the return value of `_, err = s.validateFlow(ctx, r, ctxUpdate.Flow.ID)` instead (like it was before)

selfservice/strategy/oidc/strategy_test.go
```diff
@@ -470,6 +479,67 @@ func TestStrategy(t *testing.T) {
 		return id
 	}
 
+	t.Run("case=force PKCE", func(t *testing.T) {
```
aeneasr, line 482: Here you can conveniently add tests that check if more complex flows also pass:

1. Sign in with wrong provider, then try again with another
2. Does PKCE actually get validated? Expect an error case
3. Does the automatic detection of whether to sign in or up with PKCE work as well?

aeneasr (changes requested):
Good stuff! I'm primarily concerned:

1. That the internal context gets correctly passed around and we don't lose track of the PKCE challenge or provider ID
2. That the organization flag is still working correctly
3. That we're returning the wrong flow in one instance (which could lead to some errors)
4. That the generic callback keeps track of the currently active provider ID across multiple login / sign up attempts

Please also add e2e tests verifying that PKCE works end-to-end, and that the generic callback also works correctly.

aeneasr:
Since a few other changes landed, there are a few conflicts now. Probably mostly beacuse a lot of functions got contexts now.

selfservice/strategy/oidc/strategy.go
aeneasr: Is it not possible to pass this value as an argument? The method is only used once afaik:

```
ValidateCallback(w http.ResponseWriter, r *http.Request, params httprouter.Params)
```

selfservice/strategy/oidc/strategy_helper_test.go
```diff
@@ -76,19 +76,43 @@ func (token *idTokenClaims) MarshalJSON() ([]byte, error) {
...
+			// client_secret_basic, but the token exchange fails for another
+			// reason, the error message will be swallowed and replaced with
+			// "invalid_client".
+			//
+			// Manually setting this to client_secret_post means that during
+			// tests, all token exchanges will first fail with `invalid_client`
+			// and then be retried with the correct method. This is the only way
+			// to get the actual error message from the server, however.
+			//
+			// https://github.com/golang/oauth2/blob/5fd42413edb3b1699004a31b72e485e0e4ba1b13/internal/token.go#L227-L242
```
aeneasr, line 114: Yeah, this is super annoying when trying to debug oauth2 related calls with the stdlib.

x/swagger/swagger_types_global.go
```diff
@@ -23,10 +20,7 @@ type ErrorGeneric struct {
...
 
-// Empty responses are sent when, for example, resources are deleted. The HTTP status code for empty responses is typically 201.
+// Empty responses are sent when, for example, resources are deleted. The HTTP status code for empty responses is typically 204.
 //
 // swagger:response emptyResponse
-//
-//nolint:deadcode,unused
-//lint:ignore U1000 Used to generate Swagger and OpenAPI definitions
-type emptyResponse struct{}
+type _ struct{}
```
aeneasr, line 26: unaddressable names don't work with go-swagger afaik, but I might be wrong (or it has been fixed). Easiest way is to run make sdk and check the openapi diff

alnr:
Yep, I'm rebasing.

x/swagger/swagger_types_global.go
alnr: Works!

selfservice/strategy/oidc/strategy.go
alnr: Yes, it's possible. Will address.

OskarsPakers:
> **Important**: When setting `pkce: force`, you must whitelist a different return URL for your OAuth2 client in the provider's configuration. Instead of `<base-url>/self-service/methods/oidc/callback/<provider>`, you must use `<base-url>/self-service/methods/oidc/callback` (note missing last path segment). This is to enable the use of the same OAuth client ID+secret when configuring several Kratos OIDC providers, without having to whitelist individual `redirect_uri`s for each Kratos provider config.

Would the regular callback with provider id in url work too?
If not, This way it would be mean that toggling the setting requires registering new redirect url. That would not be great.

alnr:
> > **Important**: When setting `pkce: force`, you must whitelist a different return URL for your OAuth2 client in the provider's configuration. Instead of `<base-url>/self-service/methods/oidc/callback/<provider>`, you must use `<base-url>/self-service/methods/oidc/callback` (note missing last path segment). This is to enable the use of the same OAuth client ID+secret when configuring several Kratos OIDC providers, without having to whitelist individual `redirect_uri`s for each Kratos provider config.
> 
> Would the regular callback with provider id in url work too? If not, This way it would be mean that toggling the setting requires registering new redirect url. That would not be great.

Having the generic redirect URI forces PKCE for security reasons. If you want to keep your redirect URI as-is, you'll still benefic from automatic PKCE if the provider advertises it.

selfservice/strategy/oidc/strategy_test.go
alnr: 1. We already have a bunch of these tests in the context of login hints/account linking. I've added a test which checks a mismatch between providers in URL and state is correctly detected and the flow aborted.
2. Added tests.
3. Test is already there.

selfservice/strategy/oidc/strategy_registration.go
alnr: Yes, but it did end up changing the code to get the correct OrgID directly from the provider configuration.

selfservice/strategy/oidc/strategy_settings.go
alnr: done

selfservice/strategy/oidc/strategy.go
alnr: done

selfservice/strategy/oidc/strategy_login.go
alnr: I ended up simplifying the original code and take the OrgID from the provider config directly, instead of looking at the current URL.

selfservice/strategy/oidc/strategy_registration.go
alnr: All this should continue working as before, because it's orthogonal to PKCE from what I can see in the code.

Whenever we move between flows, the OAuth2 token exchange has already taken place, so all PKCE-related stuff is already done.

Essentially, we store the verifier directly before redirecting the end user, and verify it right after he comes back (during token exchange). There are no cross-flow (`registrationToLogin` or the other way around) shenanigans to get in the way, I don't think.

alnr:
Superseeded by https://github.com/ory/kratos/pull/4078

PR #4086, opened 2024-09-10
docs: add return_to query parameter to OAS Verification Flow for Native Apps

David-Wobrock:
The idea would be to re-generate the Kratos SDK, so that one can specify a `return_to` URL when creating a Native Verification Flow using the SDK.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/flow/verification/handler.go
+++ b/selfservice/flow/verification/handler.go
@@ -123,6 +123,20 @@ func (h *Handler) NewVerificationFlow(w http.ResponseWriter, r *http.Request, ft
 	return f, nil
 }
 
+// Create Verification Flow Parameters for Native Apps
+//
+// swagger:parameters createNativeVerificationFlow
+//
+//nolint:deadcode,unused
+//lint:ignore U1000 Used to generate Swagger and OpenAPI definitions
+type createNativeVerificationFlow struct {
+	// A URL contained in the return_to key of the verification flow.
+	// This piece of data has no effect on the actual logic of the flow and is purely informational.
+	//
+	// in: query
+	ReturnTo string `json:"return_to"`
+}
+
 // swagger:route GET /self-service/verification/api frontend createNativeVerificationFlow
 //
 // # Create Verification Flow for Native Apps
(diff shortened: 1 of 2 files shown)


selfservice/flow/verification/handler.go
```diff
@@ -123,6 +123,19 @@ func (h *Handler) NewVerificationFlow(w http.ResponseWriter, r *http.Request, ft
...
+// Create Verification Flow Parameters for Native Apps
+//
+// swagger:parameters createNativeVerificationFlow
+//
+//nolint:deadcode,unused
+//lint:ignore U1000 Used to generate Swagger and OpenAPI definitions
+type createNativeVerificationFlow struct {
+	// The URL to return after the flow was completed.
+	//
+	// in: query
```
aeneasr, line 135: Does this actually work? Native flows don’t use redirects afaik?

aeneasr (changes requested):
I checked and the return_to parameter is not being used anywhere in the verification flow, so this change has no effect. We would instead need to implement return_to support.

What's your use case?

selfservice/flow/verification/handler.go
David-Wobrock: Not redirections _per se_, but this piece of information is contained in the Verification Flow.

For instance calling `http://127.0.0.1:4433/self-service/verification/api?return_to=http://127.0.0.1:4433/ok` would returns this flow content:
```
{
    "id": "689e20e6-f66c-4bcb-996c-d1f3c87e7670",
    "type": "api",
    "expires_at": "2024-09-19T14:52:43.434168067Z",
    "issued_at": "2024-09-19T13:52:43.434168067Z",
    "request_url": "http://127.0.0.1:4433/self-service/verification/api?return_to=http://127.0.0.1:4433/ok",
    "return_to": "http://127.0.0.1:4433/ok",
    "active": "code",
[...]
```

With a `return_to` key, which we can be used in our logic to know the redirection URL set at the beginning of the flow :)

David-Wobrock: I rephrased the sentence.
Let me know what you think :)

aeneasr (approved):

PR #4111, opened 2024-09-18
fix: batch identity created event

hperl:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


persistence/sql/identity/persister_identity.go
```diff
@@ -645,32 +645,37 @@ func (p *IdentityPersister) CreateIdentities(ctx context.Context, identities ...
...
 			// error occurred.
 			if err := p.DeleteIdentities(ctx, failedIDs); err != nil {
 				return sqlcon.HandleError(err)
 			}
-			// Wrap the partial error with the first error that occurred, so that the caller
-			// can continue to handle the error either as a partial error or a full error.
+
+			// Partial success: report succeeded identities as created.
+			for _, identID := range succeededIDs {
+				span.AddEvent(events.NewIdentityCreated(ctx, identID))
```
zepatrik, line 666: I don't think that we should emit the events in the transaction, as it might be retried?

aeneasr: Good question - will a retry in the transaction cause this event to be duplicated?

aeneasr: I think so, it will trace the event two times, which triggers to sends. So I agree, this should be outside of the transaction (like it was before)

alnr: Yes, this must be outside the transaction block. Could use a `defer`.

hperl: makes sense, thanks!

alnr (approved):

zepatrik (approved):

PR #4113, opened 2024-09-19
feat: delegate SMS code generation, sending and verification to an external service (PS-221)

splaunov:
Phone/sms registration and login flows are vulnerable to sms pumping attacks. 
This PR allows to delegate all the key steps of the flow to an external service which is able to detect sms pumping and throttle requests.

## Checklist

      introduces a new feature.
- [x ] I am following the
      the changes.
      works.


aeneasr:
Looks like your branch and master have divereged. We recently merged improvements to SMS which probably cause the conflicts.

splaunov:
Yes, we are a bit behind the master.
I can rebase it. But it only makes sense if you are about to accept this feature.
What would be your stance on this?

aeneasr:
The problem described is definitely something we should consider. Regarding the approach, I'm not sure if I fully understand how it works, but it does look a bit "strapped on" from first glance, with hardcoded values and some other things, which make me question if it wouldn't be better to use hooks and transient payloads for example.

PR #4115, opened 2024-09-22
feat: allow listing identities by organization ID

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


identity/handler.go
```diff
@@ -171,6 +171,12 @@ type listIdentitiesParameters struct {
 	// in: query
 	DeclassifyCredentials []string `json:"include_credential"`
 
+	// OrganizationID is the organization id to filter identities by.
+	//
+	// If `ids` is set, this parameter is ignored.
+	// required: false
```
hperl, line 177: Missing `in: query`?

identity/handler.go
```diff
@@ -211,6 +217,7 @@ func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Para
 			IdsFilter:                    r.URL.Query()["ids"],
 			CredentialsIdentifier:        r.URL.Query().Get("credentials_identifier"),
 			CredentialsIdentifierSimilar: r.URL.Query().Get("preview_credentials_identifier_similar"),
+			OrganizationID:               x.ParseUUID(r.URL.Query().Get("organization_id")),
```
hperl, line 220: Here, an invalid UUID will silently be treated as if the parameter was not set at all. This might be hard to debug.

I'd prefer a `400 Bad Request` to be returned if the `organization_id` parameter was set, but the value could not be parsed into an UUID. WDYT?

hperl (changes requested):
This makes a lot of sense. I just have one comment regarding parsing the org UUID; and some tests are still missing:

- Org ID set to a valid UUID: filters
- Org ID set to an invalid UUID: 400 Bad Request

Not sure what we want to do if the Org ID points to an org that does not exist; but I think the default behaviour (just an empty list of identitites) is fine.

identity/handler.go
jonas-jonas: Sure, SGTM.

We don't do any of this validation in the `ids` parameter, either. In fact, we don't even parse the IDs as UUIDs, and instead pass them straight to the query. That means, in practice, that an invalid UUID will return a 500 Internal Server Error. Fix for that is here: https://github.com/ory/kratos/pull/4116

identity/handler.go
jonas-jonas: Resolved

identity/handler.go
hperl: Nice catch, approved :)

jonas-jonas:
@aeneasr added.

hperl (approved):

aeneasr (approved):

aeneasr:
Index name is conflicting: https://github.com/ory/kratos/actions/runs/10992644668/job/30524759829?pr=4115#step:14:7531

persistence/sql/migrations/sql/20240923095000000001_organization_id_index.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX identities_nid_organization_id_idx ON identities (nid, organization_id);
```
alnr, line 1: This should be `... identities (organization_id,created_at);`

`organization_id` is sharded by NID anyway, so we can reduct index size here.

alnr: Maybe `(organization_id,id)`. I thought we list be created date?

alnr: We might need an index hint here, too.

aeneasr: indices sort their keys anyways, so no need to add id for sorting to the index

PR #4119, opened 2024-09-24
feat: change `method=profile:back` to `screen=previous`

aeneasr:
BREAKING CHANGE: When using two-step registration, it was previously possible to send `method=profile:back` to get to the previous screen. This feature was not documented in the SDK API yet. Going forward, please instead use `screen=previous`.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


selfservice/strategy/profile/two_step_registration.go
```diff
@@ -194,8 +205,8 @@ func (s *Strategy) displayStepTwoNodes(ctx context.Context, w http.ResponseWrite
 	regFlow.UI.Messages.Add(text.NewInfoSelfServiceChooseCredentials())
 
 	regFlow.UI.Nodes.Append(node.NewInputField(
-		"method",
-		"profile:back",
+		"action",
+		"back",
```
aeneasr, line 209: This is really the only potential breaking change. If someone relies on `profile:back` being available, then this breaks. I checked in Ory Elements where there is no reference on this. In Ory Network neither. It was also not documented. So should be fine, hopefully?

mszekiel:
Would it be possible to additionally unify the behavior between login and registration? Right now login provides `identifier_first` node which has similar role than the `profile:back` if I understand that correctly. 

But I believe this is some amount of additional work.

selfservice/strategy/profile/two_step_registration.go
hperl: I think this should be fine. I invented it for two-step logins, so it really should only have been referenced inside Kratos.

aeneasr:
> Would it be possible to additionally unify the behavior between login and registration? Right now login provides `identifier_first` node which has similar role than the `profile:back` if I understand that correctly.

That is kind of correct, I wanted to add a back button but decided against it due to time constraints, and used the identifier_first node as sort of a "workaround". But it's not good, because it adds an implicit dependency on this node.

I'm not entirely sure how to best handle this. Ideally we would have something like `login_hint` in the login flow object that is the identifier of the user, and maybe add a back button to identifier first auth.

@jonas-jonas wdyt?

PR #4131, opened 2024-09-30
fix: explicity set updated_at field when updating identity

maoanran:
### Problem

The updated_at field in the identities table is not being updated when an identity's traits are modified. This is unexpected behavior because the pop library should automatically update the `updated_at` field when a record is modified.

However, in the current implementation, the identity is being saved with the old `updated_at` timestamp, which causes the pop library to skip updating this field. As a result, the `updated_at` field does not reflect recent changes, which can lead to issues with auditing, synchronization, or downstream systems relying on the correct timestamps for updated data.

### Cause

The issue arises because the existing code inadvertently preserves the previous updated_at value during the update operation. When saving the identity, the pop library sees that the field is already set and, as a result, does not overwrite it with the current timestamp.

### Solution

This PR explicitly sets the `updated_at` field when updating an identity. By ensuring the field is updated within the code, we force the correct behavior, even if the underlying pop library does not update the field automatically.

### Steps to manually reproduce:

1.  Update the traits of a user.
2. Check the `updated_at` field in the identities table to confirm that it reflects the recent update.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/identity/manager_test.go
+++ b/identity/manager_test.go
@@ -677,6 +677,22 @@ func TestManager(t *testing.T) {
 			// That is why we only check the identity in the store.
 			checkExtensionFields(fromStore, "email-updatetraits-1@ory.sh")(t)
 		})
+
+		t.Run("case=should always update updated_at field", func(t *testing.T) {
+			original := identity.NewIdentity(config.DefaultIdentityTraitsSchemaID)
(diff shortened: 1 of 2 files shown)


aeneasr (approved):
Makes sense!

PR #4138, opened 2024-10-02
feat: change primary key on identity_verifiable_addresses and identit…

alnr:
…y_recovery_addresses

### DIFF
--- a/persistence/sql/migrations/gomigrations/identity_pk.go
+++ b/persistence/sql/migrations/gomigrations/identity_pk.go
@@ -1,0 +1,71 @@
+// Copyright © 2024 Ory Corp
+// SPDX-License-Identifier: Apache-2.0
+
+package gomigrations
+
+import (
+	"fmt"
+	"path/filepath"
+	"runtime"
+
+	"github.com/gobuffalo/pop/v6"
+	"github.com/pkg/errors"
+
+	"github.com/ory/x/popx"
+)
+
+func path() string {
+	_, file, line, _ := runtime.Caller(1)
+	return fmt.Sprintf("%s:%d", filepath.Base(file), line)
+}
+
+var ChangeAddressesPK = []popx.Migration{
+	{
+		Version:   "20241001000000000000",
+		Path:      path(),
+		Name:      "Change primary key for identity_verifiable_addresses",
+		Direction: "up",
+		Type:      "go",
+		DBType:    "cockroach",
+		RunnerNoTx: func(m popx.Migration, c *pop.Connection) error {
+			_, err := c.Store.Exec("ALTER TABLE identity_verifiable_addresses ALTER PRIMARY KEY USING COLUMNS (identity_id,id)")
+			return errors.WithStack(err)
+		},
+	},
+	{
+		Version:   "20241001000000000000",
+		Path:      path(),
+		Name:      "Revert primary key for identity_verifiable_addresses",
+		Direction: "down",
+		Type:      "go",
+		DBType:    "cockroach",
+		RunnerNoTx: func(m popx.Migration, c *pop.Connection) error {
+			_, err := c.Store.Exec("ALTER TABLE identity_verifiable_addresses ALTER PRIMARY KEY USING COLUMNS (id)")
+			return errors.WithStack(err)
+		},
+	},
+	{
+		Version:   "20241001000000000001",
+		Path:      path(),
+		Name:      "Change primary key for identity_recovery_addresses",
+		Direction: "up",
+		Type:      "go",
+		DBType:    "cockroach",
+		RunnerNoTx: func(m popx.Migration, c *pop.Connection) error {
+			_, err := c.Store.Exec("ALTER TABLE identity_recovery_addresses ALTER PRIMARY KEY USING COLUMNS (identity_id,id)")
+			return errors.WithStack(err)
+		},
+	},
+	{
+		Version:   "20241001000000000001",
+		Path:      path(),
+		Name:      "Revert primary key for identity_recovery_addresses",
+		Direction: "down",
+		Type:      "go",
+		DBType:    "cockroach",
+		RunnerNoTx: func(m popx.Migration, c *pop.Connection) error {
+			_, err := c.Store.Exec("ALTER TABLE identity_recovery_addresses ALTER PRIMARY KEY USING COLUMNS (id)")
+			return errors.WithStack(err)
+		},
+	},
+}
--- a/persistence/sql/migrations/sql/20210817181232000000_unique_credentials.sqlite3.down.sql
+++ b/persistence/sql/migrations/sql/20210817181232000000_unique_credentials.sqlite3.down.sql
@@ -1,1 +1,2 @@
+DELETE FROM "identity_credential_identifiers"; -- This migration is destructive.
 CREATE UNIQUE INDEX "identity_credential_identifiers_identifier_nid_uq_idx" ON "identity_credential_identifiers" (nid, identifier);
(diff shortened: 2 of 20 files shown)


persistence/sql/migrations/gomigrations/identity_pk.go
```diff
@@ -0,0 +1,71 @@
...
+var ChangeAddressesPK = []popx.Migration{
+	{
+		Version:   "20241001000000000000",
+		Path:      path(),
+		Name:      "Change primary key for identity_verifiable_addresses",
+		Direction: "up",
+		Type:      "go",
+		DBType:    "cockroach",
+		RunnerNoTx: func(m popx.Migration, c *pop.Connection) error {
+			_, err := c.Store.Exec("ALTER TABLE identity_verifiable_addresses ALTER PRIMARY KEY USING COLUMNS (identity_id,id)")
```
aeneasr, line 31: Move to SQL files since we don't need go for these?

alnr: Currently this is the only way to run crdb migrations outside of transactions

aeneasr: I see!

persistence/sql/migrations/sql/20210817181232000000_unique_credentials.sqlite3.down.sql
```diff
@@ -1 +1,2 @@
-CREATE UNIQUE INDEX "identity_credential_identifiers_identifier_nid_uq_idx" ON "identity_credential_identifiers" (nid, identifier);
```
aeneasr, line 1: We usually do not change existing migrations. This is fine because it's only SQLite, but was this on purpose?

persistence/sql/migrations/sql/20241001000000000000_identity_verifiable_addresses_pk.mysql.down.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE identity_verifiable_addresses
+    DROP FOREIGN KEY identity_verifiable_addresses_ibfk_1;
```
aeneasr, line 2: I guess we need this because the primary key won't be deleted if there is still a foreign key using it, correct?

persistence/sql/migrations/sql/20241001000000000000_identity_verifiable_addresses_pk.postgres.up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE UNIQUE INDEX identity_verifiable_addresses_id_uq_idx ON identity_verifiable_addresses (id);
+
+ALTER TABLE identity_verification_codes
+    DROP CONSTRAINT identity_verification_codes_identity_verifiable_addresses_id_fk;
+
+ALTER TABLE identity_verification_tokens
+    DROP CONSTRAINT identity_verification_tokens_identity_verifiable_address_i_fkey;
```
aeneasr, line 7: Do these need to be dropped and recreated?

persistence/sql/migrations/sql/20241001000000000000_identity_verifiable_addresses_pk.mysql.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE identity_verifiable_addresses
```
aeneasr, line 1: Isn't this missing a unique index for id?

persistence/sql/migrations/sql/20241001000000000001_identity_recovery_addresses_pk.mysql.down.sql
```diff
@@ -0,0 +1,24 @@
+ALTER TABLE identity_recovery_codes
+    DROP FOREIGN KEY identity_recovery_codes_identity_recovery_addresses_id_fk,
+    DROP FOREIGN KEY identity_recovery_codes_identity_id_fk;
+
+ALTER TABLE identity_recovery_tokens
+    DROP FOREIGN KEY identity_recovery_tokens_identity_id_fk,
+    DROP FOREIGN KEY identity_recovery_tokens_identity_recovery_addresses_id_fk;
```
aeneasr, line 7: Do we need to remove/recreate the child's foreign key if we change a constraint in the other table?

persistence/sql/migrations/sql/20241001000000000000_identity_verifiable_addresses_pk.mysql.down.sql
alnr: exactly.

persistence/sql/migrations/sql/20241001000000000000_identity_verifiable_addresses_pk.mysql.up.sql
alnr: added!

persistence/sql/migrations/sql/20241001000000000001_identity_recovery_addresses_pk.mysql.down.sql
alnr: This dance is only necessary on MySQL.

persistence/sql/migrations/sql/20210817181232000000_unique_credentials.sqlite3.down.sql
alnr: This may be necessary later for the PK change on identity_credential_identifiers, but it is not relevant in this PR. Removed.

persistence/sql/migrations/sql/20241001000000000000_identity_verifiable_addresses_pk.postgres.up.sql
alnr: Yes:
```
ERROR: cannot drop constraint identity_recovery_addresses_pkey on table identity_recovery_addresses because other objects depend on it (SQLSTATE 2BP01)
```

persistence/sql/migrations/sql/20241001000000000000_identity_verifiable_addresses_pk.postgres.up.sql
```diff
@@ -0,0 +1,17 @@
...
+
+ALTER TABLE identity_verifiable_addresses
+    DROP CONSTRAINT identity_verifiable_addresses_pkey,
+    ADD PRIMARY KEY (identity_id, id);
+
+ALTER TABLE identity_verification_codes
+    ADD CONSTRAINT identity_verification_codes_identity_verifiable_addresses_id_fk FOREIGN KEY (identity_verifiable_address_id) REFERENCES identity_verifiable_addresses(id) ON DELETE CASCADE;
+
+ALTER TABLE identity_verification_tokens
+    ADD CONSTRAINT identity_verification_tokens_identity_verifiable_address_i_fkey FOREIGN KEY (identity_verifiable_address_id) REFERENCES identity_verifiable_addresses(id) ON DELETE CASCADE;
```
zepatrik, line 17: As the PK changed, the foreign key should also change.
```suggestion
ALTER TABLE identity_verification_codes
    ADD CONSTRAINT identity_verification_codes_identity_verifiable_addresses_id_fk FOREIGN KEY (identity_verifiable_address_id) REFERENCES identity_verifiable_addresses(identity_id, id) ON DELETE CASCADE;

ALTER TABLE identity_verification_tokens
    ADD CONSTRAINT identity_verification_tokens_identity_verifiable_address_i_fkey FOREIGN KEY (identity_verifiable_address_id) REFERENCES identity_verifiable_addresses(identity_id, id) ON DELETE CASCADE;
```

zepatrik (changes requested):
I think that the FKs should also become composite, so we don't need an additional unique index that serves no real purpose.

PR #4139, opened 2024-10-02
feat: change primary key on identity_verifiable_addresses and identity_recovery_addresses

aeneasr:
Moves to SQL migrations


persistence/sql/migrations/sql/20241001000000000000_identity_verifiable_addresses_pk.mysql.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE identity_verifiable_addresses
```
aeneasr, line 1: Why does mysql not have a unique index for the id?

PR #4146, opened 2024-10-08
fix: use non-problematic filesystem merge lib

zepatrik:
We know that the previous lib (https://github.com/laher/mergefs) cannot properly handle recursively nesting filesystems, so we should avoid using it anywhere.
see https://github.com/ory/x/pull/709

### DIFF
--- a/driver/registry.go
+++ b/driver/registry.go
@@ -5,7 +5,6 @@ package driver
 
 import (
 	"context"
-	"io/fs"
 
 	"github.com/ory/kratos/selfservice/sessiontokenexchange"
 	"github.com/ory/x/contextx"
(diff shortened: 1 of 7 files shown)


driver/registry.go
```diff
@@ -245,12 +243,6 @@ func Inspect(f func(reg Registry) error) RegistryOption {
 	}
 }
 
-func WithExtraMigrations(m ...fs.FS) RegistryOption {
```
zepatrik, line 248: Actually an unused option.

alnr:
I have touched this code in #4138, can we get that over the line first?
Also, the `ExtraMigrations` option I would like to keep. Pretty sure that'll be useful in the future.

zepatrik:
Actually superseded by #4138

Issue #4147, opened 2024-10-08
Yandex/vk oidc: prefill phone number in claims


renom:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Describe your problem

Yandex and vk use non-default field for phone number, so it should be mapped:
* Yandex: `default_phone.number` => `phone_number`
* VK: `phone` => `phone_number`

### Describe your ideal solution

Update mapping:
* Yandex: https://github.com/ory/kratos/blob/master/selfservice/strategy/oidc/provider_yandex.go#L82-L91
* VK: https://github.com/ory/kratos/blob/master/selfservice/strategy/oidc/provider_vk.go#L84-L93

### Workarounds or alternatives

-

### Version

1.3.0

Denish3436:
hi @renom could you assign me, i would like to work on this

PR #4158, opened 2024-10-14
feat: update Yandex and VK OIDC

renom:
### VK provider: Support PKCE

It seems like VK doesn't provide `/.well-known/openid-configuration` (VK doesn't support OpenID).
So to use PKCE it's needed to be enabled with:
```
pkce: "force"
```

### VK provider: Support returning `phone_number` in `claims`

VK offers different endpoints for PKCE and non-PKCE configuration.
Token endpoint that is used in non-PKCE configuration can return `email` but not `phone`.
Extra endpoint `https://id.vk.com/oauth2/user_info` can be used to retrieve `email` and `phone`, but it doesn't accept `access_token` that was returned from non-PKCE token endpoint.
So to retrieve a phone, PKCE is required to be enabled:
```diff
  - id: vk
    provider: vk
    mapper_url: "data-mapper.jsonnet"
+   pkce: "force"
    scope:
      - email
+     - phone
```

Additional info:
I added `AccessTokenURLOptions(r *http.Request) []oauth2.AuthCodeOption` to [OAuth2Provider](https://github.com/renom/kratos/blob/update-yandex-and-vk-oidc/selfservice/strategy/oidc/provider.go#L30) interface.
It works like `AuthCodeURLOptions(r ider) []oauth2.AuthCodeOption` but it's dedicated to update a token URL.
VK requires passing `device_id` URL param (that is received by a callback) to PKCE token endpoint. Without this change it's impossible.
I updated all the other OIDC providers to comply with the updated interface.

### Yandex provider: Support returning `phone_number` in `claims`

No extra configuration is needed.

## Related issue(s)

Fixes: #4147

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

* Everything is tested manually (including backward compatibility).
* I didn't ask about approving the changes in Slack (please let me know if I need to).
* VK seems to have removed the docs about non-PKCE endpoints, so PKCE endpoints is probably the recommended way to use OIDC.

### DIFF
--- a/selfservice/strategy/oidc/provider.go
+++ b/selfservice/strategy/oidc/provider.go
@@ -27,6 +27,7 @@ type Provider interface {
 type OAuth2Provider interface {
 	Provider
 	AuthCodeURLOptions(r ider) []oauth2.AuthCodeOption
+	AccessTokenURLOptions(r *http.Request) []oauth2.AuthCodeOption
 	OAuth2(ctx context.Context) (*oauth2.Config, error)
 	Claims(ctx context.Context, exchange *oauth2.Token, query url.Values) (*Claims, error)
 }
--- a/selfservice/strategy/oidc/provider_vk.go
+++ b/selfservice/strategy/oidc/provider_vk.go
@@ -6,8 +6,11 @@ package oidc
 import (
 	"context"
 	"encoding/json"
+	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
+	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
 
@@ -18,6 +21,8 @@ import (
 
 	"github.com/ory/herodot"
 )
+
+var _ OAuth2Provider = (*ProviderVK)(nil)
 
 type ProviderVK struct {
 	config *Configuration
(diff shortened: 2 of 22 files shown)


selfservice/strategy/oidc/provider_vk.go
```diff
@@ -106,8 +130,45 @@ func (g *ProviderVK) Claims(ctx context.Context, exchange *oauth2.Token, query u
...
+				Phone string `json:"phone,omitempty"`
+			} `json:"user,omitempty"`
+		}
+
+		if err := json.NewDecoder(resp.Body).Decode(&userInfo); err != nil {
+			return nil, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("%s", err))
+		}
+
+		if len(userInfo.User.Phone) > 0 && userInfo.User.Phone[0] != '+' {
+			userInfo.User.Phone = "+" + userInfo.User.Phone
```
zepatrik, line 163: 🤔 I'm not sure if this is the correct normalization to do here. I quickly checked VK docs, but couldn't find anything about the format.
If it is not clear, I'd prefer to do nothing instead and leave it to the consumer of the data.

zepatrik (commented):
Looks very good! Just one small question.

selfservice/strategy/oidc/provider_vk.go
renom: All I've got is two examples from the docs, as well as a confirmation that it works that way (I've tested on my VK account that is registered with a Russian phone number). I didn't find any detailed description of the format either.
The examples from the docs:
1. [Example 1](https://id.vk.com/about/business/go/docs/ru/vkid/latest/vk-id/connection/api-integration/api-description#Poluchenie-nemaskirovannyh-dannyh):
```jsonc
{
    "user": {
        "user_id": "1234567890", // Идентификатор пользователя.
        "first_name": "Ivan", // Имя пользователя.
        "last_name": "I.", // Первая буква фамилии пользвателя.
        "phone": "79903922415", // Телефон пользователя.
        "avatar": "https://pp.userapi.com/60tZWMo4SmwcploUVl9XEt8ufnTTvDUmQ6Bj1g/mmv1pcj63C4.png", // Ссылка на фото профиля пользователя.
        "email": "ivan_i123@vk.com", // Адрес почты пользователя.
        "sex": 2, // Пол. Возможные значения: Возможные значения: 1 — женский, 2 — мужской, 0 — пол не указан.
        "is_verified": false, // Признак того, что аккаунт пользователя подтвержден. Если это так, возвращается параметр true.
        "birthday": "01.01.2000" // Дата рождения пользователя.
    }
}
```
2. [Example 2](https://id.vk.com/about/business/go/docs/ru/vkid/latest/vk-id/connection/start-integration/auth-without-sdk/auth-without-sdk-web#Shag-8-Opcionalno-Poluchenie-polzovatelskih-dannyh):
```jsonc
{
    "user": {
        "user_id": "1234567890",
        "first_name": "Ivan",
        "last_name": "I.",
        "phone": "79991234567",
        "avatar": "http://avatar.com/12345678",
        "email": "ivan_i123@vk.com",
        "sex": 2,
        "verified": false,
        "birthday": "01.01.2000"
    }
}
```

`79903922415` and `79991234567` are Russian phone numbers in the international format, but without a leading `+` sign.
As far as I understand, Ory Kratos uses international format, but with a leading `+` sign.
I'm not sure whether all the phones are returned in the same format or not. For example I can't test with non-Russian phones. This is why I added condition `if there's no plus sign, then add it`.
If you insist, I can remove that code block (lines 162-164). It shouldn't be the problem to do the same action within a `.jsonnet` file though.

selfservice/strategy/oidc/provider.go
```diff
@@ -27,6 +27,7 @@ type Provider interface {
 type OAuth2Provider interface {
 	Provider
 	AuthCodeURLOptions(r ider) []oauth2.AuthCodeOption
+	AccessTokenURLOptions(r *http.Request) []oauth2.AuthCodeOption
```
hperl, line 30: As this method is only relevant for VK, we could also extract it into a separate interface that only the VK provider implements. Then, in `strategy.go` we assert on the interface and add the options if the provider implements it. 

That way, we don't have to give empty implementations to every other provider. WDYT?

selfservice/strategy/oidc/provider_vk.go
```diff
@@ -34,38 +39,56 @@ func NewProviderVK(
...
 }
 
-func (g *ProviderVK) oauth2(ctx context.Context) *oauth2.Config {
-	return &oauth2.Config{
-		ClientID:     g.config.ClientID,
-		ClientSecret: g.config.ClientSecret,
-		Endpoint: oauth2.Endpoint{
+func (p *ProviderVK) oauth2(ctx context.Context) *oauth2.Config {
+	var endpoint oauth2.Endpoint
+	if p.config.PKCE == "force" {
```
hperl, line 48: Is there a reason / use case for not using PKCE with VK? Ideally, we would use PKCE by default for all providers that support it.

The explicit config flag for PKCE should only be necessary for the generic provider.

hperl (changes requested):
LGTM as well! I just wonder if we can enable PKCE for VK by default?

selfservice/strategy/oidc/provider_vk.go
renom: I can enable PKCE for `pkce == "auto"` too. I thought `auto` should enable PKCE only if `/.well-known/openid-configuration` returns that it's supported, that's why I didn't enable PKCE for `auto`.
I don't know the reason why we shouldn't use PKCE. But enabling non-PKCE configuration when `pkce` is explicitly set to `never` seems logical.

renom:
Hi guys. Any chance it'll get merged?

afansv:
Are you sure that the VK ID provider is the same as VK? Maybe we should get a separate provider, so as not to mix them, and deprecate the old one.

selfservice/strategy/oidc/provider_vk.go
```diff
@@ -34,38 +39,56 @@ func NewProviderVK(
...
+		endpoint = oauth2.Endpoint{
+			AuthURL:  "https://id.vk.com/authorize",
+			TokenURL: "https://id.vk.com/oauth2/auth",
+		}
+	} else {
+		endpoint = oauth2.Endpoint{
 			AuthURL:  "https://oauth.vk.com/authorize",
```
afansv, line 55: What is the reason? oauth.vk.com is deprecated and no longer works (https://id.vk.ru/about/faq/business/vkid/OAuth/30024)

renom: Когда делал ещё работало.

PR #4173, opened 2024-10-24
refactor: optimize identity related indices

aeneasr:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


persistence/sql/identity/persister_identity.go
```diff
@@ -956,11 +956,11 @@ func (p *IdentityPersister) ListIdentities(ctx context.Context, params identity.
 			// important to normalize the identifier before querying the database.
 
 			joins = params.TransformStatement(`
-			INNER JOIN identity_credentials ic ON ic.identity_id = identities.id
-			INNER JOIN identity_credential_identifiers ici ON ici.identity_credential_id = ic.id`)
+			INNER JOIN identity_credentials ic ON ic.identity_id = identities.id AND ic.nid = identities.nid
```
aeneasr, line 959: * [ ] This will need an update in our overrides @arekkas

persistence/sql/migrations/sql/20241023142500000001_drop_unused_indices_identity_credentials.up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE INDEX identity_credentials_nid_identity_id_v2_idx ON identity_credentials (nid ASC, identity_id ASC);
```
aeneasr, line 1: I think we might be missing a secondary index that has identity_id as the first piece, which will sped up account deletion.

persistence/sql/migrations/sql/20241023142500000002_drop_unused_indices_sessions.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE INDEX sessions_nid_expires_at_authenticated_at_idx ON sessions (nid ASC, expires_at DESC, created_at DESC); -- Used for listing sessions
```
aeneasr, line 1: This is a bit problematic because we now have to order by `expires_at, created_at` to avoid reading all rows and use the index. This is a potential breaking change as we previously only ordered by created_at and authenticated_at.

This still needs to be reflected in the persister.

Alternatively, we could consider to add a second index just for `created_at` sorting.

persistence/sql/migrations/sql/20241023142500000003_drop_unused_indices_credential_identifiers.down.sql
```diff
@@ -0,0 +1,6 @@
+DROP INDEX identity_credential_identifiers_nid_i_ici_v2_idx;
+
+CREATE INDEX identity_credential_identifiers_nid_identity_credential_id_idx ON identity_credential_identifiers (identity_credential_id ASC, nid ASC);
+CREATE INDEX identity_credential_identifiers_nid_i_ici_idx ON identity_credential_identifiers (nid ASC, identifier ASC, identity_credential_id ASC);
```
alnr, line 4: I need this one :(

persistence/sql/persister_session.go
```diff
@@ -171,7 +163,7 @@ func (p *Persister) ListSessionsByIdentity(
 		}
 		t = int64(total)
 
-		q.Order("authenticated_at DESC")
+		q.Order("created_at DESC")
```
alnr, line 166: Is there no other way to do this? IMO, `authenticated_at` is a more useful sorting order. Plus this is breaking.

persistence/sql/migrations/sql/20241023142500000002_drop_unused_indices_sessions.up.sql
alnr: Why is `expires_at` in this index?

aeneasr: It's used in list sessions and listmysessions when active is set:

```
			if *active {
				q.Where("active = ? AND expires_at >= ?", *active, time.Now().UTC())
			} else {
				q.Where("(active = ? OR expires_at < ?)", *active, time.Now().UTC())
			}
```

So we have `nid, active, expires_at` as keys and I think we need all 3 parameters to find the smallest subset of rows.

persistence/sql/persister_session.go
aeneasr: created_at == authenticated_at with maybe ~300ms clock skew

alnr (approved):

PR #4174, opened 2024-10-28
fix: account linking should only happen after 2fa when required

zepatrik:
Currently account linking happens after the first factor, even when the second factor is required for a session to be issued.

### DIFF
--- a/selfservice/flow/login/hook.go
+++ b/selfservice/flow/login/hook.go
@@ -12,7 +12,6 @@ import (
 	"github.com/gofrs/uuid"
 	"github.com/pkg/errors"
 	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/trace"
 
 	"github.com/ory/kratos/driver/config"
 	"github.com/ory/kratos/hydra"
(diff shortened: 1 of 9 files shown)


selfservice/flow/login/hook.go
```diff
@@ -133,7 +133,11 @@ func (e *HookExecutor) PostLoginHook(
 	r = r.WithContext(ctx)
 	defer otelx.End(span, &err)
 
-	if err := e.maybeLinkCredentials(r.Context(), s, i, f); err != nil {
+	// We need to set the identity here because we check the available AAL in maybeLinkCredentials.
+	s.IdentityID = i.ID
+	s.Identity = i
+
+	if err := e.maybeLinkCredentials(ctx, s, i, f); err != nil {
```
zepatrik, line 140: For me, the `PostLoginHook` in general feels a bit weird with all the order of operations.

1. webhooks fire after first factor even when 2fa is required: maybe useful?
2. 2fa requirement is not checked in the beginning or after the hook fired, but somewhere nested. I would suggest to have that as one of the very first checks.

hperl (approved):
Very nice!

PR #4181, opened 2024-10-29
feat: improve QueryForCredentials

alnr:


persistence/sql/identity/persister_identity.go
```diff
@@ -707,9 +699,9 @@ func (p *IdentityPersister) HydrateIdentityAssociations(ctx context.Context, i *
 			// from complaining incorrectly.
 			//
 			// https://github.com/gobuffalo/pop/issues/723
-			con := con.WithContext(ctx)
-			creds, err := QueryForCredentials(con,
-				Where{"(identity_credentials.identity_id = ? AND identity_credentials.nid = ?)", []interface{}{i.ID, nid}})
+			creds, err := QueryForCredentials(p.GetConnection(ctx).WithContext(ctx),
+				Where{"identity_credentials.identity_id = ?", []interface{}{i.ID}},
+				Where{"identity_credentials.nid = ?", []interface{}{nid}})
```
aeneasr, line 704: Are you sure about this? I think the bracket was there to avoid an issue with an OR statement

aeneasr (commented):
Everything LGTM but there is one change I am not confident with

persistence/sql/identity/persister_identity.go
alnr: All there `WHERE` conditions are parenthesized [here](https://github.com/ory/kratos/pull/4181/files#diff-41d54e67508bfcb98af6522993d007de47a74ada2cf8426fc871c8a8a8babaeeL791). This change is for readability.

aeneasr: I see, so before it was

```
(identity_credentials.identity_id = ? AND identity_credentials.nid = ?) AND (...)
```

and now it is

```
(identity_credentials.identity_id = ?) AND (identity_credentials.nid = ?) AND (...)
```

So i guess this is fine!

PR #4183, opened 2024-10-30
feat: allow extra go migrations in persister

alnr:


aeneasr (approved):
Would it make sense though to just add a different migration type that is sql files that just don't run in a transaction? The go migrations are quite painful to write IMO. But it's good enough.

PR #4193, opened 2024-11-05
feat: remove duplicate queries during settings flow and use better index hint for credentials lookup

aeneasr:
This patch reduces duplicate GetIdentity queries as part of submitting the settings flow, and improves an index to significantly reduce credential lookup.

For better debugging, more tracing ha been added to the settings module.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


persistence/sql/migrations/sql/20241106142200000002_identities.autocommit.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE INDEX IF NOT EXISTS identity_credential_identifiers_ici_nid_i_idx
+    ON identity_credential_identifiers (identity_credential_id ASC, nid ASC, identifier ASC);
```
aeneasr, line 2: @alnr i changed the index structure to use `identity_credential_id` as the prefix (it is also the highest cardinality) and then we can use it both for FK and the query credentials

selfservice/flow/settings/hook.go
```diff
@@ -199,23 +205,20 @@ func (e *HookExecutor) PostSettingsHook(w http.ResponseWriter, r *http.Request,
 			case "oidc":
 				group = node.OpenIDConnectGroup
 			}
-			var traits identity.Traits
-			if i != nil {
```
aeneasr, line 203: i is never nil

PR #4203, opened 2024-11-12
feat: add failure reason to events

zepatrik:
Not sure if this covers all the error types we can get, but it should be the majority. We can still add more once we stumble upon them.

### DIFF
--- a/x/events/events.go
+++ b/x/events/events.go
@@ -5,8 +5,13 @@ package events
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"net/url"
 	"time"
+
+	"github.com/ory/herodot"
+	"github.com/ory/kratos/schema"
 
 	"github.com/gofrs/uuid"
 	otelattr "go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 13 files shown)


x/events/events.go
```diff
@@ -356,3 +371,13 @@ func NewWebhookFailed(ctx context.Context, err error) (string, trace.EventOption
...
+
+func reasonForError(err error) string {
+	if ve := new(schema.ValidationError); errors.As(err, &ve) {
+		return ve.Message
+	}
+	if r := *new(herodot.ReasonCarrier); errors.As(err, &r) {
+		return r.Reason()
+	}
+	return fmt.Sprintf("reason could not be determined for error: %T", err)
+}
```
aeneasr, line 383: People will ask about "reason IDs" to find e.g. all failed logins for reason X over time. Don't our validation errors have some type of ID or error code that we could emit?

aeneasr (approved):

x/events/events.go
aeneasr: I want to avoid people relying on these errors and then we change the casing or error message and we get complaints about backwards incompatible changes.

aeneasr: Yes please

aeneasr: We can also set some default like "contact_support" or something for the reason id to make it clear that you should request a reason id if it doesn't exist (opposed to reading the reason string field)

PR #4206, opened 2024-11-12
feat: add attributes to webhook events for better debugging

zepatrik:
With this change we add two attributes to webhook events:

- `WebhookID`: a string that can be set in the config and that helps to identify the actual webhook config that was used
- `TriggerID`: a random string that helps to correlate requests across retries

### DIFF
--- a/selfservice/hook/web_hook_integration_test.go
+++ b/selfservice/hook/web_hook_integration_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/tidwall/sjson"
+	"go.opentelemetry.io/otel/attribute"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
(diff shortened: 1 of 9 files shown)


selfservice/hook/web_hook_integration_test.go
```diff
@@ -1162,32 +1167,46 @@ func TestWebhookEvents(t *testing.T) {
...
-	s := &session.Session{ID: x.NewUUID(), Identity: &identity.Identity{ID: x.NewUUID()}}
-	_ = s
 	f := &login.Flow{ID: x.NewUUID()}
 
 	webhookReceiver := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.URL.Path == "/ok" {
 			w.WriteHeader(200)
 			w.Write([]byte("ok"))
 		} else {
-			w.WriteHeader(400)
```
zepatrik, line 1174: `400` would not trigger retries.

aeneasr (approved):

PR #4083, opened 2024-09-09
feat: allow partially failing batch inserts

hperl:
When batch-inserting multiple identities, conflicts or validation errors of a subset of identities in the batch still allow the rest of the identities to be inserted. The returned JSON contains the error details that lead to the failure.

## Related issue(s)

Docs PR: https://github.com/ory/docs/pull/1864

## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ ] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [ ] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [ ] I have read the [security policy](../security/policy).
- [ ] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got the approval (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](https://github.com/ory/docs).

PR #4211, opened 2024-11-14
fix: do not roll back transaction on partial identity insert error

zepatrik:
A follow up on #4083, where the basic feature was implemented but the transaction was eventually rolled back also it seemed like it succeeded.

### DIFF
--- a/persistence/sql/identity/persister_identity.go
+++ b/persistence/sql/identity/persister_identity.go
@@ -561,7 +561,8 @@ func (p *IdentityPersister) CreateIdentities(ctx context.Context, identities ...
 		}
 	}()
 
-	return p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) error {
+	var partialErr *identity.CreateIdentitiesError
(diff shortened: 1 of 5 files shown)


persistence/sql/identity/persister_identity.go
```diff
@@ -637,7 +639,7 @@ func (p *IdentityPersister) CreateIdentities(ctx context.Context, identities ...
 				return sqlcon.HandleError(err)
 			}
 
-			return partialErr
```
zepatrik, line 640: This was the problem.

hperl: I agree that rolling back the transaction is unwanted, but we do want to return information about which identity inserts failed. Don't we lose that information here?

hperl (changes requested):

persistence/sql/identity/persister_identity.go
hperl: Ah sorry, I misread, looks good!

hperl (approved):

PR #4244, opened 2024-12-12
chore: refactor parameter parsing in ListIdentities and disallow combining filters

alnr:
The code for ListIdentities has gotten out of hand. The string-based query building is not sustainable. @aeneasr and I feel we have no choice for now but to disable combining filters (like `GET /admin/identities?organization_id=abc&credential_identifier=xyz`).

This will allow us to build a simple switch into the ListIdentities handler to go to optimized implementations for each filter in the future and in cloud.


identity/handler.go
```diff
@@ -183,11 +179,73 @@ type listIdentitiesParameters struct {
...
+		tc, ok := ParseCredentialsType(v)
+		if !ok {
+			return params, errors.WithStack(herodot.ErrBadRequest.WithReasonf("Invalid value `%s` for parameter `include_credential`.", v))
+		}
+		params.DeclassifyCredentials = append(params.DeclassifyCredentials, tc)
+	}
+
+	if requestedFilters > 1 {
+		return params, errors.WithStack(herodot.ErrBadRequest.WithReason("You cannot combine multiple filters in this API"))
+	}
```
zepatrik, line 233: We could consider to reduce the impact of this breaking change by just ignoring additional filters. If `ids=a&ids=b&organization_id=x` previously returned only `a`, it could now return `a,b` ignoring the second filter. The result should always be a super-set of the previous version, therefore significantly limiting the impact.
On the other side, someone might rely on the behavior and this could in the worst case lead to security issues if the assumption is that all returned identities are part of the given organization.
WDYT?

identity/handler.go
```diff
@@ -271,7 +282,7 @@ func (h *Handler) list(w http.ResponseWriter, r *http.Request, _ httprouter.Para
 		}
 		u := *r.URL
 		pagepagination.PaginationHeader(w, &u, total, params.PagePagination.Page, params.PagePagination.ItemsPerPage)
-	} else {
+	} else if nextPage != nil {
```
zepatrik, line 285: `nextPage` should never be nil, and the header is also relevant in case of the last page.

zepatrik (commented):
LGTM 👍 just some thoughts.

identity/handler.go
alnr: I've checked our logs and could not find anyone combining filters, so I think we're good.

Changing semantics silently is a recipe for disaster.

zepatrik (approved):

PR #4251, opened 2024-12-19
feat: rework the OTP code submit count mechanism

alnr:
Unlike what the previous comment suggested, incrementing and checking the submit count inside the database transaction is not actually optimal peformance- or security-wise.

We now check atomically increment and check the submit count as the first part of the operation, and abort as early as possible if we detect brute-forcing. This prevents a situation where the check works only on certain transaction isolation levels.


persistence/sql/persister_code.go
```diff
@@ -123,3 +111,39 @@ func useOneTimeCode[P any, U interface {
...
+	}()
+
+	nid := p.NetworkID(ctx)
+
+	if p.c.Dialect.Name() == "mysql" { // no RETURNING support
+		//#nosec G201 -- TableName is static
+		qUpdate := fmt.Sprintf("UPDATE %s SET submit_count = submit_count + 1 WHERE id = ? AND nid = ?", flowTableName)
+		//#nosec G201 -- TableName is static
+		qSelect := fmt.Sprintf("SELECT submit_count FROM %s WHERE id = ? AND nid = ?", flowTableName)
+		err = p.Transaction(ctx, func(ctx context.Context, tx *pop.Connection) error {
```
alnr, line 130: Actually this is idiotic. There's no reason for this transaction to exist. MySQL should do a one-statement increment, and then a one-statement select, outside of a transaction.

Apologies...

alnr: fixed

persistence/sql/persister_code.go
```diff
@@ -85,10 +75,8 @@ func useOneTimeCode[P any, U interface {
 
 		if err := sqlcon.HandleError(codesQuery.All(&codes)); err != nil {
 			if errors.Is(err, sqlcon.ErrNoRows) {
-				// Return no error, as that would roll back the transaction and reset the submit count.
-				return nil
+				return code.ErrCodeNotFound
```
aeneasr, line 78:
```suggestion
				return errors.WithStack(code.ErrCodeNotFound)
```

aeneasr (approved):

persistence/sql/persister_code.go
alnr: added

PR #4259, opened 2024-12-31
fix: properly setup CSRF errors

jonas-jonas:
The current implementation frequently returns errors that seem to contradict itself (reason says "no CSRF cookie sent" and detail says "CSRF cookie found but did not match). The reason is as follows: herodot errors are copied in the WithDetail method, and in theory all operations in that method happen on a new struct. The `details` field is a map[string]any, of which only the pointer to it is copied in WithDetail, not the actual contents. This means, that calling `WithDetail` on an existing error, modifies the `details` field on the original error as well. 

And in this specific case, it means that calling WithDetail on ErrInvalidCSRFTokenServer, modifies the details field for all errors that do so, leading to misleading error responses.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/driver/registry_default.go
+++ b/driver/registry_default.go
@@ -826,7 +826,7 @@ func (m *RegistryDefault) WithCSRFTokenGenerator(cg x.CSRFToken) {
 
 func (m *RegistryDefault) GenerateCSRFToken(r *http.Request) string {
 	if m.csrfTokenGenerator == nil {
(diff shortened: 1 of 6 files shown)


jonas-jonas:
Fixed via [b7b7fd4f0bdbca38eb7e3f26c641cb99ad1337de](https://github.com/ory/kratos/commit/b7b7fd4f0bdbca38eb7e3f26c641cb99ad1337de)

PR #4272, opened 2025-01-22
feat: more extension points

hperl:
This adds more extension points to the Kratos registry.


selfservice/strategy/oidc/pkce.go
```diff
@@ -48,7 +48,7 @@ func maybePKCE(ctx context.Context, d pkceDependencies, _p Provider) (verifier s
 		// autodiscover PKCE support
 		pkceSupported, err := discoverPKCE(ctx, d, p)
 		if err != nil {
-			d.Logger().WithError(err).Warnf("Failed to autodiscover PKCE support for provider %q. Continuing without PKCE.", p.Config().ID)
+			d.Logger().WithError(err).Warnf("Failed to autodiscover PKCE support for Provider %q. Continuing without PKCE.", p.Config().ID)
```
aeneasr, line 51: revert this file?

selfservice/strategy/oidc/strategy.go
```diff
@@ -742,36 +742,36 @@ func (s *Strategy) CompletedAuthenticationMethod(ctx context.Context) session.Au
...
+		return nil, errors.WithStack(herodot.ErrBadRequest.WithReasonf("The id_token claims were invalid").WithError(err.Error()))
 	}
 
 	// First check if the JWT contains the nonce claim.
 	if claims.Nonce == "" {
 		// If it doesn't, check if the provider supports nonces.
 		if nonceSkipper, ok := verifier.(NonceValidationSkipper); !ok || !nonceSkipper.CanSkipNonce(claims) {
 			// If the provider supports nonces, abort the flow!
-			return nil, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("No nonce was included in the id_token but is required by the provider"))
+			return nil, errors.WithStack(herodot.ErrBadRequest.WithReasonf("No nonce was included in the id_token but is required by the Provider"))
```
aeneasr, line 764:
```suggestion
			return nil, errors.WithStack(herodot.ErrBadRequest.WithReasonf("No nonce was included in the id_token but is required by the provider"))
```

selfservice/strategy/oidc/strategy.go
```diff
@@ -742,36 +742,36 @@ func (s *Strategy) CompletedAuthenticationMethod(ctx context.Context) session.Au
...
 		}
 		// If the provider does not support nonces, we don't do validation and return the claim.
 		// This case only applies to Apple, as some of their devices do not support nonces.
 		// https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple
 	} else if idTokenNonce == "" {
 		// A nonce was present in the JWT token, but no nonce was submitted in the flow
-		return nil, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("No nonce was provided but is required by the provider"))
+		return nil, errors.WithStack(herodot.ErrBadRequest.WithReasonf("No nonce was provided but is required by the Provider"))
```
aeneasr, line 771:
```suggestion
		return nil, errors.WithStack(herodot.ErrBadRequest.WithReasonf("No nonce was provided but is required by the provider"))
```

embedx/config.schema.json
```diff
@@ -578,6 +579,13 @@
           "type": "string",
           "enum": ["auto", "never", "force"],
           "default": "auto"
+        },
+        "fedcm_config_url": {
+          "title": "Federation Configuration URL",
+          "description": "The URL where the FedCM IdP configuration is located for the provider.",
```
aeneasr, line 585: Add network note?

spec/api.json
```diff
@@ -413,10 +413,57 @@
         },
         "type": "object"
       },
+      "Provider": {
```
aeneasr, line 416: Should this end up in the API spec?

spec/api.json
```diff
@@ -413,10 +413,57 @@
...
+            "type": "object"
+          }
+        },
+        "type": "object"
+      },
       "RecoveryAddressType": {
         "title": "RecoveryAddressType must not exceed 16 characters as that is the limitation in the SQL Schema.",
         "type": "string"
       },
+      "SubmitFedcmTokenBody": {
```
aeneasr, line 459: So far `submit` is not a keyword we use in the API spec. Is there an alternative that better aligns?

spec/api.json
```diff
@@ -905,6 +952,22 @@
         ],
         "type": "object"
       },
+      "getParametersResponse": {
```
aeneasr, line 955: Should have at least FedCM in the name, I guess

spec/api.json
```diff
@@ -2818,26 +2881,26 @@
       "updateLoginFlowWithOidcMethod": {
         "description": "Update Login Flow with OpenID Connect Method",
         "properties": {
+          "Provider": {
```
aeneasr, line 2884: This should not be upper case

aeneasr (approved):
Looks like `provider` became `Provider` in some files due to probably some refactoring. Also the API spec looks not fully correct yet. Otherwise LGTM

aeneasr (changes requested):
See previous review

spec/api.json
hperl: Yes, I think so, because it needs to be part of the SDK.

spec/api.json
hperl: A bit lengthy, but how about `createSessionFromFedCmToken`? We regularly use `create` elsewhere.

However, on registration, you really only get a session if the session hook is active, so the `createSession` part is really not correct in all cases.

selfservice/strategy/oidc/provider_netid.go
```diff
@@ -118,6 +119,63 @@ func (n *ProviderNetID) Claims(ctx context.Context, exchange *oauth2.Token, _ ur
...
+	token := struct {
+		IDToken string `json:"id_token"`
+	}{}
+
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := json.NewDecoder(bytes.NewBuffer(body)).Decode(&token); err != nil {
```
zepatrik, line 151:
```suggestion
	if err := json.NewDecoder(res.Body).Decode(&token); err != nil {
```

spec/api.json
zepatrik: In a way `getFedcmParameters` starts the FedCM login/registration flow, and `submitFedcmToken` completes it, right? So maybe use the flow terminology instead: `createFedcmFlow` and `updateFedcmFlow`?

zepatrik (approved):

PR #4277, opened 2025-01-27
fix: allow patching some /credentials sub-paths

hperl:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -916,7 +916,7 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
 
 	patchedIdentity := WithAdminMetadataInJSON(*identity)
 
-	if err := jsonx.ApplyJSONPatch(requestBody, &patchedIdentity, "/id", "/stateChangedAt", "/credentials", "/credentials/**"); err != nil {
+	if err := jsonx.ApplyJSONPatch(requestBody, &patchedIdentity, "/id", "/stateChangedAt", "/credentials", "/credentials/oidc/**"); err != nil {
 		h.r.Writer().WriteError(w, r, errors.WithStack(
 			herodot.
 				ErrBadRequest.
--- a/identity/handler_test.go
+++ b/identity/handler_test.go
@@ -1227,7 +1227,7 @@ func TestHandler(t *testing.T) {
 		}
 	})
 
-	t.Run("case=PATCH should fail to update credential password", func(t *testing.T) {
+	t.Run("case=PATCH should allow to update credential password", func(t *testing.T) {
 		uuid := x.NewUUID().String()
 		email := uuid + "@ory.sh"
 		password := "ljanf123akf"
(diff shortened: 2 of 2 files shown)


aeneasr (approved):

identity/handler.go
```diff
@@ -916,7 +916,7 @@ func (h *Handler) patch(w http.ResponseWriter, r *http.Request, ps httprouter.Pa
 
 	patchedIdentity := WithAdminMetadataInJSON(*identity)
 
-	if err := jsonx.ApplyJSONPatch(requestBody, &patchedIdentity, "/id", "/stateChangedAt", "/credentials", "/credentials/**"); err != nil {
+	if err := jsonx.ApplyJSONPatch(requestBody, &patchedIdentity, "/id", "/stateChangedAt", "/credentials", "/credentials/oidc/**"); err != nil {
```
jonas-jonas, line 919: Should we have a test, that checks if `oidc` is disallowed?

jonas-jonas (commented):
Maybe we should also have tests that check, that it's not possible to override nested objects by specifying the wrong path.

E.g.
```
PATCH 
{
	"op": "replace",
	"path": "/credentials/password",
	"value": {
		"hashed_password": "<hash>"
	},
}
```
Here, `/config/` is missing in the path. Which seems like an easy mistake to make, but would render the identity unusable? Or is this caught by the validator?

identity/handler.go
hperl: We have ;)

hperl:
> Maybe we should also have tests that check, that it's not possible to override nested objects by specifying the wrong path.
> 
> E.g.
> 
> ```
> PATCH 
> {
> 	"op": "replace",
> 	"path": "/credentials/password",
> 	"value": {
> 		"hashed_password": "<hash>"
> 	},
> }
> ```
> 
> Here, `/config/` is missing in the path. Which seems like an easy mistake to make, but would render the identity unusable? Or is this caught by the validator?

Invalid keys are ignored, and AFAIK this is how it has always been.

I agree that this would be useful, and also to extend the PATCHing to OIDC, but here I'd just like to revert to the previous behaviour first (with the added benefit of a test). Let's move this into a new issue?

jonas-jonas:
>Invalid keys are ignored, and AFAIK this is how it has always been.

Not what I meant. The key is technically valid, because the source JSON object the patch is applied against, has the following structure:

```json
{
  "credentials": {
    "password: {
	  "config": {
		"hashed_password": "<hash>",
      }
	}
  }
}
```

But after applying the faulty patch, the JSON document looks like this:

```json
{
  "credentials": {
    "password: {
	   "hashed_password": "<hash>",
	}
  }
}
```

Which is not valid, but I am not sure what would happen if you send the request. So a test for that case would be good. Again, I think that is a very common scenario that will happen easily, and we should do our best to prevent it.

alnr (changes requested):
this line of code is still broken and doesn't do what the comment says. pls fix 🙏

aeneasr:
Ideally, for both `create`, `patch`, and `PUT` operations, the following should be implemented with proper 
validation:

1. **Passwords**:  
   - Set or change passwords (with validation ensuring the hash or plaintext password is correct).  
   - Delete passwords.

2. **OIDC**:  
   - Set or change OIDC subject and provider.  
   - Delete OIDC subject and provider.

3. **WebAuthn and Passkey Credentials**:  
   - Set, change, or delete WebAuthn and passkey credentials (e.g., to import passkey users).

4. **Code Credentials**:  
   - Activate or deactivate code credentials.

5. **TOTP**:  
   - Set, change, or delete TOTP configurations.

6. **Lookup Secrets**:  
   - Set, change, or delete lookup secrets.

**Restrictions**:  
What should not be allowed is:  
- Adding arbitrary keys to the config object.  
- Changing identifiers that are typically set by the system.  
- Modifying keys that should only be set by the system.

zepatrik (approved):

hperl:
I opened https://github.com/ory/kratos/issues/4279 for a proper fix. In the meantime, reverting the forbidden paths is not more wrong / broken than what we currently have, so I'd rather have this merged right now to unblock customers.

zlodes:
> Set, change, or delete TOTP configurations.
@aeneasr is that planned?

Would be really useful to have an ability to import TOTP while migration from some other IDP.

I see that issue #4279 was closed by a bot. Shouldn't we reopen it?

PR #4286, opened 2025-02-02
fix: return `return_to` code if already authenticated

hperl:
This fixes a bug in native OIDC login and registration flows, where the user already has a session in the browser the flow is continued with (usually a web view, but depending on the platform it already has a session cookie set). In the callback, we now correctly handle the case in `alreadyAuthenticated` to return the session token exchange code. 

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


jonas-jonas (approved):
Nice, thanks! Wasn't sure if this is the correct solution, but looks good!

zepatrik (approved):

PR #4291, opened 2025-02-05
fix: ignore CSRF on all apple provider callback URLs

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


selfservice/strategy/oidc/strategy.go
```diff
@@ -176,7 +176,7 @@ func (s *Strategy) setRoutes(r *x.RouterPublic) {
 		// Hardcoded path to Apple provider, I don't have a better way of doing it right now.
 		// Also this ignore disables CSRF checks for both GET and POST requests. Unfortunately
 		// CSRF handler does not allow to define a rule based on the request method, at least not yet.
-		s.d.CSRFHandler().IgnorePath(RouteBase + "/callback/apple")
+		s.d.CSRFHandler().IgnoreGlob(RouteBase + "/callback/apple*")
```
jonas-jonas, line 179: We might even consider ignoring CSRF on all `/callback/**` URLs, even though Apple seems to be the only provider that does this weird `POST` here. 

But if we removed `/apple` from the exception, apple OIDC configurations wouldn't _need_ to start with `apple`, and could be anything really. One less bit of complexity that users need to be aware of.

hperl: I don't see why CSRF protection is necessary here given that we have the continuity containter that protects integrity for the whole OIDC flow.

hperl (approved):

selfservice/strategy/oidc/strategy.go
jonas-jonas: I don't think there is effective CSRF protection on OIDC callbacks anyway, but we have blanket CSRF protection for all POST requests, and apple _sometimes_ does a POST request here, which then trips up the CSRF handler (as there is no CSRF token in the POST body). So yea, agree, IMO we should do:

```suggestion
		s.d.CSRFHandler().IgnoreGlob(RouteBase + "/callback/*")
```

aeneasr (approved):
Please change the code doc according to what we discussed on slack - otherwise LGTM

PR #4304, opened 2025-02-11
feat: allow deleting certain first-factor credentials

zepatrik:
The admin API did not allow to delete passwords and code credentials at all. The restriction is now lifted to only block deletion of the first-factor credential if it is the last one.

### DIFF
--- a/identity/credentials.go
+++ b/identity/credentials.go
@@ -120,6 +120,7 @@ func (c CredentialsType) ToUiNodeGroup() node.UiNodeGroup {
 var AllCredentialTypes = []CredentialsType{
 	CredentialsTypePassword,
 	CredentialsTypeOIDC,
+	CredentialsTypeSAML,
 	CredentialsTypeTOTP,
 	CredentialsTypeLookup,
 	CredentialsTypeWebAuthn,
@@ -136,8 +137,8 @@ const (
 
 // ParseCredentialsType parses a string into a CredentialsType or returns false as the second argument.
 func ParseCredentialsType(in string) (CredentialsType, bool) {
-	for _, t := range []CredentialsType{
-		CredentialsTypePassword,
+	switch t := CredentialsType(in); t {
+	case CredentialsTypePassword,
 		CredentialsTypeOIDC,
 		CredentialsTypeSAML,
 		CredentialsTypeTOTP,
@@ -146,11 +147,8 @@ func ParseCredentialsType(in string) (CredentialsType, bool) {
 		CredentialsTypeCodeAuth,
 		CredentialsTypeRecoveryLink,
 		CredentialsTypeRecoveryCode,
-		CredentialsTypePasskey,
-	} {
-		if t.String() == in {
-			return t, true
-		}
+		CredentialsTypePasskey:
+		return t, true
 	}
 	return "", false
 }
--- a/identity/handler.go
+++ b/identity/handler.go
@@ -1030,16 +1030,28 @@ func (h *Handler) deleteIdentityCredentials(w http.ResponseWriter, r *http.Reque
 			h.r.Writer().WriteError(w, r, err)
 			return
 		}
-	case CredentialsTypePassword, CredentialsTypeCodeAuth:
-		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf("You cannot remove first factor credentials.")))
-		return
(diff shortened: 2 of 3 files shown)


identity/handler.go
```diff
@@ -1030,16 +1030,28 @@ func (h *Handler) deleteIdentityCredentials(w http.ResponseWriter, r *http.Reque
...
-		return
-	case CredentialsTypeOIDC:
-		if err := identity.deleteCredentialOIDCFromIdentity(r.URL.Query().Get("identifier")); err != nil {
+	case CredentialsTypePassword, CredentialsTypeCodeAuth, CredentialsTypeOIDC:
+		firstFactor, err := h.r.IdentityManager().CountActiveFirstFactorCredentials(r.Context(), identity)
+		if err != nil {
 			h.r.Writer().WriteError(w, r, err)
 			return
 		}
+		if firstFactor < 2 {
```
hperl, line 1039: Note, this could still be safe (i.e. the user can still log in), if the code strategy is enabled with `missing_credential_fallback_enabled`.

identity/handler.go
```diff
@@ -1030,16 +1030,28 @@ func (h *Handler) deleteIdentityCredentials(w http.ResponseWriter, r *http.Reque
...
+				h.r.Writer().WriteError(w, r, err)
+				return
+			}
+		}
 	default:
-		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf("Unknown credentials type %s.", cred.Type)))
+		// A bunch of credential type deletions are not yet implemented, e.g. passkeys, etc.
+		h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf("Credentials type %s cannot be deleted.", cred.Type)))
 		return
 	}
```
hperl, line 1057: Should we call `identity.Validate()` here to make sure the identity is still valid? For example, it also creates an empty password credential (if password is an identifier) with the email address so that `FindByIdentifier` still works.

identity/handler.go
```diff
@@ -1030,16 +1030,28 @@ func (h *Handler) deleteIdentityCredentials(w http.ResponseWriter, r *http.Reque
...
 			h.r.Writer().WriteError(w, r, err)
 			return
 		}
+		if firstFactor < 2 {
+			h.r.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf("You cannot remove the last first factor credential.")))
+			return
+		}
+		switch cred.Type {
+		case CredentialsTypePassword, CredentialsTypeCodeAuth:
+			identity.DeleteCredentialsType(cred.Type)
```
hperl, line 1045: If an identity as both an `oidc` credential with an identifier derived from the subject, and a password credential with an identifier set to the email address, and we delete the password credential, then the email address will not be a credential identifier any more, right? Is that what we want?

I would have expected that we just clear the password credentials config, but keep the identifier in place. Maybe `identity.Validate()` would do exactly that, but in any case we should add a test for it. WDYT?

hperl (changes requested):
LGTM, I just have a question about the consequence of deleting the pw credential.

identity/credentials.go
```diff
@@ -120,7 +120,7 @@ func (c CredentialsType) ToUiNodeGroup() node.UiNodeGroup {
 var AllCredentialTypes = []CredentialsType{
 	CredentialsTypePassword,
 	CredentialsTypeOIDC,
-	CredentialsTypeSAML,
+	// CredentialsTypeSAML, placeholder for the OEL version
```
hperl, line 123: Do we have a way do re-inject this credential type in OEL?

hperl (changes requested):

zepatrik:
@aeneasr I'm not sure about the latest changes revolving the deletion of code credentials. Happy to move that into its own PR and block code credential deletion for now (i.e. keep the behavior and only fix pw in this PR).

identity/handler.go
zepatrik: Validate is called, but that does not really fix the problem, at least not when the traits don't contain the identifier.
I have now alternatively implemented a detailed password credential deletion.

identity/credentials.go
zepatrik: The SAML type was never part of this slice. I just added it because I was confused at first why it is not part, but it really should not be.

hperl (approved):

PR #4321, opened 2025-02-26
fix: also update identifiers

hperl:
This fixes a bug where when an identity is merged into another, the identifier of the original identity was not updated.

### DIFF
--- a/selfservice/strategy/oidc/strategy_login.go
+++ b/selfservice/strategy/oidc/strategy_login.go
@@ -143,14 +143,15 @@ func (s *Strategy) handleConflictingIdentity(ctx context.Context, w http.Respons
 		if _, ok := existingIdentity.Credentials[s.ID()]; !ok {
 			existingIdentity.SetCredentials(s.ID(), *creds)
 		} else {
-
+			newCreds := existingIdentity.Credentials[s.ID()]
 			var conf identity.CredentialsOIDC
-			if err = json.Unmarshal(existingIdentity.Credentials[s.ID()].Config, &conf); err != nil {
+			if err = json.Unmarshal(newCreds.Config, &conf); err != nil {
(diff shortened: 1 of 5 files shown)


selfservice/strategy/oidc/strategy_login.go
```diff
@@ -143,14 +143,15 @@ func (s *Strategy) handleConflictingIdentity(ctx context.Context, w http.Respons
 		if _, ok := existingIdentity.Credentials[s.ID()]; !ok {
 			existingIdentity.SetCredentials(s.ID(), *creds)
 		} else {
-
+			newCreds := existingIdentity.Credentials[s.ID()]
```
aeneasr, line 146: This merge logic is no longer trivial, and it looks ok but I'm not sure. It should be a separate function with tests to cover at least all of the potential merge behaviors:

1. Identity with credentials that are not oidc is verdict merge -> new oidc credential, identifier correctly set
2. Identity with one existing oidc provider that is not autolink is merged -> this means we have two oidc providers - one is the old one, the other is the new one
3. Identity with two existing oidc providers, one of which is autolink, is merged -> we have two oidc providers, one had its autolink credential replaced

Similarly, this should/could be done with additional non-oidc credentials to make sure we don't break other credentials.

Ideally you would generate snapshots for the results to make it readable to the reviewer that the merge strategies work as expected.This merge logic is no longer trivial, and it looks ok but I'm not sure. It should be a separate function with tests to cover at least all of the potential merge behaviors:

1. Identity with credentials that are not oidc is verdict merge -> new oidc credential, identifier correctly set
2. Identity with one existing oidc provider that is not autolink is merged -> this means we have two oidc providers - one is the old one, the other is the new one
3. Identity with two existing oidc providers, one of which is autolink, is merged -> we have two oidc providers, one had its autolink credential replaced

Similarly, this should/could be done with additional non-oidc credentials to make sure we don't break other credentials.

Ideally you would generate snapshots for the results to make it readable to the reviewer that the merge strategies work as expected.

aeneasr (changes requested):

aeneasr (approved):

PR #4323, opened 2025-02-26
fix: IdentityCreated is over-reporting on error inserts

aeneasr:
`defer` was the incorrect code path here, as we should only record identity created if the transaction did not error (aka was rolled back).

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/identity/persister_identity.go
+++ b/persistence/sql/identity/persister_identity.go
@@ -553,14 +553,6 @@ func (p *IdentityPersister) CreateIdentities(ctx context.Context, identities ...
 	}
 
 	var succeededIDs []uuid.UUID
-
-	defer func() {
-		// Report succeeded identities as created.
-		for _, identID := range succeededIDs {
-			span.AddEvent(events.NewIdentityCreated(ctx, identID))
-		}
-	}()
-
 	var partialErr *identity.CreateIdentitiesError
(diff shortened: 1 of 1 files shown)


persistence/sql/identity/persister_identity.go
```diff
@@ -553,14 +553,6 @@ func (p *IdentityPersister) CreateIdentities(ctx context.Context, identities ...
 	}
 
 	var succeededIDs []uuid.UUID
-
-	defer func() {
```
aeneasr, line 557: This runs always, even if the transaction block fails.

persistence/sql/identity/persister_identity.go
```diff
@@ -651,6 +643,12 @@ func (p *IdentityPersister) CreateIdentities(ctx context.Context, identities ...
 	}); err != nil {
 		return err
 	}
+
+	// Report succeeded identities as created.
+	for _, identID := range succeededIDs {
```
aeneasr, line 648: I'm not entirely sure if this is entirely correct or if we can end up with an error in the transaction block but still have the identity partially available.

hperl: No, I think this is really the right place.

hperl (approved):

alnr (approved):

PR #4331, opened 2025-03-05
chore: minor bugs and improvements

zepatrik:


identity/identity.go
```diff
@@ -185,20 +182,11 @@ func (i Identity) TableName(context.Context) string {
 	return "identities"
 }
 
-func (i *Identity) lock() *sync.RWMutex {
-	if i.l == nil {
-		i.l = new(sync.RWMutex)
-	}
-	return i.l
-}
```
zepatrik, line 193: This was never working as it should, as the check `i.l == nil` and then the assignment is not atomic. Therefore the lock does not work for exclusive locking.
I did not find any reason for the lock in the first place, and only a hand full of methods actually used it, while a lot more should really. Therefore, I just removed it all together.

identity/identity.go
```diff
@@ -519,10 +493,9 @@ func (i *Identity) WithDeclassifiedCredentials(ctx context.Context, c cipher.Pro
...
-					var plaintext []byte
-					plaintext, err := c.Cipher(ctx).Decrypt(ctx, ciphertext)
-					if err != nil {
-						plaintext = []byte("")
+					plaintext, decryptErr := c.Cipher(ctx).Decrypt(ctx, ciphertext)
+					if decryptErr != nil {
+						plaintext = []byte{}
 					}
 					toPublish.Config, err = sjson.SetBytes(toPublish.Config, "providers."+key, string(plaintext))
 					if err != nil {
```
zepatrik, line 501: The error here would not be correctly returned because it is a different scope than the one that is finally returned.

hperl (approved):
Nice cleanups!

PR #4343, opened 2025-03-18
chore: remove fizz migration files

zepatrik:
As we'll never use them again at this point, we can clean that up a bit.


aeneasr (approved):
lgtm

PR #4345, opened 2025-03-18
chore: make tools indirect dependencies

zepatrik:


aeneasr (approved):
Can be merged once CI passes

Please be aware that module rewrites don't work for go tools

zepatrik:
> Please be aware that module rewrites don't work for go tools

I don't think this is true, at least it works to generate the SDK with the current setup.

PR #4352, opened 2025-03-20
fix: apply strategy filters in identifier first as well

aeneasr:


driver/config/testhelpers/config.go
```diff
@@ -18,13 +18,18 @@ import (
...
+		ConfigSchema []byte
 	}
 	contextKey int
 )
 
 func (t *TestConfigProvider) NewProvider(ctx context.Context, opts ...configx.OptionModifier) (*configx.Provider, error) {
-	return configx.New(ctx, []byte(embedx.ConfigSchema), append(t.Options, opts...)...)
+	schema := []byte(embedx.ConfigSchema)
+	if len(t.ConfigSchema) > 0 {
+		schema = t.ConfigSchema
```
aeneasr, line 30: Allow customizing the schema during tests

selfservice/flow/login/handler.go
```diff
@@ -204,46 +204,18 @@ func (h *Handler) NewLoginFlow(w http.ResponseWriter, r *http.Request, ft flow.T
...
-			// We only apply the filter on AAL1, because the OIDC strategy can only satsify
-			// AAL1.
-			strategyFilters = []StrategyFilter{func(s Strategy) bool {
-				return s.ID() == identity.CredentialsTypeOIDC || s.ID() == identity.CredentialsTypeSAML
-			}}
-		}
-	}
-
-	for _, s := range h.d.LoginStrategies(r.Context(), strategyFilters...) {
+	for _, s := range h.d.LoginStrategies(r.Context(), PrepareOrganizations(r, f, sess)...) {
```
aeneasr, line 207: This all moved to PrepareOrganizations()

selfservice/flow/login/handler.go
```diff
@@ -204,46 +204,18 @@ func (h *Handler) NewLoginFlow(w http.ResponseWriter, r *http.Request, ft flow.T
...
 		var populateErr error
 
 		switch strategy := s.(type) {
 		case FormHydrator:
 			switch {
 			case f.RequestedAAL == identity.AuthenticatorAssuranceLevel1:
 				switch {
-				case f.IsRefresh():
+				case f.IsRefresh() && sess != nil:
```
aeneasr, line 215: Refresh cannot be true if session is nil.

selfservice/flow/login/organizations.go
```diff
@@ -0,0 +1,55 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/ory/kratos/identity"
+	"github.com/ory/kratos/session"
+)
+
+var organizationFilter = []StrategyFilter{func(s Strategy) bool {
+	a, b := s.(interface {
+		SupportsOrganizations() bool
```
aeneasr, line 19: Let the strategy tell us if it support organizations or not, rather than inferring it.

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -432,18 +454,23 @@ func (s *Strategy) PopulateLoginMethodIdentifierFirstCredentials(r *http.Request
 		}
 
 		// We found no credentials. We remove all the providers and tell the strategy that we found nothing.
-		f.GetUI().UnsetNode("provider")
+		s.removeProviders(conf, f)
```
aeneasr, line 457: `UnsetNode` would delete all provider fields, which doesn't make sense. Instead, removeProviders now only removes providers coming from this strategy's configuration.

selfservice/strategy/oidc/strategy_login.go
```diff
@@ -432,18 +454,23 @@ func (s *Strategy) PopulateLoginMethodIdentifierFirstCredentials(r *http.Request
...
 		// Account enumeration is disabled, so we show all providers that are linked to the identity.
 		// User is found and enumeration mitigation is disabled. Filter the list!
-		f.GetUI().UnsetNode("provider")
+		s.removeProviders(conf, f)
 
 		for _, l := range linked {
 			lc := l.Config()
-			AddProvider(f.UI, lc.ID, text.NewInfoLoginWith(stringsx.Coalesce(lc.Label, lc.ID), lc.ID))
+
+			// Organizations are handled differently.
```
aeneasr, line 469: This strategy cannot handle organizations.

selfservice/flow/login/organizations.go
```diff
@@ -0,0 +1,42 @@
...
+	a, b := s.(flow.OrganizationImplementor)
+	return b && a.SupportsOrganizations()
+}}
+
+func PrepareOrganizations(r *http.Request, f *Flow, sess *session.Session) []StrategyFilter {
+	if f.RequestedAAL != identity.AuthenticatorAssuranceLevel1 {
+		return []StrategyFilter{}
+	}
+
+	if f.OrganizationID.Valid {
```
zepatrik, line 25: Is it really safe to store the organization ID in the flow and trust it? What if I initialize the flow with one organization, but then update the identifier so I use a different one?
As far as I can see, this behavior is new.

selfservice/flow/login/organizations_test.go
```diff
@@ -0,0 +1,63 @@
...
+	t.Run("should return empty filter if AAL is not AAL1", func(t *testing.T) {
+		f := &Flow{RequestedAAL: identity.AuthenticatorAssuranceLevel2}
+		r := &http.Request{URL: new(url.URL)}
+		sess := &session.Session{}
+
+		filters := PrepareOrganizations(r, f, sess)
+		assert.Empty(t, filters)
+	})
+
+	t.Run("should return empty filter if OrganizationID is not valid", func(t *testing.T) {
```
zepatrik, line 30: This case is the same as the previous, so it does not what it claims? I don't see an invalid organization ID being set.

selfservice/flow/login/organizations_test.go
```diff
@@ -0,0 +1,63 @@
...
+		r := &http.Request{
+			URL: &url.URL{
+				RawQuery: "organization=123e4567-e89b-12d3-a456-426614174000",
+			},
+		}
+		sess := &session.Session{}
+		filters := PrepareOrganizations(r, f, sess)
+		require.NotEmpty(t, filters)
+		assert.Equal(t, organizationFilter, filters)
+	})
```
zepatrik, line 62: Missing cases:
- organization ID in flow
- organization ID in query and identity (matching and non-matching)
- organization ID set but AAL2

selfservice/strategy/saml/login.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// Method to use
+	//
+	// This field must be set to `saml` when using the saml method.
+	//
+	// required: true
+	Method string `json:"method"`
+
+	// The identity traits. This is a placeholder for the registration flow.
+	Traits json.RawMessage `json:"traits"`
```
zepatrik, line 28: Why do we have the traits in login? What kind of placeholder is this?

zepatrik (changes requested):

zepatrik (approved):

selfservice/strategy/oidc/strategy.go
```diff
@@ -837,7 +846,7 @@ func (s *Strategy) linkCredentials(ctx context.Context, i *identity.Identity, to
 	creds, err := i.ParseCredentials(s.ID(), &conf)
 	if errors.Is(err, herodot.ErrNotFound) {
 		var err error
-		if creds, err = identity.NewCredentialsOIDC(tokens, provider, subject, organization); err != nil {
+		if creds, err = identity.NewOIDCLikeCredentials(tokens, s.ID(), provider, subject, organization); err != nil {
```
aeneasr, line 849: @hperl this was using OIDC credentials, the workaround was only possible in a very specific scenario prior.

selfservice/strategy/oidc/strategy.go
```diff
@@ -169,8 +173,13 @@
 				}
 
 				for _, prov := range conf.Providers {
-					if provider == prov.Provider && sub == prov.Subject &&
-						prov.Subject != "" && prov.Provider != "" {
+					if withOrgs && len(prov.Organization) == 0 {
```
hperl, line 176: Shouldn't we just count orgs that the identity is currently in here?

text/message_validation.go
```diff
@@ -290,7 +292,7 @@ func NewErrorValidationDuplicateCredentialsWithHints(availableCredentialTypes []
 			switch cred {
 			case "password":
 				humanReadable = append(humanReadable, "your password")
-			case "oidc":
+			case "oidc", "saml":
```
hperl, line 295: Is SAML really social?

hperl (approved):
LGTM, I just have a question about counting org SSO credentials.

selfservice/strategy/oidc/strategy.go
aeneasr: This specific OIDC strategy cannot deal with organizations at all - it is only for social sign in. So the correct way is to exclude any organization providers as that would be handled by a different strategy. I can link you to the code in the private repo for clarification if you want!

aeneasr: I think this is conceptually sound and we can still fix this up if not :)

PR #4361, opened 2025-03-25
feat: support importing more credentials

aeneasr:
Adds support to import SAML credentials. SAML connections are only available in Ory Enterprise License / Ory Network.


persistence/sql/migrations/sql/20241203105600000000_saml_credential_type.up.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO identity_credential_types (id, name)
+SELECT '7bddcf6c-f50e-4a18-9b0f-429114c33419', 'saml'
```
aeneasr, line 2: I moved this migration

identity/handler.go
```diff
@@ -509,6 +510,41 @@ type AdminCreateIdentityImportCredentialsOidcProvider struct {
 	//
 	// required: false
 	UseAutoLink bool `json:"use_auto_link,omitempty"`
+
+	// The organization to assign for the provider.
+	Organization uuid.NullUUID `json:"organization,omitempty"`
```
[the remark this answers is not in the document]
hperl: Yes!

identity/handler.go
```diff
@@ -451,6 +451,9 @@ type IdentityWithCredentials struct {
 
 	// OIDC if set will import an OIDC credential.
 	OIDC *AdminIdentityImportCredentialsOIDC `json:"oidc"`
+
+	// OIDC if set will import an OIDC credential.
```
hperl, line 455: SAML if set will import an SAML credential

identity/handler.go
```diff
@@ -509,6 +510,41 @@ type AdminCreateIdentityImportCredentialsOidcProvider struct {
...
+type AdminIdentityImportCredentialsSAMLConfig struct {
+	// A list of SAML Providers
+	Providers []AdminCreateIdentityImportCredentialsSAMLProvider `json:"providers"`
+}
+
+// Create Identity and Import SAML Credentials Configuration
+//
+// swagger:model identityWithCredentialsSamlConfigProvider
+type AdminCreateIdentityImportCredentialsSAMLProvider struct {
+	// The subject (`sub`) of the OpenID Connect connection. Usually the `sub` field of the ID Token.
```
hperl, line 536: Should this still reference OIDC?

hperl (approved):
LGTM! I would just not make the SAML doc strings reference OIDC, since that would be an implementation detail.

identity/handler.go
aeneasr: Good catch, I missed the code docs in copy paste.

aeneasr:
Code doc addressed!

PR #4364, opened 2025-03-27
fix: ensure context is not canceled during password hashing

zepatrik:
Especially during large imports of plaintext passwords there can be a lot of useless hashing, even after the request timed out or got canceled.

### DIFF
--- a/hash/hash_comparator.go
+++ b/hash/hash_comparator.go
@@ -60,7 +60,7 @@ func NewCryptDecoder() *crypt.Decoder {
 var CryptDecoder = NewCryptDecoder()
 
 type SupportedHasher struct {
-	Comparator func(ctx context.Context, password []byte, hash []byte) error
+	Comparator func(ctx context.Context, password, hash []byte) error
 	Name       string
 	Is         func(hash []byte) bool
 }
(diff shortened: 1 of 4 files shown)


hash/hash_comparator.go
```diff
@@ -162,11 +162,16 @@ func CompareMD5Crypt(_ context.Context, password []byte, hash []byte) error {
...
 }
 
-func CompareBcrypt(_ context.Context, password []byte, hash []byte) error {
+func CompareBcrypt(ctx context.Context, password, hash []byte) error {
 	if err := validateBcryptPasswordLength(password); err != nil {
 		return err
 	}
 
+	// ensure that the context is not canceled before doing the heavy lifting
+	if ctx.Err() != nil {
```
aeneasr, line 171: Does this actually do anything? Isn't the hashing happening sequentially?

zepatrik: I went ahead and added the context check in every place before we do any hash computation, as that's the best we can do. The comparison typically only happens during login, so there should not be more than one hash being computed, however I figured that it doesn't hurt to check there as well.

hash/hasher_bcrypt.go
```diff
@@ -45,6 +45,11 @@ func (h *Bcrypt) Generate(ctx context.Context, password []byte) ([]byte, error)
 		return nil, err
 	}
 
+	// ensure that the context is not canceled before doing the heavy lifting
+	if ctx.Err() != nil {
```
zepatrik, line 49: This is the relevant context check for the import API.

aeneasr: Ok, I see, it is called in a loop and if we don't check the context then the loop just continues. Got it!

aeneasr (approved):

PR #4373, opened 2025-04-07
feat: add email domain matcher

aeneasr:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


embedx/identity_extension.schema.json
```diff
@@ -82,6 +82,16 @@
                   "enum": ["email"]
                 }
               }
+            },
+            "organization": {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "matcher": {
```
jonas-jonas, line 90: "match_by" would align a bit more closely with the "via" keyword, I think. 

But "matcher" could also be extended to a Jsonnet (or some other scripting) based custom matcher, that decides whether a user is part of an org or not. 

How set are you on the name?

selfservice/flow/registration/strategy_form_hydrator.go
```diff
@@ -6,8 +6,12 @@ package registration
 import (
 	"encoding/json"
 	"net/http"
+
+	"github.com/pkg/errors"
 )
 
+var ErrBreakRegistrationPopulate = errors.New("skip rest of registration form population")
```
zepatrik, line 13: Use stdlib errors package or `fmt.Errorf` instead, as otherwise we get the stack trace of the init func here and later don't override it.

selfservice/strategy/profile/registration.go
```diff
@@ -213,6 +218,8 @@ func (s *Strategy) Register(w http.ResponseWriter, r *http.Request, regFlow *reg
 
 func (s *Strategy) returnToProfileForm(ctx context.Context, w http.ResponseWriter, r *http.Request, regFlow *registration.Flow, params updateRegistrationFlowWithProfileMethod) error {
 	regFlow.UI.ResetMessages()
+	regFlow.OrganizationID.Valid = false
+	regFlow.OrganizationID.UUID = uuid.Nil
```
zepatrik, line 222:
```suggestion
	regFlow.OrganizationID = uuid.NullUUID{}
```

selfservice/flow/registration/strategy_form_hydrator.go
zepatrik: Also, in general it would be nice to not use errors in cases where everything is fine and we just want to communicate a state up to the caller. But we do this elsewhere already, so I guess fine for now...

embedx/identity_extension.schema.json
aeneasr: Organizations resolver?

aeneasr: I landed on organizations matcher

selfservice/flow/registration/strategy_form_hydrator.go
aeneasr: it's now an stderr (fmt.errorf requires args which we don't have)

Issue #4381, opened 2025-04-15
Return error on code exchange native OIDC login


jonas-jonas:
### Preflight checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [x] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Describe the bug

If an error occurs at the OIDC provider, or the user rejects consent at the provider, we don't return this error in the redirect back to the mobile app. Instead, we just do a blanket return without anything, which makes the DX in these cases challenging.

### Reproducing the bug

1. integrate the code exchange native OIDC login
2. reject an OIDC provider's consent screen
3. see a redirect back to the return_to URL, without any query parameters

See https://github.com/ory/kratos/blob/master/selfservice/flow/registration/error.go#L142-L144

The intended way to deal with this is, to re-fetch the flow that was used to initialize the OIDC flow, but that's not clear from the redirect. 

Ideally, we add `continue_with_flow_id=<flow_id>&continue_with_flow_type=<registration|login>`. It might be possible to also add the error message to the redirect URL directly. 

### Relevant log output

```shell

```

### Relevant configuration

```yml

```

### Version

master

### On which operating system are you observing this issue?

None

### In which environment are you deploying?

None

aeneasr:
Which customer is this and what is the expectation around resolution time? cc @JeffreyThijs

PR #4389, opened 2025-04-22
chore: change error type to 400 if id token processing fails

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/oidc/strategy.go
+++ b/selfservice/strategy/oidc/strategy.go
@@ -796,15 +796,15 @@ func (s *Strategy) CompletedAuthenticationMethod(context.Context) session.Authen
 func (s *Strategy) ProcessIDToken(r *http.Request, provider Provider, idToken, idTokenNonce string) (*Claims, error) {
 	verifier, ok := provider.(IDTokenVerifier)
 	if !ok {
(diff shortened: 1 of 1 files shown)


aeneasr:
tests failing :(

jonas-jonas:
Hmmm, returning 400 instead of 500 causes the error handler to add the error to the flow UI, which changes the whole flow of the flow. Technically, this could be considered breaking as well, though I'd argue that the 400 case must be handled by clients anyway, so the impact is low. 

However, it does make the tests awkward because of https://github.com/ory/kratos/issues/4381 (there is no way to get the flow ID that was used and to fetch the flow to assert the error message is correct).

Sidenote: I also think, not all the possible errors are 400, but rather an expected 500 (e.g. mis config/mis use of the API). 

All in all, a bit more work to get the tests over the line, I think.

aeneasr (approved):

aeneasr:
Tests are failing otherwise LGTM

aeneasr:
I think there is another PR to add misconfiguration errors in Ory Hydra by Patrik, we can probably just use that status code here as well.

aeneasr:
This one: https://github.com/ory/kratos/pull/3248

aeneasr:
https://github.com/ory/herodot/pull/135

PR #4390, opened 2025-04-22
feat: enable JSONNet templating for password migration hook

zepatrik:
This enables JSONNet body templating for the password migration hook. There is also a significant refactoring that I stumbled into and also included.


aeneasr (commented):
Doesn’t this need changes in the config schema?

driver/registry_default_hooks.go
```diff
@@ -67,7 +70,12 @@ func (m *RegistryDefault) getHooks(credentialsType string, configs []config.Self
 		case hook.KeySessionDestroyer:
 			i = append(i, m.HookSessionDestroyer())
 		case hook.KeyWebHook:
-			i = append(i, hook.NewWebHook(m, h.Config))
+			cfg := request.Config{}
+			if err := json.Unmarshal(h.Config, &cfg); err != nil {
+				m.l.WithError(err).WithField("raw_config", string(h.Config)).Error("failed to unmarshal hook configuration, ignoring hook")
```
aeneasr, line 75: This is opaque to the user and at the very least needs an event? Before webhook mis config would cause an error in the request chain, now it’s silently ignored which makes webhooks even nastier to configure in network IMO. Maybe move this back to the request chain?

aeneasr (commented):
Otherwise LGTM.

Is this change backwards compatible so if we have an existing migration hook without body config it keeps working?

driver/registry_default_hooks.go
zepatrik: OK, I re-added this to the request chain now.

driver/registry_default_hooks.go
```diff
@@ -57,44 +63,61 @@
...
-			var found bool
 			for name, m := range m.injectedSelfserviceHooks {
 				if name == h.Name {
-					i = append(i, m(h))
-					found = true
-					break
+					if h, ok := m(h).(T); ok {
+						hooks = append(hooks, h)
+					}
+					continue allHooksLoop
```
aeneasr, line 107: I think the goto is fine, but it confused me quite a lot

aeneasr (approved):
LGTM

PR #4394, opened 2025-04-28
feat: emit event on Jsonnet claims mapping error

hperl:
We now emit an event containing the Jsonnet input and output in anonymized form when mapping the claims in the OIDC flow fails.

### DIFF
--- a/x/events/events.go
+++ b/x/events/events.go
@@ -48,11 +48,19 @@ const (
 )
 
 const (
-	AttributeKeySessionID                       semconv.AttributeKey = "SessionID"
-	AttributeKeySessionAAL                      semconv.AttributeKey = "SessionAAL"
-	AttributeKeySessionExpiresAt                semconv.AttributeKey = "SessionExpiresAt"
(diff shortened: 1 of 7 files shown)


x/events/events.go
```diff
@@ -42,6 +43,7 @@ const (
 	WebhookDelivered         semconv.Event = "WebhookDelivered"
 	WebhookSucceeded         semconv.Event = "WebhookSucceeded"
 	WebhookFailed            semconv.Event = "WebhookFailed"
+	ClaimsMappingFailed      semconv.Event = "ClaimsMappingFailed"
```
aeneasr, line 46: Any alternatives to the naming?

aeneasr (approved):
Generally LGTM - question about naming. Please add a changelog entry to both PRs and docs! :)

x/events/events.go
hperl: Hmm, prefix with `OIDC` for context? I.e. `OIDCClaimsMappingFailed` ?

PR #4404, opened 2025-05-06
feat: add LoginStarted and RegistrationStarted events

gaultier:
**Changes:**
- Add `LoginStarted` and `RegistrationStarted` events along their required attributes
- Sort all event attributes alphabetically
- Emit these events when a new login/registration flow is created, *after* basic validation passed
- It is unclear yet how many of these events will be emitted, as such it is suggested that in a first phase, they remain internal and are not yet sent externally to avoid surprises (note: sometimes, these events can be emitted without user action such as simply visiting/being redirected to the sign-in page, etc)

**Documentation PR:** [ory/docs#2144](https://github.com/ory/docs/pull/2144)

**Issue:** https://github.com/ory-corp/cloud/issues/7895

Examples in Grafana:
- LoginStarted: Screenshot 2025-05-06 at 14 54 32
- RegistrationStarted: Screenshot 2025-05-06 at 14 46 17

### DIFF
--- a/selfservice/flow/login/handler.go
+++ b/selfservice/flow/login/handler.go
@@ -13,6 +13,7 @@ import (
 	"github.com/julienschmidt/httprouter"
 	"github.com/pkg/errors"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/ory/herodot"
 	hydraclientgo "github.com/ory/hydra-client-go/v2"
@@ -270,6 +271,8 @@ preLoginHook:
 		return nil, nil, err
 	}
 
+	span := trace.SpanFromContext(r.Context())
+	span.AddEvent(events.NewLoginStarted(r.Context(), f.ID, ft.String(), f.Refresh, f.OrganizationID, string(f.RequestedAAL)))
 	return f, nil, nil
 }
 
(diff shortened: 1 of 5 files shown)


x/events/events.go
```diff
@@ -45,39 +46,40 @@ const (
 	WebhookSucceeded         semconv.Event = "WebhookSucceeded"
 	WebhookFailed            semconv.Event = "WebhookFailed"
 	JsonnetMappingFailed     semconv.Event = "JsonnetMappingFailed"
+	LoginStarted             semconv.Event = "LoginStarted"
```
aeneasr, line 49: We have RecoveryInitiated and no started yet, should we also call this LoginInitiated to reduce the number of individual namings?

selfservice/flow/login/handler.go
```diff
@@ -270,6 +271,8 @@ preLoginHook:
 		return nil, nil, err
 	}
 
+	span := trace.SpanFromContext(r.Context())
+	span.AddEvent(events.NewLoginStarted(r.Context(), f.ID, ft.String(), f.Refresh, f.OrganizationID, string(f.RequestedAAL)))
```
aeneasr, line 275: I think it makes more sense to move this event to the persister, as we might call CreateLoginFlow in more places than just here, and really the loginflow is initiated when we persist it in the DB and return it to the client

gaultier: I checked and (at least currently) the only place that calls `CreateLoginFlow` is here (excluding tests of course). So for now this is equivalent. But I'm fine either way.

alnr (approved):

PR #4407, opened 2025-05-08
feat: improve verification required flows

aeneasr:
BREAKING CHANGE: The `require_verified_address` hook no longer returns a plain error. Previously, users had to manually start the verification flow, which caused a poor experience. Now, Ory Kratos automatically creates a verification flow and redirects the user using `continue_with` or an HTTP redirect. The verification flow starts with the first verified address found for the user. This aligns the behavior of `require_verified_address` with using the `verification` and `show_verification_ui` hook combination for login.


selfservice/hook/require_verified_address.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// TODO remove once flag is removed.
+	if e.r.Config().UseLegacyRequireVerifiedLoginError(ctx) {
+		span.AddEvent(semconv.NewDeprecatedFeatureUsedEvent(ctx, "legacy_require_verified_login_error"))
+		return login.ErrAddressNotVerified
+	}
+	// END TODO
+
+	i := s.Identity
+	for k := range i.VerifiableAddresses {
```
aeneasr, line 90: I know this is duplicated from the verification hook - however the verification hook works differently than this hook. The point of this hook is to require verification for login, the verification hook instead merely sends a verification email if the user is not yet verified, but does not enforce verification checks on login.

This hook now does what the verification hook does, which is sending an email, forwarding the user agent to verification, and blocking login.

selfservice/hook/require_verified_address_test.go
```diff
@@ -0,0 +1,286 @@
...
+						Verified: false,
+						Via:      identity.VerifiableAddressTypeEmail,
+					},
+				},
+			}
+
+			// Persist identity to satisfy foreign key constraints
+			require.NoError(t, reg.IdentityManager().Create(context.Background(), identity))
+
+			sessions := &session.Session{
```
aeneasr, line 155: maybe add logic to test if the email was sent out or not

selfservice/hook/verification_test.go
```diff
@@ -115,7 +115,11 @@ func TestVerifier(t *testing.T) {
 
 						messages, err := reg.CourierPersister().NextMessages(context.Background(), 12)
 						require.NoError(t, err)
-						require.Len(t, messages, 2)
+						if enabled {
+							require.Len(t, messages, 2)
+						} else {
+							require.Len(t, messages, 1)
```
aeneasr, line 121: the new behavior sends only one email, for the first unverified email it finds. it doesn't make sense in my view to send a code to all emails, because it's unclear to the user which code they have to enter in the next screen.

selfservice/hook/verification.go
```diff
@@ -147,23 +151,20 @@ func (e *Verifier) do(
...
 			return err
 		}
 
-		if address.Value != "" && address.Via == identity.VerifiableAddressTypeEmail {
-			verificationFlow.UI.Nodes.Append(
-				node.NewInputField(address.Via, address.Value, node.CodeGroup, node.InputAttributeTypeSubmit).
-					WithMetaLabel(text.NewInfoNodeResendOTP()),
-			)
-		}
+		verificationFlow.UI.Nodes.Append(
```
aeneasr, line 158: I think this will fix inconsistencies when using SMS channel. The resend button was probably removed to solve issues around resending the flow, but it's anyways not very effective in my view and should instead be solved with rate limiting or similar

selfservice/hook/require_verified_address.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (e *AddressVerifier) ExecuteLoginPostHook(w http.ResponseWriter, r *http.Request, _ node.UiNodeGroup, f *login.Flow, s *session.Session) (err error) {
+	ctx, span := e.r.Tracer(r.Context()).Tracer().Start(r.Context(), "selfservice.hook.Verifier.do")
+	r = r.WithContext(ctx)
+	defer otelx.End(span, &err)
+
+	// TODO remove once flag is removed.
+	if e.r.Config().UseLegacyRequireVerifiedLoginError(ctx) {
+		if f.Active != identity.CredentialsTypePassword {
```
jonas-jonas, line 60: why only non-password flows?

aeneasr: I don't know - it's only for the legacy flag, where this was the case. So I kept it there as I don't want to change behavior for clients that have the flag on!

PR #4409, opened 2025-05-09
feat: emit events on jsonnet failure when templating a jwt

gaultier:
- Fix typo: parital -> partial
- Document with comments why an event is not emitted or not documented
- Emit `JsonnetMappingFailed` events on jsonnet failure when templating a jwt (see https://www.ory.sh/docs/identities/session-to-jwt-cors). After review it seems we otherwise always emit events in all the right places, except in this very case. Tested end-to-end manually with the UI.

## Related issue(s)

https://github.com/ory-corp/cloud/issues/7291

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/identity/persister_identity.go
+++ b/persistence/sql/identity/persister_identity.go
@@ -573,8 +573,8 @@ func (p *IdentityPersister) CreateIdentities(ctx context.Context, identities ...
 		p.normalizeAllAddressess(ctx, identities...)
 
 		if err = p.createVerifiableAddresses(ctx, tx, identities...); err != nil {
-			if paritalErr := new(batch.PartialConflictError[identity.VerifiableAddress]); errors.As(err, &paritalErr) {
-				for _, k := range paritalErr.Failed {
+			if partialErr := new(batch.PartialConflictError[identity.VerifiableAddress]); errors.As(err, &partialErr) {
+				for _, k := range partialErr.Failed {
 					failedIdentityIDs[k.IdentityID] = struct{}{}
 				}
 			} else {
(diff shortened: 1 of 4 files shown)


persistence/sql/identity/persister_identity.go
```diff
@@ -1129,6 +1129,9 @@ func (p *IdentityPersister) DeleteIdentity(ctx context.Context, id uuid.UUID) (e
 	return nil
 }
 
+// This function is only used internally to cleanup partially created identities,
+// when creating a batch of identities at once and some failed to be fully created.
+// This act should not be observable externally and thus we do not emit an event.
```
hperl, line 1134: In Go comments that document the function go outside the function and start with the function name.

As this reads like an internal comment, it should be moved to the function body.

Also, at some point we wanted to expose batch delete to the API, but maybe it was premature to make it public here ;).

selfservice/hook/verification.go
```diff
@@ -76,6 +76,7 @@ func (e *Verifier) ExecuteLoginPostHook(w http.ResponseWriter, r *http.Request,
 	r = r.WithContext(ctx)
 	defer otelx.End(span, &err)
 	if f.RequestedAAL != identity.AuthenticatorAssuranceLevel1 {
+		// NOTE: This is a debug event thus not using a documented event name.
```
hperl, line 79: IMHO this should not even be an event, but an attribute on the span as such:

```go
span.SetAttributes(attribute.String("skip_reason", "AAL is not 1"))
```

WDYT?

hperl (approved):
Changes LGTM! Just a nit about the function comment, and maybe we could remove the undocumented event altogether?

selfservice/hook/verification.go
gaultier: Agreed, this particular event feels out of place. I can change it to an attribute, if no one has a strong reason against it.

PR #4415, opened 2025-05-15
fix: add default issuer URL for LINE

alnr:
Fixed+expanded relevant comment.

Fixed some tracing issues.

Added error info and missing res.Body.Close() in courier.

### DIFF
--- a/courier/http_channel.go
+++ b/courier/http_channel.go
@@ -6,6 +6,7 @@ package courier
 import (
 	"context"
 	"fmt"
+	"io"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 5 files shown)


courier/http_channel.go
```diff
@@ -109,9 +112,13 @@ func (c *httpChannel) Dispatch(ctx context.Context, msg Message) (err error) {
 		"unable to dispatch mail delivery because upstream server replied with status code %d",
 		res.StatusCode,
 	)
+
+	body, _ := io.ReadAll(res.Body)
```
aeneasr, line 116: res.body has ben read already in L96, this will fail with an empty body

aeneasr (changes requested):
One comment, otherwise lgtm

courier/http_channel.go
alnr: This could be written more simply as

```go
	defer res.Body.Close()
	body := make([]byte, 1024)
	n, _ :=res.Body.Read(body)
	body = body[:n]
```

aeneasr: Ah, makes sense!

aeneasr (approved):

Issue #3543, opened 2023-10-01
Native flow ignores redirect_to in oidc flow if session in browser already established


luflow:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Ory Network Project

selfhosted

### Describe the bug

We found a strange behavior in Kratos thats somehow leading to ignoring our set `redirect_to` URL and always falling back to the `default_browser_return_url`. It happens in the case where a session in a browser is already established, but afterwards an API native flow is started with a 3rd-party Login. The app opens the browser, the session at the 3rd party is still valid so redirects are taking place and then I suddenly land on `default_browser_return_url` instead of the `redirect_to` specified in the API call.

### Reproducing the bug

Steps to reproduce behavior:

Precondition: You need a native app where the native login flow can be executed with the Kratos API

1. Setup kratos with an OIDC flow e.g. with GitHub including some simple mapping (email and firstname)
2. Setup correct `allowed_return_urls` and `default_browser_return_url` for your selfservice flows
3. Use an android device (where sessions in browser are shared between a native app and browser)
4. Open the android browser and login to your website with the GitHub OIDC integration, stay logged in
5. Go to your native app and start a login flow via the API with the GitHub OIDC integration with a `redirect_to` to `myapp://module/login`
6. After the flows redirects in the browser are done, now you will not be redirected to `myapp://module/login` but instead to `default_browser_return_url`

Somehow it looks like the redirect_to information gets lost on the way?

As soon as I logout in the browser manually and start from Step 5 again - everything works as intended and the native app flow gets completed correctly.

### Relevant log output

```shell
no errors shown in logs.
```

### Relevant configuration

```yml
selfservice:
  default_browser_return_url: https://yourwebsite.com
  allowed_return_urls:
    - https://yourwebsite.com/auth
    - https://yourwebsite.com/

    # App Links
    - myapp://
    - myapp://module/login
  methods:
[...]

PR #4486, opened 2025-10-20
fix: respond with a session token exchange code if user is already logged in

luflow:
This PR fixes an issue where native login flows (API flows) would not receive a session token exchange code when a user already has an active browser session.

In our case users have most likely a web session and will later, when they install our app, also login via e.g. OIDC in the App into the same kratos instance.

Corresponding issue #3543

### The Problem

When initiating a native login flow with `return_session_token_exchange_code=true`, the exchange code was only generated if **no active session existed**. This caused issues in the following scenario:

1. User logs in via browser flow → browser session cookie is set
2. User attempts native login (in our mobile app) and uses OIDC for that → the browser cookie is sent along with the request
3. Kratos detects the existing session and skips code generation
4. Native app receives no exchange code → login fails / login screen is shown again

The code generation was conditional on `ErrNoActiveSessionFound`, which prevented native apps from receiving the necessary exchange code when users had concurrent browser sessions.

### The Solution

Moved the session token exchange code generation before the session check. Now, whenever a native login flow requests an exchange code via `return_session_token_exchange_code=true`, the code is created regardless of existing sessions.

This ensures that:
- ✅ First-time logins receive an exchange code
- ✅ Subsequent logins with existing browser sessions also receive an exchange code
- ✅ Native apps can successfully complete authentication even when users have active browser sessions

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/selfservice/flow/login/handler_test.go
+++ b/selfservice/flow/login/handler_test.go
@@ -644,7 +644,7 @@ func TestFlowLifecycle(t *testing.T) {
 
 				apiRes, err := browserClient.Do(req)
 				require.NoError(t, err)
-				require.NoError(t, apiRes.Body.Close())
+				defer func() { _ = apiRes.Body.Close() }()
 
 				apiBody, err := io.ReadAll(apiRes.Body)
 				require.NoError(t, err)


luflow:
Not working as expected. Maybe someone else has an idea how to fix that?

PR #4532, opened 2026-01-27
fix: use proper key format in ContinuityCookieManager and prevent panic in XChaCha20Poly1305

d191:
This PR fixes two issues:

   1. ContinuityCookieManager now uses the same key format as CookieManager by applying SHA256 derivation to secrets. This enables proper key rotation support.

   2. XChaCha20Poly1305.Decrypt now checks len(rawCiphertext) instead of len(ciphertext) for minimum length validation, preventing panic on short ciphertexts.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/cipher/chacha20.go
+++ b/cipher/chacha20.go
@@ -76,7 +76,7 @@ func (c *XChaCha20Poly1305) Decrypt(ctx context.Context, ciphertext string) ([]b
 			return nil, errors.WithStack(herodot.ErrInternalServerError.WithWrap(err).WithReason("Unable to instantiate chacha20"))
 		}
 
-		if len(ciphertext) < aead.NonceSize() {
+		if len(rawCiphertext) < aead.NonceSize() {
 			return nil, errors.WithStack(herodot.ErrInternalServerError.WithReason("cipher text too short"))
 		}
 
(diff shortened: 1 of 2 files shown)


vinckr:
Hello @d191 
did you talk about this contribution with the maintainers in an issue or github discussion? 

before reviewing / commenting on a PR it would be important to make sure what problem this is solving and if it's aligned with the direction that Kratos is going currently.

d191:
Hello @vinckr 

These are runtime bugs that cause crashes - I didn't find any prior discussion about them.

1. **ContinuityCookieManager key rotation**

The issue is with how keys are passed to sessions.NewCookieStore. This function expects keys in pairs:
 - First key in pair = authentication key
 - Second key in pair = encryption key

For key rotation, the pattern is: [k1, sha256(k1), k2, sha256(k2), ...]

The old code was passing raw secrets [k1, k2, ...] which means:
 - k1 used for authentication
 - k2 used for encryption (but this is not k1's hash!)

This broke key rotation because the CookieStore couldn't properly match old keys with new ones after secret rotation.

The fix adds SHA256 derivation to create proper key pairs, matching the pattern used by CookieManager.

2. **XChaCha20Poly1305 bounds error**

When checking minimum ciphertext length, the code was using len(ciphertext) (hex string) instead of len(rawCiphertext) (decoded bytes), causing panic on short ciphertexts.

vinckr:
>These are runtime bugs that cause crashes - I didn't find any prior discussion about them.

Can you please create a Design Document issue first - this is needed for any non-trivial changes to the code base: https://github.com/ory/kratos/issues/new/choose

Will see to update CONTRIBUTING.md to make this a bit more clear that we need to have some discussion about the bug and goals here before looking at a PR like this.

Issue #4536, opened 2026-02-11
feat: Add use_oidc_discovery_issuer option for generic OIDC provider (Azure AD B2C support)


getlarge:
## Preflight checklist

- [x] I could not find a solution in the existing issues, FAQ, or Ory documentation
- [x] I agree to follow the Ory Community Code of Conduct
- [x] I have read and am following this repository's contribution guidelines
- [x] This feature request has not been reported before

## Problem

Azure AD B2C violates [OIDC Discovery spec §4.3](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation): the `issuer` in the discovery document does not match the URL used to fetch it.

```
Discovery URL:  https://<tenant>.b2clogin.com/<tenant>.onmicrosoft.com/<policy>/v2.0/.well-known/openid-configuration
Issuer returned: https://<tenant>.b2clogin.com/<tenant-id>/v2.0/
```

The discovery URL must include the B2C policy name, but the returned issuer uses the tenant GUID and omits the policy. This mismatch is unavoidable — all B2C discovery URL variants (tenant-name, tenant-id, `/tfp/` prefix) return the same GUID-based issuer.

Kratos (via go-oidc) strictly validates issuer == discovery URL and rejects the provider:

```
oidc: issuer did not match the issuer returned by provider,
expected "https://<tenant>.b2clogin.com/<tenant>.onmicrosoft.com/<policy>/v2.0"
got "https://<tenant>.b2clogin.com/<tenant-id>/v2.0/"
```

### Why existing provider types don't work

| Provider | Behavior | Result |
|----------|----------|--------|
| `microsoft` | Hardcodes `login.microsoftonline.com/{tenant}` | Redirects to generic MS login, not B2C custom flow |
| `generic` + `issuer_url` | OIDC Discovery with strict issuer validation | Issuer mismatch error |

### B2C-side mitigation

Azure Portal has a "Token compatibility > Issuer (iss) claim" setting, but the `tfp` issuer format may only be available with B2C custom policies, not standard user flows. This places the burden on the customer to migrate from user flows to custom policies — a non-trivial effort.

## Proposed solution

Add a `use_oidc_discovery_issuer` boolean field to the generic OIDC provider configuration. When `true`, use `go-oidc`'s `InsecureIssuerURLContext` to allow the discovery URL to differ from the issuer.

[...]

PR #4537, opened 2026-02-11
feat: add use_oidc_discovery_issuer option for generic OIDC provider

getlarge:
## Related issue

Closes https://github.com/ory/kratos/issues/4536

## Changes

1. **`selfservice/strategy/oidc/provider_config.go`** — add `UseOIDCDiscoveryIssuer bool` field
2. **`selfservice/strategy/oidc/provider_generic_oidc.go`** — pre-fetch discovery document to get the real issuer, pass it to `InsecureIssuerURLContext`, then call `NewProvider` with the discovery URL
3. **`embedx/config.schema.json`** — add `use_oidc_discovery_issuer` to provider schema
4. **`selfservice/strategy/oidc/provider_generic_test.go`** — 3 unit test cases
5. **`selfservice/strategy/oidc/strategy_test.go`** — 1 e2e test (full registration flow with Hydra and mismatched issuer)

## Why pre-fetch the discovery document?

`InsecureIssuerURLContext` serves two purposes in go-oidc:
1. Skip the discovery URL == issuer validation
2. Store the passed value as the provider's issuer for **ID token verification**

If we naively pass the discovery URL (e.g. `https://tenant.b2clogin.com/tenant.onmicrosoft.com/policy/v2.0`), the provider stores that as its issuer. But the ID token's \`iss\` claim contains the **real** issuer from the discovery document (e.g. `https://tenant.b2clogin.com/tenant-guid/v2.0/`). Token verification then fails with "id token issued by a different provider".

So we first fetch the discovery document to extract the real issuer, pass **that** to `InsecureIssuerURLContext`, then let `NewProvider` do its normal discovery (which is a cache hit thanks to the ristretto cache in Ory's go-oidc fork).

## Test plan

- [x] Unit tests: issuer mismatch rejected by default, accepted with flag, discovered endpoints used
- [x] E2e test: full OIDC registration flow with Hydra using mismatched issuer (localhost vs 127.0.0.1)
- [x] Manual: tested against real Azure AD B2C tenant with standard user flows

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/selfservice/strategy/oidc/provider_generic_oidc.go
+++ b/selfservice/strategy/oidc/provider_generic_oidc.go
@@ -88,7 +88,7 @@ func discoverIssuer(ctx context.Context, issuerURL string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer resp.Body.Close()
+	defer func() { _ = resp.Body.Close() }()
 	body, err := io.ReadAll(io.LimitReader(resp.Body, 1<<20))
 	if err != nil {
 		return "", err


getlarge:
If you are open to this adjustment, I can add some automated e2e tests.

PR #4543, opened 2026-03-04
fix: preserve user_handle when adding a second WebAuthn security key

### LINKED ISSUE #4519: Bug: WebAuthn `user_handle` is overwritten during Settings flow, breaking existing credentials
### Preflight checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/ory/kratos/blob/master/CODE_OF_CONDUCT.md).
- [x] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/kratos/blob/master/CONTRIBUTING.md).
- [x] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Ory Network Project

_No response_

### Describe the bug

When adding a second WebAuthn/Passkey via the Settings flow, Ory Kratos (v25.4.0) overwrites the existing `user_handle` in the identity credentials configuration. 

Per the WebAuthn specification, the `user.id` (user_handle) must be stable for the lifetime of the identity. By overwriting this handle, any authenticators registered *prior* to the update will fail to authenticate during login. This happens because the handle provided in the login challenge (the new one) no longer matches the handle stored inside the physical security key or passkey provider (the old one).

### Potential Root Cause

The issue is located in `selfservice/strategy/webauthn/settings.go`. The current implementation appears to unconditionally assign the `IdentityID` to the `UserHandle` during the settings update:

```go
// Location: selfservice/strategy/webauthn/settings.go
cc.UserHandle = ctxUpdate.Session.IdentityID[:]
```

This logic does not account for the fact that a UserHandle might already be established for that identity's credentials.

### Reproducing the bug

1. **Register** a new user using WebAuthn (Passkey) with `passwordless: true`.
2. **Verify Database:** Note the `user_handle` value in the `identity_credentials` table for the `webauthn` method.
3. **Start Settings Flow:** Add a *second* security key via the self-service settings flow.
4. **Verify Database again:** Observe that the `user_handle` has been updated to a new value (likely matching the Identity ID slice), even though it should have remained the same.
5. **Attempt Login:** Try to log in using the **first** key.
6. **Result:** The authentication fails because the `user_handle` in the challenge does not match the one stored in the first authenticator.

### Relevant log output

```shell

```

### Relevant configuration

```yml
selfservice:
  methods:
    webauthn:
      enabled: true
      config: 
        passwordless: true
```

### Version

v25.04

None

None

### Proposed Fix

The `UserHandle` should only be set if it is currently empty. This ensures that the first-generated handle persists for the life of the identity, allowing all registered keys to remain valid.

```go
if len(cc.UserHandle) == 0 {
    cc.UserHandle = ctxUpdate.Session.IdentityID[:]
}
```

or maybe following will do it also right: 
```go 
cc.UserHandle = webAuthnSess.UserID
```

Micaso:
The WebAuthn specification requires that user.id (user_handle) remains stable for the lifetime of an identity. The settings flow was unconditionally overwriting user_handle with the session identity ID each time a new WebAuthn credential was added, causing previously registered authenticators to fail during login because the stored handle no longer matched.

Only set UserHandle when it is not yet present (first credential added via settings). Existing credentials keep their original handle.

Closes #4519

## Related issue(s)

#4519

## Checklist

      introduces a new feature.
- [x] I am following the
      [security@ory.com](mailto:security@ory.com)) from the maintainers to push
      the changes.
      works.

### DIFF
--- a/selfservice/strategy/webauthn/settings.go
+++ b/selfservice/strategy/webauthn/settings.go
@@ -283,7 +283,9 @@ func (s *Strategy) continueSettingsFlowAdd(ctx context.Context, ctxUpdate *setti
 	wc.AddedAt = time.Now().UTC().Round(time.Second)
 	wc.DisplayName = p.RegisterDisplayName
 	wc.IsPasswordless = s.d.Config().WebAuthnForPasswordless(ctx)
-	cc.UserHandle = ctxUpdate.Session.IdentityID[:]
+	if len(cc.UserHandle) == 0 {
+		cc.UserHandle = ctxUpdate.Session.IdentityID[:]
+	}
 
 	cc.Credentials = append(cc.Credentials, *wc)
 	co, err := json.Marshal(cc)
--- a/selfservice/strategy/webauthn/settings_test.go
+++ b/selfservice/strategy/webauthn/settings_test.go
@@ -368,6 +368,58 @@ func TestCompleteSettings(t *testing.T) {
 		})
 	})
 
+	t.Run("case=user_handle is not overwritten when adding a second security key", func(t *testing.T) {
+		// Regression test for https://github.com/ory/kratos/issues/4519
+		// The user_handle must remain stable per the WebAuthn spec (user.id must not change).
+		var id identity.Identity
+		require.NoError(t, json.Unmarshal(settingsFixtureSuccessIdentity, &id))
+		id.NID = uuid.Must(uuid.NewV4())
+		_ = reg.PrivilegedIdentityPool().DeleteIdentity(t.Context(), id.ID)
+		browserClient := testhelpers.NewHTTPClientWithIdentitySessionCookie(t.Context(), t, reg, &id)
+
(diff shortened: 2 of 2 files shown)


Micaso:
The two failing CI checks are pre-existing issues unrelated to this fix:

- **golangci-lint**: Fails with `the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26)`. The CI pins golangci-lint `v2.4.0` (built with Go 1.25) but `go.mod` requires Go 1.26. This fails on any PR right now.

- **Anchore Scanner**: Fails with high-severity CVEs found in the base Docker image `oryd/kratos:a52cd61`. These are image/dependency vulnerabilities, not introduced by this change.

Both failures are visible on the `master` branch as well. The actual code change and its regression test pass cleanly.

PR #4597, opened 2026-08-29
fix: stop recording ErrCompletedByStrategy as a span error (closes #4109)

### LINKED ISSUE #4109: `ErrCompletedByStrategy` is not an error, but shows up in telemetry as such.
### Preflight checklist

- [X] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Ory Network Project

_No response_

### Describe the bug

Strategy functions, such as `selfservice.strategy.oidc.strategy.Login` have multiple branches where they produce the `ErrCompletedByStrategy` result.

Doing so will mark the span as errored.
image

However this is _not actually an error_, it's a control flow detail being used like a "break" on the caller of that function.

### Reproducing the bug

1. Configure an OIDC provider for Kratos v1.2.0.
2. Configure an OTLP exporter for tracing.
3. Perform a normal, successful OIDC login.
4. Multiple spans have a false-positive span error.

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

Kratos (server) v1.2.0

Linux

Kubernetes with Helm

### Additional Context

_No response_

pacocartones:
Closes #4109.

### The bug
`ErrCompletedByStrategy` is a control-flow signal (`selfservice/flow/error.go`), not a failure. The login handlers correctly suppress the *client* error for it (`if !errors.Is(err, flow.ErrCompletedByStrategy) { WriteError }`), but leave the tracked `err` set, so the deferred `otelx.End(span, &err)` records the signal as a span error/exception — a completed flow appears in telemetry as an error.

### The fix
Clear the tracked `err` where it genuinely carries the signal, in the two handlers whose deferred `otelx.End(span, &err)` tracks it: `createNativeLoginFlow` (the reported one) and `createBrowserLoginFlow`. Client behavior and API are unchanged.

Scope note (checked, to avoid a no-op): the other update handlers listed around this issue are **not** affected — `updateLoginFlow`/`updateSettingsFlow` use a loop-shadowed `err` the deferred `otelx.End` doesn't track, and the registration/recovery/verification update handlers have no `otelx.End(span, &err)`. So the fix is these two sites.

### Test
New `handler_completed_by_strategy_test.go`: installs a `tracetest` span recorder as the registry tracer, forces a strategy to return `ErrCompletedByStrategy`, drives the real API endpoint, and asserts the `createNativeLoginFlow` span is not `codes.Error` and carries no exception event. Passes; reverting only the fix makes it fail with span status `"flow response completed by strategy"`. (Focused test via in-memory mocks; the full login package's DB-backed suites need dockertest.)

---
_Disclosure: this contribution was prepared with AI assistance (Claude Code); the fix, scope, and test were verified against HEAD._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling across login, recovery, registration, verification, and settings flows when a strategy completes the response directly.
  * Prevented duplicate redirects or responses after a flow has already completed.
  * Ensured successful strategy-completed outcomes are not reported as errors in telemetry.

* **Tests**
  * Added regression coverage confirming completed authentication flows return successfully and remain absent from error and exception telemetry.


gaultier:
Thanks for the contribution. The intent is good, but there are two issues:
- Not all sites are fixed
- This error as you pointed is not a real error. 

Thus a better fix is to make this control value not an error, but instead an additional return value.

pacocartones:
Thanks @gaultier — both points addressed.

I removed `ErrCompletedByStrategy` entirely and modelled completion as an additional return value (`completed bool`) across the strategy interfaces (`Login` / `FastLogin1FA` / `FastLogin2FA`, `Register`, `Recover`, `Verify`, `Settings`) and `Handler.NewLoginFlow`, threading it through every implementation and caller. Deleting the sentinel let the compiler confirm no producer, consumer or comparison was missed.

On "not all sites": the original change only cleared the two login create-handler spans. The errored spans in your screenshot actually came from the internal OIDC strategy spans (`oidc.Login` / `processLogin`), which are now clean because completion is never an error at any layer. Where a strategy's internal call graph is deep — OIDC account-linking in `HandleError`, and the code/link recovery+verification chains — an unexported package-local marker carries completion internally and is translated to the `completed` value (and neutralised before the span's `otelx.End`) so it never surfaces as an error or an errored span.

`go build ./...` and `go vet ./...` pass, `gofmt` is clean, and the focused telemetry regression test passes; the DB-backed self-service suites will run in CI.

It's a broad change — it touches every self-service flow (44 files) — because the sentinel deletion is the only thing that unambiguously closes "not all sites". If you'd rather review it incrementally, it splits cleanly per flow (login first, then registration / recovery / verification / settings) — just say the word and I'll break it up.

pacocartones:
The updated head is now green on unit/lint, PostgreSQL and SQLite E2E, and the PostgreSQL/SQLite/CockroachDB Playwright matrices. The remaining failures appear unrelated to this change: MySQL stopped during `npm ci` with `ECONNRESET` before tests, CLI docs reproduces the same missing `ui-messages.md` failure as the previous head, and the image scanner has failed across recent neighboring runs. No additional product-code change seems warranted here; a maintainer rerun/review would be appreciated.

